Issue #4, opened 2015-07-01
Usage documentation is not correct on Windows


chinyeungli:
OS: Windows 7 - 64bits
Python 2.7
scancode version: 1.0.0

```
(scancode-toolkit-1.0.0) C:\Users\Downloads\scancode-toolkit-1.0.0>scancode --help
Usage: scancode-script.py [OPTIONS] <input> <output_file>
```

There is no `scancode-script.py` in the extracted directory.
The usage doc is incorrect. It should be
```
Usage: scancode [OPTIONS] <input> <output_file>
```

pombredanne:
This is a bug in the click library used for command line processing.  This is only on Windows though.
The click bug for reference is: https://github.com/mitsuhiko/click/issues/365

pombredanne:
This is partially fixed: when running a plain scancode command or on errors, `scancode` is displayed correctly

When running `scancode -h` , then `scancode-script.py` is still displayed.

pombredanne:
@chinyeungli can you confirm this is fixed?

chinyeungli:
@pombredanne It is fixed
```
(scancode-toolkit-master) C:\Users\CYL\Downloads\scancode-toolkit-master>scancode --help
Usage: scancode [OPTIONS] <input> <output_file>
```

pombredanne:
@chinyeungli thank you for testing!

PR #6, opened 2015-07-04
Reject running scancode --extract with other options.

pombredanne:
* also added cli tests


pombredanne:
This should fix issue #4 current problem

Issue #45, opened 2015-07-31
I can only see the resources displayed as a table instead of a tree.


linexb:
I ran: scancode --format html-app samples samples.html on my windows machine.

And the result on my machine only shows table on left resource panel. 
![sample](https://cloud.githubusercontent.com/assets/13014965/9005734/435abfc6-37b3-11e5-9341-dcfe144619e5.png)

pombredanne:
This is likely due to using backslashes vs posix paths.

pombredanne:
This is a tree alright but not broken down in nodes

pombredanne:
The issue is posix path vs. windows path separator on the one hand and also #46 absolute vs. relative path on the other hand

pombredanne:
This is fixed now

PR #49, opened 2015-08-04
#45, use posix path to replace windows path and addtest

linexb:


tests/scancode/test_cli.py
```diff
@@ -34,6 +34,8 @@
 
 from scancode import cli
 import codecs
+import ntpath
+import posixpath
```
pombredanne: Do you mind not add these two imports? They are not used anywhere

pombredanne:
Thanks! Do you mind adding a test for this?
Specifically I think the test should verify that the JSON, html and html-app outputs do not contain Windows backslash separated paths

PR #51, opened 2015-08-06
#45, fix the issue and add test cases

linexb:


tests/scancode/test_cli.py
```diff
@@ -199,4 +199,13 @@ def test_info_license_copyrights(self):
         assert 'Scanning done' in result.output
         expected = self.load_json_result(self.get_test_loc('info/all.expected.json'), test_dir)
         loaded_result = self.load_json_result(result_file, test_dir)
-        assert expected == loaded_result
\ No newline at end of file
+        assert expected == loaded_result
+
+    def test_html_app_paths_are_posix_paths(self):
```
pombredanne: Could you run this test only on copyrights and on a single file rather than multiple files? This will be faster.

Also can you check the three formats? json, html-app and html, possibly in three separate tests

pombredanne:
I am closing this in favor of #53

PR #53, opened 2015-08-07
#45, fix the posix path issue and add test cases

linexb:
Fix the posix path issue and add test cases.


pombredanne:
@linexb Thanks! I will merge this soon.

Issue #38, opened 2015-07-27
Separate JSON data into a separate file from html app


jdaguil:
The html app should have the JSON data in a separate file rather than inside the file

pombredanne:
This makes perfect sense... IMHO to work around the limitations of loading JSON files in a local HTML page and the differences between browsers you might want to take a look at using JSONP?

pombredanne:
@jdaguil I think this is all fixed, right? if so please close

PR #66, opened 2015-08-31
#38 separated json into separate file

jdaguil:
writing data.json file into the assets folder instead of into the template

### DIFF
--- a/src/scancode/format.py
+++ b/src/scancode/format.py
@@ -29,6 +29,8 @@ from operator import itemgetter
 from os.path import dirname
 from os.path import exists
 from os.path import join
+from os.path import abspath
+from os import makedirs
 
 from commoncode import fileutils
 
@@ -57,20 +59,14 @@ def get_template_dir(format):  # @ReservedAssignment
     return join(dirname(__file__), 'templates', format)
 
 
-def as_html_app(detected_data, scanned_path, output_file):
+def as_html_app(scanned_path, output_file):
     """
     Return an HTML string built from a list of results and the html-app template.
     """
     template = get_html_template('html-app')
-    import json
-    html_dirs = get_html_app_files_dirs(output_file)
-    if html_dirs:
-        _, assets_dir = html_dirs
-    else:
-        assets_dir = ''
-    return template.render(results=json.dumps(detected_data),
-                           assets_dir=assets_dir,
-                           scanned_path=scanned_path)
+    _, assets_dir = get_html_app_files_dirs(output_file)
+
+    return template.render(assets_dir=assets_dir, scanned_path=scanned_path)
 
 
 class HtmlAppAssetCopyWarning(Exception):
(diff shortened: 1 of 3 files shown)


src/scancode/format.py
```diff
@@ -81,21 +77,26 @@ class HtmlAppAssetCopyError(Exception):
...
     Return a tuple of (parent_dir, dir_name) directory named after the
     `output_file` file object file_base_name (stripped from extension) and a
-    `_files` suffix Return None if output is to stdout.
+    `_files` suffix Return empty strings if output is to stdout.
     """
+    if is_stdout(output_file):
+        return '', ''
+
     file_name = output_file.name
-    if file_name == '<stdout>':
```
pombredanne: @jdaguil can you elaborate why you updated the logic here?

jdaguil: @pombredanne The change was to make the check for '\<stdout>' more explicit. 

**BEFORE**:
<pre><code>
tgt_dirs = get_html_app_files_dirs(output_file)
if not tgt_dirs:
    raise HtmlAppAssetCopyWarning()         # not clear warning from stdout
</code></pre>
**AFTER**:

<pre><code>
if is_stdout(output_file):
    raise HtmlAppAssetCopyWarning()
tgt_dirs = get_html_app_files_dirs(output_file)         # output valid, no need to check
</code></pre>

src/scancode/format.py
```diff
@@ -105,14 +106,21 @@ def create_html_app_assets(output_file):
...
         target_dir = join(*tgt_dirs)
         if exists(target_dir):
             fileutils.delete(target_dir)
         fileutils.copytree(assets_dir, target_dir)
+
+        # write json data
+        import json
+        root_path, assets_dir = get_html_app_files_dirs(output_file)
+        with open(join(root_path, assets_dir, 'data.json'), 'w') as f:
+            f.write('data=' + json.dumps(results))
```
pombredanne: should the variable be called scandata instead ... data feels quite generic otherwise?

src/scancode/templates/html-app/template.html
```diff
@@ -113,6 +113,7 @@
   <script src="{{ assets_dir }}/jquery.min.js"></script>
   <script src="{{ assets_dir }}/jstree.min.js"></script>
   <script src="{{ assets_dir }}/jquery.dataTables.min.js"></script>
+  <script type="text/javascript" src="{{ assets_dir }}/data.json"></script>
```
pombredanne: To avoid same-origin or json vs. js issues, should the file have a .js extension instead of .json?

src/scancode/format.py
```diff
@@ -81,21 +77,26 @@ class HtmlAppAssetCopyError(Exception):
...
 def get_html_app_files_dirs(output_file):
     """
     Return a tuple of (parent_dir, dir_name) directory named after the
     `output_file` file object file_base_name (stripped from extension) and a
-    `_files` suffix Return None if output is to stdout.
+    `_files` suffix Return empty strings if output is to stdout.
     """
+    if is_stdout(output_file):
+        return '', ''
+
```
pombredanne: It might be best to return None,None rather than empty strings.

Issue #78, opened 2015-09-21
tree navigation problem in scancode results


rakeshbalusa:
While scrolling down through the scancode results in html a lot of time is wasted. If I have to check the license or copyright for the last file in the directory I must scroll down to the bottom of the html page and then scroll back to the top. If the results are fixed to a particular location in the html page a lot of time can be saved.

balusarakesh:
issue fixed in the latest version of scancode

PR #105, opened 2015-10-21
#78 Added fixed vertical height for tree view

jdaguil:
For #78 I've added a fixed vertical height for scrolling in the tree view.

### DIFF
--- a/src/scancode/templates/html-app/template.html
+++ b/src/scancode/templates/html-app/template.html
@@ -66,6 +66,11 @@
         border: 1px solid #CCC;
         border-radius: 4px;
         }
+        .tree_scrolling{
+            overflow-x:scroll;
+            overflow-y:scroll;
+            height: 500px;
+        }
     </style>
 </head>
 <body>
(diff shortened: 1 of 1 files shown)


pombredanne:
@jdaguil  Thanks for the PR!.... However I am not sure this is what should be done.
The height of the tree is not the problem , and it should IMHO match the height of the table based on the number of displayed rows.

However, the top level page bar, the whole tree itself and the header row of the table should be fixed to the top of the page, such that when you scroll, you scroll the table and not the tree not the headers.

The tree itself does not have to have a fixed size proper.

Having a fixed size constrained tree feels weird and small in most cases and I am not sure that it helps with navigation on pages with either very deep trees or many table elements displayed.

What's your take?

jdaguil:
Thanks for the description @pombredanne! what you have suggested makes sense. From the ticket (#78), I misunderstood and thought the issue had to do with scrolling in the tree view.

pombredanne:
@jdaguil The symptom is the scrolling in the tree view, but the underlying issue is that I think the tree could possibly be some iframe of sorts and things should be done such that you always have  visible on screen the tree bits and the table bits that you are looking at

pombredanne:
@jdaguil this shows up as closed, because I merged, but I then reverted the commit... So this was not merged really

Issue #123, opened 2015-11-24
Include full license texts / offline html page


ened:
During our talks with the lawyer, they requested a full _offline_ version of the HTML output in order to comply with the various legal constraints that LGPL/Apache/GPL etc bring.

That means:
* No external links for licenses
* Full copies of all license texts included in the HTML page (at the bottom)

Does the project already include the full copies of the license text somewhere? I'd like to contribute this feature.

pombredanne:
@ened What is the purpose? 
Is this to generate an attribution document to accompany redistributed software?
Or is this to collect all actual detected license texts, notices and mentions?

>Does the project already include the full copies of the license text somewhere?
All the texts are in https://github.com/nexB/scancode-toolkit/tree/develop/src/licensedcode/data

ened:
Because we are going to ship a version of the html-document with an embedded device, which MAY NOT have an active internet connection at the time of an legal check.

However, most license require full attribution and a full copy within the attribution document.

Thanks for the link to the texts. My proposal is this to add one option to ./scancode, like "--license-full-attribution", which will trigger this within the html template.

ened:
> Is this to generate an attribution document to accompany redistributed software?

Yes. For a Android system.

pombredanne:
ok, this makes sense. In earnest I never thought of this output as a possible attribution doc....
For this there is another tool called `AboutCode` https://github.com/dejacode/about-code-tool 
But it is not yet integrated in ScanCode...
That said, to collect all the texts there are a few aspects to consider:
1. The actual texts matched may need to be extracted: they may differ slightly from the reference texts  which is the whole purpose of the license detection. Since the returned license detection results contain internally the line and character offsets of the matched to regions, this can be automated.
2. These should be returned in the scan results and a custom template could then make use of them (or they could be processed in another way from the JSON data)

There are a few caveats to be aware though: 
1. the scans should always be reviewed IMHO first.
2. some of the detected licenses may be incomplete for correct attribution production. For instance say only an LGPL-3.0 notice is present and was detected. Since the LGPL-3.0 is an exception to the GPL-3.0, a correct and accurate attribution should contain eventually the notice, with eventual copyrights, and the full texts of both the LGPL-3.0 and GPL-3.0. This cannot be inferred entirely from the actual scan results nor the actual metadata available with each license..
There are likely a few more issues to pop up.

pombredanne:
@ened directly related is also the techniques used in Android for producing license attribution.
[...]

PR #128, opened 2015-11-24
Added a new "license.object" path to the format output 

ened:
=> can be used to retrieve the full license test through lazy loading

Issue discussion: #123.

### DIFF
--- a/src/scancode/format.py
+++ b/src/scancode/format.py
@@ -129,6 +129,8 @@ def as_html(detected_data):
     """
     Return an HTML string built from a list of results and the html template.
     """
+    from licensedcode.models import get_license
+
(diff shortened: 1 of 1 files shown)


pombredanne:
@ened Thanks again....  Do you mind adding a Signed-off-by to your commit? 
This is explained here: https://github.com/nexB/scancode-toolkit/blob/develop/CONTRIBUTING.rst#contributing

pombredanne:
What this means is that you agree to this http://developercertificate.org/
which is the same DCO as used in Linux and many open source projects.
You add a line at the end of your commit message saying (with your real name of course):

	Signed-off-by: Random J Developer <random@developer.example.org>

Issue #132, opened 2015-11-25
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2019' in position 14560183: ordinal not in range(128)


ened:
While analyzing the complete linux kernel copyright situation, the following error occurred while generating the template file:

```
$ scancode -c -l -f ../zlink/docs/scancode-toolkit/template.html . ~/Desktop/linux-copyright-holders.html
Scanning files...
  [####################################]  41256                                                       
Scanning done.
Traceback (most recent call last):
  File "/Users/ened/git/ened.scancode-toolkit/bin/scancode", line 9, in <module>
    load_entry_point('scancode-toolkit==1.3.8', 'console_scripts', 'scancode')()
  File "/Users/ened/git/ened.scancode-toolkit/lib/python2.7/site-packages/click/core.py", line 664, in __call__
    return self.main(*args, **kwargs)
  File "/Users/ened/git/ened.scancode-toolkit/src/scancode/utils.py", line 61, in main
    standalone_mode=standalone_mode, **extra)
  File "/Users/ened/git/ened.scancode-toolkit/lib/python2.7/site-packages/click/core.py", line 644, in main
    rv = self.invoke(ctx)
  File "/Users/ened/git/ened.scancode-toolkit/lib/python2.7/site-packages/click/core.py", line 837, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/ened/git/ened.scancode-toolkit/lib/python2.7/site-packages/click/core.py", line 464, in invoke
    return callback(*args, **kwargs)
  File "/Users/ened/git/ened.scancode-toolkit/src/scancode/cli.py", line 245, in scancode
    save_results(results, format, input, output_file)
  File "/Users/ened/git/ened.scancode-toolkit/src/scancode/cli.py", line 345, in save_results
    output_file.write(as_template(results, template=format))
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2019' in position 14560183: ordinal not in range(128)
```

Using latest GIT version (+plus patches for #123 and #127).

pombredanne:
Can you figure out which file triggered this? 
Is there any special encoding in your custom template? Did this run on a machine with a non-UTF-8 locale?

pombredanne:
Some background on this:
http://jinja.pocoo.org/docs/dev/api/#unicode
It could be that some strings in the license object are not tracked as unicode.

ened:
[...]

PR #133, opened 2015-11-25
Ensure to encode the HTML output in UTF-8

ened:
This is a initial fix for #132 .

Also WIP is a unit test for the error, which I can not reproduce easily.


pombredanne:
@ened did you had a chance to work on a test case?

pombredanne:
@ened I am closing for lack of activity. As said in #132 I was not able to reproduce.

Issue #126, opened 2015-11-24
License data for exceptions should have pointers the main license the exception relates to


pombredanne:
For instance, the Classpath exception to the GPL 2.0 should have a pointer to the GPL 2.0 license key.
And the LGPL-3.0 to the GPL-3.0. This can be useful to actually collect all the set of license texts that would need to be produced towards generating attribution for AboutCode or else in #123

ened:
That would be very helpful. It explains why the "LGPL 2.0 or later" item only contains the permission note and misses the full license text, even though I would have expected it there.

ened:
@pombredanne I could imagine this needs to work like this:

1. The definition of the main license is written in the `.yml` file, like `gpl-2.0-gcc.yml`, in a field called: `main_license`. That field is a key to the main license, like `gpl-2.0`. 
2. When scancode generates the list of licenses, it should be done in 2 passes. The first pass will go through the list of licenses, finds the `main_license` ones and adds them to the final list. The second pass should work as it is now.

Sounds good to you?

pombredanne:
@ened You wrote:
> That would be very helpful. It explains why the "LGPL 2.0 or later" item only contains the permission note and misses the full license text, even though I would have expected it there.

This is actually a whart that we need to fix. `LGPL 2.0` or later should contain the full text of the LGPL, and the rare case of `LGPL 2.0 only` should be one or more notices. I debated to the SPDX,org mailing list on this topic a week or two ago.

This should be fixed.

Now, for exceptions and this `or later` topic, there would be a few ways to go at it:
1. add a field to the License data that would point to the `main_license` ... I may prefer to name this `exception_to` .
2. or add an expression field to the License data that would be for instance for `gpl-2.0-gcc.yml`: 
 > effective_license: gpl-2.0 with gpl-2.0-gcc

and also update all the rules to point to such an expression for licenses exceptions.

pombredanne:
The new approach with #74 is going to be different after all. In particular there is no consistent way to tell which license an exception may apply too all the times. It may be more than one possibility or not defined.

- exception licenses will be flagged with an `is_exception` boolean tag.
- licenses that support `or_later` will be flagged with an `is_or_later` boolean tag
- licenses that are versioned will have a `next_version` field that points to the license key of the next version when available. This can be used for validation and possibly for a future license expression expansion (e.g. such that `GPL-1.0 or later` could be expanded to `GPL-1.0 or GPL-2.0 or GPL-3.0`) 
[...]

Issue #74, opened 2015-09-17
Report detected license expressions in ScanCode


pombredanne:
Multiple licenses can be and are detected together when appropriate rules exist. However, even though we report these with the same detected start and end line and we know internally that they are detected together and we know if they are license choices or conjunctive licenses, we only report discrete and distinct licenses, and do not provide all these detection details.

We should return all the information that we detect, especially the interesting license choices, possibly as SPDX-style license expressions.

This is especially relevant for multiple choices that exists in Qt and similar (see #73)

pombredanne:
The reporting as a license expression should be a rather straightforward addition to the license detection, since we have already all the data when we match a license RULE.
The logic will be the following: Once we have matched a text region (start/end lines) to a detection rule:

- if there is more than one license in the rule, create a license expression:
 - if the rule is a license_choice: yes, create this expression as joining with an OR all the rules licenses
 - else, create this expression as joining with an AND

At a later stage, we can replace the list of licenses in all rules (with a one time migration script)  by a plain license expression and then support more complex expressions such as (XXXX with YYY-exception) or ZZZZ

At a high level, the key thing to understand is that we already match to "expression-like" rules
we just do not expose that info in the scan reports.

pombredanne:
The latest https://github.com/nexB/license-expression/releases/tag/v0.6 now fully supports expressions "with exceptions" and expressions using license names containing "or later". This is now good enough to be used and support the implementation of expression in ScanCode

pombredanne:
The new approach is this:

- exception licenses will be flagged with an `is_exception` boolean tag.  Note that there  there is no consistent way to tell which license an exception may apply too all the times. It may be more than one possibility or not defined, hence why we only will use a flag.

- licenses that are  for this "or later" versions  will be flagged with an `is_or_later` boolean tag
- for an `is_or_later` license, the `base_license` field will point to the base license for this license. For instance `gpl-2.0-plus` will have `base_license: gpl-2.0`
- licenses that are versioned will have a `next_version` field that points to the license key of the next version when available. For instance `gpl-2.0-plus` will have `next_version: gpl-3.0-plus` and  `gpl-2.0` will have `next_version: gpl-3.0` but `gpl-3.0` will not have a `next_version` because there is no such thing yet.  `mpl-1.1` will have `next_version: mpl-2.0`

[...]

PR #137, opened 2015-11-27
[WIP] Added main_license field to license model

ened:
TODO:
- unit tests
- double check other licenses for missing `main_license` fields

### DIFF
--- a/src/licensedcode/data/licenses/lgpl-2.0-plus-gcc.yml
+++ b/src/licensedcode/data/licenses/lgpl-2.0-plus-gcc.yml
@@ -6,3 +6,4 @@ category: Copyleft Limited
 owner: Free Software Foundation (FSF)
 text_urls:
     - http://www.opensource.apple.com/source/libcppabi/libcppabi-26/include/demangle.h
+main_license: lgpl-2.0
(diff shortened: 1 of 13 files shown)


ened:
Function-wise, this is ready for review @pombredanne. Next stop is are the tests.

pombredanne:
@ened I am still pondering this in relation to #126 
We have cases and these are common where an exception can apply to multiple licenses.
For instance, there is a common exception to the GPL for openssl . See https://github.com/nexB/scancode-toolkit/blob/86283e1f0e114397dd030f4bf57067c2c47a8cdb/src/licensedcode/data/licenses/gpl-2.0-openssl.LICENSE
It is wrongly keyed after the GPL 2.0 but it applies to any GPL version.
There is a new batch of exceptions that is arriving from SPDX too, soon.
I want to make sure we come with the correct approach before adding a new attribute.

pombredanne:
So as we are adding support for license expressions, the "main" or primary license will be the first member of an expression. This should alleviate the need for a specific attribute and will be exposed as a computed property instead

pombredanne:
FYI, I pushed https://github.com/nexB/license-expression/ which is a library to parse and do some boolean analysis, comparison, normalization of license expressions. This is what I will use in ScanCode

pombredanne:
@ened if you do not mind I would like to close this since the actual implementation that will land for this #126  will be closely related to #74 and  SPDX-like license expressions. I hope you do not feel bad about this: I will credit you in the related commits in all cases!  If you want to continue the discussion we can do it on #126

Issue #158, opened 2015-12-29
unable to extract large gzip files


rakeshbalusa:
OS : windows 7 64-bit

I'm trying to extract a gzip file (`download URL` - `http://opensuse-mirror-gce-us.opensu.se/distribution/13.2/repo/oss/suse/repodata/070c71c5f39d4c84872070069631f4572f8338a843df2ae8323b6262c1f91284-other.xml.gz`) which is of `size 99.2 MB` and when I try the command `C:\*****\070c71c5f39d4c84872070069631f4572f8338a843df2ae8323b6262c1f91284-other.xml.gz` I get the following error message : 
```
ERROR extracting: C:/Users/nexB/Documents/070c71c5f39d4c84872070069631f4572f8338a843df2ae8323b6262c1f91284-other.xml.gz: ''
Extracting done.
```
The above `error message is not clear` about what the problem is and the possible solution for the above problem is to read the file in blocks of around 1 MB each. When I used the following code snippet in `uncompress` module in the the extraction was successful.
```
 with open(target_location, 'wb') as uncompressed:
            while True:
                chunk = compressed.read(1024*1024)
                if not chunk:
                    break
                uncompressed.write(chunk)
```

pombredanne:
can you open a separate issue for the error message logging?

pombredanne:
There is a patch in #159

rakeshbalusa:
yes, I will create  a new issue in a few minutes.

pombredanne:
Fixed with #159

PR #159, opened 2015-12-30
Reading files in blocks of 1 MB size

rakeshbalusa:
The following code snippet is added to `uncompress` module in the `extractcode` to perform extraction for large archives. This is a fix for #158

```
 with open(target_location, 'wb') as uncompressed:
            while True:
                chunk = compressed.read(1024*1024)
                if not chunk:
                    break
                uncompressed.write(chunk)
```

### DIFF
--- a/src/extractcode/uncompress.py
+++ b/src/extractcode/uncompress.py
@@ -63,7 +63,7 @@ def uncompress(location, target_dir, decompressor, suffix=EXTRACT_SUFFIX):
     with decompressor(location, 'rb') as compressed:
         with open(target_location, 'wb') as uncompressed:
             while True:
-                chunk = compressed.read(1024 * 1024)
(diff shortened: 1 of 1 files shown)


pombredanne:
@rakeshbalusa Thanks...  Could you make this 8MB rather than 1MB? the key point is to avoid reading large archives that could exhaust memory... but 8MB should be fine 

And I wished we could have a test for this, but I cannot fathom how we could reproduce a memory constrained test machine.

pombredanne:
I think a minor optimization would be great: Create a constant for `EIGHT_MB = 8 * 1024 * 1024` outside of the loop... This would avoid un-needed re-computation of this for each chunk, even if this is minor and likely inlined by the Python compiler.

pombredanne:
LGTM ...merging. Thank you and happy new year!

PR #164, opened 2016-01-12
new rule to detect MIT license in URL

rakeshbalusa:
Added a new rule to detect license in the URL `https://github.com/scottjehl/Respond/blob/master/LICENSE-MIT`


src/licensedcode/data/rules/mit_url_1.yml
```diff
@@ -0,0 +1,3 @@
+licenses:
+    - mit
+template: no
```
pombredanne: template: no is not needed if this is not a template

tests/licensedcode/data/licenses/mit_url_1.yml
```diff
@@ -0,0 +1,3 @@
+licenses:
+    - mit
+template: no
```
pombredanne: `template: no` is not needed if this is not a template

rakeshbalusa: ok...removing that line :+1:

pombredanne:
@rakeshbalusa do you know if this was for a specific issue? If so, please add a comment with the issue # sowe have a link.
Thanks!

rakeshbalusa:
ok, I will . I forget to put issue number sometimes.

tests/licensedcode/data/licenses/mit_url_1.yml
pombredanne: The line is still there...

PR #165, opened 2016-01-13
new license rule to detect dual licenses BSD and MIT

rakeshbalusa:
Added a new license rule to detect the dual licenses bsd-new and MIT License.


src/licensedcode/data/rules/bsd-new_and_mit_2.yml
```diff
@@ -0,0 +1,4 @@
+licenses:
+    - mit
+    - bsd-new
+
```
pombredanne: Is this an empty trailing line? It is best to avoid that in the yml files

rakeshbalusa: all the yml files have an empty line in both the `rules` and `data` (tests) folders. I thought that's a convention being followed.

Issue #173, opened 2016-01-19
GPL license not detected


rakeshbalusa:
```
------------------------------------------------------------------------------
--                                                                          --
--                         GNAT COMPILER COMPONENTS                         --
--                                                                          --
--                       S Y S T E M . V A L _ L L U                        --
--                                                                          --
--                                 S p e c                                  --
--                                                                          --
--          Copyright (C) 1992-2009, Free Software Foundation, Inc.         --
--                                                                          --
-- GNAT is free software;  you can  redistribute it  and/or modify it under --
-- terms of the  GNU General Public License as published  by the Free Soft- --
-- ware  Foundation;  either version 3,  or (at your option) any later ver- --
-- sion.  GNAT is distributed in the hope that it will be useful, but WITH- --
-- OUT ANY WARRANTY;  without even the  implied warranty of MERCHANTABILITY --
-- or FITNESS FOR A PARTICULAR PURPOSE.                                     --
--                                                                          --
-- As a special exception under Section 7 of GPL version 3, you are granted --
-- additional permissions described in the GCC Runtime Library Exception,   --
-- version 3.1, as published by the Free Software Foundation.               --
--                                                                          --
-- You should have received a copy of the GNU General Public License and    --
-- a copy of the GCC Runtime Library Exception along with this program;     --
-- see the files COPYING3 and COPYING.RUNTIME respectively.  If not, see    --
-- <http://www.gnu.org/licenses/>.                                          --
--                                                                          --
-- GNAT was originally developed  by the GNAT team at  New York University. --
-- Extensive contributions were provided by Ada Core Technologies Inc.      --
--                                                                          --
------------------------------------------------------------------------------
```
For the above text no license is detected by scancode even the text refers to `GPL version 3 or later`
RPM file : `libgnat-devel-4.4.7-4.el6.i686.rpm`
download URL - `http://vault.centos.org/6.4/cr/x86_64/Packages/libgnat-devel-4.4.7-4.el6.i686.rpm`
issue file path : `libgnat-devel-4.4.7-4.el6.i686.rpm\usr\lib\gcc\i686-redhat-linux\4.4.4\adainclude\a-assert.adb`

rakeshbalusa:
@pombredanne:  For the above issue, can I use the following rule with a template or I should I remove the template ?(I tested it and worked fine) :
```
"-- {{5 GNAT}} is free software;  you can  redistribute it  and/or modify it under --
[...]

PR #183, opened 2016-01-26
#173 new rules added to detect GPL 3.0 or later version

rakeshbalusa:
New rules to detect `GPL 3.0 or later` version of licenses are added

### DIFF
--- a/src/licensedcode/data/rules/gpl-3.0_19.yml
+++ b/src/licensedcode/data/rules/gpl-3.0_19.yml
@@ -1,0 +1,3 @@
+licenses:
+    - gpl-3.0
+template: yes
--- a/src/licensedcode/data/rules/gpl-3.0_19.yml
+++ b/src/licensedcode/data/rules/gpl-3.0_19.yml
@@ -1,3 +1,4 @@
 licenses:
     - gpl-3.0
+    - gpl-3.0-gcc
 template: yes
--- a/src/licensedcode/data/rules/gpl-3.0_20.yml
+++ b/src/licensedcode/data/rules/gpl-3.0_20.yml
@@ -1,0 +1,2 @@
+licenses:
+    - gpl-3.0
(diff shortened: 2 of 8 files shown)


src/licensedcode/data/rules/gpl-3.0_19.yml
```diff
@@ -0,0 +1,3 @@
+licenses:
+    - gpl-3.0
```
pombredanne: This is not correct: there should be two licenses: gpl-3.0 plus the other one which is an exception of sorts

src/licensedcode/data/rules/gpl-3.0_20.yml
```diff
@@ -0,0 +1,2 @@
+licenses:
+    - gpl-3.0
```
pombredanne: This is not correct: there should be two licenses: gpl-3.0 plus the other one which is an exception of sorts
This should be a template rul too.

tests/licensedcode/data/licenses/gpl-3.0_5.yml
```diff
@@ -0,0 +1,3 @@
+licenses:
+    - gpl-3.0
+template: yes
```
pombredanne: Test are not templates. Only rules are.
This is not correct: there should be two licenses: gpl-3.0 plus the other one which is an exception of sorts

tests/licensedcode/data/licenses/gpl-3.0_6.yml
```diff
@@ -0,0 +1,3 @@
+licenses:
+    - gpl-3.0
+template: yes
```
pombredanne: Test are not templates. Only rules are.
Licenses are not correct: there should be two licenses: gpl-3.0 plus the other one which is an exception of sorts

pombredanne:
LGTM. Merging...

pombredanne:
Thanks!

Issue #180, opened 2016-01-19
failed to detect GPL license 


rakeshbalusa:
```
/* Copyright (C) 1989, 2000 Aladdin Enterprises.  All rights reserved. */

/*$Id: ansi2knr.c,v 1.3 2000/04/13 03:41:48 lpd Exp $*/
/* Convert ANSI C function definitions to K&R ("traditional C") syntax */

/*
ansi2knr is distributed in the hope that it will be useful, but WITHOUT ANY
WARRANTY.  No author or distributor accepts responsibility to anyone for the
consequences of using it or for whether it serves any particular purpose or
works at all, unless he says so in writing.  Refer to the GNU General Public
License (the "GPL") for full details.

Everyone is granted permission to copy, modify and redistribute ansi2knr,
but only under the conditions described in the GPL.  A copy of this license
is supposed to have been given to you along with ansi2knr so you can know
your rights and responsibilities.  It should be in a file named COPYLEFT,
or, if there is no file named COPYLEFT, a file named COPYING.  Among other
things, the copyright notice and this notice must be preserved on all
copies.

We explicitly state here what we believe is already implied by the GPL: if
the ansi2knr program is distributed as a separate set of sources and a
separate executable file which are aggregated on a storage medium together
with another program, this in itself does not bring the other program under
the GPL, nor does the mere fact that such a program or the procedures for
constructing it invoke the ansi2knr executable bring any other part of the
program under the GPL.
*/

/*
```
GPL license not detected from the above text in the file `ansi2knr.c`
RPM : `automake-1.11.1-4.el6.noarch.rpm`
download URL : `http://mirror.centos.org/centos/6/os/x86_64/Packages/automake-1.11.1-4.el6.noarch.rpm`
issue file path : `automake-1.11.1-4.el6.noarch.rpm\usr\share\automake-1.11\ansi2knr.c`

rakeshbalusa:
new rule is working fine :+1:

PR #187, opened 2016-01-27
#180 gpl not detected in ansiknr from automake

rakeshbalusa:
New rule to detect the license GPL 3.0 or later version


src/licensedcode/data/rules/gpl-3.0-plus_22.yml
```diff
@@ -0,0 +1,2 @@
+licenses:
+    - gpl-3.0-plus
```
pombredanne: Why would this be GPL 3.0?
and further more GPL 3.0 or later?
I did not see any version mentioned in the rule.

rakeshbalusa: @pombredanne  I felt the same. The second and third paragraph seems to be unique and it is mentioned in the `COPYING` file that the license is `GPL 3.0 or later`.

pombredanne: @rakeshbalusa Things seem a bit more complicated:
from #180 the source RPM is there:
http://vault.centos.org/6.7/os/Source/SPackages/automake-1.11.1-4.el6.src.rpm
 The top level COPYING is GPL 2.0
The COPYING found side by side with ansiknr is a GPL 3.0 alright but this is really for automake.
ansiknr predates the GPL 3.0 by a long while. It would rarely be GPL-3.0 licensed if ever. 
So even though if may sound confusing to you, it is best to report this as GPL 2.0, or actually since strictly no version is stated in the file it could be technically any version.

So for now, use GPL 2.0 and add a note to the rule stating that no version is specified and that the original archive has both a GPL 2.0 and 3.0 texts. ... basically provide the explanations told here.

src/licensedcode/data/rules/gpl-3.0-plus_22.yml
```diff
@@ -0,0 +1,3 @@
+licenses:
+    - gpl-2.0
+notes: no version is specified and that the original archive has both a GPL 2.0 and 3.0 texts
```
pombredanne: Please add a bit more to the note including a link to the original sprm and the fact that ansiknr is originally GPL-2.0-licensed.

rakeshbalusa: @pombredanne  I'm not sure about the format for the `notes` parameter in a `.yml` file. can I write like this : 
```
licenses:
    - gpl-2.0
notes: first line
second line
third line
.
.
```

pombredanne: This is plain block style YAML.
https://en.wikipedia.org/wiki/YAML#Block_literals

rakeshbalusa: @pombredanne  please let me know if these notes and format are ok?
```
licenses:
    - gpl-2.0
notes: |
    no version is specified and that the original archive has both a GPL 2.0 and 3.0 texts
    srpm download URL 'http://vault.centos.org/6.7/os/Source/SPackages/automake-1.11.1-4.el6.src.rpm'
    The top level COPYING is GPL 2.0. 
    The COPYING found side by side with ansiknr is a GPL 3.0, but this is really for automake.
```

pombredanne: May be this instead?
```
licenses:
    - gpl-2.0
notes: |
    No version is specified in this notice found in ansiknr.c as redistributed
    with automake in this RPM
    
    http://vault.centos.org/6.7/os/Source/SPackages/automake-1.11.1-4.el6.src.rpm 
    
    The original tarball for automake found in this RPM contains a GPL 2.0 text
    at the top-level and a GPL 3.0 text at the same level as ansiknr.c. Since
    ansiknr.c copyright pre-dates the GPL 3.0 and the notice has no version
    stated, this rule is defined to match a GPL 2.0 rather than a GPL 3.0. GPL
    3.0 is the license for automake proper and unlikely the license for
    ansiknr.c.
```

rakeshbalusa: thanks...I'll update :+1:

src/licensedcode/data/rules/gpl-3.0-plus_22.yml
```diff
@@ -0,0 +1,14 @@
...
+    with automake in this RPM
+
+    http://vault.centos.org/6.7/os/Source/SPackages/automake-1.11.1-4.el6.src.rpm 
+
+    The original tarball for automake found in this RPM contains a GPL 2.0 text
+    at the top-level and a GPL 3.0 text at the same level as ansiknr.c. Since
+    ansiknr.c copyright pre-dates the GPL 3.0 and the notice has no version
+    stated, this rule is defined to match a GPL 2.0 rather than a GPL 3.0. GPL
+    3.0 is the license for automake proper and unlikely the license for
+    ansiknr.c.
```
pombredanne: Could you add a line return at the end of this file?

rakeshbalusa: done :+1:

pombredanne:
LGTM . Merging. Thanks!

Issue #178, opened 2016-01-19
GPL license not detected from a short GPLv2+ keyword


rakeshbalusa:
```
About: License
  This file is licensed under the GPLv2+, like the rest of Zypper.
```
GPL version 2.0 or later license not detected in the file `zypper.aug` which has the reference to license in the above text
RPM : `zypper-1.11.14-2.1.x86_64.rpm`
download URL : `https://mirror.umd.edu/opensuse/distribution/13.2/repo/oss/suse/x86_64/zypper-1.11.14-2.1.x86_64.rpm`
issue file path : `zypper-1.11.14-2.1.x86_64.rpm\usr\share\zypper\zypper.aug`

rakeshbalusa:
@pombredanne  for the above license text can I use RULE file with text `GPLv2+` or should I use `licensed under the GPLv2+` ?

pombredanne:
Always use the longest and most specific possible. Here this would be 
`This file is licensed under the GPLv2+`

rakeshbalusa:
But if I have a new text like `this program is licensed under GPLv2+`...will scancode detect the license?

pombredanne:
@rakeshbalusa have you ever seen such a statement anywhere? Do you have an example?
Assuming Google indexes a good slice of the internet see https://www.google.com/search?q="this+program+is+licensed+under+GPLv2%2B"
then there are only two mentions of this anywhere. Not worth having a specific rule for this case. It would be caught by approximate detection anyway. So unless you bring some proof that there is a reasonable number of variations for this license rule (like a sizeable number of the same sentence with program, file, code, software etc.), make it a simple text as specific as possible.

rakeshbalusa:
ok :+1:

pombredanne:
And FWIW with approx detection, this rule would match `this program is licensed under GPLv2+` with a score of 50% https://github.com/nexB/scancode-toolkit/blob/0e09d45e67ed5b8d873263f871df94ac24003719/src/licensedcode/data/rules/gpl-2.0_23.RULE

pombredanne:
And with your planned new rule here this would be a score of 83% which is plenty good enough for this very rare sentence.

rakeshbalusa:
New rule fixed the issue :+1:

PR #189, opened 2016-01-27
#178 new rule to detect license from keyword GPLv2+

rakeshbalusa:
Adding a new rule to detect license from the keyword `GPLv2+`


pombredanne:
Thanks!

Issue #177, opened 2016-01-19
LGPL 2.0 or later not detected


rakeshbalusa:
```
# This program is free software; you can redistribute it and/or modify it
# under the terms of the GNU Library General Public License as published
# by the Free Software Foundation; either version 2, or (at your option)
# any later version.
#
# This program is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
# Library General Public License for more details.
```
General Public License is referenced in the above text from the file `locale.alias` but is not detected by scancode
RPM : `glibc-common-2.12-1.149.el6_6.7.x86_64.rpm`
download URL : `http://ftp.scientificlinux.org/linux/scientific/6.6/x86_64/updates/security/glibc-common-2.12-1.149.el6_6.7.x86_64.rpm`
issue file path : `glibc-common-2.12-1.149.el6_6.7.x86_64.rpm\usr\share\locale\locale.alias`

rakeshbalusa:
New rule fixed the issue :+1:

PR #190, opened 2016-01-27
#177 adding new rule to detect LGPLv2+ license

rakeshbalusa:
New rule to detect `LPGL 2.0 or later` or later is added


src/licensedcode/data/rules/gpl-2.0-plus_53.yml
```diff
@@ -0,0 +1,2 @@
+licenses:
+    - gpl-2.0-plus
```
pombredanne: `GNU Library General Public License` is the LGPL 2.0 note the GPL 2.0

tests/licensedcode/data/licenses/gpl-2.0-plus_47.yml
```diff
@@ -0,0 +1,2 @@
+licenses:
+    - gpl-2.0-plus
```
pombredanne: `GNU Library General Public License` is the LGPL 2.0 not the GPL 2.0

rakeshbalusa: my bad..I'll update everything

pombredanne:
LGTM. Merging.

rakeshbalusa:
@pombredanne  I think we need to change this rule to a template. please check the issue #197 . Do I have to create a new ticket and pull request or is there any other way you want me to use?

Issue #176, opened 2016-01-19
Erlang Public License version 1.1 not detected


rakeshbalusa:
```
%% The contents of this file are subject to the Erlang Public License,
%% Version 1.1, (the "License"); you may not use this file except in
%% compliance with the License. You should have received a copy of the
%% Erlang Public License along with this software. If not, it can be
%% retrieved online at http://www.erlang.org/.
%% 
%% Software distributed under the License is distributed on an "AS IS"
%% basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See
%% the License for the specific language governing rights and limitations
%% under the License.
```
`Erlang Public License 1.1` not detected from the above text in the file `appmon.app`
RPM : `erlang-appmon-R16B03-0.2.el6.x86_64.rpm`
download URL : `http://packages.erlang-solutions.com/rpm/centos/6/x86_64/erlang-appmon-R16B03-0.2.el6.x86_64.rpm`
issue file path : `erlang-appmon-R16B03-0.2.el6.x86_64.rpm\usr\lib64\erlang\lib\appmon-2.1.14.2\ebin\appmon.app`

rakeshbalusa:
@pombredanne  please let me know if I have to use the second paragraph in the `license RULE` : 
```
 Software distributed under the License is distributed on an "AS IS"
 basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See
 the License for the specific language governing rights and limitations
 under the License.
```

pombredanne:
Small note: EPL is the id for the Eclipse Public License, not the Erlang license.
And yes you should use the second paragraph. Why not using it?

rakeshbalusa:
@pombredanne  thank you for updating the ticket. A quick google search on the second paragraph says it is not `completely unique`. please look at the bottom of this page - `http://www.apache.org/licenses/LICENSE-2.0` , except for a single word `warranties` the entire second paragraph is in `apache-2.0` license also. I'm not sure but this might cause a problem when `approximate license detection` is used.

rakeshbalusa:
issue fixed check #193

PR #193, opened 2016-01-28
#176 adding new rule to detect erlang license

rakeshbalusa:
New rule to detect `Erlang Public License version 1.1` is added with test cases


pombredanne:
LGTM ... Thanks!

Issue #175, opened 2016-01-19
failed to detect GPL license


rakeshbalusa:
```
StartFontMetrics 3.0
Comment Copyright URW Software, Copyright 1997 by URW
Comment Creation Date: 10/19/1999
Comment See the file COPYING (GNU General Public License) for license conditions.
FontName Dingbats
FullName Dingbats 
FamilyName Dingbats
Weight Regular
ItalicAngle 0.0
IsFixedPitch false
UnderlinePosition -72
UnderlineThickness 36
Version 001.005
Notice URW Software, Copyright 1997 by URW
EncodingScheme FontSpecific
FontBBox -1 -143 981 819
CapHeight 691
XHeight 567
Descender -143
Ascender 691
StartCharMetrics 203
```
GPL license not detected for the above text in the file `d050000l.afm`
RPM : `urw-fonts-2.4-10.el6.noarch.rpm`
download URL : `http://mirror.internode.on.net/pub/smeserver/9/smeextras/x86_64/RPMS/urw-fonts-2.4-10.el6.noarch.rpm`
issue file path : `urw-fonts-2.4-10.el6.noarch.rpm\usr\share\fonts\default\Type1d050000l.afm`

rakeshbalusa:
@pombredanne  for the above issue scancode needs to detect `GPL 2.0` as it has no reference to any version.  am I correct?
the rule I'm planning to use is `GNU General Public License) for license conditions.` please let me know if it's ok.

pombredanne:
IMHO, the rule should be `See the file COPYING (GNU General Public License) for license conditions.`
More specific is better. 

There is a also a plain `gpl` key that we could use technically when no version is stated though.
The GPL says we can pick any version in this case, so normally it would be a choice of 1.0, 2.0 or 3.0.
But in general when no version is specified erring towards 2.0 is acceptable too

mjherzog:
For detection we should show what is found - automation for concluding that GPL 2.0 is the logical choice (as in an audit) is a different step.

pombredanne:
@mjherzog Thanks.
@rakeshbalusa Use the plain `gpl` key for now then. https://github.com/nexB/scancode-toolkit/blob/c606f59785e43812be4d3e8cfabd62089c8b3157/src/licensedcode/data/licenses/gpl.yml
@mjherzog at a later stage this will be replaced by a choice expression of gpl 1, 2 or 3.

pombredanne:
[...]

PR #195, opened 2016-01-28
#175 new rule to detect gpl-2.0 license

rakeshbalusa:
New rule to detect `General Public License version 2.0` is added.


pombredanne:
LGTM. Thanks!

Issue #194, opened 2016-01-28
GPL license is not detected


rakeshbalusa:
OS : windows 7 - 64 bit
I've updated my scancode to the latest version available.

RPM file download URL - `http://bay.uchicago.edu/centos-vault/6.4/cr/x86_64/Packages/libreoffice-sdk-doc-4.0.4.2-9.el6.x86_64.rpm`

issue file path : `libreoffice-sdk-doc-4.0.4.2-9.el6.x86_64.rpm\usr\lib64\libreoffice\sdk\docs\cpp\ref\search\search.js`

In the line 6 in this text the GPLis not detected:
```
   // The code in this file is loosly based on main.js, part of Natural Docs,
// which is Copyright (C) 2003-2008 Greg Valure
// Natural Docs is licensed under the GPL.
```

rakeshbalusa:
@pombredanne  I'm trying to add a new  RULE  for the above issue `is licensed under the GPL` but it is affecting other test cases which has the same text and they are failing. what am I supposed to do?

balusarakesh:
#199 issue fixed.

PR #199, opened 2016-01-29
#194 adding a new rule to detect GPL license with no version

rakeshbalusa:
New rule to detect the license `GPL` with no version is added. Added this license to a few tests which has the same text.

### DIFF
--- a/src/licensedcode/data/rules/gpl_48.RULE
+++ b/src/licensedcode/data/rules/gpl_48.RULE
@@ -1,0 +1,1 @@
+is licensed under the GPL
--- a/src/licensedcode/data/rules/gpl_48.yml
+++ b/src/licensedcode/data/rules/gpl_48.yml
@@ -1,0 +1,2 @@
+licenses:
+    - gpl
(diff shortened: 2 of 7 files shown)


pombredanne:
Thank you!
I would need to get more details on these other tests that had the same text to make sure we are not introducing biases. Same text may not mean this is always a valid mention.
Could you paste links to the each corresponding test file snippet lines?

rakeshbalusa:
These are the `test files` with the same text which : 
- `https://github.com/nexB/scancode-toolkit/blob/develop/tests/licensedcode/data/licenses/adobe-air-sdk.txt#L18`
- `https://github.com/nexB/scancode-toolkit/blob/develop/tests/licensedcode/data/licenses/gpl-2.0-plus_and_gpl-2.0-plus_and_lgpl-2.1-plus_and_mpl-1.1_and_other.txt#L80`
- `https://github.com/nexB/scancode-toolkit/blob/develop/tests/licensedcode/data/licenses/bsd-new_53.txt#L38`

pombredanne:
Thanks: so this is about Debian packaging. Let me pounder this a bit.

pombredanne:
It would have been best to have individual commit for each test or change in the future.

Issue #101, opened 2015-10-19
licenses GPLv3+ and LGPLv3+ are not detected


rakeshbalusa:
OS: windows 7 64 bit
I've updated my scancode to the latest version available.
file : [msgbox.txt](https://github.com/nexB/scancode-toolkit/files/16173/msgbox.txt)
```
scancode output :
Path                          	            What	Start Line	End Line		Info
C:/doc/temporary/issue_files/msgbox.py     	License	         5	6		    MPL 1.1
```
In `line 3` of the above file `GPLv3+` and `LGPLv3+` are mentioned but there were no results from scancode.
`Mozilla Public License` in line 8 is ignored which is in the for a URL - `http://www.mozilla.org/MPL/` 

License text from the file : 
```
# Version: MPL 1.1 / GPLv3+ / LGPLv3+
#
# The contents of this file are subject to the Mozilla Public License Version
# 1.1 (the "License"); you may not use this file except in compliance with
# the License or as specified alternatively below. You may obtain a copy of
# the License at http://www.mozilla.org/MPL/
#
# Software distributed under the License is distributed on an "AS IS" basis,
# WITHOUT WARRANTY OF ANY KIND, either express or implied. See the License
# for the specific language governing rights and limitations under the
# License.
#
# The Initial Developer of the Original Code is
#       Laurent Godard <listes.godard@laposte.net>
# Portions created by the Initial Developer are Copyright (C) 2010 the
# Initial Developer. All Rights Reserved.
#
# Major Contributor(s):
#       Timo <timo@iera.de>
#
# For minor contributions see the git repository.
#
# Alternatively, the contents of this file may be used under the terms of
# either the GNU General Public License Version 3 or later (the "GPLv3+"), or
# the GNU Lesser General Public License Version 3 or later (the "LGPLv3+"),
# in which case the provisions of the GPLv3+ or the LGPLv3+ are applicable
# instead of those above.
```

rakeshbalusa:
@pombredanne  
For the above issue there is some rule in `scancode` which is detecting `The contents of this file are subject to the Mozilla Public License Version 1.1` as a license and ignoring the next two lines which is also license text. If I create a new rule with the entire text for `MPL 1.1` given above then two rules will be detecting `MPL 1.1` which I'm not sure is a good or bad practice. please let me know how to proceed.

rakeshbalusa:
New rule fixed the issue :+1:

pombredanne:
#200 was applied. Thanks

PR #200, opened 2016-02-02
#101 new rule to detect GPL and LGPL licenses

rakeshbalusa:
New rule to detect `gpl-3.0-plus` or `lgpl-3.0-plus` licenses is added


tests/licensedcode/data/licenses/gpl_and_lgpl_6.txt
```diff
@@ -0,0 +1,5 @@
+# Alternatively, the contents of this file may be used under the terms of
+# either the GNU General Public License Version 3 or later (the "GPLv3+"), or
+# the GNU Lesser General Public License Version 3 or later (the "LGPLv3+"),
+# in which case the provisions of the GPLv3+ or the LGPLv3+ are applicable
+# instead of those above.
```
pombredanne: Can you tell where you found this? because this is an alternative to something else (start with `Alternatively`)
So we would want to capture the larger alternative and not only the single GPL or GPL choice both in rules and tests

pombredanne:
I added more tests and fixed the rule. Merged. Thanks!

PR #201, opened 2016-02-02
#175 changing license from GPL-2.0 to GPL

rakeshbalusa:
New rule to detect `GPL` license (no-version) is added.


pombredanne:
Thanks... merging this

Issue #224, opened 2016-03-18
Recognize .deb file, 


linexb:
We should recognize of a .deb proper. Thist is likely handled nicely by the file_type, but let's double check and add test case if we have.

More info: https://en.wikipedia.org/wiki/Deb_%28file_format%29

pombredanne:
@linexb This was properly handled in the first place. And you added tests with #225 
All in. Thanks

PR #225, opened 2016-03-21
#224 deb file detection

linexb:
Add test case to test get file type function for debian file


tests/commoncode/test_filetype.py
```diff
@@ -76,6 +76,11 @@ def test_get_type(self):
 
         assert sorted(expected) == sorted(results)
 
+    def test_get_debiantype(self):
+        test_file = self.get_test_loc('filetype/debian/libjama-dev_1.2.4-2_all.deb')
+        type = filetype.get_type(test_file)
+        assert 'f' == type 
+
```
pombredanne: @linexb I think this test is not needed. It just test that a .deb package is not a special file so this is not super interesting. I think you can remove it and the corresponding test file

tests/typecode/test_contenttype.py
```diff
@@ -147,6 +147,9 @@ def test_archive_gnu_tar(self):
         assert is_archive(self.get_test_loc('contenttype/archive/e.tar'))
         assert 'posix tar archive (gnu)' == get_filetype(self.get_test_loc('contenttype/archive/e.tar'))
 
+    def test_debianfile(self):
+        assert 'debian binary package (format 2.0)' == get_filetype(self.get_test_loc('contenttype/debian/libjama-dev_1.2.4-2_all.deb'))
+
```
pombredanne: You should also test that this is recognized as an archive

pombredanne:
@linexb I merged  2c92270 with some changes. Thanks!

Issue #226, opened 2016-03-21
Need a variable to store SHA512 of Nuget packages


balusarakesh:
https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/models.py
In the Package class we don't have a variable to store `SHA512` for a given package.
Nuget API provides the SHA512 for its packages, for example - `https://api.nuget.org/v3/catalog0/data/2015.02.12.00.46.55/microsoft.aspnet.razor.3.2.3.json`

pombredanne:
We also need to add support for explicit sha512 and sha256 hashing in commoncode

pombredanne:
#229 Merged.

PR #227, opened 2016-03-21
#226 Added variable to store SHA512 of a package

balusarakesh:
Added `download_sha512` variable


pombredanne:
Thanks. This make sense.

Issue #171, opened 2016-01-15
support for scanning RPMs


rakeshbalusa:
Currently `scancode-toolkit` can scan packages or any kind of code without targeting metadata like type of package (`for example : RPM`). For any given RPM the `.spec` has most of the `metadata` like `architecture, source RPM, URL etc.,` and it is the most reliable `metadata` for that RPM. So, we need to have a separate scanning scenarios for RPMs. Everytime `scancode` enounters an RPM it should provide `scan results` using these scenarios.

pombredanne:
@rakeshbalusa actually the .spec file is rather hard to parse short of running the tools to build an RPM proper. The more reliable source of metadata is the RPM headers in an .rpm file. The .spec file itself is rarely available anyway when you deal with built RPMs.

rakeshbalusa:
@pombredanne  I guess we have to use those headers info instead of `.spec` file in case of an RPM.

balusarakesh:
@pombredanne 
I think the line - https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/recognize.py#L57 should be `        if location.endswith(tuple(package.extensions)):`. I mean we should look for extensions not metafiles, but this will be a problem for NPM packages as they have `package.json` as metafile which we must look for. I'm not sure how to solve this issue.

pombredanne:
We are scanning RPMs alright and there is no plan for now to scan .spec files yet, though I am renaming this issue to do that

xsuchy:
FYI if you want to implement it you can use https://github.com/packit/specfile to extract the data from SPEC file.

PR #228, opened 2016-03-22
#171 add support for RPM packages

balusarakesh:
Adding support to RPM pacakges in `scancode`


src/packagedcode/rpm.py
```diff
@@ -26,9 +26,9 @@
 
 from collections import namedtuple
 
-from packagedcode.models import AssertedLicense
+from packagedcode.models import AssertedLicense, Package
```
pombredanne: Please use one package import per line and sort your imports

src/packagedcode/rpm.py
```diff
@@ -38,27 +38,33 @@
...
+    'patch',
+    'url',
+    'source_rpm',
+    'source_package',
+    'dist_url',
+    'bin_or_src',
+)
+
+
+class EnhancedRpmPackage(RpmPackage):
```
pombredanne: Why this subclass?
You want rather to update the RpmPackage class, eventually moving it there but not create an EnhancedRpmPackage subclass

balusarakesh: mybad, it is not supposed to be there. 
deleting it.

pombredanne:
Thanks. Let me review this
For the commit message `Extracted epoch from RPM` did you mean rather:
`Added epoch to tags collected from RPM headers` ?

src/packagedcode/pyrpm/rpm.py
```diff
@@ -3,17 +3,17 @@
 # vim:ts=4:sw=4:et
 
 # Copyright (c) M�rio Morgado
-#
```
pombredanne: It is usually best to avoid mixing formatting changes with actual code changes. It make the patches easier to review. ;)

Issue #113, opened 2015-11-09
redhat logos license not detected


rakeshbalusa:
OS : Windows 7 (64-bit)
I've updated my `scancode` to the latest version available.
file :  [COPYING.txt](https://github.com/nexB/scancode-toolkit/files/30267/COPYING.txt)

When i try to run scancode on the above file for licenses there were no results found. I think the license text is similar to that of the license text in this URL - `http://www.sslug.dk/~chlor/rpm3html/pack/redhat-logos.html`
The license name for this license text is not clear in this URL. please check the `license` metadata (`License: Copyright © 1999 Red Hat, Inc.  All rights reserved.`) in the above URL it looks like a `copyright `

pombredanne:
Applied and verified. Thanks.

PR #230, opened 2016-03-30
#113 Added new rule to detect red-hat-logos license

balusarakesh:
New rule added to detect redhat-logos license.


pombredanne:
Thanks!

Issue #100, opened 2015-10-19
wrong gpl license detected for a license text


rakeshbalusa:
OS: windows 7 64 bit
I've updated my scancode to the latest version available.
file : [ref_array.txt](https://github.com/nexB/scancode-toolkit/files/16170/ref_array.txt)
```
scancode output : 
Path	                                         What	Start Line	End Line	Info
C:/doc/temporary/issue_files/ref_array.h	     License	12	       15		GPL
```
The above file has license text for LGPL 3.0 or later in lines 8 to 17 but GPL license is detected by scancode.

pombredanne:
All merged. Thanks

PR #231, opened 2016-03-30
#100 Added new rule to detect LGPL license

balusarakesh:
New rule to detect lgpl-3.0-plus is added.


pombredanne:
Fixed and merged. Thanks!

Issue #93, opened 2015-10-19
license text for GPL is ignored in the comments section


rakeshbalusa:
OS: windows 7 64 bit
I've updated my scancode to the latest version available.
file : [hashtable.txt](https://github.com/nexB/scancode-toolkit/files/15761/hashtable.txt)

For the file attached no license is detected from lines 6 to 24. The license text belongs to `GPL 3.0 or Later`

rakeshbalusa:
I am assuming that these lines are ignored because the license text is placed line by line using `single line comments`

rakeshbalusa:
@pombredanne  yes, the license text is placed line by line using single line comments

pombredanne:
Fixed

PR #232, opened 2016-03-30
#93 Added new rule to detect GPL license with GCC exception

balusarakesh:
Added new rule to detect gpl-3.0-gcc license.


pombredanne:
Thank you :)

Issue #91, opened 2015-10-19
Kerberos license text ignored


rakeshbalusa:
OS: windows 7 64 bit
I've updated my scancode to the latest version available.
file : [NOTICE.txt](https://github.com/nexB/scancode-toolkit/files/15753/NOTICE.txt)
For the text file attached there is `MIT license text` from lines 10 to 21 and was ignored by scancode.
```
WITHIN THAT CONSTRAINT, permission to use, copy, modify, and distribute
this software for any purpose and without fee is hereby granted,
provided that the above copyright notice appear in all copies and that
both that copyright notice and this permission notice appear in
supporting documentation, and that the name of M.I.T. not be used in
advertising or publicity pertaining to distribution of the software
without specific, written prior permission.  Furthermore if you modify
this software you must label your software as modified software and not
distribute it in such a fashion that it might be confused with the
original MIT software.  M.I.T. makes no representations about the
suitability of this software for any purpose.  It is provided "as is"
without express or implied warranty.
```

rakeshbalusa:
@pombredanne  I'm using the following text as a RULE, please let me know if that's not ok : 
```
Copyright (C) 1985-2012 by the Massachusetts Institute of Technology.

All rights reserved.

     Export of software employing encryption from the United States of
     America may require a specific license from the United States
     Government.  It is the responsibility of any person or organization
     contemplating export to obtain such a license before exporting.

WITHIN THAT CONSTRAINT, permission to use, copy, modify, and distribute
this software for any purpose and without fee is hereby granted,
provided that the above copyright notice appear in all copies and that
both that copyright notice and this permission notice appear in
supporting documentation, and that the name of M.I.T. not be used in
advertising or publicity pertaining to distribution of the software
without specific, written prior permission.  Furthermore if you modify
this software you must label your software as modified software and not
distribute it in such a fashion that it might be confused with the
original MIT software.  M.I.T. makes no representations about the
suitability of this software for any purpose.  It is provided "as is"
without express or implied warranty.
```

pombredanne:
1. do not put the copyright in the rule
2. Is this really an MIT license?
In particular if the last part is MIT, the addition of that:
```
Export of software employing encryption from the United States of
America may require a specific license from the United States
Government.  It is the responsibility of any person or organization
contemplating export to obtain such a license before exporting.
```
may make this a new license . @DennisClark what do you think?

pombredanne:
[...]

PR #233, opened 2016-03-30
#91 Added new rule to detect Kerberos license

balusarakesh:
New rule to detect kerberos license is added.


pombredanne:
Merged. Thank you

Issue #77, opened 2015-09-18
Unable to detect Microsoft .NET Library License 


rakeshbalusa:
OS: Windows 7 64-bit
Python version 2.7 (for windows 64 bit)
I am using the latest version of `scancode`.
URL for the license txt - `http://www.microsoft.com/web/webpi/eula/net_library_eula_enu.htm`

I am expecting scancode to detect the license `src\licensedcode\data\licenses\ms-net-library.LICENSE` from scancode license libraries

I copied the following text into a text file called `license.txt`
```
MICROSOFT SOFTWARE LICENSE TERMS
MICROSOFT .NET LIBRARY
These license terms are an agreement between Microsoft Corporation (or based on where you live, one of its affiliates) and you. Please read them. They apply to the software named above, which includes the media on which you received it, if any. The terms also apply to any Microsoft

·         updates,

·         supplements,

·         Internet-based services, and

·         support services

for this software, unless other terms accompany those items. If so, those terms apply.

BY USING THE SOFTWARE, YOU ACCEPT THESE TERMS. IF YOU DO NOT ACCEPT THEM, DO NOT USE THE SOFTWARE.

IF YOU COMPLY WITH THESE LICENSE TERMS, YOU HAVE THE PERPETUAL RIGHTS BELOW.

1.    INSTALLATION AND USE RIGHTS.

a.    Installation and Use. You may install and use any number of copies of the software to design, develop and test your programs.  You may modify, copy, distribute or deploy any .js files contained in the software as part of your programs.

b.    Third Party Programs. The software may include third party programs that Microsoft, not the third party, licenses to you under this agreement. Notices, if any, for the third party program are included for your information only.

2.    ADDITIONAL LICENSING REQUIREMENTS AND/OR USE RIGHTS.

a.    DISTRIBUTABLE CODE.  In addition to the .js files described above, the software is comprised of Distributable Code. “Distributable Code” is code that you are permitted to distribute in programs you develop if you comply with the terms below.

i.      Right to Use and Distribute.

·         You may copy and distribute the object code form of the software.

·         Third Party Distribution. You may permit distributors of your programs to copy and distribute the Distributable Code as part of those programs.

ii.    Distribution Requirements. For any Distributable Code you distribute, you must

·         use the Distributable Code in your programs and not as a standalone distribution;

·         require distributors and external end users to agree to terms that protect it at least as much as this agreement;

·         display your valid copyright notice on your programs; and

[...]

PR #234, opened 2016-03-30
#77 Added new rule to detect Microsoft .NET Library License

balusarakesh:
Added new rule to detect ms-net-library license.


pombredanne:
Thanks!

Issue #236, opened 2016-03-30
mit or gpl-2 license ignored


balusarakesh:
created from ticket #118 
file: 
[html5.js.txt](https://github.com/nexB/scancode-toolkit/files/196498/html5.js.txt)

the above file has license in line 1 but no license is detected by scancode

pombredanne:
Fixed and merged.

PR #238, opened 2016-03-30
#236 Added new rule to detect MIT and GPL licenses.

balusarakesh:
New rule to detect mit or gpl-2.0 is added.


pombredanne:
@balusarakesh Thanks. The order matters `MIT/GPL2` is mit or gpl-2,0 in this order, nor gpl then mit

pombredanne:
Thanks! I fixed the license ordering

Issue #237, opened 2016-03-30
postgresql license ignored


balusarakesh:
created from #118 
file: [Statement.pm.txt](https://github.com/nexB/scancode-toolkit/files/196511/Statement.pm.txt)

the above file has license in line 3(PostgreSQL License) but no license is detected by scancode

pombredanne:
Merged

PR #239, opened 2016-03-30
#237 Added new rule to detect PostgreSQL license.

balusarakesh:
New rule to detect postgresql license is added.


src/licensedcode/data/rules/postgresql_5.RULE
```diff
@@ -0,0 +1,2 @@
+This program is open source, licensed under the PostgreSQL Licence.
+For license terms, see the LICENSE file.
```
pombredanne: @rakeshbalusa I did update the rule text with the longest text that should be matched, not the shortest

Issue #241, opened 2016-03-31
Enforce the correctness of Package objects


pombredanne:
The Package object holds quite a bit of data, and it is pretty loose right now in checking what is expected.
You can construct an incorrect one pretty easily.
We need some validation of sorts, possibly by defining some schema/validation library.
There are many such libraries so the hard part is picking a sensible and simple one.

pombredanne:
Of these, http://awesome-python.com/#data-validation  this one http://schematics.readthedocs.org/en/latest/ seems the most promising at first glance.

balusarakesh:
planning to use `http://docs.python-cerberus.org/en/stable/usage.html#validation-rules` as the validator.

pombredanne:
@balusarakesh can you explain your rationale?
cerberus seems to be fine to validate arbitrary data, but does not seem to help define correct object models like schematics does. It has not been designed for that at all, whereas schematics has from what I can see

balusarakesh:
When Model class (schematics package) is inherited into Package class it doesn't let us create new variables and throws an exception like 
```
raise ModelConversionError(errors, data)
          ModelConversionError: {'release': 'Rogue field', 'epoch': 'Rogue field', 'arch': 'Rogue field'}
```

pombredanne:
So now that we have used schematics, I am not super happy with the way this works. In particular anything that would be stored in the DB in a near future would require having yet another schema being defined in yet another library. I made a few local tests storing things as raw JSON in a DB but that raise another problem of schema migrations.
I am considering using Django's ORM to define the models instead. I will try out some tests in a branch. The semantics are essentially the same as schematics or marshmallow (or the other way around) and there could be some side benefits to store an-flight scan in a temp sqlite DB. I will work out something.
Feedback welcomed

pombredanne:
Moving this comment to #275 and closing this

PR #249, opened 2016-04-11
#241 new package model validation

balusarakesh:
Enforced the correctness of package objects.


src/packagedcode/models.py
```diff
@@ -117,14 +117,6 @@ class AssertedLicense(Model):
     notice = StringType(default=None)
     url = StringType(default=None)
 
-    def as_dict(self):
```
pombredanne: What is this replaced by? the equivalent schematics function?
If this is the case, as_dict is rather clear as a name, I would not want to have the schematics specific to percolat in the codebase everywhere. Instead we should have an as_dict function that wraps schematics calls.

balusarakesh: there is an inbuilt function `to_primitive` which can serialize an object...so, I removed `as_dict`

pombredanne: ok, so do not remove `as_dict`. Make it available on the top level Package class and just have it call `to_primitive` : the point is that we do not want the external access to Packages to be tied to schematics.

pombredanne: Related we should not depend on the default constructor for schematics models to pass a dict vs named arguments: we should instead abstract that and accept **kwargs and pass the kwargs to schematics

balusarakesh: @pombredanne  in order to override the default constructor in `schematics` we need to edit the `Model` class in `schematics.models.py`. But everytime we configure `scancode` we get a fresh install of `schematics

balusarakesh: maybe we can edit the `schematics` wheel file...is that a good idea?

pombredanne: > maybe we can edit the schematics wheel file...is that a good idea?
Not really a good idea.

pombredanne: > in order to override the default constructor in schematics we need to edit the Model class in schematics.models.py. But everytime we configure scancode we get a fresh install of schematics and we lose the changes. how can we solve this problem?

do not ever change schematics locally. We can subclass and compose, not override....

src/packagedcode/models.py
```diff
@@ -297,21 +334,21 @@ def compare_version(self, package, package_version=True):
 
         For example:
 
-        >>> q=Package(version='2')
-        >>> p=Package(version='1')
+        >>> q=Package(dict(versioning=Version(dict(version='2'))))
```
pombredanne: This is the part we would not want: a direct depdency on the way schematics demands a dict as args.
We want instead to do:
`Package(versioning=Version(version='2'))`
or `Package(versioning=Version('2'))`

tests/packagedcode/test_package_models.py
```diff
@@ -38,7 +41,35 @@ class TestModels(FileBasedTesting):
...
+            vcs_tool='git',
+            asserted_licenses=[
+                AssertedLicense(dict(
+                    license='apache-2.0'
+                    )
+                )
+            ],
+            )
+        )
+        assert 'Sample' == package.name
```
pombredanne: Could you test the whole result data rather than just some fields?

thirdparty/prod/schematics.ABOUT
```diff
@@ -0,0 +1,11 @@
+about_resource: schematics-1.1.1-cp27-none-any.whl
+version: 1.1.1
+
+name: schematics
+home_url: http://github.com/schematics/schematics
+owner: James Dennis
+dje_license: bsd-new
+notice_file: schematics.NOTICE
```
pombredanne: We need also a `license_text_file: schematics.LICENSE` and the  file and a `copyright:`

setup.py
```diff
@@ -93,6 +93,7 @@ def read(*names, **kwargs):
 
         # packagedcode
         'requests >= 2.7.0, < 3.0.0',
+        'schematics',
```
pombredanne: You need to put a version range constraint, we do not accept any schematics version but the current up to the next major excluded

src/packagedcode/models.py
```diff
@@ -94,14 +103,80 @@
...
+    party_org = 'organization'
+
+    PARTY_TYPES = (party_person, party_project, party_org,)
+
+    type = StringType(choices=PARTY_TYPES)
+
+
+class Dependency(EnhancedModel):
+    """
+    A dependency points to a Package via a package id and version, and jhas a version
```
pombredanne: There is a typo in  `jhas`

src/packagedcode/models.py
```diff
@@ -135,153 +210,121 @@ class Package(object):
     DEPENDENCY_GROUPS = (dep_runtime, dep_dev, dep_optional, dep_test,
                          dep_build, dep_ci, dep_bundled)
 
+    # path to a file or directory
+    location = None
+
+    # the id of the package
+    id = StringType()
```
pombredanne: I am not sure we had an id originally? did we? can you refresh me on the purpose?

src/packagedcode/models.py
```diff
@@ -135,153 +210,121 @@ class Package(object):
...
+    versioning = ModelType(Version)
+    name = StringType(required=True)
+    # this is a "short" description.
```
pombredanne: Is there a way in schematics to add doc strings to the field directly rather than using comments?

src/packagedcode/models.py
```diff
@@ -135,153 +210,121 @@ class Package(object):
...
+    keywords_doc_url = URLType()
+
+    # paths to metadata files for this package, if any
+    # can be the same as the package location (e.g. RPMs)
+    metafile_locations = ListType(StringType(), default=[])
+
+    # URLs to metadata files for this package.
+    metafile_urls = ListType(URLType(), default=[])
+
+    homepage_url = URLType()
```
pombredanne: What are the constraints of a URL type? in some case we have a home that is an FTP url. Does it support this? If not then we should use a plain string instead

src/packagedcode/models.py
```diff
@@ -135,153 +210,121 @@ class Package(object):
...
+    notes = StringType()
+
+    # one or more direct download urls, possibly in SPDX vcs url form
+    # the first one is considered to be the primary
+    download_urls = ListType(URLType(), default=[])
```
pombredanne: Same comment as above: we cannot use a URL type if this is limited to  HTTP schemes only.

src/packagedcode/models.py
balusarakesh: we have an `id` in the original Pacakge class `https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/models.py#L152`

src/packagedcode/models.py
```diff
@@ -135,153 +210,121 @@ class Package(object):
...
+    bug_tracking_url = URLType()
+
+    # strings (such as email, urls, etc)
+    support_contacts = ListType(StringType(), default=[])
+
+    # a URL where the code can be browsed online
+    code_view_url = URLType()
+
+    # one of git, svn, hg, etc
+    vcs_tool = StringType(choices=['git', 'svn', 'hg'])
```
pombredanne: add also `cvs`

src/packagedcode/models.py
```diff
@@ -94,14 +103,80 @@
   -- code in source or compiled form or both.
 """
 
-class Package(object):
+
+class EnhancedModel(Model):
+    def __init__(self, deserialize_mapping=None, strict=True, **kwargs):
```
pombredanne: What are `deserialize_mapping=None, strict=True` about? 
Also the class should be rather called `BaseModel` IMH

src/packagedcode/models.py
```diff
@@ -94,14 +103,80 @@
...
+        Model.__init__(self, raw_data=kwargs, deserialize_mapping=deserialize_mapping, strict=strict)
+
+
+class Version(EnhancedModel):
```
pombredanne: I am a tad confused about Version being a model vs being what used to be called versioning? can you explain how you see this work in practice?

src/packagedcode/models.py
```diff
@@ -94,14 +103,80 @@
...
+    version = StringType()
+
+
+class AssertedLicense(EnhancedModel):
+    """
+    As asserted in a package
+    """
+    license = StringType(default=None)
```
pombredanne: Why do these strings have a default to None and most other strings have no defaults? what is the purpose?

src/packagedcode/models.py
```diff
@@ -94,14 +103,80 @@
...
+    version_constraint = StringType()
+
+    # a normalized list of version constraints for this dep.
+    # this is package indepdenent
+    normalized_version_constraints = ListType(StringType(), default=[])
+
+    # is the dep up to date and the latest available?
+    is_latest = BooleanType(default=False)
+
+    def resolve_and_normalize(self):
```
pombredanne: What is the purpose of this new method?

src/packagedcode/models.py
```diff
@@ -94,14 +103,80 @@
...
+    id = StringType()
+    # the effective or concrete resolved and used version
+    version = StringType()
+
+    # the potential constraints for this dep.
+    # this is package format specific
+    version_constraint = StringType()
+
+    # a normalized list of version constraints for this dep.
+    # this is package indepdenent
```
pombredanne: There is typo in the comment. But I would need to get some explanation on how you see versioning working first

src/packagedcode/models.py
balusarakesh: it was there in the original Package class `https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/models.py#L421`

thirdparty/prod/schematics.ABOUT
balusarakesh: Added `schematics.LICENSE` and copyrigths.

src/packagedcode/models.py
balusarakesh: fixed the typo

src/packagedcode/models.py
balusarakesh: removed URLType and using StringType instead

src/packagedcode/models.py
balusarakesh: renamed class `Version` to `Versioning` to make it more readable

src/packagedcode/models.py
balusarakesh: removed default values when not required

src/packagedcode/models.py
balusarakesh: fixed the typo

src/packagedcode/models.py
balusarakesh: using StringType instead of URLType

tests/commoncode/test_fileutils.py
```diff
@@ -318,14 +318,14 @@ def test_os_walk_with_unicode_path(self):
     def test_fileutils_walk(self):
         test_dir = self.get_test_loc('fileutils/walk')
         base = self.get_test_loc('fileutils')
-        result = [(as_posixpath(t.replace(base, '')), d, f,) for t, d, f in fileutils.walk(test_dir)]
+        result = [(as_posixpath(t.replace(base, '')), d, sorted(f),) for t, d, f in fileutils.walk(test_dir)]
```
pombredanne: How could a change to these tests be related and part of the package model validation?

tests/packagedcode/test_package_models.py
```diff
@@ -22,23 +22,146 @@
...
     test_data_dir = os.path.join(os.path.dirname(__file__), 'data')
 
     def test_model_creation_and_dump(self):
-        aap = models.AndroidAppPackage()
+        aap = models.AndroidAppPackage(name='someAndroidPAcakge')
         result = aap.as_dict()
-        expected = OrderedDict([('type', 'Android app'),
-                                ('packaging', 'archive'),
-                                ('primary_language', 'Java')])
+        expected = {
```
pombredanne: We need the as_dict() function to return an OrderedDict with a stable an well defined order. Otherwise the results of a identical scan may not be the same text-wise. We want them to be stable and in a predictable order.
Could you return an OrderedDict again?

thirdparty/prod/schematics.LICENSE
```diff
@@ -0,0 +1,27 @@
+(The BSD License)
```
pombredanne: You probably want to delete this file and rename the .NOTICE to .LICENSE

tests/commoncode/test_fileutils.py
balusarakesh: the tests failed in my windows PC when I removed the `sorted()` function.

balusarakesh: because the `result` attribute have different values in Ubuntu and Windows : 
windows : 
```
expected = [
            ('/walk', ['d1'], ['f', 'unicode.zip']),
            ('/walk/d1', ['d2'], ['f1']),
            ('/walk/d1/d2', ['d3'], ['f2']),
            ('/walk/d1/d2/d3', [], ['f3'])
        ]
```

Ubuntu : 
```
expected = [
            ('/walk', ['d1'], ['unicode.zip', 'f']),
            ('/walk/d1', ['d2'], ['f1']),
            ('/walk/d1/d2', ['d3'], ['f2']),
            ('/walk/d1/d2/d3', [], ['f3'])
        ]
```
So, I put a `sorted()` which fixed the issue.

tests/packagedcode/test_package_models.py
balusarakesh: the `as_dict()` function returns `OrderedDict`, please check this URL - `https://github.com/nexB/scancode-toolkit/blob/241-new-package-model/src/packagedcode/models.py#L306`

tests/commoncode/test_fileutils.py
pombredanne: ok, this make sense. Can you put that in a completely different pull request?
I would rather not mix this with the package model. They are completely unrelated things.

tests/packagedcode/test_package_models.py
pombredanne: But if you return an ordered dict, why did you change the test to use a bare dict?

tests/commoncode/test_fileutils.py
balusarakesh: Ok

tests/packagedcode/test_package_models.py
pombredanne: Actually, this is unlikely yield a stable ordering:
https://github.com/nexB/scancode-toolkit/blob/829ea55b4d11e4c8ad379a2e081cd7f590b07e81/src/packagedcode/models.py#L306.
If `self.to_primitive().items()` return a bare dict, then you are not really creating an ordereddict with a stable order. You can wrapping a random dict in an ordereddict. Still random in the end.

thirdparty/prod/schematics.LICENSE
balusarakesh: fixed it

tests/packagedcode/test_package_models.py
balusarakesh: fixed it - please check - https://github.com/nexB/scancode-toolkit/blob/241-new-package-model/src/packagedcode/models.py#L305

pombredanne:
I am merging this in develop now. Thanks @rakeshbalusa !
In hindsight the use of schematics may not be my best choice, but we can fix that later

PR #256, opened 2016-05-02
fixed fileutils test case by sorting the results obtained

balusarakesh:
For some tests like - https://github.com/nexB/scancode-toolkit/blob/241-new-package-model/tests/commoncode/test_fileutils.py#L358 this, results obtained in operating systems like Windows and Ubuntu are not in order. These tests are fixed by sorting the outputs obtained.


pombredanne:
Thanks. Merging

PR #261, opened 2016-05-09
Add SPDX keys even for (L)GPL licenses that do not specify a version

sschuberth:
In that case, choose the minimal applicable version including the "+"
modifier denoting "this or a later version".


pombredanne:
Simple and neat! Thank you. Merged

PR #263, opened 2016-05-10
Ignore .repo directories created by the git-repo tool

sschuberth:
git-repo [1] is used to manage multiple Git repositories. The .repo
directory contains the consolidated Git histories of the managed projects
(which are symlinked from the individual projects' .git repositories), so
we should ignore the .repo directory just like the .git directory.

However, the .repo directory also contains the source code to the git-repo
tool itself. One might argue that this source code should be picked up
during a scan, but then again the code of git-repo is not integrated into
the deliverable / what is being built from the source code as git-repo is
just a VCS tool, so a user is probably not interested in any license
findings in git-repo.

[1] https://code.google.com/p/git-repo/


pombredanne:
This make perfect sense. This quasi VCS metadata to me alright.
Do you mind adding a quick test to https://github.com/nexB/scancode-toolkit/blob/cc0c972cb1a3e743749b746e5a923ce719075395/tests/commoncode/test_ignore.py ?
I can handle it otherwise, but I am lazy ;)

Note also that at some point in the near future we would need to make the `ignore` configurable with a config file ala .gitignore

pombredanne:
note also that the pattern for these tests is to put tests files in a tarball, otherwise they could be ignored in a local git config and not committed

sschuberth:
@pombredanne Done.

pombredanne:
This has been merged with https://github.com/nexB/scancode-toolkit/commit/36b86c3d32e2691e4137885c38ed737f215e57ea 
I amended your commit to make the test file a tad smaller as it did not need a full repo tool proper. Just the right paths and files.
Thank you!

Issue #281, opened 2016-07-25
develop branch: Failure to scan NPM package.json


pombredanne:
```
 ./scancode --package tests/packagedcode/data/npm/as_installed/package.json 
Scanning files...
  [------------------------------------]  1  Scanning: package.json
Scanning done.
Traceback (most recent call last):
  File "/home/me/scancode-toolkit/bin/scancode", line 9, in <module>
    load_entry_point('scancode-toolkit==1.6.3', 'console_scripts', 'scancode')()
  File "/home/me/scancode-toolkit/local/lib/python2.7/site-packages/click/core.py", line 664, in __call__
    return self.main(*args, **kwargs)
  File "/home/me/scancode-toolkit/src/scancode/utils.py", line 61, in main
    standalone_mode=standalone_mode, **extra)
  File "/home/me/scancode-toolkit/local/lib/python2.7/site-packages/click/core.py", line 644, in main
    rv = self.invoke(ctx)
  File "/home/me/scancode-toolkit/local/lib/python2.7/site-packages/click/core.py", line 837, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/me/scancode-toolkit/local/lib/python2.7/site-packages/click/core.py", line 464, in invoke
    return callback(*args, **kwargs)
  File "/home/me/scancode-toolkit/src/scancode/cli.py", line 248, in scancode
    results = scan(input, copyright, license, package, email, url, info, verbose, quiet)
  File "/home/me/scancode-toolkit/src/scancode/cli.py", line 315, in scan
    scan_result.update(scan_one(res, scanners))
  File "/home/me/scancode-toolkit/src/scancode/cli.py", line 332, in scan_one
    scan = scan_func(input_file)
  File "/home/me/scancode-toolkit/src/scancode/api.py", line 178, in get_package_infos
    package = recognize_packaged_archives(location)
  File "/home/me/scancode-toolkit/src/packagedcode/recognize.py", line 58, in recognize_packaged_archives
    return package.get_package(location)
AttributeError: type object 'NpmPackage' has no attribute 'get_package'
```

majurg:
Side note: running `scancode --package tests/packagedcode/data/npm/as_installed/package.json` on 86-approximate-license-detection does not generate this error. 

It does still lack package information, however:

majurg:
I believe this can be closed. Node packages throughout the file tree are correctly scanned and identified when using the `--package` flag.

majurg:
[...]

PR #292, opened 2016-08-09
#281: added explicit test cases for npm package scanning

majurg:
Added two test cases for `--package` flag scanning, specifically for an `npm` package. Feedback appreciated.


pombredanne:
Thanks! mering in.

PR #308, opened 2016-08-24
184 Added new rule to detect LGPL 2.1 license

balusarakesh:
Added new rule and tests to detect LGPL 2.1 license


src/licensedcode/data/rules/lgpl-2.1_39.RULE
```diff
@@ -0,0 +1,18 @@
+/*
+ * libuci - Library for the Unified Configuration Interface
+ * Copyright (C) 2008 Felix Fietkau <nbd@openwrt.org>
```
pombredanne: Let me fix this even though I merged: we never put copyrights in license rules.

PR #333, opened 2016-10-12
If the platform is unsupported, actually disclose its name

sschuberth:
This is useful to know for debugging.

### DIFF
--- a/etc/conf/base.py
+++ b/etc/conf/base.py
@@ -6,8 +6,8 @@ import sys
 Check that we run a supported OS and architecture.
 """
 
-def unsupported():
-    print('Unsupported OS/platform.')
+def unsupported(platform):
+    print('Unsupported OS/platform "%s".' % platform)
     print('See https://github.com/nexB/scancode-toolkit/ for supported OS/platforms.')
     print('Enter a ticket https://github.com/nexB/scancode-toolkit/issues asking for support of your OS/platform combo.')
     sys.exit(1)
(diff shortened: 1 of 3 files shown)


pombredanne:
Looking thanks....  a minor comment: if you use `%r` instead of `"%s"` you will get the same effect with a slightly improved readability (`%r` is to get the "representation" of an object, which in cases of a string will be a quoted string.)

pombredanne:
Just curious: which OS/arch combo did scancode fail on?

sschuberth:
"win64", as I accidentally had the 64-bit version of Python in PATH. Likewise just curious, when will ScanCode run with 64-bit Python on Windows?

pombredanne:
running on 64bits should not be a big issue, this is just about recompiling all the thirdparty and ensuring the tests pass: I have started working on getting 64 bits Python wheels here https://github.com/pombreda/thirdparty to have a fully automated rebuild of all thirddparties. There are some of them that can compile only under Mingw and I did not come down to these yet. If you are interested to chip in I could guide you alright.  This is is mostly about native libraries and testing: the candidates are likely libarchive, libmagic and 7zip and it is very possible that no recompilation is needed at all since these are either spawned (7zip) or use with a ctypes binding (libarchive, libmagic)

pombredanne:
Merged with 6bbd4e5088f15ca082682a112eecc246e39ba880
Thank you!

Issue #143, opened 2015-12-04
Incorrect license detection


chinyeungli:
https://github.com/vmware/open-vm-tools/blob/master/open-vm-tools/lib/string/convertutf.c

scancode returns "unicode" , but the actual license should be "unicode-mappings"

```
./scancode -l convertutf.c 
Scanning files...
  [####################################]  1                        
Scanning done.
{
  "scancode_notice": "Generated with ScanCode and provided on an \"AS IS\" BASIS, WITHOUT WARRANTIES\nOR CONDITIONS OF ANY KIND, either express or implied. No content created from\nScanCode should be considered or used as legal advice. Consult an Attorney\nfor any legal advice.\nScanCode is a free software code scanning tool from nexB Inc. and others.\nVisit https://github.com/nexB/scancode-toolkit/ for support and download.", 
  "scancode_version": "1.4.0", 
  "resource_count": 1, 
  "results": [
    {
      "location": "convertutf.c", 
      "licenses": [
        {
          "key": "unicode", 
          "short_name": "Unicode Inc License Agreement", 
          "category": "Attribution", 
          "owner": "Unicode Consortium", 
          "homepage_url": "http://unicode.org/", 
          "text_url": "http://unicode.org/copyright.html", 
          "dejacode_url": "https://enterprise.dejacode.com/license_library/Demo/unicode/", 
          "spdx_license_key": "", 
          "spdx_url": "", 
          "start_line": 5, 
          "end_line": 23
        }
      ]
    }
  ]
}
```

pombredanne:
@JonoYang I am reopening as this is missing a test

pombredanne:
All set. @JonoYang Thank you!

PR #343, opened 2016-10-24
#143 incorrect license detection

JonoYang:
@pombredanne I have modified `unicode_2.yml` to detect only `unicode-mappings` instead of `unicode`, as the license text for `unicode_2.RULE` matches that of `unicode-mappings`. Also, I deleted the copyright line from `unicode_2.RULE`.


pombredanne:
@JonoYang Thanks. Simple and effective!

pombredanne:
@JonoYang I am reopening as this is missing a test

Issue #92, opened 2015-10-19
incorrect license GPL is detected


rakeshbalusa:
OS: windows 7 64 bit
I've updated my scancode to the latest version available.
file : [README.txt](https://github.com/nexB/scancode-toolkit/files/15755/README.txt)

scancode output :
```
Path	                                What	Start Line	End Line		Info
C:/doc/temporary/issue_files/README	License	     28	28		GPL 2.0 
```
In the `README` file attached there is license text which says `BSD Unix, and not under the GNU General Public License (GPL).` in the line 28. This means the correct license is `BSD` but no `GPL`

pombredanne:
@JonoYang I am reopening as this is missing a test

pombredanne:
@JonoYang ready for your review. Here is what I did:
1. the negative rule you added was really a positive rule: I renamed it
2. there were several ambiguities introduced by flex-2.5 and other rules that did not contribute to accurate license detection. I made quite a few updates accordingly
3. finally, I edited the branch history to rebase cleanly on develop rather than merging and forced pushed the branch. This is preferred for feature branches to avoid polluting the history with too many un-needed  merge commits. (forced pushing is never nor allowed on the master and develop branches FWIW)

pombredanne:
See #349

pombredanne:
All set.

PR #344, opened 2016-10-24
#92: Add 'not-a-license' rule to remove GPL false positive

JonoYang:
@pombredanne I created a new 'not-a-license' rule for the flex README to prevent scancode from detecting a GPL license. The flex license is properly detected, but the 'Brian Gladman 3-Clause License' (a bsd license derivative) is now being detected also.


pombredanne:
@JonoYang I added a note explaining the rationale in the rule YAML file. Look good.merging

pombredanne:
@JonoYang I am reopening as this is missing a test

PR #348, opened 2016-10-25
#143 Add test for unicode-mappings license

JonoYang:
@pombredanne I added the unicode-mappings license text as a test.


pombredanne:
@JonoYang perfect. The tests look good on 1/2 of the slaves so no need to wait. Merging. Thanks!

PR #349, opened 2016-10-26
#92 incorrect license gpl detected

pombredanne:
@JonoYang Here is what I did:
1. the negative rule you added was really a positive rule: I renamed it
2. there were several ambiguities introduced by flex-2.5 and other rules that did not contribute to accurate license detection. I made quite a few updates accordingly
3. finally, I edited the branch history to rebase cleanly on develop rather than merging and forced pushed the branch. This is preferred for feature branches to avoid polluting the history with too many un-needed merge commits. (forced pushing is never nor allowed on the master and develop branches FWIW)


pombredanne:
@JonoYang You can check the details, but this good to merge now.

PR #350, opened 2016-11-01
Add some convenience around creating the result file

sschuberth:


pombredanne:
Thanks! I get the change for https://github.com/nexB/scancode-toolkit/pull/350/files#diff-1e37ecfb9a8c3fccd8ecdfb2ce51f591R212 .... 

But what is the problem you faced that triggered https://github.com/nexB/scancode-toolkit/pull/350/files#diff-1e37ecfb9a8c3fccd8ecdfb2ce51f591R366 ?

src/scancode/cli.py
```diff
@@ -362,6 +363,14 @@ def save_results(scanned_files, format, input, output_file):
     """
     Save results to file or screen.
     """
+    dirname = os.path.dirname(output_file.name)
+    if not os.path.isdir(dirname):
+        try:
+            os.makedirs(dirname)
```
pombredanne: Do you think you could use a call to `commoncode.fileutils.create_dir` instead?
See https://github.com/nexB/scancode-toolkit/blob/086addda656150b7feed6da438fe1413091145a2/src/commoncode/fileutils.py#L60

The rationale is that its needs to eventually support long Windows path using the UNC `\\?\` syntax so it is best to control places where directories are created.

sschuberth:
It's basically just convenience. We're scanning several projects in a loop and the individual result files need to go to a separate directory structure that is based on the project name, scan date etc. Instead of having to create this empty directory structure beforehand, it's easy to just let ScanCode create it when writing the result files.

sschuberth:
I'll use the commoncode function you suggested in the next iteration.

pombredanne:
@sschuberth you wrote:
> Instead of having to create this empty directory structure beforehand, it's easy to just let ScanCode create it when writing the result files. 

This makes perfect sense!

pombredanne:
and you can see the appevyor checks now properly in Github for this PR.

sschuberth:
@pombredanne I've updated the PR.

pombredanne:
Thanks. Looking good. Will merge as soon as the Ci checks are completed.

src/scancode/cli.py
```diff
@@ -209,6 +209,7 @@ class ScanCommand(utils.BaseCommand):
 formats = ('json', 'html', 'html-app',)
 
 def validate_formats(ctx, param, value):
+    value = value.lower()
```
pombredanne: I wonder if the new code should not be more or less like this:
```
def validate_formats(ctx, param, value):
    value_lower = value.lower()
    if value_lower not in formats:
        # render using a user-provided custom format template
        if not os.path.isfile(value):
            raise click.BadParameter('Invalid template file: "%(value)s" does not exists or is not readable.' % locals())
    else:
        value = value_lower
    return value
```

The rationale being that the value could be either a format or a file path... If you lower the path on a case-sensitive OS it may not longer be valid.

src/scancode/cli.py
```diff
@@ -362,6 +363,9 @@ def save_results(scanned_files, format, input, output_file):
     """
     Save results to file or screen.
     """
+    if output_file != sys.stdout:
```
pombredanne: This may be missing an `import sys` at the top?

src/scancode/cli.py
```diff
@@ -362,6 +363,9 @@ def save_results(scanned_files, format, input, output_file):
...
+        fileutils.create_dir(os.path.dirname(output_file.name))
```
pombredanne: You may want to consider the case where the path is does not contain any directory but is just a plain file. In case, the parent directory name may be empty. So may be something like that may help:
```
        parent_dir = os.path.dirname(output_file.name)
        if parent_dir:
            fileutils.create_dir(parent_dir)
```
There is may be another case to consider for userdir expansion (e.g. if some `~/some/path/to/file` was used). In that case  then using this may be better:
```
from os.path import expanduser
from os.path import abspath
```
```
        parent_dir = os.path.dirname(output_file.name)
        if parent_dir:
            fileutils.create_dir(abspath(expanduser(parent_dir)))
```

src/scancode/cli.py
sschuberth: You're absolutely right. However, I "Inverted" the logic to reduce indentation and the number of code branches.

src/scancode/cli.py
sschuberth: Thanks for this, I'll go with the latter.

pombredanne:
LGTM. Merging

pombredanne:
Thanks!

Issue #267, opened 2016-05-23
segmentation fault error while running scans in an ubuntu instance in AWS


balusarakesh:
I cloned into the github repo `scancode-toolkit` in an Ubnuntu instance `(AWS-EC2 AMI - ami-06116566)` and used the command `./scancode --help` to configure scancode for the first time.
The next time when I try to run a command `./scancode --license file.txt` I'm getting the error

```
/scancode: line 16:  8253 Segmentation fault      (core dumped) $SCANCODE_ROOT_DIR/bin/scancode "$@"
```

pombredanne:
Do you have some more details on the OS version, architecture and version of Python used there?

balusarakesh:
`Ubuntu Server 14.04 LTS (HVM)`

balusarakesh:
@pombredanne  Python version is 2.7.6

pombredanne:
And which architecture? 32 or 64 bits?

balusarakesh:
64 bit

balusarakesh:
2.7 version of Python

yahalom5776:
Sorry to bring this up again but I seem to have a very similar issue when scanning a large project:
`./scancode: line 16:  2977 Killed $SCANCODE_ROOT_DIR/bin/scancode "$@"`

It is the current development branch installed on Debian 8 running in a VM with 2GB of RAM.

Python:
`python --version: 
Python 2.7.9`

OS:
`uname -a: 
Linux debian 3.16.0-4-amd64 #1 SMP Debian 3.16.36-1+deb8u1 (2016-09-03) x86_64 GNU/Linux`

Let me know if you need more information. Thank you.

pombredanne:
@yahalom5776 Thanks for the report. Can you tell me which command line options you used?

My best guess is that you have been prockilled' rather than segfaulted' based on your error message. This is something that should go away pretty soon. 
The issue is simple: at the moment the results of scanning each file is accumulated in RAM for all files.
Eventually when scanning larger codebase, this will exhaust available RAM and the kernel may prockill the scancode process in this case. The solution is going to be saving on disk the scan results of each file on the go and then streaming the final JSON at the end such that the scan results are never fully held in memory.
This is something I am working on in the orm-models branch https://github.com/nexB/scancode-toolkit/tree/orm-models 

The short term solution is to scan smaller chunks of a larger codebase. I am hoping to get some fix sometimes this week. In any case I will ping you to test the fix ass soon as this is ready.

[...]

PR #369, opened 2016-11-14
Long scan timeout for #267

pombredanne:
In #267 the key issue is some pathological, hard to diagnose scan that takes a long time and eventually uses a lot of RAM to get prockilled  (at least on Linux).

This should both help handle the issue and help diagnose it:

* a scan will be interrupted if it takes too long and exceeds a timeout: so if there is a problem somewhere, it should timeout rather than suck all the memory and get proc killed. The approach is to run each scan in a thread with a timeout using code from @dano
The timeout is computed for each file being scanned. It has a minimum of 60 seconds and a maximum of 600 seconds.  Each file bigger than 1MB gets an extra 30 of timeout for each extra MB up to the max  600s.

* a scan can run on multiple processes with the `-n or --processes` option (this is a nice side effect of handling timeouts and can also speed up scanning of course, but would use more RAM). When running on multiple process the memory used by each process may vary but would be about 600MB + size of the scanned file in memory + some extras.

* -`-verbose` output is always send to stdout, as you requested by @sschuberth which will help to diagnose issues for background, CI-like scan jobs. So running a --verbose scan on this would likely tell exactly which is the file that is causing an error.

* a summary of the scan statistics is displayed at the end of a scan: this helps diagnose errors as it does list all the paths that failed to scan alright or did timeout. It also provide some timing and scanning speed details.


pombredanne:
The latest commits now make the memory check the default

src/scancode/cli.py
```diff
@@ -262,102 +300,201 @@ def scancode(ctx, input, output_file, copyright, license, package,
...
+            _progress_line = verbose and _scanned_path or fileutils.file_name(_scanned_path)
+            return style('Scanning: ') + style(_progress_line, fg=_scan_success and 'green' or 'red')
+
+    scanning_errors = []
+    files_count = 0
+    with utils.progressmanager(scanned_files, item_show_func=scan_event, show_pos=True, verbose=verbose, quiet=quiet) as scanned:
+        while True:
+            try:
+                result = scanned.next()
+                scan_sucess, scanned_rel_path = result
```
sschuberth: Minor typo: Should say "scan_success".

sschuberth:
Looks good to me from a cursory view, thanks for working on this. On the CI side, the one Travis build seems to have timed out and probably just needs to be retriggered. For AppVeyor things look a bit worse, probably some tweaking of `TEST_MAX_MEMORY` is required?

pombredanne:
@sschuberth I fixed the typos and a few more. Thanks ++ for this review! I need to fix the CI failures now.

pombr:
And at least one of the builds passed on Mac too: https://travis-ci.org/nexB/scancode-toolkit/builds/176613254

pombr:
Linux and Windows seem OK

pombr:
All tests pass now: :green_apple: :green_heart:

sschuberth:
Let's get it merged for some wider exposure! :-)

pombredanne:
@sschuberth agreed. Further fixes if needed can branch as needed.

PR #371, opened 2016-11-18
#267 Add command options for --timeout and --max-memory

pombredanne:
* defaults are 120 seconds and 1000 MB
 * this replaces a dynamic computation based on file size entirely
 * this allow to control explicitly and strictly the runtime quotas
   and adjust them to constrained environments such as VMs.

### DIFF
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -68,8 +68,9 @@ from commoncode import filetype
 from scancode import __version__ as version
 
 from scancode.interrupt import interruptible
-from scancode.interrupt import compute_memory_quota
-from scancode.interrupt import compute_timeout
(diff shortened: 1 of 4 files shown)


sschuberth:
The only thing I'm wondering about is whether it would make sense to still compute the default values dynamically, and only for constrained environments use the fixed values from the command line, if any.

pombredanne:
@sschuberth you have a good point but I thought this became too complicated
Especially since I finally found the memory leak that is the root cause for #267 ... I should have a new commit tomorrow.

sschuberth:
Sounds reasonable!

pombredanne:
ok, so I am merging this and I will start a separate branch to push the rest of the fixes

PR #376, opened 2016-11-24
#267 Remove memory leak from license detection

pombredanne:
... and other refinements and fixes detailed in the commit messages.
This should address all remaining issues in #267


pombredanne:
@sschuberth @yahalom5776 This is passing all checks and there quite a few changes I would like to build future code on. So barring a negative feedback, I will merge this over the wee-end at the latest.

sschuberth:
Even if I did not have the time to review the code yet, I did give it a try by running it on one of our projects. And there is good news: One project that was previously causing the scanner to hang (I aborted the scan after 19 hours) completed after 30 minutes with this branch merged to the current `develop` branch.

pombredanne:
@sschuberth you wrote:
> And there is good news: One project that was previously causing the scanner to hang (I aborted the scan after 19 hours) completed after 30 minutes with this branch merged to the current develop branch.

Good news! I am not entirely surprised. I removed the caching of license matches and the caching of scans has been entirely refactored to use plain JSON files instead of a sqlite db (through diskcache). Both were using sqlite and sqlite was a major source of slowdown and was making the scan super I/O intensive for no reason. On a single process, I saw that this was at least about 4 times faster than before on my laptop. On multiple processes I achieved up to 10 files/sec for a license+copyright and 200 files/sec for just info. So one million files should take about 3 hours now and possibly much less on a faster-than-laptop-with-ssd server.

pombredanne:
We should create some benchmarks to watch for any perf changes btw.

yahalom5776:
@pombredanne looks good so far! Thank you.

sschuberth:
> We should create some benchmarks to watch for any perf changes btw.

That would be nice to have indeed, i.e. to have tests to identify performance regressions (or improvements).

As a first step, looking at the times Travis CI builds take could be useful. Of course, this includes the setup and build times while we're only interested in the time to run tests, and the tests might be too many small projects whereas scanning performance probably is better measured with a single big project, but still.

So I was quickly giving http://scribu.github.io/travis-stats/#nexB/scancode-toolkit a look, but unfortunately it does not render properly for me.

pombredanne:
Thank you for the feedback. I think this is good enough for me to merge in now!

Issue #342, opened 2016-10-24
Bare CPOL license detection rule detection issue


pombredanne:
See https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/rules/cpol-1.02_2.RULE
Its text is only `CPOL` and this creates several kind of false positive detections. This license rule should be removed entirely as this is not discriminant enough

pombredanne:
Actually the issue is not so much with the rule but rather with the fact that the score presented when `CPOL` is matched  is 100%. The thing is this is a weak clue.

After some chats, @DennisClark suggested that we could help determine if a license match was strong by returning (and displaying in the AboutCode Manager) the actual text that is matched (there are #260 and #304 on that topic) 

@mjherzog suggested that we need to provide a better score with some sliding scale of the weigth of a match. Here is the weight should be small.... but a bare work or short rule could be quite strong as in `license: gpl2` so @DennisClark suggested that we should be able to override any automated strength computation for these cases. @JonoYang suggested a metric or strength indication could be called `confidence` . One my side I think that there is likely something we could compute from the data: for instance a matched rule text even if matched 100% may represent a small portion of and overall license text and therefore offer less confidence about the detection.

pombredanne:
Some examples of true and false positives for CPOL in C/C++/C# code:
- https://github.com/search?l=C%2B%2B&q=cpol&type=Code
- https://github.com/search?l=C&q=cpol&type=Code 
- https://github.com/search?l=C%23&q=cpol&type=Code

pombredanne:
Here is what I suggest:
Add a new `confidence` attribute to each license match that is computed as a value between 0 (low confidence that this a good match) and 100 (high confidence that this a good match).

Computation would be based on :
- the length of a rule text compared to the length of the corresponding full license text as a float value between 0 and 1 (and in the rare cases where the rule text would bigger than the license text use 1) and would be the rule `relevance`. 
- the `score` of a match (which is already computed internally) as the ratio of the match length to the matched rule text full length.
- (and with some special corner cases for rule relevance and other minor factors where we could set the relevance manually) 

The rule `relevance` would be multiplied by the match `score` to provide a match `confidence` with a formula roughly like:
`Rule relevance = Rule length / License length`
`Match score =  Match length / Rule length`
`Match confidence =  Match score * Rule relevance`
[...]

PR #383, opened 2016-12-01
#342 Improved license match scoring with new score based on rule relevance

pombredanne:
This implements the new scoring systems proposed in #342 

I squeezed a few unrelated commits in there and I could extract them if this is an issue.


pombredanne:
Ok, since there is no push back, I will merge this.

Issue #339, opened 2016-10-20
Upgrade json2csv.py script to handle the new ABCD format.


majurg:
Currently, the json2csv.py script cannot convert ABCD formatted scan results from JSON to csv.

For example, running the current develop branch (w ABCD formatting) results in this error:

```
# scancode -ilpc samples/ ~/ABCD_scan_results.json && python etc/scripts/json2csv.py ~/ABCD_scan_results.json ~/ABCD_scan_results.csv
Scanning files...
  [####################################]  43                                     
Scanning done.
Traceback (most recent call last):
  File "etc/scripts/json2csv.py", line 345, in <module>
    cli()
  File "/Users/sesser/Documents/src/scancode/lib/python2.7/site-packages/click/core.py", line 664, in __call__
    return self.main(*args, **kwargs)
  File "/Users/sesser/Documents/src/scancode/lib/python2.7/site-packages/click/core.py", line 644, in main
    rv = self.invoke(ctx)
  File "/Users/sesser/Documents/src/scancode/lib/python2.7/site-packages/click/core.py", line 837, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/sesser/Documents/src/scancode/lib/python2.7/site-packages/click/core.py", line 464, in invoke
    return callback(*args, **kwargs)
  File "etc/scripts/json2csv.py", line 341, in cli
    json_scan_to_csv(json_input, csv_output, strip=strip)
  File "etc/scripts/json2csv.py", line 119, in json_scan_to_csv
    scan_results = load_scan(json_input)
  File "etc/scripts/json2csv.py", line 109, in load_scan
    scan_results = scan_results['results']
KeyError: 'results'
```

And running on the v1.6.3 release (old scan format) generates a correct `csv` output file:

```
# scancode -ilpc samples/ ~/old_scan_results.json && python etc/scripts/json2csv ~/old_scan_results.json ~/old_scan_results.csv
```

JonoYang:
Fixed in #399

PR #399, opened 2016-12-09
#339 Upgrade json2csv.py

JonoYang:
`json2csv.py` has been updated to handle the latest ScanCode scan results. All options (copyrights, licenses, file info, packages, emails, URLs) are supported. New tests have been created to reflect the change in ScanCode output.


etc/scripts/json2csv.py
```diff
@@ -106,239 +75,148 @@ def load_scan(json_input):
...
 
 
-def scan_as_list(scan, strip=0):
+def flatten_scan(scan):
     """
-    Return a list of ordered dictionaries of key/values flattening the data and
-    keying always by location, given a ScanCode scan results list.
+    Yield ordered dictionaries of key/values flattening the data and
+    keying always by path, given a ScanCode scan results list.
     Optionally strip up to `strip` path segments from the location paths.
```
pombredanne: With the `strip` arg gone, the doc is obsolete, right?

etc/scripts/json2csv.py
```diff
@@ -106,239 +75,148 @@ def load_scan(json_input):
...
+
+    for scanned_file in scan:
+        path = scanned_file['path']
+        path = path if path.startswith('/') else '/' + path
+
+        if scanned_file.get('type', '') == 'directory':
+            if not path.endswith('/'):
+                path = path + '/'
+
+        path = '/code' + path
```
pombredanne: Could this be something that could be passed as an arg or option? Let me think

etc/scripts/json2csv.py
```diff
@@ -106,239 +75,148 @@ def load_scan(json_input):
...
+
+        file_info = OrderedDict(Resource=path)
+        info_details = OrderedDict(((k, v) for k, v in scanned_file.items() if k != 'path' and not isinstance(v, list)))
+        file_info.update(info_details)
+        yield file_info
+
+        error_info = OrderedDict(Resource=path)
```
pombredanne: Since there is only one scan_errors attribute for any given file, why not yielding this with the file_info above?

etc/scripts/json2csv.py
```diff
@@ -106,239 +75,148 @@ def load_scan(json_input):
...
+        # Scan errors are to be joined in a multi-line cell
+        errors = [error for error in scanned_file.get('scan_errors', [])]
```
pombredanne: Why using a list comprehension here? `errors = scanned_file.get('scan_errors', [])` is all that is needed IMHO

etc/scripts/json2csv.py
```diff
@@ -106,239 +75,148 @@ def load_scan(json_input):
...
+        error_info['scan_errors'] = '\n'.join(errors)
```
pombredanne: ... meaning you could write here:
`error_info['scan_errors'] = '\n'.join(scanned_file.get('scan_errors', []))`

etc/scripts/json2csv.py
```diff
@@ -106,239 +75,148 @@ def load_scan(json_input):
...
+                    else:
+                        pack[nk] = ''
+            yield pack
+
+
+def collect_header_keys(scan_data):
+    """
+    Return a list of keys collected from a list of scanned data dictionaries.
+    """
+    keys = OrderedSet()
```
pombredanne: If this is the only place where you use an OrderedSet you could get rid of the OrderedSet class and fold the bits of code from it in that function for simplicity

etc/scripts/json2csv.py
```diff
@@ -106,239 +75,148 @@ def load_scan(json_input):
...
 @click.argument('csv_output', type=click.Path(exists=False, readable=True))
-@click.option('-s', '--strip', help='Number of leading path segments to strip from location paths', type=click.INT, default=0)
 @click.help_option('-h', '--help')
-def cli(json_input, csv_output, strip=0):
+def cli(json_input, csv_output):
     """
     Convert a ScanCode JSON scan file to a nexb-toolkit-like CSV.
-    Optionally strip up to `strip` leading path segments from the location paths.
 
     JSON_INPUT is either a ScanCode json format scan or the data.json file from a ScanCode html-app format scan.
```
pombredanne: Is this still True? or are only the plain JSON results handled? Did you add a test for both cases?

etc/scripts/json2csv.py
```diff
@@ -106,239 +75,148 @@ def load_scan(json_input):
...
+                            'keywords_doc_url',
+                            'download_sha1',
+                            'download_sha256',
+                            'download_md5',
+                            'code_view_url',
+                            'vcs_tool',
+                            'vcs_revision',
+                            'license_expression')
+
+        for package in scanned_file.get('packages', []):
```
pombredanne: Can you make sure that for the package version field you always pad it with `v space` if present .... otherwise loading a version in a spreadsheet like Excel or LibreOffice will mistake a `1.10` version for `1.1` when loading?

JonoYang:
Thanks! I'll make the changes.

etc/scripts/json2csv.py
JonoYang: This is still true. `json2csv.py` works for both regular JSON ScanCode results and the data.json file from the html-app data files.

pombredanne:
all tests are passing except for one pending on Mac release.... Looking good and merging. Thanks for making the updates too!

Issue #395, opened 2016-12-06
Add support for PHP Composer packages detection


pombredanne:
As part of supporting many package formats , composer is commonly used with PHP code https://getcomposer.org/

The format of the manifest is documented here: https://getcomposer.org/doc/04-schema.md

Some pointers that may be of some vague use:
https://github.com/cloudfoundry/php-buildpack/blob/master/extensions/composer/extension.py

pombredanne:
This is now fixed in develop

PR #403, opened 2016-12-13
#395 Add the parser for PHP composer

linexb:
Add the parser for PHP composer


src/packagedcode/phpcomposer.py
```diff
@@ -0,0 +1,403 @@
...
+
+from packagedcode import models
+
+"""
+Handle PHP composer packages
+per https://getcomposer.org/
+"""
+
+"""
+To check
```
pombredanne: This is for NPMs not composer. Strike this comment out

src/packagedcode/phpcomposer.py
```diff
@@ -0,0 +1,403 @@
...
+
+
+logger = logging.getLogger(__name__)
+# import sys
+# logging.basicConfig(level=logging.DEBUG, stream=sys.stdout)
+# logger.setLevel(logging.DEBUG)
+
+
+class PHPComposerPackage(models.Package):
+    metafiles = ('package.json')
```
pombredanne: This would be `composer.json` not `package.json`

src/packagedcode/phpcomposer.py
```diff
@@ -0,0 +1,403 @@
...
+    filetypes = ('.json',)
+    mimetypes = ('application/x-tar',)
```
pombredanne: Where is this from? Is there such a mimetype that can be used for composer?

src/packagedcode/phpcomposer.py
```diff
@@ -0,0 +1,403 @@
...
+    repo_types = (models.repo_phpcomposer,)
+
+    type = models.StringType(default='phpcomposer')
+    primary_language = models.StringType(default='JavaScript')
```
pombredanne: The language is PHP not JavaScript

src/packagedcode/phpcomposer.py
```diff
@@ -0,0 +1,403 @@
...
+def parse(location):
+    """
+    Return a Package object from a composer.json file or None.
+    """
+    if not is_phpcomposer_json(location):
+        return
+    # mapping of top level composer.json items to the Package object field name
+    plain_fields = OrderedDict([
+        ('name', 'name'),
+        ('description', 'summary'),
```
pombredanne: What about the version? Why do you handle it differently  below than these other attributes?

src/packagedcode/phpcomposer.py
```diff
@@ -0,0 +1,403 @@
...
+        ('keywords', 'keywords'),
+        ('homepage', 'homepage_url'),
+    ])
+
+    # mapping of top level package.json items to a function accepting as arguments
+    # the package.json element value and returning an iterable of key, values Package Object to update
+    field_mappers = OrderedDict([
+        ('authors', author_mapper),
```
pombredanne: Is there anything extra available? such ass the VCS/git repo?

src/packagedcode/phpcomposer.py
```diff
@@ -0,0 +1,403 @@
...
+        ('require', dependencies_mapper),
+        ('require-dev', dev_dependencies_mapper),
+        ('repositories', repository_mapper),
+        ('support', support_mapper),
+    ])
+
+    with codecs.open(location, encoding='utf-8') as loc:
+        data = json.load(loc, object_pairs_hook=OrderedDict)
+
+    if not data.get('name') or not data.get('description'):
```
pombredanne: I get that a name would not make sense.... bu something without a description is perfectly usage... why ignore this?

src/packagedcode/phpcomposer.py
```diff
@@ -0,0 +1,403 @@
...
+        # a package.json without name and description is not a usable PHP composer package
+        return
+
+    package = PHPComposerPackage()
+    # a composer.json is at the root of a PHP composer package
+    base_dir = fileutils.parent_directory(location)
+    package.location = base_dir
+    # for now we only recognize a package.json, not a node_modules directory yet
+    package.metafile_locations = [location]
+    package.version = data.get('version')
```
pombredanne: See my comment above about  version

src/packagedcode/phpcomposer.py
```diff
@@ -0,0 +1,403 @@
...
+    for source, target in plain_fields.items():
+        value = data.get(source)
+        if value:
+            if isinstance(value, basestring):
+                value = value.strip()
+            if value:
```
pombredanne: What are the cases where the attribute is not a string?

src/packagedcode/phpcomposer.py
```diff
@@ -0,0 +1,403 @@
...
+        "license": [
+           "LGPL-2.1",
+           "GPL-3.0+"
+        ]
+        """
+    if not licenses:
+        return package
+
+    if isinstance(licenses, basestring):
+        if '(' in licenses and ')' in license and 'or' in license:
```
pombredanne: Is there any case where we have parens? I reckon this was the case fro NPMs, but composer is not NPM.

pombredanne: Also at this stage, let's not try to parse license expressions yet. I reckon this is a tad odd, but we would need to do this with https://github.com/nexB/license-expression/ to be correct.

src/packagedcode/phpcomposer.py
```diff
@@ -0,0 +1,403 @@
...
+                    package.vcs_tool = 'hg'
+                elif repo_url.startswith('cvs'):
+                    package.vcs_tool = 'cvs'
+                else:
+                    package.vcs_tool = 'git'
+                package.vcs_repository = parse_repo_url(repo.get('url'))
+    return package
+
+
+VCS_URLS = (
```
pombredanne: Are these listed in the composer doc?

src/packagedcode/phpcomposer.py
```diff
@@ -0,0 +1,403 @@
...
+    'git+http://',
+    'svn+https://',
+    'svn+http://',
+    'svn://',
+)
+
+
+def parse_repo_url(repo_url):
+    """
+    Validate a repo_ulr and handle shortcuts for GitHub, GitHub gist,
```
pombredanne: repo_url not repo ulr

src/packagedcode/phpcomposer.py
```diff
@@ -0,0 +1,403 @@
...
+    else:
+        package.dependencies[resolved_type] = dependencies
+    return package
+
+
+dependencies_mapper = partial(deps_mapper, field_name='dependencies')
+dev_dependencies_mapper = partial(deps_mapper, field_name='devDependencies')
+
+
+person_parser = re.compile(
```
pombredanne: I reckon this was needed for the weirdness of NPMs, but why would this apply to composer?

src/packagedcode/phpcomposer.py
```diff
@@ -0,0 +1,403 @@
...
+    Both forms are equivalent.
+    """
+    if isinstance(persons, list):
+        for person in persons:
+            # ensure we have our three values
+            name = person.get('name')
+            email = person.get('email')
+            url = person.get('homepage')
+            yield name and name.strip(), email and email.strip('<> '), url and url.strip('() ')
+    else:
```
pombredanne: Is this a good idea to exit so early  if some person cannot be parsed/made sense of?

tests/packagedcode/data/phpcomposer/a-timer/composer.json.expected
```diff
@@ -0,0 +1,82 @@
+{
+  "type": "phpcomposer", 
+  "name": "jandreasn/a-timer",
```
pombredanne: Is there some meaning in the composer world to have xxxx/yyyyy ? do xxx and yyy mean something? or the whole is the package name?
Is this related to github's onwer/repo somehow?

tests/packagedcode/data/phpcomposer/a-timer/composer.json.expected
```diff
@@ -0,0 +1,82 @@
+{
+  "type": "phpcomposer", 
+  "name": "jandreasn/a-timer", 
+  "version": null, 
+  "primary_language": "JavaScript",
```
pombredanne: PHP, Not JavaScript

tests/packagedcode/data/phpcomposer/a-timer/composer.json.expected
```diff
@@ -0,0 +1,82 @@
+{
+  "type": "phpcomposer", 
+  "name": "jandreasn/a-timer", 
+  "version": null, 
+  "primary_language": "JavaScript", 
+  "packaging": null,
```
pombredanne: What could be the proper packaging? not an archive, but may be a directory?

src/packagedcode/phpcomposer.py
linexb: Not listed, but the repository in composer doc mentions: 

> vcs: The version control system repository can fetch packages from git, svn, fossil and hg repositories.

https://getcomposer.org/doc/04-schema.md#repositories

src/packagedcode/phpcomposer.py
linexb: This is Composer style, but not NPM one. Please refer to: https://getcomposer.org/doc/04-schema.md#license, the example is:

{
    "license": "(LGPL-2.1 or GPL-3.0+)"
}

tests/packagedcode/data/phpcomposer/a-timer/composer.json.expected
linexb: From the doc it means "vendor/package"

src/packagedcode/phpcomposer.py
linexb: Good catch, I think we should use "repr" instead.

src/packagedcode/phpcomposer.py
pombredanne: ok, this is something entirely different we have here, e.g. a license expression. For now I would prefer that no parsing takes places on the asserted license which is the raw as-asserted license statement found in the package manifest. 

So best to leave it unchanged.

The parsing of this statement in a normalized expression will take place using license detection proper (and possibly the upcoming support for expression with https://github.com/nexB/license-expression/

tests/packagedcode/data/phpcomposer/a-timer/composer.json.expected
pombredanne: Then this mean that the `vendor` part could be a vendor Party for that package?

src/packagedcode/phpcomposer.py
linexb: Yes, I will keep the original format. I will keep "(" and ")" too just make sure the forward license detection will handle it.

src/packagedcode/phpcomposer.py
linexb: OK.

src/packagedcode/phpcomposer.py
linexb: Yes, version could be handled the same with the plain field. I have refined it.

src/packagedcode/phpcomposer.py
```diff
@@ -55,13 +55,13 @@
 
 
 class PHPComposerPackage(models.Package):
-    metafiles = ('package.json')
+    metafiles = ('composer.json')
     filetypes = ('.json',)
-    mimetypes = ('application/x-tar',)
+    mimetypes = ('text/xml',)
```
linexb: I didn't find a case that composer.json is archived in an archive file. So I believe it the mimetype should be directory, do you agree? We could use " ('application/x-directory', 'text/directory', 'inode/directory')" instead, OK?

src/packagedcode/phpcomposer.py
linexb: description is a required field in the schema.

tests/packagedcode/data/phpcomposer/a-timer/composer.json.expected
linexb: Yes, I agree. I will add the vendor fetch and set code.

src/packagedcode/phpcomposer.py
linexb: Yes, it could be more clear, writing code on this.

src/packagedcode/utils.py
```diff
@@ -0,0 +1,100 @@
+#
```
pombredanne: Do you you mind adding tests for this?

pombredanne:
I am merging and you can enter another ticket for the tests of `packagedcode/utils.py 
Thanks!

Issue #396, opened 2016-12-07
gpl-2.0 detected for a jar file from apache


chinyeungli:
Not sure how the `pdfbox-app-1.8.2.jar` being detected as gpl-2.0

https://repo1.maven.org/maven2/org/apache/pdfbox/pdfbox-app/1.8.2/pdfbox-app-1.8.2.jar

pombredanne:
This is a fluke. Somewhere in the binary there is a GPL string. Thanks for the report.... We should NOT scans compressed binary files at all...  This confirms the suspicion of @yahalom5776 in #374

pombredanne:
Note that at least that the reported score is very low... e.g. 5/100 as it should

pombredanne:
This is now fixed in develop.

Issue #374, opened 2016-11-24
Ensure that extracted archives are not double scanned


pombredanne:
from #267 by @yahalom5776
> I would just like to add another data point to the discussion. My suspicion (not verified so far) is that the scan of large archive files is the culprit. In my test case I had many nested archives extracted by extractcode and when I ran the scan I realized the archives were still there and ScanCode probably processed them.

> So could we perhaps improve the situation also by adding a new functionality to extractcode: "Delete archive(s) after successful extraction"?

pombredanne:
@yahalom5776 the reason why this is happening is because of this code here which is falling back always on attempting to squeeze some text out of binaries
https://github.com/nexB/scancode-toolkit/blob/6aa76df41304330381b6c77831e36654ce54130e/src/textcode/analysis.py#L96

I think a saner approach would be to exclude archives, compressed files and media/image files as they would not contain any interesting provenance clues. Would this work?

pombredanne:
This is happening indeed and this is a bug. @chinyeungli reported a false license detected in #396 too.

pombredanne:
This is now fixed in develop.

PR #404, opened 2016-12-14
#374 #396 Do not scan archives and compressed files

pombredanne:
This deals with #374 where @yahalom5776 reported that extracted archives were likely also scanned in their unextracted version and #396 where @chinyeungli  reported a false positive license detection in a  a jar file (which is a zip).

The new approach is to only scans things in a file if it `contains_text` which is an attribute of the typecode file type registry. Archives, compressed files, some common package archives (e.g. RPMs, deb) and a few other types (such as PNG) are now considered as not having any text content and `get_type().contains_text` returns False.  This is then used as a test in `textcode.analysis` to return no text if a file does not `contains_text`. This is also used as an early test in  `licensedcode.query` to return an empty query in these cases. The typecode detection has been refined and several new tests added. Same for the textcode text extraction.
As a notable side effect, some non-compressed archives such as AR archives (e.g. used mostly for .lib and .d static libraries) are no longer scanned unless their payload is extracted first with extractcode (in which case the payload will be scanned)


pombredanne:
If there is no objection since all tests pass, I will merge this.

Issue #409, opened 2016-12-19
ScanCode does not detect bpmn-io license


JonoYang:
A rule for the bpmn-io license is needed.

pombredanne:
@JonoYang Thanks. It is this: http://bpmn.io/license/ ?
And if so is it not detected at all, or rather detected as MIT?

JonoYang:
It is not detected at all. I ran ScanCode on https://github.com/bpmn-io/bpmn-js, only a JSON license was found in the license file.

JonoYang:
Merged in #410

PR #410, opened 2016-12-19
#409 Added LICENSE file and test for bpmn-io license

JonoYang:
The bpmn.io license was not being detected at all, despite similarities with the MIT license. A new LICENSE file and test were created to detect the bpmn.io license.


pombredanne:
LGTM.. Thank you!

Issue #411, opened 2016-12-20
json2csv.py crashes when an asserted license's license value is null


JonoYang:
In the case where the "license" name of a license in "asserted_licenses" is null, as shown below, json2csv.py crashes as a string is expected, but a NoneType was found. 
```
"asserted_licenses": [
  {
    "license": null,
    "url": "license.txt",
    "text": null,
    "notice": null
  }
]
```

JonoYang:
Merged in #412

PR #412, opened 2016-12-20
#411 Return an empty string for package license name values that are null

JonoYang:
I have modified the list comprehension that collects the license names from a package's "asserted_licenses" key to return an empty string when a license's name is null. I have also created a new test for this issue.

### DIFF
--- a/etc/scripts/json2csv.py
+++ b/etc/scripts/json2csv.py
@@ -163,7 +163,7 @@ def flatten_scan(scan):
                             pack[nk] = '\n'.join(val)
                         if k == 'asserted_licenses':
                             # All license names are to be joined in a multi-line cell
-                            licenses = [license_info['license'] for license_info in val]
+                            licenses = [license_info.get('license', '') or '' for license_info in val]
                             pack[nk] = '\n'.join(licenses)
                     else:
                         pack[nk] = ''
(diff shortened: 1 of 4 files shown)


etc/scripts/json2csv.py
```diff
@@ -163,7 +163,7 @@ def flatten_scan(scan):
                             pack[nk] = '\n'.join(val)
                         if k == 'asserted_licenses':
                             # All license names are to be joined in a multi-line cell
-                            licenses = [license_info['license'] for license_info in val]
+                            licenses = [license_info.get('license', '') or '' for license_info in val]
```
pombredanne: The or in `license_info.get('license', '') or ''` is not needed ....
Use instead IMHO:
`license_info.get('license', '')`
this is enough

JonoYang: My first modification was to use `license_info.get('license', '')`, but this returns `None` instead of an empty string when `license` is `null`. I had to use the or to use the empty string when `None` was returned.

pombredanne: good point! so for simplicity and clarity you could use instead this:
`license_info.get('license') or ''`
e.g. if license is really a `None` or not present, you will get a empty string

JonoYang: I'll make the change.

pombredanne:
Thanks!

Issue #16, opened 2015-07-05
Check the tree view for long path (deep path) and long file/directory name on Windows


pombredanne:
reported by @chinyeungli based on bugs reported in AboutCode https://github.com/dejacode/about-code-tool/issues/143

Windows needs special care to handle deep paths.  We should test and verify that ScanCode support such deep path

pombredanne:
Since we now use the workbench I am closing this as the treeview of the HTML output is not a priority

PR #414, opened 2016-12-21
#413 no crash on weird file names

### LINKED ISSUE #413: KeyError exception in multiprocessing module
Using ScanCode 1ec43b7 on Linux with Python 2.7 I get:

      File "/var/lib/jenkins/jobs/scancode/workspace/tools/scancode/bin/scancode", line 9, in <module>
        load_entry_point('scancode-toolkit', 'console_scripts', 'scancode')()
      File "/var/lib/jenkins/jobs/scancode/workspace/tools/scancode/local/lib/python2.7/site-packages/click/core.py", line 664, in __call__
        return self.main(*args, **kwargs)
      File "/var/lib/jenkins/jobs/scancode/workspace/tools/scancode/src/scancode/utils.py", line 64, in main
        standalone_mode=standalone_mode, **extra)
      File "/var/lib/jenkins/jobs/scancode/workspace/tools/scancode/local/lib/python2.7/site-packages/click/core.py", line 644, in main
        rv = self.invoke(ctx)
      File "/var/lib/jenkins/jobs/scancode/workspace/tools/scancode/local/lib/python2.7/site-packages/click/core.py", line 837, in invoke
        return ctx.invoke(self.callback, **ctx.params)
      File "/var/lib/jenkins/jobs/scancode/workspace/tools/scancode/local/lib/python2.7/site-packages/click/core.py", line 464, in invoke
        return callback(*args, **kwargs)
      File "/var/lib/jenkins/jobs/scancode/workspace/tools/scancode/src/scancode/cli.py", line 312, in scancode
        diag, timeout, max_memory)
      File "/var/lib/jenkins/jobs/scancode/workspace/tools/scancode/src/scancode/cli.py", line 409, in scan
        result = scanned.next()
      File "/var/lib/jenkins/jobs/scancode/workspace/tools/scancode/local/lib/python2.7/site-packages/click/_termui_impl.py", line 238, in     
        return next(self.iter)
      File "/var/lib/jenkins/jobs/scancode/workspace/tools/scancode/src/scancode/cli.py", line 41, in wrap
        return func(self, timeout=timeout or 1e10)
      File "/usr/lib64/python2.7/multiprocessing/pool.py", line 668, in next
        raise value
    KeyError: 'sha1'

The command line used was

    scancode --timeout 180 -n 2 -f html . /var/lib/jenkins/jobs/scancode/workspace/results/github.com/remy/nodemon/2016-10-06_2cd85b1/nodemon-scancode-1ec43b7.html

and the project scanned was https://github.com/remy/nodemon/commit/2cd85b1f609e9749d19afcdbd71368e6ab063f9d.

pombredanne:
This should fix #413 but the test may fail at first on Windows or Mac.


pombredanne:
AS expected the test fail on Windows (cannot extract the test archive with weird names using the plain tar from Python) and on Mac (weird and slightly different behaviour on file types and mime types). That will be easy to fix.

sschuberth:
@pombredanne Let me know if I can help to resolve the remaining issues.

pombredanne:
@sschuberth I have been working on a fix which just a tad more involved than I thought. ETA likely later today

pombredanne:
This latest push should fix all the remaining bits now.
The main (significant) change is handling of unicode paths when extracting archives using a more robust transliteration to ascii and several update to the handling of various paths transformations such as resolution and transformation to "safe paths".

I may make a few adjustments if some tests  fail on Mac or Windows

pombredanne:
I am still getting some test failures on Windows. This requires a tad more involved surgery in handling paths. I am also hitting some issues with #16   in https://ci.appveyor.com/project/nexB/scancode-toolkit/build/337/job/jxkfds6dp1mkf57q (even though it does not show in the logs, this is the underlying issue...) I may set this issue with windows long paths handling aside to close this issue quick and come back on that right after. There is a solution there: https://github.com/nexB/attributecode/blob/be36cc0668e5db3f47a7a3d0a38fe1f1af8f3674/src/attributecode/util.py#L400 and also in the Python 3 pathlib (and its backport to Python 2 https://github.com/mcmtroffaes/pathlib2/  )

The key issue that while the stdlin and other libraries help handle correctly paths correctly on each OS, there is no clean solution that helps to handle paths from other OSes on a given OS (e.g. handle windows paths on POSIX and the other way around) and scancode does most of it correctly except for a few corner cases that we are hitting here. We do have this needs as archives can contain Windows or POSIX paths and must be handled correctly on all OSes and the main issue is that of path style detection, lengths, normalization, joining and filesystem encodings. (nothing big of its own, but borderline intractable when taken together). 

`<rant> I envy the warm and fuzzies to have to code only for a single OS </rant> `

sschuberth:
> I envy the warm and fuzzies to have to code only for a single OS.

As a side note, so far I've found Python to be quite bad at abstracting platform-specific *semantics*. Sure, you get a common API, but that's just a superficial translation leaving out the bigger workflow picture.

pombredanne:
@sschuberth I would be interested if you can elaborate a bit, just curious :)

On my side here the issue is not so much Python, but rather some (fixable) mistakes in ScanCode and the fact that we scan arbitrary files (and especially archives) coming from arbitrary OS, filesystem path encodings and file encodings, that are often providing no clue about what this encoding could be. And we process that using eventually multiple tools in some case native tools (especially on the file type and archive extraction) and on multiple OSes that each have their own conflicting filesystem path encodings and limitations: each of Mac, Linux and Windows have a very different take and each native tool we use makes it own interpretation on top of this. If scanning was supported only on a single OS, things would be much simpler.  With that said there are also self-inflicted wounds ... Not using unicode throughout in ScanCode as a default is a source of further issues and entirely my bad and a part of history too... This is something that I will address soon and will pave the way to a Python 3 migration.

The crux of some of these issues is therefore the combination of all of the possible cases:
`scanning OS(Windows, Linux, MacOSX)` X `source OS(Windows, Linux, MacOSX)` X `Python(Windows, Linux, MacOSX)` X `7Zip(Windows, Linux, MacOSX)` X `libarchive(Windows, Linux, MacOSX)` X `libmagic(Windows, Linux, MacOSX)`

AFAIK, things would not be any better with Java or Ruby to handle all these variations especially the combo of native and multiple source and scanning OSes... For instance  in particular libarchive does not handle correctly all archive quirks on all source OSes and scanning OSes and was designed on FreeBSD/POSIX first and not Windows. Same for sevenzip which was squarely designed for Windows and ported on POSIX.  Each make assumptions about scanning/running OSes which are not consistent but both are needed to get a comprehensive support of common archive types across OSes as each handles some archives better than the other ins some corner cases. 

With all that said, we are getting and I should have this stable today as there are only a few minor bits remaining to fix.

pombredanne:
We are almost there. Only 3 Windows tests failures and another 10 trivial tests to fix left (also on Windows)...

sschuberth:
Thanks for your huge efforts in this!

pombredanne:
@sschuberth This was badly needed. The worst thing is when a scan cannot carry on or if it were to skip some file silently. It would be the end of the world :P

pombredanne:
We are down to the 10 trivial failures only on Windows... tackling this now

pombredanne:
At last Windows is all green now. As soon as all is passing I will merge in develop

pombredanne:
Alright. Merging now! at last.

sschuberth:
@pombredanne Since you were asking, let me comment on your post, even if a bit off-topic. We could continue by mail afterwards :-)

> archive extraction) and on multiple OSes that each have their own conflicting filesystem path encodings and limitations: each of Mac, Linux and Windows have a very different take and each native tool we use makes it own interpretation on top of this.

My point is that a cross-platform scripting language like Python should make it easy to solve these kind of problems by abstracting semantics, not only syntax (in the form of platform-spefic APIs). It should make it easy to identify paths that are not portable across platforms, and it should make it hard to construct such paths.

> Not using unicode throughout in ScanCode as a default is a source of further issues and entirely my bad and a part of history too...

This is also something where Python could have made it easier by not having both `str` *and* `unicode` types, and in particular strings not being Unicode. The way e.g. Java does handle strings is more sane, IMO.

> AFAIK, things would not be any better with Java or Ruby to handle all these variations especially the combo of native and multiple source and scanning OSes...

Except Unicode strings maybe, see above ;-)

> For instance in particular libarchive does not handle correctly all archive quirks on all source OSes and scanning OSes and was designed on FreeBSD/POSIX first and not Windows.

That's one of my points: If you use a cross-platform scripting language, how can you design something for only a subset of the supported OSes? That's making nonsense out of cross-platform development.

pombredanne:
Thanks. You bring all good points. Frankly the unicode issue is partly historic as ScanCode codebase was started in the early Python 2.5 days. And this is somewhat a self-inflicted wound to some degree and a "technical debt" we will get rid of soon. There are still corner cases that are somewhat unique when you both deal with files and archives at a lower level that come from arbitrary places with eventually unknown file system encodings and when you want to have a consistent behavior on POSIX Linux and Mac and Windows.  And on the topic of cross platform, we do resort to native libraries that have no counterpart otherwise in pure Python (and no counterpart either in any other interpreted language)  and that's both a curse and a blessing: a blessing because rewriting these from scratch would be several immense projects of their own and a curse because we have to deal with their quirks for corner cases.

The point I was making mostly is that Python or not, supporting multiple runtime OS for ScanCode is hard.

If only Linux was supported, handling arbitrary and unknown encodings for paths and archives created on unknown OSses and file systems would be a much simpler thing.  

So this is not so much an issue with Python but more about the OS on which ScanCode runs. The core of the issue is that files and archives being scanned come from any OS and filesystem with any possible weird encodings and each OS supported to run ScanCode has specificities and limitations in its ability to handle or not handle at all some of these weirdnesses.

ScanCode is providing a mostly universal and consistent support for these features on Windows, Mac and Linux and this is what makes it difficult: to the best of my knowledge this is something that very few tools --if any-- do support in any programming language for these lower level things.

Issue #417, opened 2016-12-28
The `--quiet` option doesn't suppress progress messages


akaihola:
The `scancode --help` instructions advise to use the `--quiet` option to skip printing progress messages:

    $ ./scancode --help
    ...
      --quiet                  Do not print progress messages.
    ...

However, that option has no effect. I still get the progress messages printed on stdout before the JSON output. This prevents me from running `scancode` as a subprocess and capturing its output as JSON:

    $ ./scancode --quiet -l /tmp/LICENSE
    Scanning files for: licenses with 1 process(es)...
    Building license detection index...Done.
    Scanning files...
    Scanning done.
    Scan statistics: 1 files scanned in 3s.
    Scan options:    licenses with 1 process(es).
    Scanning speed:  0.57 files per sec.
    Scanning time:   1s.
    Indexing time:   1s.
    Saving results.
    {... result JSON ...}

pombredanne:
Thank you for reporting this! This is closely related to #388 from @sschuberth that is fixed but still needs work to implement the following scheme:

1. if no output file is specified, send the scan results to `stdout` 
2. with no option: messages, progress bar and summary sent to `stderr`. 
3. with the option `--quiet` no messages, no progress bar, no summary 
4. with the option `--verbose` messages, per-file log and  summary sent to `stderr`

The key would be to always send messages to `stderr` and the scan output to `stdout`
@akaihola would this work for you alright?

akaihola:
@pombredanne, yes, that would be perfect. 

Actually, before noticing the `--quiet` option, I was expecting to be able to read the JSON output from stdout and discard messages in stderr.

I'm currently working around this by doing the reverse: specify `/dev/stderr` as the output file, read JSON output from stderr and discard messages in stdout.

pombredanne:
@akaihola No nee to work around... I am working on a  proper fix and will push it in a few

pombredanne:
@akaihola @sschuberth  The branch `388-417-consistent-stderr-stdout-outputs` contains a proper implementation of the proposed update to outputs redirection for here and #388.
Feedback welcomed.

pombredanne:
@akaihola Just curious you said:
> This prevents me from running scancode as a subprocess and capturing its output as JSON

what is your use case there? I am always interested by new ways to use scancode!

akaihola:
@pombredanne Our use case is a Python script which
[...]

Issue #388, opened 2016-12-02
Still no --verbose output on Jenkins CI


sschuberth:
Although cf024ac88462b79fc5b7a00207d7f8fcc8a854e7 says "--verbose output is always send to stdout", I still do not see the individual files being scanned printed to the console / log when running ScanCode as part of a Jenkins CI job.

pombredanne:
@sschuberth sorry for that, I thought I fixed it, but I did not add a test. There is something fishy happening in the guts of the click ProgressBar. I think I will just skip using a progress bar subclass ProgressLogger for verbose output entirely i.e. not use at all https://github.com/nexB/scancode-toolkit/blob/3d89c2871543a61576cd485e2335da93e96a8c98/src/scancode/utils.py#L83

sschuberth:
On the other hand, it probably does not make sense to duplicate too much code for verbose vs. non-verbose output, and I'd be curious what's the cause for ProgressLogger not showing output for a non-tty stdout.

pombredanne:
Actually not using ProgressLogger is likely to mean less code than using it... so no worries there

pombredanne:
So after all the issue what that ProgressLogger was not forcing itself to be visible at all times (with the `hidden` attribute) which this fixed: https://github.com/nexB/scancode-toolkit/commit/376cb9dbe77e0df2caf5fdbd6b1a3bba4c9bcea3#diff-7412a01bdb89dd7c7021bf5530be25a7R97

The logic is that `--verbose` forces the output to go to `stdout`. There is one case where this is may be awkward: If a CLI run does not specify an output file then the scan results are also printed to `stdout` and when combined with a `--verbose` option then the per-file progress and the scan results will both go to `stdout` which can be a tad messy but is corner case: the general approach should be to use an explicit file.

sschuberth:
Another alternative would be to always send "primary" output (i.e. the scan results) to *stdout*, but all other output (warnings, errors, progress, which includes output from `--verbose`) to *stderr*. That would sort of solve the problem when specifying `--verbose` but no output file: While it still looks messy / interleaved in the terminal, you could easily separate the streams, e.g. by redirecting only *stderr* to a file.

pombredanne:
Of course! And this much much cleaner!
So the behavior should be:
1. if no output file is specified, send the scan results to `stdout` 
2. no option: messages, progress bar and summary sent to `stderr`. 
3. `--quiet` no messages, no progress bar, no summary 
[...]

PR #418, opened 2016-12-28
#388 and #417 Ensure that a --quiet run is really quiet

pombredanne:
* and also ensure that all messages are sent to stderr and not stdout

This implement these changes:

1. if no output file is specified, send the scan results to `stdout` 
2. with no option: messages, progress bar and summary sent to `stderr`. 
3. with the option `--quiet` no messages, no progress bar, no summary 
4. with the option `--verbose` messages, per-file log and  summary sent to `stderr`

The key would be to always send messages to `stderr` and the scan output to `stdout`

### DIFF
--- a/tests/scancode/test_cli.py
+++ b/tests/scancode/test_cli.py
@@ -464,3 +464,22 @@ def test_scan_can_handle_licenses_with_unicode_metadata(monkeypatch):
     result = runner.invoke(cli.scancode, ['--license', test_dir, result_file], catch_exceptions=True)
     assert result.exit_code == 0
     assert 'Scanning done' in result.output
+
+
+def test_scan_quiet_does_not_echo_anything(monkeypatch):
+    monkeypatch.setattr(click._termui_impl, 'isatty', lambda _: True)
+    test_dir = test_env.extract_test_tar('info/basic.tgz')
+    runner = CliRunner()
+    result1_file = test_env.get_temp_file('json')
+    result1 = runner.invoke(cli.scancode, ['--quiet', '--info', test_dir, result1_file], catch_exceptions=True)
+    assert result1.exit_code == 0
+    assert not result1.output
+
+    # also test with an output of JSON to stdout
+    runner2 = CliRunner()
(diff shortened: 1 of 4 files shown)


pombredanne:
@sschuberth @akaihola .... here it is for your review!

pombredanne:
Unless there is an objection, I will merge this in soon enough... at the latest tomorrow.

sschuberth:
Looks good to me, just some nit-picking: You're using comments like "on terminal stderr" or "the terminal stdout". Strictly speaking, you don't know at that time  whether a terminal is attached to stdout / stderr, they could also be redirected e.g. to a file. I'd simply say "to stderr" or "to stdout" instead.

akaihola:
Just one detail in the test suite in case you want to strive for perfection there... :)
The `test_scan_quiet_does_not_echo_anything()` test runs `invoke()` for two different runners. If the first one fails, the second one isn't tested at all. I would divide these into two separate tests.

See also my note about the progress bar in #417.

pombredanne:
@sschuberth thanks... re `to stderr" or "to stdout" ` ... I will update these

tests/scancode/test_cli.py
```diff
@@ -464,3 +464,22 @@ def test_scan_can_handle_licenses_with_unicode_metadata(monkeypatch):
...
+    assert not result1.output
+
+    # also test with an output of JSON to stdout
+    runner2 = CliRunner()
+    result2 = runner2.invoke(cli.scancode, ['--quiet', '--info', test_dir], catch_exceptions=False)
+    assert result2.exit_code == 0
+
+    # outputs to file or stdout should be identical
+    result1_output = open(result1_file).read()
+    assert result1_output == result2.output
```
pombredanne: @akaihola You quite rightly suggest to split the test `test_scan_quiet_does_not_echo_anything` in tow functions, but there is a reason why they are in a same function: this assert checks that both outputs to a file or to stdout are identical... I am not sure I could achieve this but splitting this in two tests?

pombredanne:
@sschuberth @akaihola I have addressed your comments and feedback with the latest commit

pombredanne:
Ok, merging now then. We can handle subsequent updates in another PR.

Issue #253, opened 2016-04-13
Recognize Python packages (wheels, eggs, various dist as archives or installed)


balusarakesh:
scancode must be able to detect Python packages used by in a project by parsing files like `setup.py`, `.whl` files, `.eggs` files, `requirements.txt` or `pip_requirements.txt`.

sschuberth:
Just for the record, while PR #421 is thankfully merged, support for recognizing Python packages based on `*requirements*.txt` files is still missing.

pombredanne:
@sschuberth indeed. There are several limitation in the current code.

1. support for requirements is one: https://github.com/landscapeio/requirements-detector may be what the doctor ordered, may be together with `pip` proper used as a library

2. support for parsing `setup.py` is somewhat weak too. These could be some pointers:
* https://github.com/msabramo/flake8-exact-pin/blob/master/flake8_exact_pin.py
* https://gitlab.com/abre/setup-freeze/blob/master/setup_freeze.py
* https://github.com/msabramo/setuppycheck/blob/master/setuppycheck.py
* https://github.com/inveniosoftware/requirements-builder/blob/master/requirements_builder/requirements_builder.py
* https://github.com/specialunderwear/setupreader/blob/master/setupreader.py
* https://github.com/pombredanne/setupreader
* https://github.com/ashwoods/python-packr/blob/master/setup_parser.py
* https://github.com/openembedded/openembedded-core/blob/f3e753372baac43d0921186340cf260df056de20/scripts/lib/recipetool/create_buildsys_python.py#L181
* https://github.com/yosi-dediashvili/SubiT/blob/ee154c43c3c0170e82750d428688b4d441f95134/build/setup.py
* https://github.com/mk-fg/fgtk/blob/master/dev/distribute_regen
* https://github.com/noisy/parse_setup.py
* http://stackoverflow.com/questions/24236266/how-to-extract-dependencies-information-from-a-setup-py
* http://stackoverflow.com/questions/27790297/parse-setup-py-without-setuptools
* https://github.com/andymckay/baked/blob/master/baked.py
* https://github.com/mgood/quipu/blob/master/quipu.py
* https://github.com/newville/asteval

3. Parsing actual packages as installed could likely benefit from code/functions existing in `distutils`, `pip` and `setuptools`

pombredanne:
The basic is there. Addintional updates will be for post 2.0

pombredanne:
[...]

PR #421, opened 2016-12-30
#253 python packages

pombredanne:
Add initial support to detect Python packages for #253
 * handle some level of detection using setup.py, wheels and eggs metadata

### DIFF
--- a/src/packagedcode/pypi.py
+++ b/src/packagedcode/pypi.py
@@ -1,0 +1,85 @@
+#
+# Copyright (c) 2015 nexB Inc. and others. All rights reserved.
+# http://nexb.com and https://github.com/nexB/scancode-toolkit/
+# The ScanCode software is licensed under the Apache License version 2.0.
+# Data generated with ScanCode require an acknowledgment.
+# ScanCode is a trademark of nexB Inc.
+#
+# You may not use this software except in compliance with the License.
+# You may obtain a copy of the License at: http://apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software distributed
+# under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+#
+# When you publish or redistribute any data created with ScanCode or any ScanCode
+# derivative work, you must accompany this data with the following acknowledgment:
+#
+#  Generated with ScanCode and provided on an "AS IS" BASIS, WITHOUT WARRANTIES
+#  OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+#  ScanCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  ScanCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/scancode-toolkit/ for support and download.
+
+from __future__ import absolute_import
+from __future__ import print_function
+
+import logging
+import re
+
+from packagedcode.models import AssertedLicense
+from packagedcode.models import PythonPackage
+
+"""
+Handle Python PyPi packages
+"""
+
+
+logger = logging.getLogger(__name__)
+# import sys
+# logging.basicConfig(level=logging.DEBUG, stream=sys.stdout)
+# logger.setLevel(logging.DEBUG)
+
+
+def get_attribute(setup_location, attribute):
+    """
+    Return the value specified for a given 'attribute' mentioned in a 'setup.py'
+    file.
+    Example :
+    setup(
+        name='requests',
+        version='1.0',
+        )
+    'requests' is returned for the attribute 'name'
+    """
+    setup_text = open(setup_location, 'rb').read()
+    setup_text = setup_text.replace('\n', '')
+    # FIXME Use a valid parser for parsing 'setup.py'
+    values = re.findall('setup\(.*?'+attribute+'=[\"\']{1}.*?\',', setup_text)
+    if len(values) > 1:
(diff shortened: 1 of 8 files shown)


sschuberth:
From reading the code, my understanding is that this relies on the meta-data specified as part of `setup.py` an reports whatever license is declared there. More concretely, this does *not* download the source code of a Python package to run ScanCode over it. This should be made very clear as this means cases where the license from the meta-data is wrong compared to the `LICENSE` file in the source code will not get detected. Moreover, licenses from transitive dependencies are not taking into account.

@pombredanne Are my statements correct?

pombredanne:
@sschuberth the goal of this PR (and of all the code in the `packagedcode` package which is invoked with the `-p` or `--package` scan is to:

1. **detect the presence of a package** in a codebase based on its manifest, it file or archive type. Typically it would be a third party package but it may be yours. Here this is for Python and they can exists in multiple forms: 
1.a) a **source checkout** (or some source archive such as what is called a source distribution or an `sdist`) where the presence of a `setup.py` or some `requirements` file is the key marker. For Maven it would be a `pom.xml` or a `.gradle` file, for Ruby a `Gemfile` or `Gemfile.lock`, and so on, eventually covering **ALL** the packages formats/types that are out there. 
1.b) an **installable archive** such as a Pypi `wheel` or `egg`, a Maven `jar`, a Ruby `gem`, a `.nupkg` for a Nuget, etc... Here the type, shape and name structure of an archive as well as possibly some its file content is the key marker for detection.
1.c) the **package as-installed** such as when you `pip install` or `bundle install` or `npm install` one or packages. Here the key markers may be some combo of a directory layout and specific files (such as the metadata installed with a Python `wheel`, a `vendor` directory for Ruby, some `node_modules` tree of sorts for npm, a `.rpm` or `.deb` Linux package, etc.

2. **parse and collect the package manifest(s)** metadata. For Python, this means extracting name, version, authorship, asserted licensing and declared dependencies as found in the any of the package files (`setup.py` and/or `requirements` file(s) and/or any of the `*-dist-info` dir files such as a `metadata.json`). Other package formats have their own metatada more or less comprehensive (e.g. `.nuspec`, `package.json`, `bower.json`, Godeps, etc...)

These two will be then injected in a `packages` scan section. 

What code in `packagedcode` is not meant to do:

A. download packages from a thirdparty repository: there is code I have in an unpushed repo that is specifically dealing with this and also handles collecting the metadata as served by a package repository (which are in most cases --but not always-- the same as what is declared in the manifest). I will publish this sometimes in Q1 hopefully. 

B. resolve dependencies: the focus of scancode is on a purely static analysis that does not rely on any network access for running by design. To scan for actually used dependencies the process is to instead scan for an as-built or as-installed or as-deployed application where the dependencies have already been provisioned and installed and there scancode would detect these. I also have an unpushed prototype for a dynamic multi-package dependencies resolver that actually runs live the proper tool to resolve and collect dependencies (e.g. effectively running Maven, bundler, pip, npm, gradle, bower, go get/dep, etc). This will be a tool separate from scancode as this requires having several/all package managers installed (and possibly multiple versions) and does run things and accesses the network. It may be best exposed as a web service that can take in a manifest and package and run safely the dep resolution in an isolated environment (e.g. a chroot jail or docker container) and return the collected deps.

C. match packages (and files) to actual repositories or registries, e.g. given a scan detecting packages actually looking them up in a remote package repository and then using A. and/or B. additionally if needed. I have unpushed code for this too that will eventually land on Github and handles also building an index of actual registries/repositories and matching using hashes and fingerprints.

Now the goal of all is simple: detect a package, determine its deps, detect its asserted license (at the metadata level) and its actual licenses (at the scan level, and these may differ, conflict at times)

So to finally answer your questions:

> More concretely, this does not download the source code of a Python package to run ScanCode over it.

Correct. The assumption with ScanCode proper (aside of the other not-yet-published tools that I mentioned
above) is that the deps have been fetched in the code you scan if you want to scan for deps. Packages will be detected with their declared deps but the deps will neither be resolved nor fetched (but as a second step I plan also to have a check to verify that all the declared deps are also present in the scanned code as detected packages too) 

> This should be made very clear as this means cases where the license from the meta-data is wrong compared to the LICENSE file in the source code will not get detected. 

Both the metadata and the file level licenses (such as a header comment or a `LICENSE` file of sorts) are be detected by scancode here: the interesting thing thanks to this is that eventual conflicts can then be analyzed and a deduction should be doable automatically: given a scan with packages and licenses and copyrights, do the metadata match the actual detected licenses? if not this could be reported as some "error" condition... Furthermore, this could be refined based on classification of the files: a package may declare an `MIT` license and use a GPL-licensed build script. By knowing the build script is a build script we could then report that the GPL detected in such script is not conflicting with the overall MIT license of the package. 

> Moreover, licenses from transitive dependencies are not taking into account.

If the transitive dependencies have been resolved and their code present in the codebase, then they would be caught by a static ScanCode scan and eventually scanned both for package metadata or actual license detection. There are some caveats that would need to be dealt with of course as some tools (e.g. Maven) may not store locally (e.g. side-by-side with a given checkout) the corresponding artifacts/Jars and use instead a `~/user` "global" dot directory to store a cache.

Beyond this, the goal is to use in the future the other tools that I mentioned above for actual dependency resolution of a single package or a complete manifest

sschuberth:
Thanks @pombredanne for the sophisticated write up. I believe it's too valuable to just sit here, so would you mind adding it to a `README.md` in the `packagedcode` directory?

pombredanne:
@sschuberth good point! Will do today.

pombredanne:
@linexb I wonder if you could help there. The code needs some love and has many tests failing.

pombredanne:
@linexb Thanks.... Looking good!

Issue #405, opened 2016-12-15
truncation of copyright


pkunz:
For this file header, 

> /*
 * Copyright 2007-2010 the original author or authors.
 * 
 * Licensed under the Apache License, Version 2.0 (the "License"); you may not u

the scan resulted in 

> Copyright 2007-2010 the

pombredanne:
Thanks. That's a good catch

JonoYang:
Fixed in #432

PR #432, opened 2017-01-05
#405 Rule to capture \`Copyright 2007-2010 the original auth…

JonoYang:
A new rule has been added to capture `Copyright 2007-2010 the original author or authors.`.


pombredanne:
Looking good.... No need to wait for the last Ci slave to finish as this testing the same things twice in this case. Thanks!

Issue #384, opened 2016-12-01
Limit number of urls and emails reported in a single file 


pombredanne:
At times some file -such as a test file- may contain a large number of emails and/or urls. Millions of them. We should limit the number of returned items per file in these cases to a sensible number like a few 100's and report a warning that not everything was scanned...

sam09:
Anyone working on this? I would love to give it a try

pombredanne:
@sam09 Go for it . Thanks!

sam09:
I think we could probably limit the list of emails and urls to a 1000?
The get_urls and get_emails method can be modified to limit the number.
https://github.com/nexB/scancode-toolkit/blob/develop/src/scancode/api.py.

pombredanne:
@sam09 anything with more than a 100 emails or urls is likely a data file with a huge amount of these. Actually 50 would already be a huge lot. 

One consideration is how to report such cases of partial detection? some warning in the scan errors? some what to state like: 123673 URLs were detected in this file. Only the first 100 are reported... TBD. Suggestions  are welcomed.

Also it could make sense to make the max number a CLI option and have a default to 50 or 100 for instance. In which case the issue of reporting partial detection may be a lesser issue.

pombredanne:
and also since both urls and email would need a limit, having a single wrapper function that works for both would the way to go imho.... e.g. some generic function that accepts an iterable and yields a truncated iteable.   It may even exists in the stdlib itertools.

pombredanne:
e.g. https://docs.python.org/2.7/library/itertools.html?highlight=itertools#itertools.islice

pombredanne:
so with this and a `functools.partial` you can create simple wrappers and use at first a hardcoded limit of say 50, and later a CLI option for the limit, where using `0` would mean return all.

inishchith:
@pombredanne I'd like to work on this , can you assign this to me ?

majurg:
@inishchith I believe you can assign yourself :)

majurg:
@inishchith Assignees can only be apart of our GitHub org for now, but you are free to contribute to this (and any other) issue. :)

inishchith:
Oh , I'll it a try  👍

inishchith:
@majurg do i restrict the number to 100 per file for now ? and report as per above comments by @pombredanne ?

majurg:
@inishchith 100 per file is a good place to start for now.

inishchith:
[...]

PR #435, opened 2017-01-06
cli: Clarify how default command line options are enabled

sschuberth:


pombredanne:
Good catch. This was not enough indeed. In fact I truly think that all scans may be best always enabled by default and that if you pass a specific scan option, then only that scan would be used.
Now to for this to be practical #384 would need to be there first as otherwise there are cases not so uncommon where you end up with data files with 1000's of urls or emails and that does not make sense to capture them all by default.

pombredanne:
Merging.... Thanks!  .... btw @sschuberth do you want commit rights directly on the repo? that could make it easier? your call! you definitely earned it in any case.

sschuberth:
> do you want commit rights directly on the repo?

I'd indeed appreciate that as it would simply the workflow a bit by me creating branches in your repo instead of my fork.

pombredanne:
@sschuberth great. Invite sent! :bowing_man:

pombredanne:
of note: force pushing is disabled/protected (for everyone) on the `master` and `develop` branches for sanity. This is fair game in any other branches otherwise.

sschuberth:
Sure thing!

pombredanne:
This is just an FYI... force pushing would fail if attempted on `master` and `develop` :)

pombredanne:
even by me... I do not trust me in particular ;)

Issue #408, opened 2016-12-19
Change the returned license category from `Attribution` to `Permissive`


chinyeungli:
The scancode toolkit is currently return `Attribution` as one of the license categories. However, we do not have a License Category of `Attribution`, it should be `Permissive`

pombredanne:
This makes sense and is easy enough. Would you mind a pull request?

chinyeungli:
I can do that

pombredanne:
@chinyeungli I think we are all set now. Thank you!

PR #440, opened 2017-01-10
#408 Change license category from Attribution to Permissive

pombredanne:
* the category changes are:
  * category: "Attribution Restricted" changed to: "Free Restricted"
  * category: "Atribution" changed to: "Permissive"


pombredanne:
The reason why the tests pass on Windows and not on Linux/Mac is because the py.test config skips `etc/scripts` but only on Windows... I will fix this and then the json2csv expectations can be fixed for Windows alright.

pombredanne:
@JonoYang I rebased this branch and forced pushed.... You may need to do some manual cherry picking on your side. The new CI setup will show you early in the build if etc/scripts tests failed and how.

pombredanne:
@chinyeungli @JonoYang The commit I added should fix all the remaining json2csv issues. @chinyeungli I think you had try to update the expected CSV files with LibreOffice or similar and it had damaged the dates to convert them to spreadsheet dates. A similar thing may have happened to license score... Hence the failures. I fixed the tests accordingly

pombredanne:
The tests runs are looking good and on their way to all pass

chinyeungli:
thanks

pombredanne:
@chinyeungli Thank you!

PR #441, opened 2017-01-10
Work around spdx-tools assuming a "str" file instead of "unicode"

sschuberth:


pombredanne:
@sschuberth Thanks... let me review this. I had made only some quick tests and without keeping much history of it and only on Linux.

pombredanne:
It is likely either because of https://github.com/sschuberth/scancode-toolkit/blob/daf67e313f0b47e39e911bd47782b8f9224b4322/src/scancode/cli.py#L28 and/or StringIO returning bytes and encoding OK in this case because of the default encoding.... Not so sure. But you should add some tests with and without non-ASCII content in the scans here to validate the behavior IMHO.

pombredanne:
Thanks! we will still need some tests though ;)

PR #449, opened 2017-01-12
Resolve various path issues with SPDX output

sschuberth:
This fixes two issues at once: By reusing the existing SHA1 from `info`, we can easily skip directories when writing SPDX file entries as for these the SHA1 is empty.

### DIFF
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -312,6 +312,10 @@ def scancode(ctx, input, output_file, copyright, license, package,
         license = True
         package = True
 
+    # A hack to force info being exposed for SPDX output in order to reuse calculated file SHA1s.
+    if format in ('spdx-tv', 'spdx-rdf'):
+        info = True
+
     scans_cache_class = get_scans_cache_class()
     try:
         files_count, results = scan(input_path=input,
@@ -678,7 +682,7 @@ def save_results(files_count, scanned_files, format, input, output_file):
             output_file.write(unicode(json.dumps(meta, separators=(',', ':'), iterable_as_array=True, encoding='utf-8')))
         output_file.write('\n')
 
-    elif format == 'spdx-tv' or format == 'spdx-rdf':
+    elif format in ('spdx-tv', 'spdx-rdf'):
         from spdx.checksum import Algorithm
         from spdx.creationinfo import Tool
         from spdx.document import Document, License
@@ -695,10 +699,15 @@ def save_results(files_count, scanned_files, format, input, output_file):
         doc.package = Package(input, NoAssert())
 
         for file_data in scanned_files:
-            file_entry = File(file_data['path'])
-            # FIXME: should we really compute the checksum here rather than getting it from the scan?
-            file_entry.chk_sum = Algorithm('SHA1', file_entry.calc_chksum())
-            for file_license in file_data['licenses']:
+            file_sha1 = file_data.get('sha1')
+            if not file_sha1:
+                # Skip directories.
+                continue
+
+            file_entry = File(file_data.get('path'))
+            file_entry.chk_sum = Algorithm('SHA1', file_sha1)
+
+            for file_license in file_data.get('licenses'):
                 spdx_id = file_license.get('spdx_license_key')
                 # TODO: we should create a "LicenseRef:xxx" identifier 
                 # if the license is not known to SPDX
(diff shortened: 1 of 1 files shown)


pombredanne:
You should not fetch things from the cache. This is an internal stuff and is meant to change at any time. You should instead use a scan with `info`.

sschuberth:
Maybe I'm misunderstanding you here, but I need this to work also for scans that did not get the `-i` option passed on the command line...

pombredanne:
ok then is this case, you want to force returning `info` if an SPDX output has been selected. And FWIW, IMHO `info` should be part of the default in all cases. Let me paste a snippet that does this

pombredanne:
```patch
diff --git a/src/scancode/cli.py b/src/scancode/cli.py
index dc9664b..e36c798 100644
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -262,6 +262,10 @@
         license = True
         package = True
 
+    # hack to force collecting info for SPDX outputs
+    if format in ('spdx-tv', 'spdx-rdf',):
+        info = True
+
     try:
         files_count, results = scan(input_path=input,
```

pombredanne:
And then you can get the`sha1` directly this way:
```patch
diff --git a/src/scancode/cli.py b/src/scancode/cli.py
index dc9664b..62d669e 100644
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -646,8 +646,11 @@
 
         for file_data in scanned_files:
             file_entry = File(file_data['path'])
-            # FIXME: should we really compue the checcksum here rather than get it from the scan?
-            file_entry.chk_sum = Algorithm('SHA1', file_entry.calc_chksum())
+            # get the sha1 from scan, it is part of the info
+            file_sha1 = file_data.get('sha1')
+            if file_sha1:
+                # do not add a sha1 if not present (e.g. not in scan or for a directory) 
+                file_entry.chk_sum = Algorithm('SHA1', file_sha1)
             for file_license in file_data['licenses']:
                 spdx_id = file_license.get('spdx_license_key')
                 # TODO: we should create a "LicenseRef:xxx" identifier 
```

pombredanne:
This would be much better than trying to get this from the cache which is not something that should be relied on or used for anything except caching.

pombredanne:
To more explicit, I would rather not rely on implementations details of the cache for any collected scan data: this is a better separation of concerns. It does today use a sha1 from the scan and may not tomorrow. If you were to use the sha1 as used in the cache internal cooking, you are making your scan output dependent on the caching implementation details and this would soon break when we use plugins for formats.

sschuberth:
Thanks. Let me try to clarify once more: Doing the "hack to force collecting info for SPDX outputs" does *not* cause any additional scans to be performed. As such this is not really "*collecting* info", but more like "*exposing* info" that was already collected internally. Is that correct?

pombredanne:
You wrote:
> Doing the "hack to force collecting info for SPDX outputs" does not cause any additional scans to be performed. As such this is not really "collecting info", but more like "exposing info" that was already collected internally. Is that correct?

This is correct, the `info` is not part of the "scans" in https://github.com/nexB/scancode-toolkit/blob/4df6e424ed22b0b8dcc5203f4228cdab7d7bfab3/src/scancode/cli.py because it is always done as part of the internals of caching here: https://github.com/nexB/scancode-toolkit/blob/4df6e424ed22b0b8dcc5203f4228cdab7d7bfab3/src/scancode/cli.py#L465 and the collected info are only it is injected in the returned results if `info` was True here: https://github.com/nexB/scancode-toolkit/blob/4df6e424ed22b0b8dcc5203f4228cdab7d7bfab3/src/scancode/cli.py#L437 otherwise the cache does not return that data, even though if was always generated

sschuberth:
Thanks, then I'm confident that this is not as bad a hack as I thought in the beginning, and I just removed the "[WIP]" from the title.

sschuberth:
@pombredanne Does this look good to you now?

pombredanne:
@sschuberth looks all good, yet do you think you could add one or a couple tests? I can do it otherwise.

sschuberth:
I didn't forget about the tests and I agree we need some, but I first wanted to fix the stuff that I know to be broken. But on top of that I have to admit that due to my limited time it might take me a while to add these tests.

pombredanne:
@sschuberth fair enough. Merged.!

PR #451, opened 2017-01-13
cli: Write SPDXNone if no licenses are found

sschuberth:
The field is mandatory, so we must write it.


sschuberth:
@pombredanne BTW, I was thinking about implementing the logic to write `SPDXNone()` in case of an empty list of licenses in the [tools-python](https://github.com/spdx/tools-python) package, but I came to the conclusion that it's better to require `SPDXNone()` to be explicitly written by the user of the package as the user might also want to write `NoAssertion()` instead if "the SPDX file creator has made no attempt to determine this field".

pombredanne:
@sschuberth this make sense. Thanks!

PR #459, opened 2017-01-17
cli: Add a fallback to calculate the SHA1 if it is not present in the cache

sschuberth:
@pombredanne I've updated the PR with a commit that addresses your `TODO` of writing a `LicenseRef` for licenses not known to SPDX.

### DIFF
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -679,12 +679,15 @@ def save_results(files_count, scanned_files, format, input, output_file):
             if file_licenses:
                 for file_license in file_licenses:
                     spdx_id = file_license.get('spdx_license_key')
-                    # TODO: we should create a "LicenseRef:xxx" identifier
-                    # if the license is not known to SPDX
                     if spdx_id:
                         spdx_license = License.from_identifier(spdx_id)
                         file_entry.add_lics(spdx_license)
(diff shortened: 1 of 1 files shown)


pombredanne:
@sschuberth Thanks! 
I am sorry that you have to resort to this... I think instead that I may prefer always adding the checksums (even if the file is empty) and instead add a new file-level flag `is_empty1 that would state the file is essentially empty junk: e.g. really empty, size is 0, or composed only of a single character (e.g. ----- or zeroes) or composed only of white spaces (spaces, tabs, lf, etc).... Let me think about this overnight: this is small and easy change and might be much cleaner.

sschuberth:
I agree that adding hashes even for empty files is more consistent. However, you also told me I shouldn't rely on cache internals, and maybe at some point the info does not expose the SHA1 anymore, so having a fallback to calculate it might make sense anyway.

WRT storing info about empty files, maybe instead of introducing a whole new flag, consider adding another type in `filetype.py`?

pombredanne:
The info will always expose a sha1 (the fact that it is also used as a cache key is the only "internal" implementation part).

And you got it right: that would be a new flag in `filetype.py`

pombredanne:
@sschuberth I am merging this for now so this is not in your way and I created #461 to update the behavior

Issue #464, opened 2017-01-23
Option to only report first occurrence of a license of a kind


sschuberth:
For license provenance purposes it often is enough to prove that at least one file under a specific license is present. In order to make the result set smaller and easier to review, how about adding an option to only keep the first occurrence of a license of a kind?

For example, if file `A` is *GPL*, and file also `B` is *GPL*, only report file `A`.

In other cases, I'm not yet entirely sure how the feature should work: If file `A` has both *BSD* and *Apache-2.0*, file `B` has only *BSD*, and file `C` has only *Apache-2.0*, would you only report file `A` as that's sufficient to cover the list of different licenses, or should you report `B` and `C` to clarify the licenses might appear in different files, or report all three files in this case?

pombredanne:
This makes a lot of sense and as you explain the devils is in the details... IMHO this would be best handled as part #377 ... would you agree? e.g. the goal would not be to make the scan results smaller per se, but to make them easier to review hence apply some smarts on top of results to get some sorts of summarization (one of which could be the one you mentioned here)

pombredanne:
Note that if you want to be comprehensive, you may also need to collect the copyrights. And there are cases where you will have various combos of licenses and copyrights present or not in various files of the same codebase. Again, all these small cases matter I think.

sschuberth:
> IMHO this would be best handled as part #377 ... would you agree?

That sounds reasonable, yes.

> the goal would not be to make the scan results smaller per se, but to make them easier to review hence apply some smarts on top of results

How would you do this e.g. for SPDX tag-value output? IMO you cannot, so you would really have to omit data from the result file in this case.

> various combos of licenses and copyrights present

Good point. So you would basically need to de-duplicate on copyright-license tuples instead of just licenses.

pombredanne:
Since we now report license expressions, this is IMHO no longer relevant. Would you agree?

sschuberth:
Just to document the current state: `license_expressions` (plural) are currently reported once per file, and multiple findings of the exact same license are de-duplicated in that list. So that's different from what I was originally asking in this ticket.

However, I hereby withdraw my original proposal as I believe ScanCode should always provide the full picture and report all findings. If picking an (arbitrary or specific) occurrence of a license finding is enough for someone's use case, that picking should be done by some custom tooling provided by the user as a post-processor to ScanCode's results.

Issue #377, opened 2016-11-25
Proposal: Scan deduction and summarization


pombredanne:
## Context
Scanning operates at the file level. This is good but in many cases a scan reports too much data at a too detailed level.  This happens when related clues are detected across files or inside the same file.

## Problem
**Multiple related clues in different files**

For instance, if every file in a directory tree has the same license and copyright statements, then the license and origin information could be **rolled up** at the level of this directory and the file details could be omitted.

Or say that a scanned directory only contains a COPYING file with a license and notice and that all the files in that directory have no license or copyright. Then the license and origin information could be extended from the COPYING to all the files in that tree.

Or say that a scanned directory only contains a README file with a license and notice and that all the files in that directory have a comment `See README for licensing`. Then the license and origin information could be extended from the README to all the files in that tree that carry this comment.

Or say that a Package is detected (such as Maven Jar or an NPM or else) and that the package-level metadata described correctly the licensing of all the files for this package and that the scan of the files in this package does not bring new details. Then only the license and origin information from the package could be kept and the file details omitted.

Or say that a directory contains code in a mix of programming languages: the primary or main language or language stats could be rolled up at the directory level. 

Or say that a directory contains both code and build scripts and that the license for the build scripts is different from that of the code (say this is some autotools MIT or FSF notice). Then the licenses for the directory could be summarized based on a classification of the code in code proper and build scripts and the build scripts license being marked as such would not be reported as the directory or package license in this case.

**Multiple related clues in the same file**

Some scans operate on the same data in a given file and this may trigger reporting extra or spurious clues and could be instead considered together. 

For instance a license text may contain a copyright statement for the text of the license and URLs and emails. Detecting licenses, copyrights, emails and urls could report four different clues in same scanned file and scanned text region when this is may be instead a single clue for the license that should be reported and not four clues.

Or a package metadata file would typically contains origin and license information and these would end up reported twice both as package attributes and individual detection for license, copyright and urls.

## Solutions elements

A comprehensive solution may cover some or all of these:

- determine where to summarize and roll up clues. For instance, rolling everything at the root directory level would rarely make sense; instead rolling things up at a package level and finding what would be a good directory level to use as a break point would be important
- implement some classification of files such as test, code proper, build scripts, test code, etc.
- implement some stats, rules and/or machine learning to summarize and deduct proper higher-level origin and license. 
- may be scan all the clues at all time to be able to combine (and filter) them properly
- combine package detection with license and copyright

[...]

PR #476, opened 2017-02-06
Omit files without findings

sschuberth:
@pombredanne What do you think about this? It sort of tries to help solving the problem mentioned in #464 in the sense that it tries to also reduce the amount of results being generated to the most useful ones.

### DIFF
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -244,9 +244,12 @@ def validate_formats(ctx, param, value):
 @click.option('--timeout', is_flag=False, default=DEFAULT_TIMEOUT, type=int, show_default=True, help='Stop scanning a file if scanning takes longer than a timeout in seconds.')
 @click.option('--max-memory', is_flag=False, default=DEFAULT_MAX_MEMORY, type=int, show_default=True, help='Stop scanning a file if scanning requires more than a maximum amount of memory in megabytes.')
 
-def scancode(ctx, input, output_file, copyright, license, package,
-             email, url, info, license_score, license_text, format,
-             verbose, quiet, processes,
+def scancode(ctx,
+             input, output_file,
+             copyright, license, package,
+             email, url, info,
+             license_score, license_text,
+             format, verbose, quiet, processes,
              diag, timeout, max_memory,
              *args, **kwargs):
     """scan the <input> file or directory for origin clues and license and save results to the <output_file>.
@@ -255,26 +258,43 @@ def scancode(ctx, input, output_file, copyright, license, package,
     Error and progress is printed to stderr.
     """
 
+    possible_scans = OrderedDict([
+        ('infos', info),
+        ('licenses', license),
+        ('copyrights', copyright),
+        ('packages', package),
+        ('emails', email),
+        ('urls', url)
+    ])
+
     # Use default scan options when no options are provided on the command line.
-    possible_scans = [copyright, license, package, email, url, info]
-    if not any(possible_scans):
-        copyright = True
-        license = True
-        package = True
+    if not any(possible_scans.values()):
+        possible_scans['copyrights'] = True
+        possible_scans['licenses'] = True
+        possible_scans['packages'] = True
 
     # A hack to force info being exposed for SPDX output in order to reuse calculated file SHA1s.
     if format in ('spdx-tv', 'spdx-rdf'):
-        info = True
+        possible_scans['infos'] = True
+
+    get_licenses_with_score = partial(get_licenses, min_score=license_score, include_text=license_text, diag=diag)
+
+    # List of scan functions in the same order as "possible_scans".
+    scan_functions = [
+        None, # For "infos" there is no separate scan function, they are always gathered, though not always exposed.
+        get_licenses_with_score,
+        get_copyrights,
+        get_package_infos,
+        get_emails,
+        get_urls
+    ]
+
+    scanners = OrderedDict(zip(possible_scans.keys(), zip(possible_scans.values(), scan_functions)))
 
     scans_cache_class = get_scans_cache_class()
     try:
         files_count, results = scan(input_path=input,
-                                    copyright=copyright,
-                                    license=license,
-                                    package=package,
-                                    email=email,
-                                    url=url,
-                                    info=info,
+                                    scanners=scanners,
                                     license_score=license_score,
                                     license_text=license_text,
                                     verbose=verbose,
@@ -298,8 +318,7 @@ def scancode(ctx, input, output_file, copyright, license, package,
 
 
 def scan(input_path,
-         copyright=True, license=True, package=True,
-         email=False, url=False, info=True,
+         scanners,
          license_score=0, license_text=False,
          verbose=False, quiet=False,
          processes=1, timeout=DEFAULT_TIMEOUT, max_memory=DEFAULT_MAX_MEMORY,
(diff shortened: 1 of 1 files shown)


pombredanne:
@sschuberth Thanks! Yes it makes perfect sense. I wonder though if we could find a better name for the option? I was actually thinking about having that and also another option to scan only certain files (as opposed to excludes).

Also I will look at the last commit separately if this is OK with you as this does not seem directly related ?
 https://github.com/nexB/scancode-toolkit/pull/476/commits/1647c9fabe2dfd6203bcd35725b7e2c28980b420

sschuberth:
That commit is a more or less required preparation to more elegantly get to know the scans that were done, when writing the results.

I'm perfectly fine with naming the option differently. I also find it a bit long but could not think of anything shorter that still is precise.

pombredanne:
Naming is always difficult... so beside `--omit-files-without-findings` what about some other names?

May be there is a way to avoid the double negation?
hum... tough. The concept is to not return records for files or directories that have no data for the requested scans. 

Or stated an other way, return records for files or directory with data vs. returning all records.

so what about these: `Only return files or directory with scan data for the requested scans`.
- `--with-data`
- `--with-clues` 
- `--with-findings`

or shorter but still with a double negation: `Do not return files or directory without scan data for the requested scans`. which is just a shorter take on `--omit-files-without-findings`:
- `--no-empty`
- `--skip-empty`
- `--omit-empty`
- `--suppress-empty`

sschuberth:
I was thinking about a naming that involves "empty", too, but discarded it as it might be confused too easily with empty files being scanned.

From your other suggestions I like "--with-findings" best. To be more precise, I would call it "--with-findings-*only*", or do you think that's again too long and the original is clear enough?

When it comes to the help message, I believe we can omit the "or directories" part, as scan results only list files, never directories, if I'm not mistaken.

pombredanne:
What about `--only-findings` then? and `Only return files or directory with scan findings for the requested scans. Files without findings (e.g. no license, no copyright) are skipped.`

FWIW, the directories are returned when you scan for `info`, and they may become ever more important with a directory-level roll-up deduction in a near future with #377

pombredanne:
What I like with `--only-xxx` is that there could be other options of the same style in the future e.g. `--only-summary` or else.

sschuberth:
I like the "--only-xxx" pattern, will update shortly.

sschuberth:
Done. I just tweaked the help wording a bit as I believe "requested scans" is not always true, as more scans than requested might be run (think of the "info" scan which gets automatically activated for SPDX output), and I also skipped the "e.g." part as I felt it to be a bit unclear that files with no license *and* no copyright are only omitted if these were the only active scans, i.e. no other scans returned any findings.

pombredanne:
@sschuberth Thanks!  I will merge after review

src/scancode/cli.py
```diff
@@ -579,10 +583,24 @@ def scan_one(input_file, scanners, diag=False):
...
+    return False
+
+
+def save_results(scanners, only_findings, files_count, scanned_files, format, input, output_file):
     """
     Save scan results to file or screen.
     """
+
+    if only_findings:
+        active_scans = [k for k, v in scanners.items() if v[0] and v[1]]
```
pombredanne: what are v[0] and v[1] ? I find it is usually best to avoid indexed access and use names instead, it is less brittle and clearer.

src/scancode/cli.py
```diff
@@ -255,26 +262,42 @@ def scancode(ctx, input, output_file, copyright, license, package,
...
 
     # A hack to force info being exposed for SPDX output in order to reuse calculated file SHA1s.
     if format in ('spdx-tv', 'spdx-rdf'):
-        info = True
+        possible_scans['infos'] = True
+
+    get_licenses_with_score = partial(get_licenses, min_score=license_score, include_text=license_text, diag=diag)
+
+    scan_functions = [
+        None,
```
pombredanne: What does this `None` means?

src/scancode/cli.py
```diff
@@ -255,26 +262,42 @@ def scancode(ctx, input, output_file, copyright, license, package,
     Error and progress is printed to stderr.
     """
 
+    possible_scans = OrderedDict([
```
pombredanne: I think I get where you are going with this, but we end having eventually three bits of data: the scan function, the name of the scan and its option flag as true or false.
This likely reaches the point where a small object would be best and more explicit and would avoid any indexed property access or zipping for parallel sequences realignments.

src/scancode/cli.py
```diff
@@ -572,11 +583,25 @@ def scan_one(input_file, scanners, diag=False):
...
+
+
+def save_results(scanners, only_findings, files_count, scanned_files, format, input, output_file):
     """
     Save scan results to file or screen.
     """
-    # note: in tests, sys.sdtout is not used, but some io wrapper with no name attributes
+
+    if only_findings:
+        active_scans = [k for k, v in scanners.items() if v[0] and v[1]]
```
pombredanne: To the point I was making above, it sounds like you are carrying a tuple in the value and have to resort to an index access. I usually try to stay away from indexed access as this does not make super explicit what is going on.

src/scancode/cli.py
```diff
@@ -572,11 +583,25 @@ def scan_one(input_file, scanners, diag=False):
...
+        scanned_files = [file_data for file_data in scanned_files if has_findings(active_scans, file_data)]
+        files_count = len(scanned_files)
```
pombredanne: We would have to check if these two do not force loading the whole scan results in memory which had been replaced by loading a stream from the disk cache.   AFAICR `scanned_files` is the iterator loading from the cache.

pombredanne:
@sschuberth btw, since you can commit directly, in the future you could also push a branch here directly. This way I could commit in there too (otherwise I could instead push a PR to your PR.... though that may be a tad too many PRs stacked ;) )

sschuberth:
> This way I could commit in there too

You're right, sorry. I'm still stuck to my old setup, but let me fix that now.

src/scancode/cli.py
sschuberth: I agree having a dedicated object instead of a tuple as the value in the dict would be more readable. However, it would probably not get rid of the `zip`ping as I do want to be able to access all first items of the tuple as a list to easily check them with `any`. Let me play around a bit with this...

src/scancode/cli.py
sschuberth: I've added a comment for now.

src/scancode/cli.py
sschuberth: I've added a comment for now.

src/scancode/cli.py
sschuberth: What"s the proper way to check this in Python? Is there a more elegant way than using a memory profiler?

sschuberth:
> otherwise I could instead push a PR to your PR...

Actually, have you tried using GitHub's "Allow edits from maintainers" feature? I agree using just your repo still is more convenient, but I'm just curious whether it works for you.

src/scancode/cli.py
sschuberth: So I've played with using a generator to filter the list. But the problem then is to calculate the number of file entries the generator would "remove" in order to correct the `files_count` variable. Any idea how to address this?

sschuberth: It's probably better to not add files without findings in the first place (if that option is given on the command line) rather than filtering the results afterwards...

pombredanne:
@sschuberth Thanks for looking into all this. ""Allow edits from maintainers" "--> Let me try this too.
BTW, I wonder if the feature would not be best pushed down to the cache loading streaming function, but if I can build that on top of your PR, you can see the proposed changes alright

pombredanne:
I might have used a slightly different approach to implement the "no empty" which is the main purpose of this PR. But my command line handling code is a pile of ugly spaghetti code that I created and that I need to clean up and unwind.
So for now, I think I will be merging this in, and will add a small fix to ensure empty results are not returned from the cache (and handle this in the cache rather than in the cli)
And I will cleanup my spaghetti separately in #489

pombredanne:
Actually for now `--only-findings` will load all the scan results in memory, just so you are aware of this

sschuberth:
> Actually for now `--only-findings` will load all the scan results in memory, just so you are aware of this

Correct. For the record, I've pushed my (incomplete) idea to use a generator as PR #491.

pombredanne:
This has introduced a regression... License indexing always takes place when the index needs to even if a license scan was not requested. This is minor in practice but annoying in development where any codechange and test run will trigger reindexing. I will handle this in #489

sschuberth:
@pombredanne Sorry for that regression. As a side note, I had the feeling that even before license indexing was triggered unnecessarily. Under what circumstances *should* license indexing happen?

pombredanne:
@sschuberth do not worry I will fix this. My code is real mess :) Reindexing is triggered automatically when trying to get an index if there is any code or data change since last indexing, but should be done only if `--license` was requested as this is not needed otherwise.

PR #486, opened 2017-02-09
Use "NOASSERTION" for disabled scans in SPDX output

### LINKED ISSUE #483: Distinguish between a scan not being performed and a scan not returning results
In particular for SPDX output, it is important to distinguish between a scan not being performed and a scan not returning results in order to write either `NOASSERTION` or `NONE`. Currently, no such distinction is made when writing SPDX output.

sschuberth:
Please have a look at the individual commit messages for the details. Fixes #483.


pombredanne:
Thanks. All merged.

Issue #489, opened 2017-02-13
Clean up command line handling code


pombredanne:
This is a follow up to #476 
I have grown `cli.py` into some very ugly spaghetti code that is hard to understand and evolve.

How ugly? what about this summary:
```
 scancode(): receives flags
 1. determine which scan flag is really activebased on default.
 2. call scan(), passing a scan cache class
  2.1. build a template results mapping keyed by scan name
  2.2. build scanning functions for each active scan flag
  2.3. if license is requested, runs indexing first
  2.4. build a scanit() wrapper function to _scanit() this is a simple closures on _scanit args
  2.5. using a process Pool, for each resource  calls scanit()
  2.5.1. scanit() scans one single file
     2.5.1.1. collect info, always   
     2.5.1.2. check in the cache if we have scanned that file
     2.5.1.3. if yes, just return a code and the scanned path
     2.5.1.4. if no:
       2.5.1.4.1. build a scans_runner wrapper to scan_one() such that all its args are closed.
       2.5.1.4.2. run that scans_runner wrapper as an interruptible function, getting the results from it
       2.5.1.4.3. puts the results in the cache
     2.5.1.5. scan_one() loops on the scanners provided for one file:
       2.5.1.5.1. execute the scan function
       2.5.1.5.2. return aggregated scan results (and possible errors) from scanning that ilfe
 3. save results
 4. clears cache
 ```
 
So in recap the flow looks more or less like this:
```
scancode()
  scan()
   for each resource, spawn process:
     scanit()
      _scannit()
       check cache and return if cached
       interruptible()
         scan_runner()
          scan_one()
            for each scan function:
              call func, update results
          return results
       cache results
       return code
  save_results()
   invoke "output" function, passing an interator on the cache
  clear cache 
  ```
So this is deeply nested and complex.
Beside, the code to save results in different format just compounds this all.

pombredanne:
#885 being marged we now have a fairly clean CLI code. Closing.

Issue #275, opened 2016-06-26
Refine and simplify Package models


pombredanne:
Things that need to be corrected:
- the use of subclassing for Package does not make sense as implemented. We need instead some composition IMHO: the current way smells funny. 
- And we need a plugin approach such that new package formats are implement as external libraries, eventually using different languages than Python to reuse the many parsers natively available (Ruby, Js, Java, etc). This would likely make the pure subclassing likely naturally vanish. 
- schematics is likely not the right library for our use case: attrs or marshmallow need to be reviewed. The schematics-based code smells funny too

We need a few more types and change some types:
- Project: several packages are part of a larger project that defines many packages attributes
- Repo: a package name/version is typically unique only within the confine of a repo
- FQN: fully qualified name such as in repo:name:version (and possibly os/arch for many packages beyond deb and rpms) and defining some kind of global identification
- Grouping: several packages are often derived for different architecture/platforms (Java for Ruby, egg/wheels/sdist/bdist on Python, various os/arch on RPMs and debs, etc.) or different artifacts are available for a package (src/bin/doc/pom for Maven)
- Dependencies: while they live in some group, a flat sequence is likely better than a by-group tree. Also there is always a notion of potential deps vs resolved, locked deps. This needs to be more simply modelled.
- Downloads: we have a 1-to-1 between a download and a package (even though each download can be several urls, this is for the same exact file). This could be refined in one package -> many downloads where each download is for the same package version but packaged differently (arch/os/platform/source/binary/etc)

Some extra data fields needed too:
- documentation_url
- the web page if any for a package release, as is typical for many package repos
- beyond keywords, support for Trove-like classifiers (in sf.net or Pypi)
- download stats when provided?
- file size and filename
- release or upload date

Or should be retired:
- keywords_doc_url

pombredanne:
We need also a `documentation_url`

pombredanne:
From #241 
So now that we have used schematics, I am not super happy with the way this works. In particular anything that would be stored in the DB in a near future would require having yet another schema being defined in yet another library. I made a few local tests storing things as raw JSON in a DB but that raise another problem of schema migrations.
I am considering using Django's ORM to define the models instead. I will try out some tests in a branch. The semantics are essentially the same as schematics or marshmallow (or the other way around) and there could be some side benefits to store an-flight scan in a temp sqlite DB. I will work out something.
Feedback welcomed

pombredanne:
Some thought on refining the models and processing:
[...]

PR #491, opened 2017-02-13
WIP: Use a generator to reduce memory consumption when filtering results

sschuberth:
This is still work-in-progress as a follow-up to PR #476. DO NOT MERGE YET.

### DIFF
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -591,6 +591,16 @@ def has_findings(active_scans, file_data):
     return any(file_data.get(scan_name) for scan_name in active_scans)
 
 
+def filter_scanned_files(scanned_files, active_scans, keep_condition):
+    """
+    A generator to filter the list of scanned files on demand to reduce memory usage.
+    """
+
+    for file_data in scanned_files:
+        if keep_condition(active_scans, file_data):
+            yield file_data
+
+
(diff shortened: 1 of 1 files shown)


pombredanne:
Thanks for giving it a try!
The key thing is that `file_count` needs to iterate the results and eventually consume iterators.
I eschewed this originally because we did not care originally about things were with or without findings.

I think the proper implementation may be one of these:

- ensure that when the scans are run and cached we also track if there were "findings" for the requested scans, and propagate this flag up such that the main loop maintains two counts: one for the total number of files and one for the files with actual results for the requested scans. This way there would not be a need to iterate the whole results to get the counts and iterating from cache could receive a flag to optionally skip things that do not contain findings.

- use the upcoming database-backed storage with sqlite which is prototyped on packages first in #275. With this the actual results will be stored in a proper DB that can be queried alright and doing a `count` will be a no brainer while not its primary purpose which is rather to have: 1. cleaner models for Packages 2. a db storage to support future deductions for #377 that by nature may require access to the results of several files (typically a whole tree or subtree) to apply smarts and heuristics to create the deduction properly.

I tend to favor the second solution which is better in the long run, but I may handle the first one first as part of #489

sschuberth:
Will be handled as part of #489.

Issue #381, opened 2016-11-27
Pluggable output formats


pombredanne:
Following discussions with @sschuberth in #375 we should have a simplified and better to handle output formats. The outline would be this:

1. the only standard format for output would be JSON
2. through a simple plugin mechanism and/or conventions, this could be transformed into other formats such as:
- SPDX #338
- HTML #375 #324
- HTML-APP #320 #271#154 #16
- CSV #339
- a database (from @yayahalom5776 in #273) 
- other formats as needed.

pombredanne:
So for plugins basics there are a few contenders: 
- pluggy (used in in py.test) https://github.com/pytest-dev/pluggy
- stevedore (used in openstack) https://github.com/openstack/stevedore
- pluginbase (from Flask author) https://github.com/mitsuhiko/pluginbase
- roll our own simple stuff with setuptools entry points

All have some pros and cons. In all cases, combining these with setuptools entry points (even if built in) is the way to go IMHO.

sschuberth:
Creating other formats as a post-processing step from JSON also has the advantage of easily gettin scan results in different formats from the same scan. Currently, if you would need both JSON and SPDX, you would need to run the scan twice.

pombredanne:
Note: A plugin architecture is what is needed here. Some refinements on the formats is welcomed but is not the purpose. The purpose is to refactor the existing code for plugins.

pombredanne:
and each plugin for format should ideally be also runnable as its own CLI, with input = scan json (e.g. ABC Data format), output = other format 

See for ABCD:
https://github.com/nexB/aboutcode/blob/a84e207bbe579dd9b419cc52bbfd00239f5beeb9/aboutcode-data/README.rst

pombredanne:
So format plugins have landed in `develop` thanks to @yashdsaraf  with #639
There is no new features there for now except that there is a new CSV output plugin.
Beside this we need documentation on how to create plugins and likely implement the JSON-> other format conversion as discussed here.

pombredanne:
This has been completed and merged in develop.

@sschuberth you wrote:

> Creating other formats as a post-processing step from JSON also has the advantage of easily gettin scan results in different formats from the same scan. Currently, if you would need both JSON and SPDX, you would need to run the scan twice.

Just ping as if we support #789 in the future, then we would need to have one option for each format such that we can specify which file each output should be written to.

sschuberth:
[...]

Issue #375, opened 2016-11-24
"File Information" table present in HTML output even if -i is not given


sschuberth:
Scanning for file information (`-i`) is disabled by default, yet the (empty) "File Information" table is included in the HTML output. I was looking into fixing this, but I could not get my head around where the actual information is added to the info dictionary. In [here](https://github.com/nexB/scancode-toolkit/blob/develop/src/scancode/format.py#L211) only the file name seems to be added, but where is the `md5` etc. added before feeding it to the HTML template? I know the info is calculated [here](https://github.com/nexB/scancode-toolkit/blob/develop/src/scancode/api.py#L149), but I do not see how this gets fed into the HTML template.

pombredanne:
@sschuberth in the latest code for  #267 now mostly in develop the file information is always collected and reported in the final JSON only if effectively requested. This is because the info is used for the cache keys generation. Now the nature of how the data is store for info has also changed and this is no longer a separate data block for a file but instead plain attribute at the file level which probably added to the confusion.

Now, the html output (the the html-app) is entirely driven by a Jinja template found here https://github.com/nexB/scancode-toolkit/blob/ca640b92d9cb82c3523aa229dd26c76907b1a17b/src/scancode/templates/html/template.html and this has likely not been updated correctly when info have been rolled up as file attribute. 

I hope this makes sense for a start.... more to come tomorrow

sschuberth:
Thanks @pombredanne, I'm aware that you use Jinja to fill in the template. But I don't get where the [row.type](https://github.com/nexB/scancode-toolkit/blob/ca640b92d9cb82c3523aa229dd26c76907b1a17b/src/scancode/templates/html/template.html#L128) etc. are filled in the Python code. I do get that `row.name` comes from [here](https://github.com/nexB/scancode-toolkit/blob/develop/src/scancode/format.py#L211), but what about all the other file info properties?

pombredanne:
@sschuberth you stumbled on the kludgy part... It felt not right when I wrote this and I knew this would boomerang: you have eagle eyes!

The info are no longer under an explicit `infos` attribute: instead they are each a direct attribute of a file.
And they are not collected explicitly but instead implicitly by exclusion.

So in this loop https://github.com/nexB/scancode-toolkit/blob/ca640b92d9cb82c3523aa229dd26c76907b1a17b/src/scancode/format.py#L207

```Python
        # this is klunky: we need to drop templates entirely
        converted_infos[path] = OrderedDict()
[...]

Issue #338, opened 2016-10-19
Support SPDX tag-value as an output format


sschuberth:
It would be great if the SPDX tag-value format would be supported as an output format, see [this example](https://spdx.org/spdx-tagvalue-example).

pombredanne:
@sschuberth this makes perfect sense.  (FWIW, I am after all one of the co-founders of SPDX: http://web.archive.org/web/20160617012420/https://spdx.org/about-spdx/what-is-spdx )

That said, @ah450 wrote this Python library for SPDX for the Google Summer of Code a few yeasr back (I was the mentor). See:
- http://git.spdx.org/?p=spdx-tools-python.git;a=summary
- https://github.com/ah450/spdx-tools-python

This is not up entirely to date with the latest spec and would likely require some work first but is a great base to start from.

And  the general idea would be to have probably an external utility that takes scan results (or more generally the upcoming ABC data format to be pushed soo to https://github.com/nexB/aboutcode ) and converts that to SPDX (and the possibly the other around too).

@sschuberth Would you have some cycles to chip in may be?
@ah450 Do you think you could update your library to support the latest SPDX spec, at least the tag-value format?

sschuberth:
@pombredanne I'd love to help, but don't have the time currently. Anyway it's good to see that you acknowledge this to be a meaningful feature.

pombredanne:
@sschuberth no problem. In any case this is definitely in the roadmap: I just updated it: https://github.com/nexB/scancode-toolkit/wiki/Roadmap#data-exchange

sschuberth:
@pombredanne I'm currently thinking about implementing this as a post-processing step to JSON output, not as a strictly separate format. That would have the advantage of being able to convert existing JSON files to SPDX files (if we expose the conversion via command line options). What you you think?

pombredanne:
@sschuberth This makes perfect sense :+1:

sschuberth:
@pombredanne I've started to add some *very* basic SPDX tag / value output support based on https://github.com/spdx/tools-python/issues/9 to ScanCode. Locally, I've simply copied the [spdx](https://github.com/spdx/tools-python/tree/master/spdx) directory into ScanCode's [src](https://github.com/nexB/scancode-toolkit/tree/develop/src) directory. But in order to submit my changes to ScanCode we need to come up with something better. I guess working on https://github.com/spdx/tools-python/issues/2 to publish the spdx-tools to PyPI is the way to go?

[...]

PR #492, opened 2017-02-13
Be consistent about when to sort the SPDX list of licenses / copyrights

sschuberth:
Do not sort or deduplicate findings on the file level as maintaining the
original order might be useful for provenance analysis. On the package level,
we anyway do not know from which file a particular finding originates, so sort
the list of unique findings to get a better overview.

### DIFF
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -687,6 +687,8 @@ def save_results(scanners, only_findings, files_count, scanned_files, format, in
         doc.creation_info.set_created_now()
 
         doc.package = Package(os.path.basename(input_path), NoAssert())
+
+        # Use a set of unique copyrights for the package.
         doc.package.cr_text = set()
 
         all_files_have_no_license = True
@@ -722,6 +724,8 @@ def save_results(scanners, only_findings, files_count, scanned_files, format, in
                         license_key = 'LicenseRef-' + file_license.get('key')
                         spdx_license = License(file_license.get('short_name'), license_key)
 
+                    # Add licenses in the order they appear in the file. Maintaining the order
+                    # might be useful for provenance purposes.
                     file_entry.add_lics(spdx_license)
                     doc.package.add_lics_from_file(spdx_license)
             else:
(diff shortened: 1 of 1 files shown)


pombredanne:
FWIW, there is a pattern to keep unique instances of items while keeping order something like:
```
def ordered_unique(seq):
    """
    Return unique items in a sequence preserving the original order.
    """
    if not seq:
        return []
    seen = set()
    uniques = []
    for item in seq:
        if item in seen:
            continue
        uniques.append(item)
        seen.add(item)
    return uniques
```

You can either yield or accumulate a unique list (like here) and the backing set handles unicity.
Not sure that this is what you wanted here though

sschuberth:
> Not sure that this is what you wanted here though

Actually it is not :-) The idea is to keep the order *and* duplicates on the **file** level (for both licenses and copyrights) as that might provide some additional info that can be useful during provenance analysis. On the **package** level, however, it is more convenient to get an overview and have a sorted list with unique items. Even if we kept the order / duplicates on the package level, SPDX provides no means to map them back to files, so there really is no point in keeping that information, IMO.

pombredanne:
I am merging this in.
@sschuberth I want to split anything about formatting results output in a separate file and out of `cli.py` towards #489 . Do you have any other pending changes that would not be a in a PR? This is to avoid any pain in merging or rebasing.

sschuberth:
@pombredanne No, I don't have any pending changes anymore right now.

But FYI, one thing that I'm looking into right now is to *not* use SPDX output directly from ScanCode (mainly because the [python-tools](https://github.com/spdx/tools-python) have too many limitations, like being somewhat limited to SPDX 1.2, having no control over adding comments etc.), but to do like I proposed [here](https://github.com/nexB/scancode-toolkit/issues/338#issuecomment-261914210) and [here](https://github.com/nexB/scancode-toolkit/issues/375#issuecomment-262971097), and create our "custom flavor" of SPDX as a post-processing step from JSON output.

I realize you're already tracking this idea as part of #381.

PR #494, opened 2017-02-16
Use SPDX NOASSERTION for unknown licenses

sschuberth:
The spec says that NOASSERTION should be used if "the SPDX file creator
has intentionally provided no information (no meaning should be implied
by doing so)".

### DIFF
--- a/src/licensedcode/data/licenses/unknown.yml
+++ b/src/licensedcode/data/licenses/unknown.yml
@@ -2,3 +2,4 @@ key: unknown
 short_name: unknown
 name: Unknown license detected, but not recognized
 category: Unknown License
+spdx_license_key: NOASSERTION


pombredanne:
@sschuberth be careful there: the `unknown` license key is what is reported when there is a license that was detected BUT what this license is is not clear. Which is IMHO quite different from NOASSERTION because some information has been provided, but what it is is not well defined.

sschuberth:
@pombredanne At first I also though `NOASSERTION` should only be used when no effort was done to determine the license (see case `(i)` in the spec). But as I've written in the commit message, the spec also mentions a second case:

> (ii) the SPDX file creator has intentionally provided no information (no meaning should be implied
by doing so).

And one reason why no information is provided might be that a license was detected, but is known.

pombredanne:
@sschuberth I have given a good to this overnight, and I think it does not make sense to assign such an SPDX id so early on in the process. Rather if you want to treat the detection of an "unknown" license as a `NOASSERTION` this should be done when creating an SPDX formatted output instead and even there, this likely may need to be an option as "unknown" means: a license was found but ScanCode cannot tell which one it is for sure. And NOASSERTION means: I, as the creator of this SPDX document, have decided intentionally to provide no information about the license of this file/dir/package.... The two concepts are IMHO too far apart to be conflated. Where someone could elect to go with NOASSERTION in one case, someone else may elect to report this as `LicenseRef:unknown` instead or not report it at all. Would you agree?

sschuberth:
I'm not fully convinced yet, but I have to admit that the SPDX spec is not really clear on how to handle findings which we know to be a license, but we don't know which one. I still believe `NOASSERTION` would be a valid choice with how the spec is currently worded, but using `LicenseRef-unknown` (and maybe more specific IDs like `LicenseRef-fsf-unknown` if we at least know it's an FSF license) might be another good option.

I'm going to close this as we either implement the use of `NOASSERTION` in our internal `json2spdx` script, or prepend `LicenseRef-` to any unknown license finding. So there's nothing to address upstream in ScanCode anymore.

Issue #484, opened 2017-02-08
Memory leaks on large scans


pombredanne:
This was reported to me by email:
The ScanCode Toolkit provides an API to set limits to memory and time consumptions for each scan (the `--timeout` and `--max-memory` arguments).

Looking at `interruptible(func, *args, **kwargs)` function in [`src/scancode/interrupt.py`](https://github.com/nexB/scancode-toolkit/blob/f7434cc72437124605982c55b9a0c9e00b2ac01c/src/scancode/interrupt.py#L106), you start two threads (or execution units as you called it): `(func, args,)` and `(time_and_memory_guard, [max_memory, timeout],)`.

After that, you wait for the first thread to finish by doing `threads.next(timeout)`. Actually, that is a pretty smart strategy, I like that.

In the [`_finally_`](https://github.com/nexB/scancode-toolkit/blob/f7434cc72437124605982c55b9a0c9e00b2ac01c/src/scancode/interrupt.py#L165) section of that `_try_` block, you call `pool.terminate()`, however there might be threads that keep running:

- a. If `(func, args,)` returns first, then `(time_and_memory_guard, [max_memory, timeout],)` will keep running for at least _timeout_ seconds;

- b. If `(time_and_memory_guard, [max_memory, timeout],)` returns first, then `(func, args,)` will keep running until the end of the scanning process (which actually, doesn’t protect against memory consumption).

Finally, the behavior of `pool.terminate()` is to keep the threads running until the entry point finishes.

Net effect: based on this excellent we have a leak on the threading for multiprocessing that needs to be fixed ASAP.

pombredanne:
humm.. So killing threads sounds like a no-go based on this http://stackoverflow.com/questions/323972/is-there-any-way-to-kill-a-thread-in-python and http://code.activestate.com/recipes/496960-thread2-killable-threads/ unless ctypes is used.

More pointers: http://stackoverflow.com/questions/2564137/python-how-to-terminate-a-thread-when-main-program-ends

And something to consider if signals may be a cure is that https://docs.python.org/2/library/signal.html
> On Windows, signal() can only be called with SIGABRT, SIGFPE, SIGILL, SIGINT, SIGSEGV, or SIGTERM. A ValueError will be raised in any other case.

Now the whole point of using threads here instead of processes is that the memory checks need to be on the same process that runs the actual scan function so it knows about itself, not in another process.... but threading here shows some limits.... Still digging.

pombredanne:
[...]

Issue #378, opened 2016-11-26
Scan Benchmarks


pombredanne:
Following on the discussion in https://github.com/nexB/scancode-toolkit/pull/376  we should have and run scan benchmarks.

 @pombredanne commented:

>    We should create some benchmarks to watch for any perf changes btw.

 @sschuberth commented:
> That would be nice to have indeed, i.e. to have tests to identify performance regressions (or improvements).

>As a first step, looking at the times Travis CI builds take could be useful. Of course, this includes the setup and build times while we're only interested in the time to run tests, and the tests might be too many small projects whereas scanning performance probably is better measured with a single big project, but still.

>So I was quickly giving http://scribu.github.io/travis-stats/#nexB/scancode-toolkit a look, but unfortunately it does not render properly for me. See scribu/travis-stats#12.

sschuberth:
FYI, https://scribu.github.io/travis-stats/#nexB/scancode-toolkit/develop looks much better now after @scribu was very quick at fixing some isses.

pombredanne:
Very nice. Not sure which story this tells yet, but it looks quite decent.

pombredanne:
The goal is to create a benchmark between scancode and other similar tools.
1. on license detection first
2. on copyright detection second

I have listed some of the tools to consider here: https://wiki.debian.org/CopyrightReviewTools

Some of the things to consider:
1. how to create a comprehensive test data set? How to get feedback on it? how to allow the community to add new tests? (web app?, posts on license-discuss, spdx? )
2. how to set the ground truth of what a perfect detection would be?
3. what would be the metrics to use (precision? recall? F1?)
4. how and where to run the benchmarks?
5. which tools to consider in the benchmarks?
6. how to map the results of each tool to the standard expectations? (e.g different ways of naming licenses, etc)
7. how to publish and publicize the results of the benchmarks? automated web app?
8. how to involve these other tools projects in the effort?
9. how to make the benchmark as serious, unbiased, scientific and factual as possible?
10.  how to keep things up to date and rerun the benchmark on a regular basis?
11. which os top run the benchs on? IMHO Linux only. case solved.

pombredanne:
FYI with the merge of #885 in develop here are some tidbits. The license scan enjoys a ~ 20% speedup!
Not too bad.

```
Both are using a Linux kernel 4.1.5 rc2 from https://git.kernel.org/torvalds/t/linux-4.15-rc2.tar.gz

###############################################
With develop, 2180-02-07
[...]

PR #502, opened 2017-02-17
#484 New implementation of a killable thread with a process pool

pombredanne:
* pool.py implements a get_pool to obtain a patched pools of processes.
 * thread2.py has a utility function to kill threads.
 * interrupt.py "interruptible" has been updated to work with
   a single runner thread and a timeout and memory guard loop.
   This new implementation is vastly simplified with fewer moving parts
   and possibly slightly faster.
 * cli.py now uses an outer pool of processes and a thread-based
   interruptible. It should leak no of few threads since they are
   now killed alright. Leakage is verified in the "interruptible" tests.

### DIFF
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -27,31 +27,12 @@ from __future__ import print_function
 from __future__ import division
 from __future__ import unicode_literals
 
-###########################################################################
-# Monkeypatch Pool iterators so that Ctrl-C interrupts everything properly
-# derived from https://gist.github.com/aljungberg/626518
-# FIXME: unknown license
-###########################################################################
-from multiprocessing.pool import IMapIterator, IMapUnorderedIterator
+# Import early because this import has monkey-patching side effects
+from scancode.pool import get_pool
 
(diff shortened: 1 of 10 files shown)


pombredanne:
@hugo-bmwcarit I would be interested to get your feedback on this. I verified on a few large scans that it does not leak threads anymore and the implementation is tighter and simplified and eventually a tad faster too, though I would need to setup some benchmarks for it per #378

pombredanne:
@sschuberth if you can have a quick look too, this would be welcomed!

hugo-bmwcarit:
Hi @pombredanne, sorry for the late response, I'm in a hurry this week. I already had a look at the code but I hadn't time to test it. I will try to do that tomorrow and then let you know and write some comments about the code also. Thank you very much.

pombredanne:
let me rebase on the latest develop and force push

pombredanne:
@hugo-bmwcarit I am merging this as I need to make further code changes in the same area. You can provide additional feedback in a new ticket.

Issue #465, opened 2017-01-24
Make scancode --version report the Git commit


sschuberth:
When using ScanCode's `develop` branch and updating often, the version number in the source code / last tag in Git stays the same although potentially quite a lot of code has changed. A such it would be beneficial if `scancode --version` would report its Git commit in some way in order to determine the exact version, maybe by appending the output of `git describe --tags`. That would also solve the problem of keeping the source code in sync with Git tags, as the current source code reports "ScanCode version 2.0.0rc3" although the latest Git tag is "v2.0.0.rc2". Maybe baking the version info into ScanCode could be done as part of running `configure`?

pombredanne:
@sschuberth I agree with you. The only difficulty is to get something that works all the time when you are either running:
1. from a `git clone` where git is available or not
2. from a Github tar/zipball download
3. from a release download (e.g. a Python sdist)

There are many tools available that may help... but I do not know if they can support all three cases without requiring to have Git installed as a dependency. This would require some research. 

- https://pypi.python.org/pypi?%3Aaction=search&submit=search&term=setuptools%20Git
- https://pypi.python.org/pypi?%3Aaction=search&submit=search&term=setuptools%20scm
- https://pypi.python.org/pypi?%3Aaction=search&submit=search&term=setuptools%20vcs

pombredanne:
This is fixed. Right @sschuberth ?

sschuberth:
Yes, thanks @pombredanne!

PR #510, opened 2017-02-22
#465 Use setuptools-scm to get version with a git tag

pombredanne:


pombredanne:
@sschuberth for your review for #465

sschuberth:
Cool, just today I was missing that feature again! But even after reading about the [scheme](https://pypi.python.org/pypi/setuptools_scm) I'm having difficulties to understand how exactly it works.

For my local repo (with some custom merges) I get `2.0.0rc3.dev101+ng2ec2cb59`. But where does the `2.0.0rc3` part come from, as there is no such tag, and the latest tag is [v2.0.0.rc2](https://github.com/nexB/scancode-toolkit/tree/v2.0.0.rc2)?

Also, even after creating local (lightweight and annotated) tags, I still see the "distance" part of "101", and the tag name not being picked up.

Finally, I believe in our case the "dev" prefix is misleading as one could misunderstand it to refer to the "develop" branch, which it does not. I'd prefer to make the scheme match exactly the one of `git describe`, i.e. "{tag}-{distance}-{revision hash}".

Note that there seems to be a bug in the documentation of setuptools_scm as `{revision hash}` is not exactly the Git revision hash, but the Git revision hash prefixed by "g" as `git describe` would use it. That's why with the default scheme you get the awkward "ng" (which sounds like "next generation") prefix for the hash.

pombredanne:
2.0.0rc3 comes either from the `version` in setup.py to which things are added as I did not remove that. Or it is setuptools_scm that seem to always bump the minor or patch segment from the latest tag e.g. `2.0.0rc2` if the workspace is "dirty" e.g not on the tag.

sschuberth:
The [setuptools_scm docs](https://pypi.python.org/pypi/setuptools_scm) say the default scheme to use `{next_version}` if there is a *distance*. I don't like this "auto-bumping" because I prefer to refer to the app's revision using *existing* tags, not some that may or may not exist in the future. For example, how's the bumping logic supposed to now if after 2.0.1 comes 2.0.2, or 2.1.0 as you've a big number of features?

pombredanne:
@sschuberth would you know what to do to get this behavior then?

sschuberth:
After reading through [this post](http://www.danielstender.com/blog/setuptools-scm/) it seems the solution is to not set `use_scm_version` to a *bool*, but to *dict* overriding `version_scheme` with one of the [available implementations](https://github.com/pypa/setuptools_scm#version-number-construction) like

    use_scm_version={
        'version_scheme': 'post-release'
    },

But that still does not seem to work fully. While I now see strings like

    Requirement already satisfied (use --upgrade to upgrade): SPARQLWrapper in ./lib/python2.7/site-packages (from rdflib->spdx-tools>=0.4.1->scancode-toolkit==2.0.0rc2.post102+ngd137d38e.d20170224

when running `./configure`, and the contained `2.0.0rc2.post102+ngd137d38e.d20170224` looks better WRT to not using "rc3", `./scancode --version` still shows

    ScanCode version 2.0.0rc3.dev102+ngd137d38e.d20170224

Not sure why there's `rc3` (and also `dev` instead of `post`) still in there.

sschuberth:
FYI, I've also added a comment about the misleading "g" [here](https://github.com/pypa/setuptools_scm/issues/152#issuecomment-282271402).

pombredanne:
@sschuberth Thanks. wrt to the `scancode --version`, it comes from elsewhere. I will sync that up once we agreed on something decent.

pombredanne:
@sschuberth setuptools_scm is too sophisticated. I just added a few extra lines in setup.py to the same (controllable) effect.

setup.py
```diff
@@ -16,6 +16,59 @@
...
+
+def get_git_version():
+    """
+    Return version parts from Git or raise an exception.
+    """
+    from subprocess import check_output, STDOUT
+    # this may fail with exceptions
+    cmd = 'git', 'describe', '--tags', '--long', '--dirty',
+    version = check_output(cmd, stderr=STDOUT).strip()
+    dirty = version.endswith('-dirty')
```
sschuberth: Funny, that's basically exactly the same that I've implemented [meanwhile in setuptools_scm](https://github.com/pypa/setuptools_scm/pull/154/commits/08bb6392ed5d765544b517e2853fb9844c8693be#diff-7a9c0c3ec91c440b46ba9f3506777bf0R104) :-)

src/scancode/__init__.py
```diff
@@ -41,3 +43,11 @@
...
 if not exists(scans_cache_dir):
     fileutils.create_dir(scans_cache_dir)
+
+
+from pkg_resources import get_distribution, DistributionNotFound
+try:
+    __version__ = get_distribution('scancode-toolkit').version
+except DistributionNotFound:
+    # package is not installed ??
+    __version__ = '2.0.0rc3'
```
sschuberth: Is there a way to use the version from `setup.py` so the fallback is only hard-coded once?

pombredanne: This is what should happen, but I am playing it safe. The versions are updated at once in all places with `bumpversion` in anycase

setup.py
pombredanne: ah! sorry... I had not tracked you were submitting a PR there. If we can switch back to `setuptools_scm` later this would be great. Though it does not handle any fallback to a default version...

sschuberth: No, I was no meaning to say you should continue using `setuptools_scm`. It will take me more time (and be convincing) to shape it into what I'd expect ;-)

src/scancode/__init__.py
sschuberth: Ok.

sschuberth:
So I just tried the latest state. With this, the version may look like

    ScanCode version 2.0.0rc2.dev105+31d696e5

or

    ScanCode version 2.0.0rc2.dev105+31d696e5.dirty.20170224170011

Personally, I don't like the `.` before `dev`. I'd prefer `-` to visually separate more from `.` in the version number. Also, I don't like `dev` as it might again be confused with the `develop` branch. Likewise for the `+`, let's be consistent with the separators and only use `-`. But I *do* like that the date is only added in case of `dirty`.

So, bottom line, how about this instead:

    ScanCode version 2.0.0rc2-post105-31d696e5

and

    ScanCode version 2.0.0rc2-post105-31d696e5-dirty-20170224170011

?

pombredanne:
Please update as you please ... The only constraint would be to ensure that this ends up being compliant with https://www.python.org/dev/peps/pep-0440/ and that its sorts well.

pombredanne:
I made a few tests with your proposed scheme:
`The version specified ('2.0.0.rc2-post111-0df8a27-dirty-20170226174708') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.`

and

`The version specified ('2.0.0.rc2.post111-0df8a27-dirty-20170226174805') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.`

and

`Normalizing '2.0.0.rc2.post111+0df8a27-dirty-20170226174839' to '2.0.0rc2.post111+0df8a27.dirty.20170226174839'`

So net-net is that dashes are in all cases replaced by dots or creating a non-compliant version  wrt. PEP440.

'2.0.0rc2.post111+0df8a27.dirty.20170226175122' is the scheme that works and sorts. Let's go with this for now.

Issue #495, opened 2017-02-16
File type information in JSON output should be "false" instead of "null"


sschuberth:
In JSON output with `--info`, we have e.g.

    ...
    "is_binary": null,
    "is_text": true,
    ...

It seems odd that `is_binary` is `null` instead of `false`, as it should be the opposite of `is_text`'s value. Likewise for the other `is_*` key, they never seem to be `false`, just `null`.

pombredanne:
Let me dig why this is the case: I may wanted these to be used as tri-boolean, but I doubt and this is more likely a bug. Good catch! In Python `false` is `False` and `null` is `None` and they both result to `False` in an `if XXX` like expression, but this may not be the case in other languages.... and to your point it does not look good nor consistent.

pombredanne:
All merged in develop. @sschuberth Thanks for the report!

PR #511, opened 2017-02-22
#495 Return only True or False for "is_*" flags in get_file_infos

pombredanne:


pombredanne:
@sschuberth for your review for #495

src/scancode/api.py
```diff
@@ -183,12 +183,12 @@ def get_file_infos(location, as_list=True):
     infos['mime_type'] = is_file and T.mimetype_file or None
     infos['file_type'] = is_file and T.filetype_file or None
     infos['programming_language'] = is_file and T.programming_language or None
-    infos['is_binary'] = is_file and T.is_binary or None
-    infos['is_text'] = is_file and T.is_text or None
-    infos['is_archive'] = is_file and T.is_archive or None
-    infos['is_media'] = is_file and T.is_media or None
-    infos['is_source'] = is_file and T.is_source or None
-    infos['is_script'] = is_file and T.is_script or None
+    infos['is_binary'] = bool(is_file and T.is_binary)
```
sschuberth: It's probably correct, but why is the cast to `bool` necessary here? Can any of the arguments be `None`?

tests/scancode/data/unicodepath/unicodepath.expected-linux.json
```diff
@@ -59,20 +57,19 @@
       "type": "file", 
       "name": "%CF%A8%E1%BD%80%E2%8C%A8a", 
       "extension": "", 
-      "date": "2016-12-05",
```
sschuberth: Why's this change necessary in this context? If it was wrong, it should have been wrong before already, or?

sschuberth:
Oh, and do [json2csv.py](https://github.com/nexB/scancode-toolkit/blob/develop/etc/scripts/json2csv.py) or [test_json2csv.py](https://github.com/nexB/scancode-toolkit/blob/develop/etc/scripts/test_json2csv.py) needs adjustments in this context?

src/scancode/api.py
pombredanne: `bool` is not absolutely necessary, but I would rather be explicit on the api side and not rely of the values returned as the `typecode` implementation could evolve. Internally `typecode` uses to emulated tri-booleans (e.g. None, True or False) and this is not super clean and will need some refactoring.

tests/scancode/data/unicodepath/unicodepath.expected-linux.json
pombredanne: I just regenerated the test scans hence the new date. In most cases the dates are ignored in tests with this https://github.com/nexB/scancode-toolkit/blob/917b9db7e4fc3bd91811a82e34e50d4d56067f3e/src/scancode/cli_test_utils.py#L51 ... otherwise the tests would never pass

pombredanne:
@sschuberth you wrote:
> Oh, and do json2csv.py or test_json2csv.py needs adjustments in this context?

It does not look like any adjustment is needed..... Though the scans used by `json2csv` tests are not the current scans... Good catch. Let me add a test

sschuberth:
Thanks for you answers, @pombredanne, but you've missed:

> What happened to file tests/scancode/data/api/info/test.txt as part of this PR? The diff seems to be empty.

pombredanne:
@sschuberth you wrote
> What happened to file tests/scancode/data/api/info/test.txt as part of this PR? The diff seems to be empty.

Actually the latest commit explains why: https://github.com/nexB/scancode-toolkit/pull/511/commits/a415561dda5ec9a690e8360263ba729e50146a09
The empty file is empty on purpose as this is the goal of the text: e.g. get some info that would be otherwise None before the updates in this PR.

pombredanne:
This has been merged after a rebase.

Issue #505, opened 2017-02-21
Collect more data from Maven POMs


linexb:
"dependency_version", "dependency_scope", "contributor_email", "issue_url"  fields are not captured yet in the Maven parser: and we should to support these fields

pombredanne:
@linexb  Could you look instead in a completely different implementation of parsing Maven POMs? with https://github.com/sassoftware/pymaven  and in a maven2.py script? I think this would be much better to get actual deps. And eventually we can patch pymaven and submit upstream our improvements as it may support dependencies well but not basic information fields.

See also https://github.com/sassoftware/pymaven/issues/3

pombredanne:
These are captured now but may not be full added to packages as part of #529

pombredanne:
This is merged in develop.

PR #517, opened 2017-02-24
#505 Add more fields definition for maven in packagedcode

linexb:
Add more POM supported field to get more rich data from pom file


pombredanne:
@linexb Thanks! Let me review this

src/packagedcode/maven.py
```diff
@@ -92,37 +92,49 @@ def recognize(cls, location):
     ('component_parent_version', '/project/parent/version'),
     ('component_packaging', '/project/packaging'),
     ('component_version', '/project/version'),
-
+
```
pombredanne: You have a few extra spaces there.

tests/packagedcode/data/maven/expectations/activemq-camel-pom.xml.json
```diff
@@ -31,10 +34,25 @@
   "maven_component_group_id": [
     "org.apache.activemq"
   ], 
+  "maven_contributor_email": [], 
   "maven_distribution_management_group_id": [], 
   "maven_scm_developer_connection": [], 
   "maven_component_extend_1": [], 
   "maven_developer_name": [], 
+  "maven_contributor_organization": [], 
+  "maven_dependency": [
```
pombredanne: I think the approach we have with XPATH in the current maven code is flawed (not fault from you but from me). We are missing the actual groupid, the scope, optional, etc for each dep. This needs to be a proper data structure and not a plain aligned lists....

pombredanne:
@linexb Note that I rebased your branch on the latest develop

src/packagedcode/maven.py
linexb: I have removed the extra spaces.

pombredanne:
@linexb I squashed and reworded the commits but I am not sure I grok exactly where we are going with this. For sure this maven module is a mess (that I created) and we need to clean it up. Piling up up more features on top can only look weirder.

pombredanne:
This has been replaced by #561and #529 which are merged in develop. So I am closing this without merging.

Issue #447, opened 2017-01-11
Scan output doesn't came up with a root directory


nspsjsu:
I ran scans on a directory named gulp-3.9.1 by using the latest version of scancode-toolkit to get a .json & html output. Everything went well other than having a root directory! As a user i was expecting an overall root directory out of the scan result rather than having sub directories & files at the top level tree view!

pombredanne:
@nspsjsu yes, there is no root directory indeed. And it did not used to be the case. I have been thinking about the fact it is indeed a tad weird.  

Do you think we should always have one? and if so what would it be:
- If scan `/this/that/foo/` would this be `foo`?
- If I scan a single file `/this/that/foo/file.txt` would this be `foo` too?
- If I scan the root directory (rare) `/` would this be `/`?

pombredanne:
Yet, even if no root feels a tad weird, I am still wondering what is the benefit of a root directory?

pombredanne:
ah... I think I get it: if there is no root directory for a scan, I cannot ever summarize all of a scan under a single license and origin if that's the case, say for a package, as there is root to summarize to.

nspsjsu:
Yes, you're right. Sometimes it also helps us to determine the correct version & name.

jdaguil:
somewhere between `rc1` and `rc2` the `path`  value changed. For example: 

`"path": "/Users/jilliandaguil/Downloads/gulp-3.9.1/LICENSE", `

is now

`"path": "LICENSE",`

pombredanne:
@jdaguil @nspsjsu Thanks... this makes sense. Anyone else? with an opinion on the topic?

mjherzog:
I thought we had agreed to re-establish a root somehow - the issue is not summarizing at the level - the issue is very simply context

majurg:
I thinking having the 'root' makes a lot of sense and for me, its much easier to navigate around in the node view.

pombredanne:
@sschuberth any preference?

sschuberth:
I'd prefer to have relative paths WRT the directory to scan (as given on the command line) in the output. Scan results with absolute paths are not "portable" and might disclose unwanted information, like the user name. The caller always knows the root directory as that's what's being passed to ScanCode, and could amend the scan report with that information if required.

pombredanne:
[...]

PR #518, opened 2017-02-24
#447 restore default behavior of v1.x to return a root path by default

pombredanne:
* the new option "--strip-root" does not add a root directory segment


pombredanne:
@jdaguil @nspsjsu Ready for you to test!

jdaguil:
@pombredanne LGTM! :star:

pombredanne:
@sschuberth heads up! This is being merged tomorrow. Note that with #520 you should be able to have a config file that could always `--strip-root` if this is what you prefer.

sschuberth:
Thanks for the notice. I'll take care of adding `--strip-root` on our side.

pombredanne:
@sschuberth thank you for the quick reply! I can merge now then.

Issue #525, opened 2017-03-01
The UI of html-app produced by the scancode needs some improvement


singh1114:
The UI of html-app produced by the command "./scancode --format html-app ...." needs some improvement.

mjherzog:
This app is being deprecated and replaced by AboutCode Manager - https://github.com/nexB/aboutcode-manager

majurg:
I agree with @mjherzog, we are phasing out the usage of `html-app` in favor of `aboutcode-manager` https://github.com/nexB/aboutcode-manager.

Minor changes can certainly be made, but sweeping UI/UX improvements are not what we want to do here. 

@singh1114 What in particular do you wish was different about the `html-app` UI?

singh1114:
@majurg 
These are just the minor changes a few changes like the position of the of few things to make them more readable. You can see the #526 for the for the image of the changes.

majurg:
@singh1114 I am not opposed to the general direction of your changes. However, you need some UI improvement. 

For instance, you would need to make sure the 'bar' object containing the pathname is the same dimensions as the tree-view box below it. 

I also think the the text 'scan results for: ' should be on the same line as the path. In fact, change that would probably make the tests pass without having to change the test cases themselves. 

All in all, these UI changes would need a more 'polish' to be considered an improvement from their current state.

singh1114:
@majurg: I agree on what you suggest about the bar length and the same line thing.

What changes do you suggest for making it "polished"?

pombredanne:
I am closing this and further work on the UI can take place in  https://github.com/nexB/aboutcode-manager that should be used for visualization of scans instead.

PR #526, opened 2017-03-01
UI: changes the UI of the html-app created by the --format html-app ...command

singh1114:
Signed-off-by: Ranvir Singh <ranvir.singh1114@gmail.com>

### DIFF
--- a/src/formattedcode/templates/html-app/template.html
+++ b/src/formattedcode/templates/html-app/template.html
@@ -59,16 +59,13 @@
       </div>
       -->
       <div class="row">
-        <div class="row">
-          <div class="col-md-12" style = "margin: 0px 0px 20px 10px;">
-             <strong>scan results for:
-               <br>
-               <div class = "col-md-4" style="border: 1px solid #ccc; border-radius: 5px; padding: 2px 20px 2px 20px; margin-top:5px; ">
-                 {{ scanned_path }}
-               </div>
-             </strong>
+        <div class="row" id = "scan-result-header">
+          <div class="col-md-4">
+            <strong>scan results for:</strong>
           </div>
-
+          <div class = "col-md-8" id = "file-path">
+            <p>{{ scanned_path }}</p>
+          </div>
         </div>
       <div class="col-md-4" id="leftCol">
         <div id="tree" style="overflow-x:scroll; overflow-y:auto;"></div>
--- a/src/formattedcode/templates/html-app/template.html
+++ b/src/formattedcode/templates/html-app/template.html
@@ -37,7 +37,7 @@
     <link href="{{ assets_dir }}/simple-sidebar.css" rel="stylesheet"/>
 </head>
 <body>
-  <nav class="navbar navbar-inverse navbar-static-top">
+  <div class="navbar navbar-inverse navbar-static-top">
     <div class="container-fluid">
       <div class="navbar-header">
           <span class="navbar-brand icon-primary"><a href="https://github.com/nexB/scancode-toolkit/" target="_blank"><strong>ScanCode</strong></a></span>
@@ -47,7 +47,7 @@
           <li><a href="http://www.nexb.com" target="_blank">Made by nexB</a></li>
        </ul>
     </div>
-  </nav>
+  </div>
 
   <div class="container-fluid">
     <div class="row">
--- a/src/formattedcode/templates/html-app/template.html
+++ b/src/formattedcode/templates/html-app/template.html
@@ -37,59 +37,70 @@
     <link href="{{ assets_dir }}/simple-sidebar.css" rel="stylesheet"/>
 </head>
 <body>
-
-<div class="container-fluid">
-    <nav class="navbar navbar-default">
-      <div class="container-fluid">
-        <div class="navbar-header">
-            <span class="navbar-brand icon-primary"><a href="https://github.com/nexB/scancode-toolkit/" target="_blank">ScanCode</a> scan results for: {{ scanned_path }}</span>
-        </div>
-          <ul class="nav navbar-nav navbar-right">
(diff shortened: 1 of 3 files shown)


pombredanne:
@singh1114 Thanks! ... Some tests are afiling though. Did you run the tests locally?

pombredanne:
Also it would be best to avoid keeping or adding dead code in commented HTML blocks.
And I would need to understand what these change do: can you explain this and eventually adding a screenshot before and after?

singh1114:
@pombredanne  
That code was already there so, I thought that it was there for some reason. DO you want me to remove that? I don't know much about the tests I will study about them soon.
scancode

pombredanne:
@singh1114 dead code has never a good reason to say.
Could you post a screenshot of what the UI was before? I am hard pressed to grok what is changed.

singh1114:
@pombreda: I will remove that code in the next commit. 

Screenshot of the UI before:
screenshot

pombredanne:
Please always prefix your commits with the ticket number they refer to, e.g. #525

pombredanne:
Also please craft proper commit message explaining what the commit does, in imperative tense.
not `changes the UI to match the suggestions in #526`
but rather
```
#526 refine html-app UI
 * add XYX
 * move foo by one pixel to the left

Signed-off-by: your name <your email>
```

You can amend/rewrite these messages and force push this branch then.

pombredanne:
So in a nutshell explain what your commit is about :)

singh1114:
Please review and point them out the mistakes.

pombredanne:
@singh1114 Thanks! this starts to look good. I still would like to have a screenshot of the same screen before and after side by side in a comment here to understand visually what you did change exactly.

pombredanne:
These screenshots should me made on your machine for the same data set of course.

singh1114:
Yes! you are right. Here is the screenshot.
scancode

And have you asked for a code review? If yes, then after doing the force push I am unable to see what you have asked for. So please ask for the code review again.

pombredanne:
So this is the screenshot after? what would be the exact same data and page before, on your machine?
I want to see thing side by side :)

singh1114:
Ok Here is it.

Before:
scancodeanother

After:
scancode

pombredanne:
Thanks. LGTM, I will review in details tomorrow.

src/formattedcode/templates/html-app/template.html
```diff
@@ -37,33 +37,33 @@
...
-            <span class="navbar-brand icon-primary"><a href="https://github.com/nexB/scancode-toolkit/" target="_blank">ScanCode</a> scan results for: {{ scanned_path }}</span>
-        </div>
-          <ul class="nav navbar-nav navbar-right">
-            <li><a href="{{ assets_dir }}/help.html" target="_blank">Help</a></li>
-            <li><a href="http://www.nexb.com" target="_blank">Made by nexB</a></li>
-         </ul>
-      </div>
-    </nav>
-    <!--
+  <nav class="navbar navbar-inverse navbar-static-top">
```
pombredanne: I never saw this <nav> tag before? Is this standard on all browsers starting IE8 and up? afaik this is only HTML5. If yes, we cannot use this.

singh1114: We can HTMLshiv to make bootstrap work on IE8+.
https://github.com/aFarkas/html5shiv

pombredanne:
@singh1114 I would like to avoid adding more JS deps to this if possible. Cant's you use non-HTML5 tags instead?

singh1114:
@pombredanne Yes, we can do that. We can keep the same thing and write some special css for IE's.

For example: 
``

source: http://www.monitis.com/blog/cross-browser-compatibility-tips-part-1/

src/formattedcode/templates/html-app/template.html
pombredanne: @singh1114 could you use a plain div instead? this would be simpler and more compatible, right?

singh1114:
@pombredanne We can use a div inside the nav tag. Along with that we can use a special CSS file for IE<9.

I don't know how can I test my changes on IE<9. I have checked on google but there is no server that can test the changes. Do you have any idea about this.

pombredanne:
I meant it would be simpler to eschew  using a nav tag entirely

singh1114:
@pombredanne Ok, I will give it a try and submit a commit by tonight.

tests/formattedcode/test_writers.py
```diff
@@ -142,7 +142,8 @@ def test_scanned_path_is_present_in_html_app_output(monkeypatch):
     assert 'Scanning done' in result.output
     html_file = open(result_file).read()
     assert '<title>ScanCode scan results for: %(test_dir)s</title>' % locals() in html_file
-    assert 'ScanCode</a> scan results for: %(test_dir)s</span>' % locals() in html_file
+    assert '<strong>scan results for:' % locals() in html_file
+    assert '<p>%(test_dir)s</p>' % locals() in html_file
```
pombredanne: I think it would make sense to assert a bit more in a test. What about check the whole HTML file rather that a few tidbit? or at least something more meaty?

pombredanne:
Thanks. This is mostly ready to merge except for some deeper test assertions. I am always going slow on a UI change that may impact many users.

tests/formattedcode/test_writers.py
singh1114: Checking the whole HTML file doesn't make sense. Let's insert something meaty in there. Just a question, How to include next line in the assert statement? I was having some trouble in that part.

pombredanne:
@singh1114 Thanks. LGTM! Will merge soon.

singh1114:
@pombredanne, Which checks are failing?

pombredanne:
There is something failing but it is not your fault.

PR #528, opened 2017-03-01
#36 Add option to ignore files while scanning

### LINKED ISSUE #36: configurable ignores with a `--ignore` option
This is a  follow up of #35 from @K-Rex
We should add an --ignore command line option to point to an ignore file

yashdsaraf:
Fixes #36


src/scancode/cli.py
```diff
@@ -91,6 +91,7 @@
 acknowledgment_text = delimiter + acknowledgment_text
 
 notice = acknowledgment_text.strip().replace('  ', '')
+user_ignored = {}
```
pombredanne: We cannot use a global like that

src/scancode/cli.py
```diff
@@ -185,6 +186,11 @@ def validate_formats(ctx, param, value):
         raise click.BadParameter('Invalid template file: "%(value)s" does not exists or is not readable.' % locals())
     return value
 
+def ignore_files(ctx, param, value):
+    for file in value.split(','):
+        user_ignored.update({
```
pombredanne: same thing: no global please.

src/scancode/cli.py
```diff
@@ -228,6 +234,7 @@ def validate_formats(ctx, param, value):
 @click.option('--diag', is_flag=True, default=False, help='Include additional diagnostic information such as error messages or result details.')
 @click.option('--timeout', is_flag=False, default=DEFAULT_TIMEOUT, type=int, show_default=True, help='Stop scanning a file if scanning takes longer than a timeout in seconds.')
 @click.option('--max-memory', is_flag=False, default=DEFAULT_MAX_MEMORY, type=int, show_default=True, help='Stop scanning a file if scanning requires more than a maximum amount of memory in megabytes.')
+@click.option('-I', '--ignore', is_flag=False, default='', callback=ignore_files, help='Do not scan <input> for these files')
```
pombredanne: please no `-I` uppercase option. Just a single `--ignore` is enough.
Also we need an option that accepts only a single pattern not comma-separated and that can be repeated multiple times.

src/scancode/cli.py
```diff
@@ -532,7 +539,8 @@ def resource_paths(base_path):
     base_path = os.path.abspath(os.path.normpath(os.path.expanduser(base_path)))
     base_is_dir = filetype.is_dir(base_path)
     len_base_path = len(base_path)
-    ignored = partial(ignore.is_ignored, ignores=ignore.ignores_VCS, unignores={})
+    user_ignored.update(ignore.ignores_VCS)
```
pombredanne: No globals please.

src/scancode/cli.py
```diff
@@ -185,6 +186,11 @@ def validate_formats(ctx, param, value):
         raise click.BadParameter('Invalid template file: "%(value)s" does not exists or is not readable.' % locals())
     return value
 
+def ignore_files(ctx, param, value):
+    for file in value.split(','):
```
pombredanne: Do not use a comma separated single option, instead use the proper click option that can accept `--ignore "abc*" --ignore "src/*/foobar` ...

pombredanne:
See also the comments in #36 :)

pombredanne:
To write good commit messages, please read https://github.com/nexB/aboutcode/wiki/Writing-good-commit-messages

src/scancode/cli.py
yashdsaraf: How about I define a callback in the ignore.py https://github.com/nexB/scancode-toolkit/blob/develop/src/commoncode/ignore.py, which can modify a dict in the same file and can be later imported through this function https://github.com/nexB/scancode-toolkit/blob/develop/src/scancode/cli.py#L521
Would this be a viable solution to **not** defining globals?
This will also

pombredanne: I do not think so , but I am not exactly what you mean. Globals are bad in all cases, callback to update a global is equally bad if not worse.

ignores are a plain dict, just use plain dict  and update it in the `scancode()` main function based on user options.

yashdsaraf: Alright, I guess I got that, I'll give it a try and let you know 👍

yashdsaraf: @pombredanne I've [pushed all the changes](https://github.com/yashdsaraf/scancode-toolkit/commit/6401c5b406be5e6bcfb8b2c3246eae14d3782bc7) you suggested, there's just one issue that `fnmatch` does not support double wildcard, is it alright if I add a third party library [glob2](https://github.com/miracle2k/python-glob2) to enable the double wildcard support?

pombredanne:
@yashdsaraf you wrote:
>  is it alright if I add a third party library glob2 to enable the double wildcard support?

Possibly but it does not seem super actively maintained, is it? On the other this seems like  backport of Python3. May be there are other libraries on Pypi for this? To add a library see how this is done in setup.py and the thirdparty dir. I als made a post on the ML commenting on some of this

src/scancode/cli.py
pombredanne: You are still using a global. No globals. Use a function with args

src/scancode/cli.py
pombredanne: No comma separated values

src/scancode/cli.py
pombredanne: You still use globals atm.

pombredanne: Ah, You need to re-push your branch.... https://github.com/yashdsaraf/scancode-toolkit/commit/6401c5b406be5e6bcfb8b2c3246eae14d3782bc7 is not showing up in this PR. Hence all my comments because things here are still looking like you had made no changes :)

yashdsaraf: Oops, I've pushed the changes to the proper branch now :P

src/scancode/cli.py
```diff
@@ -216,6 +216,8 @@ def validate_formats(ctx, param, value):
               help=('Set <output_file> format <style> to one of the standard formats: %s '
                     'or the path to a custom template' % ' or '.join(formats)),
               callback=validate_formats)
+@click.option('--ignore', default='', multiple=True, show_default=True, metavar='<pattern>',
+              help=('Ignore files mentioned in <pattern> while scanning'))
```
pombredanne: You might want to be explicit about what a pattern is? a glob? a path regex?

src/scancode/cli.py
```diff
@@ -279,6 +282,9 @@ def scancode(ctx,
 
     scans_cache_class = get_scans_cache_class()
 
+    # Convert ignore (user defined ignored files list) to a dict
+    user_ignore = dict((patt, 'User ignore: Supplied by --ignore') for patt in ignore)
```
pombredanne: What is `'User ignore: Supplied by --ignore' about?
Also using a dict comprehension `{x: y for x, y in foo}` is more explicit

src/scancode/cli.py
```diff
@@ -313,7 +320,8 @@ def scan(input_path,
          processes=1, timeout=DEFAULT_TIMEOUT, max_memory=DEFAULT_MAX_MEMORY,
          diag=False,
          scans_cache_class=None,
-         strip_root=False):
+         strip_root=False,
+         ignore={}):
```
pombredanne: never ever use a mutable as a default arg in a function, unless you know what you are doing. This is a recipe for a catastrophy:  http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

src/scancode/cli.py
yashdsaraf: Will do. :)

src/scancode/cli.py
yashdsaraf: 'User ignore: Supplied by --ignore' is returned when [debugging](https://github.com/nexB/scancode-toolkit/blob/develop/src/commoncode/fileset.py#L34) is turned on, I took a reference from [these descriptions](https://github.com/nexB/scancode-toolkit/blob/develop/src/commoncode/ignore.py#L95), I know it's not the best description to go with, perhaps you could suggest something better?
About dict comprehensions, that really makes more sense, (I actually didn't know about them), thanks for the suggestion 👍

src/scancode/cli.py
yashdsaraf: That's a very helpful link, I'll make the changes ASAP.

pombredanne:
@yashdsaraf ;)

unit-tests

yashdsaraf:
@pombredanne I'm already done with the cli tests, could you please point me towards a sample unit test in the project for reference.

pombredanne:
@yashdsaraf you can look in many places outside of test_cli.py
The point is that you should have a function and a test that verifies only this function is working as expected.

tests/scancode/test_ignore_files.py
```diff
@@ -0,0 +1,64 @@
+#
```
pombredanne: Looking good! I will review in details later

tests/scancode/test_ignore_files.py
```diff
@@ -0,0 +1,64 @@
...
+from unittest import TestCase
+from commoncode.ignore import is_ignored
+
+class TestIgnoreFiles(TestCase):
+
+    def test_ignore_glob_path_1(self):
+        test = is_ignored('common/src/test/sample.txt', {
+            '*/src/test/*': 'test ignore'
+            }, {})
+        self.assertTrue(test)
```
pombredanne: we prefer using plain `assert` style for tests.

tests/scancode/test_ignore_files.py
```diff
@@ -0,0 +1,64 @@
...
+#  ScanCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/scancode-toolkit/ for support and download..
+
+from unittest import TestCase
+from commoncode.ignore import is_ignored
+
+class TestIgnoreFiles(TestCase):
+
+    def test_ignore_glob_path_1(self):
+        test = is_ignored('common/src/test/sample.txt', {
```
pombredanne: It is best to align all the args.

tests/scancode/test_ignore_files.py
```diff
@@ -0,0 +1,64 @@
...
+            'sample.txt': 'test ignore'
+            }, {})
+        self.assertTrue(test)
+
+    def test_ignore_glob_file(self):
+        test = is_ignored('common/src/test/sample.txt', {
+            '*.txt': 'test ignore'
+            }, {})
+        self.assertTrue(test)
```
pombredanne: You are missing a line return

tests/scancode/test_ignore_files.py
pombredanne: e.g. something like:
```
test = (
    'common/src/test/sample.txt', 
    {'sample.txt': 'test ignore'},
    {}
)
assert is_ignored(*test)
```

tests/scancode/test_ignore_files.py
yashdsaraf: @pombredanne I've made the changes you requested. Just a humble reminder :)

src/scancode/cli.py
```diff
@@ -357,7 +365,9 @@ def scan(input_path,
 
     # maxtasksperchild helps with recycling processes in case of leaks
     pool = get_pool(processes=processes, maxtasksperchild=1000)
-    resources = resource_paths(input_path)
+    if not ignore:
```
pombredanne: Use rather: `ignore = ignore or {}`

pombredanne:
Also can you squash this PR in a single clean commit?

yashdsaraf:
@pombredanne Yes I'll post the latest commit here with the changes you requested and if it doesn't require any more changes, I'll open a new PR with all these changes squashed into a single commit.

yashdsaraf:
Yes I'll do it in the next commit. Thanks for the review!

src/scancode/cli.py
```diff
@@ -518,7 +541,7 @@ def _scanit(paths, scanners, scans_cache_class, diag,
     return success, rel_path
 
 
-def resource_paths(base_path):
+def resource_paths(base_path, user_ignores):
```
pombredanne: Could you add a few few unit tests for this function, especially now that there would be ignores support?

yashdsaraf: @pombredanne Is it alright if I modify the existing methods in the test_ignore_files.py https://github.com/yashdsaraf/scancode-toolkit/blob/develop/tests/scancode/test_ignore_files.py with asserts for this function? Or is it preferable to write separate methods?

yashdsaraf:
@pombredanne I'll squash the latest commit after your review, so any changes (if required) will be easier to check. 👍

tests/scancode/test_ignore_files.py
```diff
@@ -24,6 +24,13 @@
 
 from unittest import TestCase
 from commoncode.ignore import is_ignored
+from scancode.cli import resource_paths
+from commoncode.testcase import FileDrivenTesting
+from os import path
+
+test_env = FileDrivenTesting()
```
pombredanne: This pattern of using file-driven testing should not be used except in test_cli.py. Subclass instead FileBasedTesting rather than the stdlib TestCase anywhere else.

tests/scancode/test_ignore_files.py
```diff
@@ -74,3 +81,45 @@ def test_ignore_glob_file(self):
             {}
         )
         assert is_ignored(*test)
+
+    def test_resource_paths_1(self):
```
pombredanne: Rather than _1 could you use a prefix in your test name that explains what you are testing?

tests/scancode/test_ignore_files.py
```diff
@@ -74,3 +81,45 @@ def test_ignore_glob_file(self):
...
+            test_dir,
+            {'*.doc': 'test ignore'}
+        )
+        expected = [
+            'user',
+            'user/src',
+            'user/src/test',
+            'user/src/test/sample.txt'
+        ]
+        for abs_path, rel_path in resource_paths(*test):
```
pombredanne: Why using tuple unpacking rather than explicit named args? Also why using a global test_dir? There is no reason that each unit tests would be dependent on a global and dependent on each other.

pombredanne: Also why testing in a loop? Make this a list and assert that the list is what you expect as a whole.

yashdsaraf: The generator object returned from the resource_paths object seems to have different order of elements on different operating systems.
For e.g one of the previous CI tests failed due to this
`['user','user/src','user/src/test.txt'] != ['user','user/src/test.txt','user/src']`
This test passed in windows and mac systems but failed on linux.

pombredanne: > The generator object returned from the resource_paths object seems to have different order of elements on different operating systems.

Yes, the order is not stable. So you can just sort your results as here we do not care for ordering. And therefore make it a simple sorted list-based test

src/scancode/cli.py
```diff
@@ -131,6 +131,20 @@ def print_about(ctx, param, value):
...
     scancode -f json -l -c samples/zlib/ > scan.json
 
+Scan a directory while ignoring a single file. Print scan results to stdout as JSON:
+
+    scancode --ignore README samples/
+
+Scan a directory while ignoring all files with txt extension. Print scan results to
+stdout as JSON:
+
+    scancode --ignore *.txt samples/
```
pombredanne: Are you sure this would work? a *nix shell would expand this pattern. Actually you should add a test doing just that and it will likely fail if there are more than one .txt file in the directory though the tests may be hard to exactly reproduce unquoted shell expansion. You can verify this in any case by trying this command your self with a dir that has more than one .txt file.

pombredanne: This applies to your other examples BTW>

src/scancode/cli.py
```diff
@@ -216,6 +230,8 @@ def validate_formats(ctx, param, value):
               help=('Set <output_file> format <style> to one of the standard formats: %s '
                     'or the path to a custom template' % ' or '.join(formats)),
               callback=validate_formats)
+@click.option('--ignore', default='', multiple=True, show_default=True, metavar='<pattern>',
```
pombredanne: Why default to an empty strings instead of None?

src/scancode/cli.py
```diff
@@ -131,6 +131,20 @@ def print_about(ctx, param, value):
 
     scancode -f json -l -c samples/zlib/ > scan.json
 
+Scan a directory while ignoring a single file. Print scan results to stdout as JSON:
+
+    scancode --ignore README samples/
+
+Scan a directory while ignoring all files with txt extension. Print scan results to
+stdout as JSON:
```
pombredanne: We need some help text that explain which glob characters are allowed in a pattern and how they behave.

src/scancode/cli.py
```diff
@@ -279,6 +296,9 @@ def scancode(ctx,
 
     scans_cache_class = get_scans_cache_class()
 
+    # Convert ignore (user defined ignored files list) to a dict
```
pombredanne: That comment is not needed

src/scancode/cli.py
yashdsaraf: Thanks for pointing that out, I was doing most of the test on BASH which by default did not expand globs, but it failed in a ZSH. I'll change the examples to use quoted patterns.
Perhaps adding a note about the quotes would help in the `examples_text` doc?

src/scancode/cli.py
yashdsaraf: A Nonetype object was throwing errors back when I added the ignore option to be parsed as a comma-separated list, I forgot to change it back in later commits. I'll change it in the next commit. :)

src/scancode/cli.py
```diff
@@ -131,6 +131,29 @@ def print_about(ctx, param, value):
 
     scancode -f json -l -c samples/zlib/ > scan.json
 
+Scan a directory while ignoring a single file. Print scan results to stdout as JSON:
```
pombredanne: One question, and this may need to be explained here and you may need a test for it: is the README ignored at the root of the project only? or anywhere in the scanned tree? and can I use a path? e.g. is the glob applied to the path as a string? or to the filenames only?

src/scancode/cli.py
```diff
@@ -216,6 +239,8 @@ def validate_formats(ctx, param, value):
               help=('Set <output_file> format <style> to one of the standard formats: %s '
                     'or the path to a custom template' % ' or '.join(formats)),
               callback=validate_formats)
+@click.option('--ignore', default=None, multiple=True, metavar='<pattern>',
+              help=('Ignore files mentioned in the GLOB <pattern> while scanning. Check --examples for more information'))
```
pombredanne: What about this:
`Ignore files matching <pattern>.`
The check example part is not needed.

src/scancode/cli.py
yashdsaraf: It will be ignored anywhere in the scanned tree. I'm sorry I don't fully understand what meant by this `is the glob applied to the path as a string? or to the filenames only?`.
Glob patterns can be applied to path and filenames i.e */src/test/* and *.doc both will work as expected (if that's what your were asking) :)

src/scancode/cli.py
yashdsaraf: Should I mention more about GLOB support here?

src/scancode/cli.py
pombredanne: eg. if I use `--ignore "foo*bar"` and I have a path `jogn/foo/fubar/bar` will this be ignored? This is what you need to clarify for the help examples.

src/scancode/cli.py
pombredanne: glob is a somewhat arcane name. Mention it instead in the help examples, possibly with a link to https://en.wikipedia.org/wiki/Glob_(programming)

src/scancode/cli.py
yashdsaraf: Ah, I get it now. No the glob patterns cannot be applied to path as strings.
I'll add a note for the same in the examples.

pombredanne:
Please make sure there is a tests for that. If not, then you should add one

pombredanne:
@yashdsaraf thanks for the rebase. I will merge soon

pombredanne:
@yashdsaraf I rebase it all and will merge shortly now

pombredanne:
This has been merged in develop

pombredanne:
See https://github.com/nexB/scancode-toolkit/commit/80935d973d4f1061949b680b62882e241d35d350

PR #530, opened 2017-03-02
Fix typos

jwilk:
Found using [mwic](http://jwilk.net/software/mwic).


pombr:
Thanks! every little bit counts.

pombredanne:
@jwilk Do you create these PR by hand btw?

jwilk:
Yes.

Issue #524, opened 2017-02-28
json2csv.py should not prepend a root directory on the resource path


JonoYang:
Now that root paths are reintroduced into ScanCode resource paths, json2csv.py does not need to prepend the resource path with `/code` anymore.

pombredanne:
I think this should become/stay as an option. The output may of may not have a root dir as this is as option on the scancode side.

JonoYang:
I agree that this should become an option. By default, json2csv.py should not add a root directory, as the new version of ScanCode reintroduces root directories. In the cases where the scan result does not have a root directory, then json2csv.py should have an option, like `--prepend-root`.

yasharmaster:
I want to work on this issue. Can someone guide me where to get started?

pombredanne:
@yasharmaster Thanks. You would need to wait for input from @JonoYang for this.

JonoYang:
@yasharmaster 

`json2csv.py` converts ScanCode scan result data from JSON to CSV. For legacy compatibility, the path value is prepended with `/code`. Since the newer releases of ScanCode, we do not need to prepend `/code` anymore, but this should be an option if we ever need to do so. 

You should start by finding the script here:
- https://github.com/nexB/scancode-toolkit/blob/develop/etc/scripts/json2csv.py

You should look at the `flatten_scan()` and `cli()` functions to see how the path value is being changed and to see where to add in options.

kartiksibal:
@JonoYang When exactly is **json2csv.py** being called?

jack17529:
@kartiksibal  dude I have nearly solved it please take another issue ,I gonna submit P.R.

pombredanne:
@kartiksibal @jack17529 there is already a work in progress PR by @yasharmaster who is working on this: a different PR is always welcomed, but duplicating work does not make sense especially on such a small feature. 

You should collaborate instead such as providing comments and feedback and review on the current PR.

pombredanne:
And yet another PR #541 that I had to close ass duplicate from @sudeepb02 
... Please have mercy on us: one at a time!

kartiksibal:
Haha. Also, @pombredanne It'd be awesome if you could assign the issue, to whoever is working on it. It'd avoid such confusions in the future. Cheers! :-D

pombredanne:
@kartiksibal issues in GH can only be assigned to actual project collaborator, e.g. these that have earned comitter right from the project standpoint. This is not great but this is the way for now.
[...]

PR #535, opened 2017-03-03
#524 Do not prefix resource paths with a "/code" root directory

agneet42:
yet to put in the conditions  where '/code' will be pre-appended to path.

### DIFF
--- a/etc/scripts/test_json2csv.py
+++ b/etc/scripts/test_json2csv.py
@@ -45,7 +45,7 @@ def load_csv(location, ignore_date=False):
     with codecs.open(location, 'rb', encoding='utf-8') as csvin:
         reader = unicodecsv.DictReader(csvin)
         fields = reader.fieldnames
-        values = list(reader)
+        values = sorted(reader)
         return fields, values
 
 


pombredanne:
@agneet42 Thanks for this PR. Do you min rewording your commit as per https://github.com/nexB/aboutcode/wiki/Writing-good-commit-messages ? Thanks!

agneet42:
on it.

agneet42:
please review the changes @pombredanne. also only pre-appendment of '/code' has been taken care of. conditions about when '/code' is to be appended haven't been imposed yet.

pombredanne:
@agneet42 Thanks for the updates: we do not use UPPERCASE subject in commits.... we use Capitalized subject. Also do not refereencec the PR number in the commit and PR title, but the corresponding issue number.
So rather than:
`#535 REMOVE PRE-APPENDMENT OF '/CODE' TO PATH`
your subject could be something like:
`#524 Do not prefix resource paths with a "/code" root directory`

pombredanne:
@JonoYang yours to review otherwise :)

pombredanne:
@agneet42 also the tests are failing i.e. you broke the build in this branch and this will need to be feixed before it can be merged

etc/scripts/json2csv.py
```diff
@@ -246,7 +244,7 @@ def cli(json_input, csv_output):
 
     JSON_INPUT is either a ScanCode json format scan or the data.json file from a ScanCode html-app format scan.
 
-    Paths will be prefixed with '/code/' to provide a common base directory for scanned resources.
+    #Paths will be prefixed with '/code/' to provide a common base directory for scanned resources.
```
JonoYang: There should be an option to prepend the path value with `/code` in case the path values in a scan does not contain a root directory.

JonoYang:
@agneet42 @pombredanne 

You have made the correct change. Since the path values are not prepended with `/code` anymore, you will need to update the tests so that the expected path values do not start with `/code`. 

The tests are found here: https://github.com/nexB/scancode-toolkit/blob/develop/etc/scripts/test_json2csv.py

As I mentioned in #524, one thing that I think is a good idea is an option to prepend `/code` to path values. This is for the case where we have a scan with no root path.

etc/scripts/json2csv.py
jack17529: @JonoYang so if there is a directory named home then do nothing to path or else prepend it by /code.
If that you mean in issue 524 then I have done it

JonoYang: @jack17529 No. This option, which for now I will call `--prepend-root`, should prepend the path value with `/code` when the option is called. When the `--prepend-root` option is not called, the path value should not be modified.

jack17529: @JonoYang  so make function named --prepend-root  and with some condition.

Then which condition should I put? Please help

pombredanne:
@agneet42 I think you still need to refine and amend your commit message and the title of this PR. And you have not yet addressed the review comments from @JonoYang nor added tests. And some tests are not yet passing. Just in case you missed something

agneet42:
Yes sir. I'll be doing it just a few hours.
Sorry for the delay.

pombredanne:
no rush. I just wanted to make sure you would be following up after all the drama of so many folks wanting to work on this PR and ticket on your behalf ;)

pombredanne:
@agneet42 The tests are not passing on Linux and Mac: https://travis-ci.org/nexB/scancode-toolkit/builds/207482261

agneet42:
@pombredanne I am trying to resolve it.

pombredanne:
@agneet42 any update?

agneet42:
@pombredanne , I am working on a few other issues here itself, do you want me to finish this and do the others, later?

pombredanne:
@agneet42 The comments that I made in #562 on the commit messages apply here too. And you still have the tests not passing here. You should not have multiple PR for the same issue.

jack17529:
why am I getting updates everyday I am already done with the project

PR #538, opened 2017-03-04
#389 Prevent progressbar spill over new line

### LINKED ISSUE #389: Incorrect progress bar display
Sometimes the progressbar goes bad and display each message on a new line
It could be because the file name displayed exceed the terminal width and spills over the next line, detroying the progressbar curses sync along the way.
:
```
[-----------------------------------#]  95981  Scanned: c                                                                                                       [-----------------------------------#]  95982  Scanned: f                                                                                                       [-----------------------------------#]  95983  Scanned: 2                                                                                                       [-----------------------------------#]  95984  Scanned: 6                                                                                                       [-----------------------------------#]  95985  Scanned: b                                                                                                       [-----------------------------------#]  95986  Scanned: build                                                                                                   [-----------------------------------#]  95987  Scanned: release                                                                                                 [-----------------------------------#]  95988  Scanned: conf                                                                                                    [-----------------------------------#]  95989  Scanned: scripts                                                                                                 [-----------------------------------#]  95990  Scanned: MANIFEST.in.dev                                                                                         [-----------------------------------#]  95991  Scanned: configure.py                                                                                            [-----------------------------------#]  95992  Scanned: scancode-toolkit.spec                                                                                   [-----------------------------------#]  95993  Scanned: release.sh                                                                                              [-----------------------------------#]  95994  Scanned: setup.cfg.dev                                                                                           [-----------------------------------#]  95995  Scanned: MANIFEST.in.release                                                                                     [-----------------------------------#]  95996  Scanned: dev                                                                                                     [-----------------------------------#]  95997  Scanned: base.txt                                                                                                [-----------------------------------#]  95998  Scanned: base.py                                                                                                 [-----------------------------------#]  95999  Scanned: base.txt                                                                                                [-----------------------------------#]  96000  Scanned: irc-notify.py.ABOUT                                                                                     [-----------------------------------#]  96001  Scanned: test_json2csv.py                                                                                        [-----------------------------------#]  96002  Scanned: irc-notify.py                                                                                           [-----------------------------------#]  96003  Scanned: README.rst                                                                                              [-----------------------------------#]  96004  Scanned: dejacode.py                                                                                             [-----------------------------------#]  96005  Scanned: json2csv.py                                                                                             [-----------------------------------#]  96006  Scanned: testdata                                                                                                [-----------------------------------#]  96007  Scanned: json2csv                                                                                                [-----------------------------------#]  96008  Scanned: no_keys.json                                                                                            [-----------------------------------#]  96009  Scanned: minimal.json                                                                                            [-----------------------------------#]  96010  Scanned: gpl-2.0.LICENSE                                                                                         [-----------------------------------#]  96011  Scanned: minimal.csv                                                                                             [-----------------------------------#]  96012  Scanned: setup.cfg.release                                                                                       [-----------------------------------#]  96013  Scanned: key_order.json                                                                                          [-----------------------------------#]  96014  Scanned: minimal.json-expected                                                                                   [-----------------------------------#]  96015  Scanned: minimal.json-strip-expected                                                                             [-----------------------------------#]  96016  Scanned: format_html_app_data.json                                                                               [-----------------------------------#]  96017  Scanned: format_json_scan.json                                                                                   [-----------------------------------#]  96018  Scanned: full.json                                                                                               [-----------------------------------#]  96019  Scanned: full.json-expected                                                                                      [####################################]  96019                        
```

yasharmaster:
Fixes #389 
Signed-off-by: Yash Sharma yasharmaster@gmail.com

### DIFF
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -407,6 +407,8 @@ def scan(input_path,
                 if item:
                     _scan_success, _scanned_path = item
                     _progress_line = verbose and _scanned_path or fileutils.file_name(_scanned_path)
+                    if not verbose:
+                    	_progress_line = shorten_filename(_progress_line)
                     return style('Scanned: ') + style(_progress_line, fg=_scan_success and 'green' or 'red')
 
             scanning_errors = []
@@ -465,6 +467,26 @@ def scan(input_path,
     cached_scan = scans_cache_class()
     root_dir = _get_root_dir(input_path, strip_root)
     return files_count, cached_scan.iterate(scans, root_dir)
+
+
+def shorten_filename(filename='', max_length=25, number_of_dots=3):
+    """
+    Return the original filename if it is short enough
+    otherwise return its shortened version
+    """
+    if filename is None:
+        return ''
+    filename = fileutils.file_name(filename)
+    if len(filename) <= max_length:
+        return filename
+    base_name, extension = fileutils.splitext(filename)
+    remaining_length = max_length - len(extension) - number_of_dots
+    prefix_and_suffix_length = abs(remaining_length // 2)
+    prefix = base_name[:prefix_and_suffix_length]
+    ellipsis = number_of_dots*'.'
+    suffix = base_name[-prefix_and_suffix_length:]
+    shortened_filename = "{prefix}{ellipsis}{suffix}{extension}".format(**locals())
+    return shortened_filename
 
 
 def _get_root_dir(input_path, strip_root=False):
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -469,10 +469,9 @@ def scan(input_path,
     return files_count, cached_scan.iterate(scans, root_dir)
 
 
-def shorten_filename(filename='', max_length=25, number_of_dots=3):
+def shorten_filename(filename, max_length=25):
     """
-    Return the original filename if it is short enough
-    otherwise return its shortened version
+    Return a truncated `filename` usable for fixed width display if `filename` is longer than `max_length`.
     """
     if filename is None:
         return ''
@@ -480,6 +479,7 @@ def shorten_filename(filename='', max_length=25, number_of_dots=3):
     if len(filename) <= max_length:
         return filename
     base_name, extension = fileutils.splitext(filename)
+    number_of_dots = 3
     remaining_length = max_length - len(extension) - number_of_dots
     prefix_and_suffix_length = abs(remaining_length // 2)
     prefix = base_name[:prefix_and_suffix_length]
--- a/tests/scancode/data/long_file_name/0123456789012345678901234567890123456789.c
+++ b/tests/scancode/data/long_file_name/0123456789012345678901234567890123456789.c
@@ -1,0 +1,1 @@
+/* Copyright © 2000 ACME, Inc., All Rights Reserved */
--- a/tests/scancode/test_cli.py
+++ b/tests/scancode/test_cli.py
@@ -28,6 +28,7 @@ from __future__ import division
 
 import json
 import os
+from unittest import TestCase
 from unittest.case import skipIf
 
 import click
@@ -102,6 +103,17 @@ def test_verbose_option_with_copyrights(monkeypatch):
     assert 'copyright_acme_c-c.c' in result.output
     assert os.path.exists(result_file)
     assert len(open(result_file).read()) > 10
+
+
+def test_scan_can_handle_long_file_names(monkeypatch):
+    monkeypatch.setattr(click._termui_impl, 'isatty', lambda _: True)
+    test_dir = test_env.get_test_loc('long_file_name/0123456789012345678901234567890123456789.c')
+    runner = CliRunner()
+    result_file = test_env.get_temp_file('json')
+    result = runner.invoke(cli.scancode, ['--copyright', test_dir, result_file], catch_exceptions=True)
+    assert result.exit_code == 0
+    assert 'Scanning done' in result.output
+    assert '0123456789...0123456789.c' in result.output
 
 
 def test_license_option_detects_licenses(monkeypatch):
@@ -452,3 +464,41 @@ def test_scan_can_run_from_other_directory():
         print(stderr)
     assert rc == 0
     check_scan(test_env.get_test_loc(expected_file), result_file, strip_dates=True, regen=False)
+
+
+class TestShortenFilename(TestCase):
+
+    def test_filename_is_shortened(self):
+        test = cli.shorten_filename(filename='012345678901234567890123.c')
+        expected = '0123456789...4567890123.c'
+        assert expected == test
+
+
+    def test_filename_is_not_shortened(self):
+        test = cli.shorten_filename(filename='01234567890123456789012.c')
+        expected = '01234567890123456789012.c'
+        assert expected == test
+
+
+    def test_with_empty_args(self):
+        test = cli.shorten_filename()
+        expected = ''
+        assert expected == test
+
+
+    def test_with_none_filename(self):
+        test = cli.shorten_filename(filename=None)
+        expected = ''
+        assert expected == test
+
+
+    def test_without_extension(self):
+        test = cli.shorten_filename(filename='012345678901234567890123456')
+        expected = '01234567890...67890123456'
+        assert expected == test
+
+
+    def test_with_full_path(self):
+        test = cli.shorten_filename(filename='C/Documents_and_Settings/Boki/Desktop/head/patches/drupal6/drupal.js')
+        expected = 'drupal.js'
+        assert expected == test
--- a/tests/scancode/test_cli.py
+++ b/tests/scancode/test_cli.py
@@ -468,37 +468,47 @@ def test_scan_can_run_from_other_directory():
 
 class TestShortenFilename(TestCase):
 
-    def test_filename_is_shortened(self):
-        test = cli.shorten_filename(filename='012345678901234567890123.c')
-        expected = '0123456789...4567890123.c'
+    def test_filename_larger_than_max_length_shortened(self):
+        test = cli.shorten_filename(filename='0123456789012345678901234.c')
+        expected = '0123456789...5678901234.c'
         assert expected == test
 
-
-    def test_filename_is_not_shortened(self):
+    def test_filename_at_max_length_not_shortened(self):
         test = cli.shorten_filename(filename='01234567890123456789012.c')
         expected = '01234567890123456789012.c'
         assert expected == test
 
-
-    def test_with_empty_args(self):
-        test = cli.shorten_filename()
-        expected = ''
+    def test_filename_smaller_than_max_length_not_shortened(self):
+        test = cli.shorten_filename(filename='0123456789012345678901.c')
+        expected = '0123456789012345678901.c'
         assert expected == test
-
 
     def test_with_none_filename(self):
         test = cli.shorten_filename(filename=None)
         expected = ''
         assert expected == test
 
-
-    def test_without_extension(self):
+    def test_filename_without_extension(self):
         test = cli.shorten_filename(filename='012345678901234567890123456')
         expected = '01234567890...67890123456'
         assert expected == test
 
-
-    def test_with_full_path(self):
+    def test_posix_path_without_shortening(self):
         test = cli.shorten_filename(filename='C/Documents_and_Settings/Boki/Desktop/head/patches/drupal6/drupal.js')
         expected = 'drupal.js'
+        assert expected == test
+
+    def test_posix_path_with_shortening(self):
+        test = cli.shorten_filename(filename='C/Documents_and_Settings/Boki/Desktop/head/patches/drupal6/012345678901234567890123.c')
+        expected = '0123456789...4567890123.c'
+        assert expected == test
+
+    def test_non_posix_path_without_shortening(self):
+        test = cli.shorten_filename(filename='C\\Documents_and_Settings\\Boki\\Desktop\\head\\patches\\drupal6\\drupal.js')
+        expected = 'drupal.js'
+        assert expected == test
+
+    def test_non_posix_path_with_shortening(self):
+        test = cli.shorten_filename(filename='C\\Documents_and_Settings\\Boki\\Desktop\\head\\patches\\drupal6\\012345678901234567890123.c')
+        expected = '0123456789...4567890123.c'
         assert expected == test
--- a/tests/scancode/test_cli.py
+++ b/tests/scancode/test_cli.py
@@ -512,3 +512,6 @@ class TestShortenFilename(TestCase):
         test = cli.shorten_filename(filename='C\\Documents_and_Settings\\Boki\\Desktop\\head\\patches\\drupal6\\012345678901234567890123.c')
         expected = '0123456789...4567890123.c'
         assert expected == test
+
+    def test_with_no_args(self):
+        self.assertRaises(Exception, cli.shorten_filename)


yasharmaster:
@pombredanne I have hard coded the allowed size of a filename to be 20 characters.
Please review the changes.

src/scancode/cli.py
```diff
@@ -382,6 +382,9 @@ def scan_event(item):
                     return ''
                 if item:
                     _scan_success, _scanned_path = item
+                    if len(_scanned_path)>20:
```
pombredanne: Your spacing is not right. We use PEP8 for code standard. eg. `len(scanned_path) > 20`

src/scancode/cli.py
```diff
@@ -382,6 +382,9 @@ def scan_event(item):
...
+                    	current_filename, current_file_extension = os.path.splitext(_scanned_path)
```
pombredanne: Please consider the use of the `commoncode.fileutils` functions instead of splitext

src/scancode/cli.py
```diff
@@ -382,6 +382,9 @@ def scan_event(item):
...
+                    	_scanned_path = current_filename[:3] + '...' + current_filename[-3:] + current_file_extension
```
pombredanne: keeping only three characters at start and end does not sound right.
IMHO you should also isolate this in its own function so it can be tested.

yasharmaster:
@pombredanne please review the changes. I have -
[*] used `commoncode.fileutils` instead of `splitext`
[*] implemented file name shortening in a separate function
[*] set max filename size to be 25 chars, more than 25 was spilling progress bar in my system
[*] improved shortened filename by allowing more than 3 chars from beginning and the end

src/scancode/cli.py
```diff
@@ -443,6 +446,20 @@ def scan_event(item):
     return files_count, cached_scan.iterate(scans, root_dir)
 
 
+def _shorten_file_path(file_path, max_length):
```
pombredanne: do not use a _ prefix here.

pombredanne:
ping.

yasharmaster:
@pombred

yasharmaster:
I am travelling right now. Will finish these edits by tonight.

yasharmaster:
@pombredanne Default terminal width is 80 chars across all OS and the text before the filename spans across 55 characters. So, only 25 chars can be safely allowed for the filename if we hardcode the value. Otherwise we can do something like this https://gist.github.com/jtriley/1108174 to find current terminal width. What do you think?

yasharmaster:
@pombredanne please review

src/scancode/cli.py
```diff
@@ -44,6 +44,8 @@
 import click
 from click.termui import style
 
+from math import floor
```
pombredanne: This is a stdlib import and should go with other stdlib imports, before `import os`

src/scancode/cli.py
```diff
@@ -382,6 +384,7 @@ def scan_event(item):
                     return ''
                 if item:
                     _scan_success, _scanned_path = item
+                    _scanned_path = shorten_filename(_scanned_path)
```
pombredanne: this should only be done if we are not verbose. Note also that you are never passing a filename but a path. So your assumptions are not right in your function.

src/scancode/cli.py
```diff
@@ -443,6 +446,26 @@ def scan_event(item):
     return files_count, cached_scan.iterate(scans, root_dir)
 
 
+def shorten_filename(filename, max_length=25):
+	"""
+	Return the original filename if it is short enough
+	otherwise return its shortened version 
+	"""
+	if len(filename) <= max_length:
+		shortened_filename = filename
```
pombredanne: why not just return early?

src/scancode/cli.py
```diff
@@ -443,6 +446,26 @@ def scan_event(item):
...
+	else:
```
pombredanne: if you return early above, `else` is no longer needed

src/scancode/cli.py
```diff
@@ -443,6 +446,26 @@ def scan_event(item):
...
+		file_base_name = fileutils.file_base_name(filename)
+		file_extension = fileutils.file_extension(filename)
+		number_of_dots = 3
+		remaining_length = max_length - len(file_extension) - number_of_dots
+		number_of_beginning_chars = int(floor(remaining_length/2))
```
pombredanne: pep8 spacing please. using abs and  a floor `//` division would  be simpler and clearer and more explicit:
`abs(remaining_length // 2)`

Also I am not sure that `number_of_beginning_chars` is a good variable name as you use it to compute both the prefix and suffix length to keep from the base name

src/scancode/cli.py
```diff
@@ -382,6 +384,7 @@ def scan_event(item):
                     return ''
                 if item:
                     _scan_success, _scanned_path = item
+                    _scanned_path = shorten_filename(_scanned_path)
                     _progress_line = verbose and _scanned_path or fileutils.file_name(_scanned_path)
```
pombredanne: This code cannot work  correctly with the new code that you injected?

src/scancode/cli.py
```diff
@@ -443,6 +446,26 @@ def scan_event(item):
     return files_count, cached_scan.iterate(scans, root_dir)
 
 
+def shorten_filename(filename, max_length=25):
+	"""
+	Return the original filename if it is short enough
```
pombredanne: You are receiving a path and not a name. So your docstring is misleading

src/scancode/cli.py
```diff
@@ -443,6 +446,26 @@ def scan_event(item):
...
+	otherwise return its shortened version 
+	"""
+	if len(filename) <= max_length:
+		shortened_filename = filename
+	else:
+		file_base_name = fileutils.file_base_name(filename)
+		file_extension = fileutils.file_extension(filename)
```
pombredanne: use instead  something like 
`base_name, ext = fileutils.splitext(filename)` .... but again you did not receive a filename but a path (though it does not matter to splitext)

src/scancode/cli.py
```diff
@@ -443,6 +446,26 @@ def scan_event(item):
...
+		number_of_dots = 3
+		remaining_length = max_length - len(file_extension) - number_of_dots
+		number_of_beginning_chars = int(floor(remaining_length/2))
+		shortened_filename = (file_base_name[:number_of_beginning_chars] + 
+							number_of_dots*'.' + 
+							file_base_name[-number_of_beginning_chars:] +
```
pombredanne: Your pep8 spacing is not correct in this code block

tests/scancode/data/long_file_name/0123456789012345678901234567890123456789.c
```diff
@@ -0,0 +1,1967 @@
+/* deflate.c -- compress data using the deflation algorithm
```
pombredanne: Why use such a big file for tests when you only care about the name? and not the content? This is wasted space.

tests/scancode/data/long_file_name/expected.json
```diff
@@ -0,0 +1,36 @@
+{
```
pombredanne: You should not care about the detailed scan results when you tests name shortening in the CLI ui.

tests/scancode/test_cli.py
```diff
@@ -169,6 +169,15 @@ def test_scan_info_does_collect_infos_with_root(monkeypatch):
...
 
+def test_scan_can_handle_long_file_names(monkeypatch):
+    monkeypatch.setattr(click._termui_impl, 'isatty', lambda _: True)
+    test_dir = test_env.get_test_loc('long_file_name/0123456789012345678901234567890123456789.c')
+    runner = CliRunner()
+    result_file = test_env.get_temp_file('json')
+    result = runner.invoke(cli.scancode, ['--copyright', test_dir, result_file], catch_exceptions=True)
+    assert result.exit_code == 0
+    assert 'Scanning done' in result.output
+    check_scan(test_env.get_test_loc('long_file_name/expected.json'), result_file, regen=False)
```
pombredanne: This is not needed. We do not care about the scan. On the other hand you make no assertions on the shortening of the name in the UI, so you are not testing anything there

src/scancode/cli.py
pombredanne: You should prefer a more explicit interpolation or format. Something like `'{start}...{end}'.format(**locals())` were you compute `start` and `end` before.

pombredanne:
@yasharmaster When you are done with updating your code, you will also need to squash your commit in one clean commit. But only then.

yasharmaster:
@pombredanne Please review. Should I send a new PR with this last single commit ?

pombredanne:
@yasharmaster nope, you can just rebase, squash and force push to that same branch of yours. This will land right here in the same PR aqlright.

yasharmaster:
@pombredanne Can you please point me towards the existing unit tests in the repo? Also the location where unit tests of shorten filename should go? Thanks

pombredanne:
Just look anywhere in the tests/ directory. There are like 1000's of them. You can add your tests alright in the same test_cli.py module
An example of a simple unit tests could be this: https://github.com/nexB/scancode-toolkit/blob/21b6c8456f6cfd42ffc07411312cc489adec26d5/tests/commoncode/test_fileutils.py#L297

yasharmaster:
@pombredanne Are these unit tests sufficient? I have now addressed all the comments from your previous review.

tests/scancode/test_cli.py
```diff
@@ -463,3 +464,17 @@ def test_scan_can_run_from_other_directory():
         print(stderr)
     assert rc == 0
     check_scan(test_env.get_test_loc(expected_file), result_file, strip_dates=True, regen=False)
+
+
+class TestCLI(TestCase):
+
+	def test_shorten_filename(self):
```
pombredanne: You should use 4 spaces. not 8

tests/scancode/test_cli.py
```diff
@@ -463,3 +464,17 @@ def test_scan_can_run_from_other_directory():
...
+	    assert '0123456789...0123456789.c' == cli.shorten_filename('0123456789012345678901234567890123456789.c')
+	    assert '012345.c' == cli.shorten_filename('012345.c')
+	    assert '012345678901234567890.cpp' == cli.shorten_filename('012345678901234567890.cpp')
```
pombredanne: you should add tests for limit args: empty, None, list, string of a length that is and a length that is not shortened right around your 25 limit.  It may help to use variables and give names to make things more explicit. Also you need to test things without extensions and full paths.
In general it is better to create one test function for one test, though the example I gave you did not follow that pattern.

src/scancode/cli.py
```diff
@@ -383,6 +384,8 @@ def scan_event(item):
                 if item:
                     _scan_success, _scanned_path = item
                     _progress_line = verbose and _scanned_path or fileutils.file_name(_scanned_path)
+                    if not verbose:
+                    	_progress_line = shorten_filename(_progress_line)
```
yasharmaster: @pombredanne The logic here is such that `_progress_line` will contain the file path only if `verbose` is set to True. In that case `shorten_filename()` will not be called at all. So argument passed to  `shorten_filename()` will always be filename only and not the file path. So why are we testing `shorten_filename()` with a file path?

tests/scancode/test_cli.py
yasharmaster: @pombredanne what should be the output of shorten filename if a list is passed as the filename?

src/scancode/cli.py
pombredanne: good point.

tests/scancode/test_cli.py
pombredanne: what do you think? probably an exception

yasharmaster:
@pombredanne I have merged all commits into one!

tests/scancode/test_cli.py
```diff
@@ -104,6 +105,17 @@ def test_verbose_option_with_copyrights(monkeypatch):
...
 
+def test_scan_can_handle_long_file_names(monkeypatch):
+    monkeypatch.setattr(click._termui_impl, 'isatty', lambda _: True)
+    test_dir = test_env.get_test_loc('long_file_name/0123456789012345678901234567890123456789.c')
+    runner = CliRunner()
+    result_file = test_env.get_temp_file('json')
+    result = runner.invoke(cli.scancode, ['--copyright', test_dir, result_file], catch_exceptions=True)
+    assert result.exit_code == 0
+    assert 'Scanning done' in result.output
+    assert '0123456789...0123456789.c' in result.output
```
pombredanne: Perfect. You are getting it.

tests/scancode/test_cli.py
```diff
@@ -452,3 +464,41 @@ def test_scan_can_run_from_other_directory():
...
     check_scan(test_env.get_test_loc(expected_file), result_file, strip_dates=True, regen=False)
+
+
+class TestShortenFilename(TestCase):
+
+    def test_filename_is_shortened(self):
+        test = cli.shorten_filename(filename='012345678901234567890123.c')
+        expected = '0123456789...4567890123.c'
+        assert expected == test
+
```
pombredanne: Use only one blank line between class methods. Two lines is only between top level module members

tests/scancode/test_cli.py
```diff
@@ -452,3 +464,41 @@ def test_scan_can_run_from_other_directory():
...
+
+    def test_without_extension(self):
+        test = cli.shorten_filename(filename='012345678901234567890123456')
+        expected = '01234567890...67890123456'
+        assert expected == test
+
+
+    def test_with_full_path(self):
```
pombredanne: Good test. You should also have a test with a non-posix path

pombredanne: And tests with our without posix paths that do shortening or not

tests/scancode/test_cli.py
```diff
@@ -452,3 +464,41 @@ def test_scan_can_run_from_other_directory():
         print(stderr)
     assert rc == 0
     check_scan(test_env.get_test_loc(expected_file), result_file, strip_dates=True, regen=False)
+
+
+class TestShortenFilename(TestCase):
```
pombredanne: You should add tests that test at the boundaries: e.g. when you have exactly the max length that will be shorted., then that minus one and that plus one. This way this would spec your function fully. You may have them already, but if this is the case, the test function names are not explicit enough

src/scancode/cli.py
```diff
@@ -467,6 +469,26 @@ def scan_event(item):
     return files_count, cached_scan.iterate(scans, root_dir)
 
 
+def shorten_filename(filename='', max_length=25, number_of_dots=3):
+    """
+    Return the original filename if it is short enough
```
pombredanne: What about rather:
`Return a truncated `filename` usable for fixed width display if `filename` is longer than `max_length`.`
Also why do we need a default for `filename=''`? Calling this function without args doe not make sense. It should just error out. Nothing special needed. `number of dots` should not be an arg. This is not something I would need to ever modify.

yasharmaster: @pombredanne If we don't keep `filename=''` as default, test with empty args will fail! Should I remove the test with empty args?

pombredanne: Or you can test ass an exercise that the proper exception is raised in this case.

yasharmaster:
@pombredanne Since `shorten_filename()` will not be called without args, therefore I have removed the test with no args. I have added new tests and have renamed them to be more explicit.

pombredanne:
@yasharmaster you wrote in #aboutcode IRC:
> Could anyone please let me know if should raise an exception if the function is called with no args or let it just error out? 

Just let it error out alright. My only point is that if you want to have a comprehensive coverage, then you could test that an exception is thrown in these cases. Not really needed in general but a possible useful excercise for you

yasharmaster:
@pombredanne please review

pombredanne:
This looks good to me. I am merging soon.

pombredanne:
as a side note, I wished we would have something based on the TTY width, such as with `blessings` instead of computing things based on a guessed fixed width.

yasharmaster:
@pombredanne Since the code has been merged after rebase, this can be closed as well.

pombredanne:
@yasharmaster closing as this was merged. Thanks++

PR #541, opened 2017-03-04
#524 Add option to prepend root directory '/code' for legacy support

sudeepb02:
This PR solves issue #524


pombredanne:
@sudeepb02 Thank!. but I am not sure your code can do anything as it is now. Also a good idea before starting to work on a PR is to chime in on the ticket with a comment to ask if none is working on this before. There is already a PR https://github.com/nexB/scancode-toolkit/pull/535 on this exact same ticket, it might be best to avoid duplicated work there...

pombredanne:
Please provide feedback and comment on the PR in progress in #535 instead

Issue #14, opened 2015-07-05
Transparent extraction of archives


pombredanne:
As noted in #3, we do not extract and scan at the same time.

A better way would be to handle internally an archive as if it were a special type of directory (both contain files after all), and when a single archive scan is requested (or when archives are found in a larger scan) we could extract these temporarily to a temp directory,  scan the extract and return the results. This would require a bit more thinking to get it right. 
At a high level a tree with archives would be considered the same as a tree with directories. Archives would become just a special type directory-like containers for more files.

We could expose an `os.walk`-like function that would transparently extract archives to a temp directory and yield a real path and the temp location of a given file

pombredanne:
I need some input on the best way transparent extraction would behave when reporting paths to files inside extracted archives and whether to handle extraction outside or inside/in-place of the scanned tree.

In the current way, `extractcode` needs to be launched separately prior to a scan, an each archive is extracted inside the scanned tree in a new directory created side-by-side with its archives and named after the archive file name with an `-extract` suffix added.
The paths reported in the scan are both the archive path and separately the path to the `-extract` directory and its contained files.

With a transparent archive extraction, we could either extract in place or not:
1. keep the same approach, where the extraction is just integrated in the scan and the archive physically extracted in-place as when running `extractcode`

2. during the scan, extract each archive to a temporary location instead (outside of the scanned tree).
The corresponding extracted files would not exist in the scanned tree but instead extracted out of tree in temp files that would no longer be available once the scan is complete which may be either a good thing or a confusing thing.

With a transparent archive extraction, we could report a real or a "virtual" path to an extracted file in the scanned results:

1. Report `path/to/archive.zip-extract/some/dir/and/file` and also report  `path/to/archive.zip`
2. Report `path/to/archive.zip/some/dir/and/file` without an `-extract` suffix as if an archive were a directory-like structure. And we would also report  `path/to/archive.zip`. Here the path would be "virtual"

And if the extraction is not done in-place, then the paths would always be kinda virtual an absent of the scanned tree.

So this about either extracting in place or not and which path to report in the scan results.

@akintayo @chinyeungli @majurg @nakami @rakeshbalusa @sschuberth @yahalom5776 Each of you has been involved with issues related to archive extraction. What would be your take and preference? What could be alternative ways?

Thanks for your input!

sschuberth:
[...]

Issue #458, opened 2017-01-17
Extract a specific paths from an archive


pombredanne:
It would be useful (and a prep for #14) to be able to extract a single path or a list of path from a given archive rather than everything all the times.

ashutoshsaboo:
I'll work on this as well in addition to the new PR that I just posted - #544 . Kindly give your suggestions towards the same. Thanks!

PR #544, opened 2017-03-07
Implements extraction of Archives before Scanning

ashutoshsaboo:
Signed-off-by: Ashutosh Saboo <ashutosh.saboo96@gmail.com>

@pombredanne Sir. This is a step towards implementing extraction of archives for scanning of licenses within the files of the archives as well. 

I feel, instead of extracting to a temporary location, as @pombredanne you were discussing in #14 , we must instead extract to the original location itself, since that will also not hinder the performance of the scanning engine as well. Scancode then itself, searches for licenses in the extracted files, and then we could delete the extracted files as well.

I have been looking at Scancode's source code since 2-3 days now, and I would be willing to work for implementing this. Any suggestions on this @pombredanne Sir? I'll issue new commits with the remaining TODOs as well.

Achieved:
1. Extraction of all archives in the folder tree under a specified location before starting the scanning process.
2. Silent extraction of archives.

Currently TODO's:
1. Delete extracted files after scanning is complete
2. Also implement points mentioned in #458 for specific archive extraction.

### DIFF
--- a/src/extractcode/extract_directory.py
+++ b/src/extractcode/extract_directory.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2015 nexB Inc. and others. All rights reserved.
+# Copyright (c) 2017 nexB Inc. and others. All rights reserved.
 # http://nexb.com and https://github.com/nexB/scancode-toolkit/
 # The ScanCode software is licensed under the Apache License version 2.0.
 # Data generated with ScanCode require an acknowledgment.
@@ -47,54 +47,6 @@ if TRACE:
     logger.setLevel(logging.DEBUG)
 
 
-"""
-Extract archives and compressed files recursively to get the file content available for
-further processing. This the high level extraction entry point.
-
-This is NOT a general purpose un-archiver. The code tries hard to do the right
-thing, BUT the extracted files are not meant to be something that can be
-faithfully re-archived to get an equivalent archive. The purpose instead is
-to extract the content of the archives as faithfully and safely as possible to
-make this content available for scanning: some paths may be altered. Some files
-may be altered or skipped entirely.
-
-In particular:
-
- - Permissions and owners stored in archives are ignored entirely: The extracted
-   content is always owned and readable by the user who ran the extraction.
-
- - Special files are never extracted (such as FIFO, character devices, etc)
-
- - Symlinks may be replaced by plain file copies as if they were regular files.
-   Hardlinks may be recreated as regular files, not as hardlinks to the original
-   file.
-
- - Files and directories may be renamed when their name is a duplicate. And a
-   name may be considered a duplicate ignore upper and lower case mixes even
-   on case-sensitive file systems. In particular when an archive contains the
-   same file path several times, every paths will be extracted with different
-   files names, even though using a regular tool for extraction would have
-   overwritten previous paths with the last path.
-
- - Paths may be converted to a safe ASCII alternative that is portable across
-   OSes.
-
- - Symlinks, relative paths and absolute paths pointing outside of the archive
-   are replaced and renamed in such a way that all the extract content of an
-   archive exist under a single target extraction directory. This process
-   includes eventually creating "synthetic" or dummy paths that did not exist in
-   the original archive.
-"""
-
-
-"""
-An ExtractEvent contains data about an archive extraction progress:
- - `source` is the location of the archive being extracted
- - `target` is the target location where things are extracted
- - `done` is a boolean set to True when the extraction is done (even if failed).
- - `warnings` is a mapping of extracted paths to a list of warning messages.
- - `errors` is a list of error messages.
-"""
 def extract_directory(location, kinds=extractcode.default_kinds, recurse=False):
     """
     Walk and extract any archives found at `location` (either a file or
@@ -127,5 +79,5 @@ def extract_directory(location, kinds=extractcode.default_kinds, recurse=False):
         if archive.can_extract(file):
             #TODO : Implement Extraction of Files without printing to the Terminal
             for y in extract(file, kinds=default_kinds, recurse=True):
-                print (y)
+                # print (y)
                 pass
--- a/src/extractcode/extract_directory.py
+++ b/src/extractcode/extract_directory.py
@@ -47,7 +47,7 @@ if TRACE:
     logger.setLevel(logging.DEBUG)
 
 
-def extract_directory(location, kinds=extractcode.default_kinds, recurse=False):
+def extract_directory(resources, kinds=extractcode.default_kinds, recurse=True):
     """
     Walk and extract any archives found at `location` (either a file or
     directory). Extract only archives of a kind listed in the `kinds` kind tuple.
@@ -70,14 +70,8 @@ def extract_directory(location, kinds=extractcode.default_kinds, recurse=False):
     if recurse and a nested archive is found, it is extracted to full depth
     first before resuming the file system walk.
     """
-    abs_input = fileutils.as_posixpath(os.path.abspath(os.path.expanduser(location)))
-    original_input = fileutils.as_posixpath(location)
-    abs_input = fileutils.as_posixpath(os.path.abspath(os.path.expanduser(location)))
-    ignored = partial(ignore.is_ignored, ignores=ignore.ignores_VCS, unignores={})
-    resources = fileutils.resource_iter(abs_input, ignored=ignored)
     for file in resources:
         if archive.can_extract(file):
             #TODO : Implement Extraction of Files without printing to the Terminal
-            for y in extract(file, kinds=default_kinds, recurse=True):
-                # print (y)
+            for y in extract(file, kinds=kinds, recurse=recurse):
                 pass
--- a/src/extractcode/extract_directory.py
+++ b/src/extractcode/extract_directory.py
@@ -65,13 +65,15 @@ def extract_directory(resources, kinds=extractcode.default_kinds, recurse=True):
     If `recurse` is True, extract recursively archives nested inside other
     archives If `recurse` is false, then do not extract further an already
     extracted archive identified by the corresponding extract suffix location.
-
+for y in
     Note that while the original file system is walked top-down, breadth-first,
     if recurse and a nested archive is found, it is extracted to full depth
     first before resuming the file system walk.
     """
     for file in resources:
         if archive.can_extract(file):
-            #TODO : Implement Extraction of Files without printing to the Terminal
             for y in extract(file, kinds=kinds, recurse=recurse):
                 pass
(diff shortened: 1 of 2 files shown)


pombredanne:
Thanks for the input. This should go to a temporary location and not in-place+delete.
Also it should be implemented as an alternative resources iterator that receives a stream of resources and handles extraction on the fly if needed and yields more resources. So it should be its own function. I suggest you start with tests too.

ashutoshsaboo:
@pombredanne I am sorry, could you explain what do you mean by temporary location - not in - place+delete? I understand your alternate resources iterator point though. I'll try working towards it.

ashutoshsaboo:
I was thinking of something of this kind, for a revamped iterator that'll handle, extracting the files part too. 

1)But then when we try to `import extractcode` within the `fileutils.py` file, it pops with an error. Any suggestion on how to combat that? 

2)And just one thing, the extract() function actually gives returns an iterable. Is there any functionality implemented as of now, that'll directly extract the location that we specify recursively, without returning the iterable, because then we can avoid the unnecessary printing of extract statements on the terminal ?


```
def resource_iter_extractable(location, ignored=ignore_nothing, with_files=True, with_dirs=True, with_extract=True):
    """
    Return an iterable of resources at `location` recursively.

    :param location: a file or a directory.
    :param ignored: a callable accepting a location argument and returning True
                    if the location should be ignored.
    :param with_dirs: If True, include the directories.
    :param with_files: If True, include the  files.
    :return: an iterable of file and directory locations.
    """
    assert  with_dirs or with_files, "One or both of 'with_dirs' and 'with_files' is required"
    for top, dirs, files in walk(location, ignored):
        if with_files:
            for f in files:
                if archive.can_extract(f) and with_extract:
                    for y in extract(f, kinds=default_kinds, recurse=True):
                        print (y)
                    yield os.path.join(top,f+'-extract')
                else:
                    yield os.path.join(top, f)
        if with_dirs:
            for d in dirs:
                yield os.path.join(top, d)

```

pombredanne:
do not modify fileutils. You are creating circular imports. extracting and print are unrelated. Check the code in extractcode

ashutoshsaboo:
@pombredanne Yes I already checked the code in extractcode. I'm looking into it. One thing, I noticed after I ran the test cases on my local system is that, the reason for the failing tests was basically related to the archives. It basically compares the output of Scancode, with a pre-defined json usually saved in the info folder. Now it contains, params like number of files scanned and others as well, which are bound to increase after we, scan the files in the archives as well. 

So, I'll have to change the test cases first, by modifying the pre-defined jsons. So, that seems to be correct right? @pombredanne

pombredanne:
You did not get my point:  start a new script in `src/extractcode` ... do not modify existing code for now. Create new tests as needed.

ashutoshsaboo:
@pombredanne Thanks for your input. I've added a new script that achieves extracting all the archives within a directory, before starting the scanning process. Please provide your input on the same. 

I've also updated the TODOs in the main issue, which I'll be working on shortly.

ashutoshsaboo:
I just checked out your wiki here - https://github.com/nexB/aboutcode/wiki/GSOC-2017 . I'd like to work on this full time, to resolve #14 completely, if that's fine. Would that be fine @pombredanne ?

src/extractcode/extract_directory.py
```diff
@@ -0,0 +1,131 @@
+#
+# Copyright (c) 2015 nexB Inc. and others. All rights reserved.
```
pombredanne: I guess we are no longer on 2015, are we?
Be careful to not copy extra things around without checking first.

src/extractcode/extract_directory.py
```diff
@@ -0,0 +1,131 @@
...
+TRACE = False
+
+if TRACE:
+    import sys
+    logging.basicConfig(stream=sys.stdout)
+    logger.setLevel(logging.DEBUG)
+
+
+"""
+Extract archives and compressed files recursively to get the file content available for
```
pombredanne: How is this doc relevant to this new script? Pleas remove

src/scancode/cli.py
```diff
@@ -70,6 +70,11 @@
 
 from scancode import utils
 
+from extractcode.extract import extract
+from extractcode import default_kinds
+from extractcode import archive
+from extractcode.extract_directory import extract_directory
+
```
pombredanne: None of these import are useful at this stage

src/extractcode/extract_directory.py
```diff
@@ -0,0 +1,131 @@
...
+
+    Note that while the original file system is walked top-down, breadth-first,
+    if recurse and a nested archive is found, it is extracted to full depth
+    first before resuming the file system walk.
+    """
+    abs_input = fileutils.as_posixpath(os.path.abspath(os.path.expanduser(location)))
+    original_input = fileutils.as_posixpath(location)
+    abs_input = fileutils.as_posixpath(os.path.abspath(os.path.expanduser(location)))
+    ignored = partial(ignore.is_ignored, ignores=ignore.ignores_VCS, unignores={})
+    resources = fileutils.resource_iter(abs_input, ignored=ignored)
```
pombredanne: I am confused as to why you are creating a resource iterator here. This does not sound right.
A transparent extraction iterator would receive a stream or resources and yield more resources after extraction.

src/scancode/cli.py
ashutoshsaboo: Resolved this, in the latest commit. :+1:

src/extractcode/extract_directory.py
ashutoshsaboo: Resolved this, in the latest commit. :+1:

src/extractcode/extract_directory.py
ashutoshsaboo: Oops. Sorry for the mistake. Resolved this, in the latest commit. :+1:

src/extractcode/extract_directory.py
ashutoshsaboo: Resolved this in the latest commit as well. :+1: Thanks for your input. :smile:

ashutoshsaboo:
@pombredanne thanks a lot for your latest comment here #14 . I have pushed a few more commits to this PR , which works on basically accepting a resources iterator as input, and extract all the files that are extractible before beginning with the scanning process. This works in the short run as of now, and it suffices the need very well, because the latest commit, will basically take the resource iterator as an input, and extract all the extractible files, to ensure that the tree is built perfectly during scanning.

I understand your point, about the File object thing as well. I have been thinking of the same thing for my GSoC proposal as well, since implementing that with the correct unit tests should take a considerable amount of time as well. 

Basically once, we extract an archive we'll have to report both the **real path** as well as the **extracted/virtual path** as well. So, I'll think about implementing that functionality too before writing my proposal. 

Any other inputs from your end, which I could include in my proposal? I'll communicate the points that I have thought of as well, shortly. Also, is there any other requirement that you would want me to meet before sending the proposal? I have been in good touch with the code-base in general, and I've pushed #544 and #547 till now. I'll be working pushing more commits shortly as well.

Is there any specific/additional thing that you would want me to work towards to be a worthy GSoC candidate for resolving #14 completely? I am open to working full time for this as well. 

Would be good to have your input on this! :smile:

pombredanne:
The approach you detailed in https://github.com/nexB/scancode-toolkit/pull/544#issuecomment-284671722 seems fine. You just need to not print things. instead you further recursively iterate once you have extracted something.

pombredanne:
The outline of what do to would be:
```
def iter_extract(resources)
    for resource in resources:
        if is extractible:
            extract(target  dir)
            return iter_extract(target dir)
        else:
            yield resource
```

pombredanne:
> Is there any specific/additional thing that you would want me to work towards to be a worthy GSoC candidate for resolving #14 completely? I am open to working full time for this as well.

You are almost there. So finishing this would make sense if you can.

pombredanne:
ping?

ashutoshsaboo:
@pombredanne Caught up with exams. Will work on this ASAP, the exams get over. Thanks :+1: .

ashutoshsaboo:
@pombredanne Sorry for the delay, I'm caught up with exams, which is taking most of my time. I've tried to make some changes, to make the new extract iterator yield extracted files also on the fly. Please let me know your feedback on the latest commit. Thanks :smile:

ashutoshsaboo:
@pombredanne Why do the Appveyor tests fail, while the Travis tests pass? Could you help with this?
Also, any comments on the recent commit? Seems fine?

pombredanne:
The tests failure are not yours but mine!

src/scancode/cli.py
```diff
@@ -280,6 +285,24 @@ def scancode(ctx,
     scans_cache_class = get_scans_cache_class()
 
     try:
+        #We could call the below function if required to first extract all the archives, and then start scanning
+
+        '''
+        abs_input = fileutils.as_posixpath(os.path.abspath(os.path.expanduser(location)))
```
pombredanne: What is this comment code?

ashutoshsaboo: Yes I'll remove it.

ashutoshsaboo: Removed in the latest commit. :+1:

ashutoshsaboo:
@pombredanne Let me know once your tests pass, and if this looks perfect? :smile:

src/scancode/cli.py
```diff
@@ -280,6 +285,22 @@ def scancode(ctx,
     scans_cache_class = get_scans_cache_class()
 
     try:
+        '''
```
pombredanne: Why put a string instead of code? Please either delete or implement it

ashutoshsaboo:
Should I add a separate option called ``--extract`` that would search for licenses within the archives as well? Since, in case if we don't add any such option, it'd require modifying all the other tests completely, which can lead to failing tests as well, and cause other problems. So, adding a new option like ``--extract`` would seem to be fine? @pombredanne

pombredanne:
@ashutoshsaboo a new option is fine. You should model it after the extractcode cli options IMHO

ashutoshsaboo:
@pombredanne Thanks a lot. I'll work on it some time shortly. :smile:

pombredanne:
@ashutoshsaboo I am leaving this open as I might cherry pick some of it later when we come to implement this for real (eventually as a plugin).

majurg:
@ashutoshsaboo @pombredanne Ping. Not sure if we want this open still or if I should close.

pombredanne:
I am closing this at this stage, but I kept a copy of your branch here in https://github.com/nexB/scancode-toolkit/tree/ashutoshsaboo/archive_extraction so we can reuse it when we come to implement this after all!.
Thank you ++

PR #563, opened 2017-03-20
#524 Remove prepend of root directory from the source of path

agneet42:
'/code' removed from json2csv.py.

'/cod' removed from tests:

test_json2csv.py
full.csv
 full.json-expected
minimal.csv
minimal.json-expected
package_license_value_null.json-expected.


agneet42:
@pombredanne @JonoYang , yours to review.
Thanks.

etc/scripts/json2csv.py
```diff
@@ -114,9 +114,6 @@ def collect_keys(mapping, key_group):
             if not path.endswith('/'):
                 path = path + '/'
 
-        # alway create a root directory
-        path = '/code' + path
```
pombredanne: There should still be a CLI option to prefix the path if a user wishes to do so.

agneet42: Do you want me to put a display message such as "Do you want to prefix the path"?

JonoYang: @agneet42 There should not be any sort of message. This should be a command line option that the user sets, e.g. `python json2csv.py --prepend-root in.json out.csv`

pombredanne: `--prefix-path` might be better? naming things is always difficult!

agneet42: #JonoYang @pombredanne, as there is a user option, then the '/code' that I've removed from the test files, need to have an if-else option too, right?

JonoYang: @agneet42 You would have to add in a new test to make sure the option works correctly, that it prefixes `/code` when the option is selected.

agneet42: @JonoYang okay.

pombredanne:
@agneet42 did you read the instructions on commit messages https://github.com/nexB/aboutcode/wiki/Writing-good-commit-messages ?

pombredanne:
ping?

pombredanne:
@agneet42 I am closing this as this was eventually fixed by @JonoYang with fc8ad9bfb0a414bb130ce49a54d94285f643ed91

Thanks a lot!

Issue #513, opened 2017-02-22
Identify that information are from scancode or another tool


pombredanne:
For results such as packages, there is no direct way to determine from the JSON that this came from scancode or another tool. This is implicit. We should have a way to list which tool provided data explicitly.

pombredanne:
A way to track which tools have touched a JSON could be to include some history/log so we know this has been touched by scancode, then aboutcode manager then something else, etc.

patniharshit:
Hi, I am new here. Looking to contribute. Can you guide me on how to start with this issue?

kartiksibal:
@pombredanne I'd like to resolve this, if no one is working on it. 

If no one is. I was thinking of making a log file, wherein, whenever a tool makes changes to the JSON. There is an explicit record. Which can maybe be done, by creating a wrapper/decorator? I am not sure if I am expressing my thought correctly here. Please do share your approach.

Thanks :-)

pombredanne:
@patniharshit @kartiksibal  I guess only one of the two of you should work on this.

 @patniharshit what's your approach? did you start? @kartiksibal same question. 

 @kartiksibal Your approach sounds ok. Show me some code so I can understand what you mean exactly.

kartiksibal:
@pombredanne Sure. I'll do that.

kartiksibal:
@pombredanne I am thinking of making a `@log`, decorator, for this, which will write to a file whenever a change to JSON has been made, hence containing the tool used. 

Am I somewhere near, to what your expectations are with this? :-)

patniharshit:
@pombredanne I didn't had any ideas in particular. I was hoping to get started here if somebody could guide me on what to do. 
But since kartiksibal has something, I am cool with he going ahead with this.

kartiksibal:
@patniharshit I think, what matters is coming up with an efficient solution to the problem, it doesn't matter who does it. We are here to improvise the product and learn. :-D

I'd love to know how you were thinking of approaching this issue. :-)

pombredanne:
@kartiksibal I can comment when I see some code.

kartiksibal:
@pombredanne I am on it.

pombredanne:
@kartiksibal you wrote:

> I have made a very minuscule change in /src/scancode/cli.py. Just when we dump the JSON data, I have added this . Is this somewhere on the lines of what you were expecting with this enhancement? :-)

[...]

PR #576, opened 2017-03-27
#513 Log to track which tools provided data

kartiksibal:
Signed Off By: Kartik Sibal <kartiksibal@gmail.com>

### DIFF
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -34,6 +34,7 @@ from scancode.pool import get_pool
 from collections import OrderedDict
 from functools import partial
 import os
+import datetime
 from os.path import expanduser
 from os.path import abspath
 import sys
@@ -442,7 +443,14 @@ def scan(input_path,
 
     files_scanned_per_second = round(float(files_count) / scanning_time , 2)
     scan_summary['files_scanned_per_second'] = files_scanned_per_second
-
+    
+    if not scanning_errors:
+        #Log file containing timestamp & Project's name.
+        log = open (input_path + "_log.txt", "a+")
+        day = datetime.datetime.today().replace(microsecond=0)
+        msg = ("ScanCode processed this on: {} \n").format(day)
+        log.write (msg)
+    
     if not quiet:
         # Display stats
         ##########################
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -10,10 +10,10 @@ If you are new to ScanCode and want to find easy tickets to work on, check https
 When contributing to ScanCode (such as code, bugs, documentation, etc.) you
 agree to the Developer Certificate of Origin http://developercertificate.org/
 and the ScanCode license (see the NOTICE file).  The same approach is used
-by the Linux Kernel developers.
+by the Linux Kernel developers and several other projects.
 
-For commits, simply add a line like this to your commit message, with your
-name and email::
+For commits, it is best to simply add a line like this to your commit message,
+with your name and email::
 
     Signed-off-by: Jane Doe <developer@example.com>
     
@@ -25,7 +25,7 @@ See https://github.com/nexB/aboutcode/wiki/Writing-good-commit-messages for deta
 Feature requests and feedback
 =============================
 
-To send feedback, file an issue at
+To send feedback or ask a question file an issue at
 https://github.com/scancode/scancode-toolkit/issues
 
 If you are proposing a feature:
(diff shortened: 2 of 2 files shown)


src/scancode/cli.py
```diff
@@ -442,7 +443,14 @@ def scan_event(item):
 
     files_scanned_per_second = round(float(files_count) / scanning_time , 2)
     scan_summary['files_scanned_per_second'] = files_scanned_per_second
-
+    
+    if not scanning_errors:
+        #Log file containing timestamp & Project's name.
+        log = open (input_path + "_log.txt", "a+")
```
pombredanne: you should use a `with` statement (to close the file). You are leaking an open file

src/scancode/cli.py
```diff
@@ -442,7 +443,14 @@ def scan_event(item):
...
+        day = datetime.datetime.today().replace(microsecond=0)
```
pombredanne: Check commoncode. There is likley some utility function there or elswehere to get a timestamp as a string

src/scancode/cli.py
```diff
@@ -34,6 +34,7 @@
 from collections import OrderedDict
 from functools import partial
 import os
+import datetime
```
pombredanne: we like imports to be grouped, then sorted. This should go right after the collections import

pombredanne:
Thanks! I provided some comments for you consideration.

src/scancode/cli.py
kartiksibal: Noted.

src/scancode/cli.py
kartiksibal: That was really stupid of me, sorry.

src/scancode/cli.py
kartiksibal: Noted.

kartiksibal:
@pombredanne I'll re-structure it as you have guided. I have a small query. 
 
If suppose, our objective was to go ahead with my approach. 
Have I coded it, in an efficient way, then? In what way could I have changed my approach angle. You don't have to elaborate, just a subtle hint will be awesome! _Apart from the mistake on the file closing_. 

PS: I do realise the code is not enough to judge or comment. But, I'd love any input.
  
Thanks

pombredanne:
@kartiksibal what is missing in all cases is tests first and then some docs. 

- Tests are needed to set the specs of your code in stone: I would have no idea how to use your approach as it is and what it means and what you expect it to be. Tests is the doc you write for the other developers.
- Doc is needed to explain a user-visible feature: a user would likely be puzzled as to why a `*_log.txt` would have popped up in the scanned code directory and what it meant. Rescanning the same directory a second time would yield different results as your log file would be now scanned that second time.

So the net net: when you write code, there are two goals:
1. your code will be read eventually 100 of times by other devs.  Therefore, you write first for your code it to be red and understood first. Hence the need for tests and at times some comments for non-obvious parts.
2. your code will be used eventually 100 of times by users.Therefore your need a user experience and documentation that makes it easy to understand, as self-evident as possible and non-surprising.

pombredanne:
As a side note your use of a `patch-1` branch suggests that you may have used the online Github editor for this. This is in practice OK for small trivial typo edits. I do not think this is something that works for anything non-trivial involving multiple files, tests and more such as eventual amend and squash that may be needed. There are (many) better tools out there to edit code locally and the git CLI and GUI tools are your friends there.

kartiksibal:
@pombredanne That's just the advise I needed! 
I had actually thought of adding tests later, once you were satisfied with my approach. I'll make sure I add them hand in hand.

I am completely aware about the online editor, used it only since the change was minuscule. :-)

Thank You so much! :grin:

pombredanne:
So in terms of the approach, we cannot use a log file. We need this as some history list in the scan header instead. Can you rework that along these lines? Thanks!

kartiksibal:
@pombredanne Sure, I'll get onto it once the application period is over?

src/scancode/cli.py
kartiksibal: @pombredanne I digged a little commoncode is somewhat eventually doing exactly this. What would you suggest?

kartiksibal:
FML! 
Would you mind if I close this PR and start a new one.

src/scancode/cli.py
pombredanne: tell me what you found there. I am sure I wrote such a function in the past in there

kartiksibal: In `commoncode/timeutils.py`, there's a function `tstamp2time` which takes a UTC time stamp as input and converts to a datetime object & a function `time2tstamp` which returns a UTC time stamp. So, if we feed the latter to the former. It'll return a timestamp. 
In such a case, is it better to call `commoncode`?

pombredanne:
>  Would you mind if I close this PR and start a new one.

Sure. Make sure you reference this PR in the body of the new PR.... for reference

pombredanne:
@kartiksibal I am closing this as requested and will wait for the new one instead.

kartiksibal:
@pombredanne Shoot! I got too engrossed in the project. 😅 
Will come back to the issues, now. 😄

pombredanne:
feel free to reopen this one if you want

Issue #564, opened 2017-03-20
Add coverage and coding standard checks to CI runs


pombredanne:
In order to ease the review of PRs, we should add new checks for code coverage and coding standard.
DRF does a decent job there as pointed to me by @tdruez : https://github.com/tomchristie/django-rest-framework/blob/master/runtests.py

pombredanne:
I wonder if we can also have checks for the format of the commit messages too. I am sure this exists somewhere.

yasharmaster:
@pombredanne 
I would like to work on this. How should I get started?

michaelrup:
Here are a few links, which I think are interesting to look at:
- code coverage: [codecov](https://codecov.io/) or [coveralls](https://coveralls.io/) 
- coding standard: [flake8](http://flake8.pycqa.org/en/latest/) and [isort](https://pypi.python.org/pypi/isort) as seen in [DRF](https://github.com/tomchristie/django-rest-framework/blob/master/runtests.py) 
- commit messages: [gommit](https://github.com/antham/gommit) looks interesting

singh1114:
@pombredanne For the commit messages, can we use something like this:

https://github.com/commitizen/cz-cli
I think the repo is MIT licensed.

all3fox:
About code coverage:

Since `scancode-toolkit` already runs on travis-ci (has a `.travis.yml`) and runs unittests with pytest, then theoretically code coverage should not be hard:

1. Enable travis for the repository (already done)
2. Enable `coveralls.io` (say) for the repository
3. Install `pytest-cov`, `pytest-xdist` and `coveralls` in `.travis.yml`
4. Add an `after-success` wich will send collected stats with `coveralls` to `coveralls.io` 

So, something like this (simplifying, `.travis.yml` for scancode-toolkit is more complicated and has a matrix):

```
install:
  - pip install pytest pytest-cov pytest-xdist coveralls
script:
  - pytest --cov=src tests
after_success:
  - coveralls
```

Applying this approach did not quite work for me, see [my branch][1] and [resulting travis builds][2] (failing).

1. Tried `pip install` just after configure:

```
install:
  - ./configure
  - pip install pytest-cov pytest-xdist coveralls
```

That failed as [first series of builds][3]

2. Fiddled with `--cov-append` and ignoring/including directories, not really the problem right now. Second [series of builds][2]

3. Tried to install `pytest-cov` from within `./configure`, modified `etc/conf/dev/base.txt` for that to look like this:

```
[...]

PR #577, opened 2017-03-27
#564 add code coverage to .travis-ci

all3fox:
I have figured how to add `pytest-cov`, `coveralls` and its dependency `docopt` to the configure system you described [in the issue][1] (will add .ABOUT and .LICENSE files later. Also, `docopt` lacks a wheel so created it.):

Now, when I locally do:

```
./configure
```
Then `type pytest` gives `pytest not found` which means there is no globally installed pytest on my machine. Then I do:

```
source bin/activate
type pytest
type coveralls
```

Which gives me:

```
pytest is /home/alex/Code/scancode-toolkit/bin/pytest
coveralls is /home/alex/Code/scancode-toolkit/bin/coveralls
```

This means that a local `./configure` installs `pytest` and `coveralls` just fine for me. However, when I use the following `.travis.yml` on CI, it tells me 

```
/Users/travis/build.sh: line 62: bin/pytest: No such file or directory
```

So, it looks like pytest does not get installed and the build fails. Any ideas why?

Update: for this PR I changed the order of `type pytest` and `type coveralls`, so travis will probably be missing coveralls, but you get the idea. Here is a link to my ci runs that miss pytest:

https://travis-ci.org/all3fox/scancode-toolkit/jobs/215599442

[1]: https://github.com/nexB/scancode-toolkit/issues/564#issuecomment-289366836

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,35 +1,35 @@
-language: cpp
+language: python
 
-os: 
-    - osx
-    - linux 
+python:
+  - 2.7.13
 
-env:
-    matrix:
-        - TEST_SUITE="bin/py.test -n 2 -vvs --ignore=tests/scancode --ignore=tests/extractcode --ignore=tests/licensedcode --ignore=tests/cluecode --ignore=tests/packagedcode"
-        - TEST_SUITE="bin/py.test -n 2 -vvs tests/scancode"
-        - TEST_SUITE="bin/py.test -n 2 -vvs tests/extractcode"
-        - TEST_SUITE="bin/py.test -n 2 -s tests/licensedcode"
-        - TEST_SUITE="bin/py.test -n 2 -s tests/cluecode"
-        - TEST_SUITE="bin/py.test -n 2 -s tests/packagedcode"
-        - TEST_SUITE="./etc/release/release.sh"
-
+matrix:
+  include:
+    - os: osx
+      language: generic
+      python:
+      osx_image: xcode7.3
 
 install:
-    - ./configure
+  - ./configure
+  - ./bin/pip install pytest-cov coveralls
 
 script:
-    - $TEST_SUITE
+  # If debugging, use a subset of tests to reduce wait:
+  # - ./bin/py.test -n 2 -vvs --cov=src/scancode tests/scancode
+  - ./bin/py.test -n 2 -s --cov=src
+
+after_success:
+  - ./bin/coveralls
+  - ./etc/release/release.sh
 
 notifications:
-    irc:
-        channels:
-          - "chat.freenode.net#aboutcode"
-    on_success: change
-    on_failure: always
-    use_notice: true
-    skip_join: true
-    template:
-        - "%{repository_slug}#%{build_number} (%{branch}-%{commit}:%{author})-%{message}- %{build_url}"
-        - 
-        - 
+  irc:
+    channels:
+      - "chat.freenode.net#aboutcode"
+  on_success: change
+  on_failure: always
+  use_notice: true
+  skip_join: true
+  template:
+    - "%{repository_slug}#%{build_number} (%{branch}-%{commit}:%{author})-%{message}- %{build_url}"
--- a/README.rst
+++ b/README.rst
@@ -1,23 +1,22 @@
-===============================
+================
 ScanCode toolkit
-===============================
+================
 
 
 Build and tests status
 ======================
 
-+-------+-----------------------------------------------------------------------------+-----------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------+
-|Branch |                         **Linux (Travis)**                                  |                         **MacOSX (Travis)**                                 |                         **Windows (AppVeyor)**                                                |
-+=======+=============================================================================+=============================================================================+===============================================================================================+
-|       |.. image:: https://api.travis-ci.org/nexB/scancode-toolkit.png?branch=master |.. image:: https://api.travis-ci.org/nexB/scancode-toolkit.png?branch=master |.. image:: https://ci.appveyor.com/api/projects/status/4webymu0l2ip8utr/branch/master?png=true |
-|Master |   :target: https://travis-ci.org/nexB/scancode-toolkit                      |   :target: https://travis-ci.org/nexB/scancode-toolkit                      |   :target: https://ci.appveyor.com/project/nexB/scancode-toolkit                              |
-|       |   :alt: Linux Master branch tests status                                    |   :alt: MacOSX Master branch tests status                                   |   :alt: Windows Master branch tests status                                                    |
-+-------+-----------------------------------------------------------------------------+-----------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------+
-|       |.. image:: https://api.travis-ci.org/nexB/scancode-toolkit.png?branch=develop|.. image:: https://api.travis-ci.org/nexB/scancode-toolkit.png?branch=develop|.. image:: https://ci.appveyor.com/api/projects/status/4webymu0l2ip8utr/branch/develop?png=true|
-|Develop|   :target: https://travis-ci.org/nexB/scancode-toolkit                      |   :target: https://travis-ci.org/nexB/scancode-toolkit                      |   :target: https://ci.appveyor.com/project/nexB/scancode-toolkit                              |
-|       |   :alt: Linux Develop branch tests status                                   |   :alt: MacOSX Develop branch tests status                                  |   :alt: Windows Develop branch tests status                                                   |
-+-------+-----------------------------------------------------------------------------+-----------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------+
-
++-------+--------------------------------------------------------------------------------------------+-----------------------------------------------------------------------------+-----------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------+
+|Branch |                                        **Coverage**                                        |                         **Linux (Travis)**                                  |                         **MacOSX (Travis)**                                 |                         **Windows (AppVeyor)**                                                |
++=======+============================================================================================+=============================================================================+=============================================================================+===============================================================================================+
+|       |.. image:: https://coveralls.io/repos/github/nexB/scancode-toolkit/badge.svg?branch=master  |.. image:: https://api.travis-ci.org/nexB/scancode-toolkit.png?branch=master |.. image:: https://api.travis-ci.org/nexB/scancode-toolkit.png?branch=master |.. image:: https://ci.appveyor.com/api/projects/status/4webymu0l2ip8utr/branch/master?png=true |
+|Master |   :target: https://coveralls.io/github/nexB/scancode-toolkit?branch=master                 |   :target: https://travis-ci.org/nexB/scancode-toolkit                      |   :target: https://travis-ci.org/nexB/scancode-toolkit                      |   :target: https://ci.appveyor.com/project/nexB/scancode-toolkit                              |
+|       |   :alt: Linux Master branch test coverage                                                  |   :alt: Linux Master branch tests status                                    |   :alt: MacOSX Master branch tests status                                   |   :alt: Windows Master branch tests status                                                    |
++-------+--------------------------------------------------------------------------------------------+-----------------------------------------------------------------------------+-----------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------+
+|       |.. image:: https://coveralls.io/repos/github/nexB/scancode-toolkit/badge.svg?branch=develop |.. image:: https://api.travis-ci.org/nexB/scancode-toolkit.png?branch=develop|.. image:: https://api.travis-ci.org/nexB/scancode-toolkit.png?branch=develop|.. image:: https://ci.appveyor.com/api/projects/status/4webymu0l2ip8utr/branch/develop?png=true|
+|Develop|   :target: https://coveralls.io/github/nexB/scancode-toolkit?branch=develop                |   :target: https://travis-ci.org/nexB/scancode-toolkit                      |   :target: https://travis-ci.org/nexB/scancode-toolkit                      |   :target: https://ci.appveyor.com/project/nexB/scancode-toolkit                              |
+|       |   :alt: Linux Develop branch test coverage                                                 |   :alt: Linux Develop branch tests status                                   |   :alt: MacOSX Develop branch tests status                                  |   :alt: Windows Develop branch tests status                                                   |
++-------+--------------------------------------------------------------------------------------------+-----------------------------------------------------------------------------+-----------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------+
 
 
 ScanCode is a suite of utilities used to scan a codebase for license, copyright


pombredanne:
@all3fox you wrote
> I have figured how to add pytest-cov, coveralls and its dependency docopt to the configure system you described in the issue (will add .ABOUT and .LICENSE files later. Also, docopt lacks a wheel so created it.):

You rock.

pombredanne:
> So, it looks like pytest does not get installed and the build fails. Any ideas why?

I will dig in this later today.

all3fox:
Just before you go:

I tried moving `./configure` call from the `install:` section down to `scripts:` section of `.travis-yml`:

```
scipts:
  - ./configure
  - $TEST_SUITE
```

In the hopes that it would somehow pick up the virtual environment. It however showed another problem: I missed an the `coverage` dependency for `pytest-cov`.

I am trying to build a universal wheel out of it right not, the `py2.py3-none-any` type of wheel (I do not know much about wheels now, reading up on that). However, I am slightly stuck at that. And even though the [pypi page][1] gives a huge list of wheels for `coverage`, they are all not universal (probably?)

If you immediately see that I am missing something, tell me. Otherwise will keep reading about wheels.

[1]: https://pypi.python.org/pypi/coverage/4.3.4

all3fox:
> I have figured how to add pytest-cov, coveralls and its dependency docopt to the configure system

Ha-ha, spoke too soon. Basically, as far as I understand, `coverage` is a python package that is a dependency for `pytest-cov` and has C extensions (have never written one myself). Looks like that means that wheels for this package can only be made "platform specific": for a particular python and particular OS. So, no universal `py2.py3-none-any` wheels and I don't know how to add those packages under `thirdparty/dev`.

I have been trying alternative approaches with `pip install`ing whatever I need directly in `.travis.yml` However, since travis container images use python `2.7.9` by default, it causes a warning:

The log itself is, for instance, line 1342 here: https://travis-ci.org/all3fox/scancode-toolkit/jobs/215734761

Searching leads to this meantion:

https://github.com/python-telegram-bot/python-telegram-bot/issues/346


Basically, it causes the test coverage to *not* be sent to coveralls.io

My current idea is to use the `sudo: true` travis mirrors that allow installing packages to the ubuntu image. The link above indicates that `requests[security]` somehow might do the trick. But I am quite clueless here, any idea-bouncing appreciated.

all3fox:
And it finally works on my branch! See commit message for a succint yet complete desciption of how I changed `.travis.yml`. In short:

1. Could not make a universal wheel for `coverage` dependency, so ended up manually `pip installing` in `.travis.yml`
2. Could not use default python `2.7.9` that comes with travis images. So, changed language and fixed python version to `2.7.13` to make coveralls work (which relies on requests/urllib3)

README.rst is rendered here: https://github.com/all3fox/scancode-toolkit/tree/issue-564-ci-poking
coveralls.io page is here: https://coveralls.io/github/all3fox/scancode-toolkit

Everything needs to be checked/reviewed, happy to discuss!

all3fox:
Marking WIP again because I somehow missed MacOS which is still failing. Will see tomorrow.

pombredanne:
I can help you on the wheels side. If we need OS-specific wheels I have these build loops: https://github.com/pombreda/thirdparty and https://github.com/pombreda/thirdparty-manylinux/ that can help build these on Travis and Appveyor for Linux, macOS and WIndows and then upload the built wheels and their deps to http://dl.bintray.com/pombreda/thirdparty/

all3fox:
@pombredanne I understand now the idea behind https://github.com/pombreda/thirdparty and https://github.com/pombreda/thirdparty-manylinux/

You choose the package that you need to build the wheels for, add it to the CI script of the repository and push. That triggeres CI that builds the platform-specific wheel. The wheel gets uploaded to bintray, where I can now download it back to my machine and put right into the project.

That way, if I have no access to a MacOS or Windows box, I still can build wheels for those, right?

all3fox:
@pombredanne is there a specific reason why most of your `.travis.yml` scripts have a `language: cpp` requirement?

For example, current `[.travis.yml][1]` for `scancode-toolkit` does, even though it is very much a python project.

[1]: https://github.com/nexB/scancode-toolkit/blob/c67488aa4aa8fa0900deb91b2ea19884a2b8d52d/.travis.yml

all3fox:
TIL: looks like doing `language: python` and then asking Travis for a macos build never works:

Many versions of python2 [failed here][2] and the [travis docs][1] mention "unsupported" and "`language: generic` trick".

Turning back to `language: cpp` or no language line at all (which defaults to ruby) actually gives you a `python 2.7.12` on `OS X 10.11` and `Xcode 7.3.1`

In the spirit of making things repeatable and predictable, I am thinking of building a `matrix: include:` structure that would fix version for MacOS (and its tools) and maybe even linux.

Then will try to eliminate `pip install` calls from `.travis.yml` by using the wheel repositories you mentioned above. Also, note that here is a similar wheel-generating repo:

https://github.com/matthew-brett/multibuild/tree/e6ebbfa42588c26579eff99c50e84326dd9f6b3e

It has impressive docs and is a dependency for many projects (see end of its README). So, might be superior to the repos you mentioned above but I am not sure.

In the meanwhile will update this PR shortly, as MacOS builds no work.   

[1]: https://docs.travis-ci.com/user/multi-os/#Python-example-(unsupported-languages)
[2]: https://travis-ci.org/all3fox/scancode-toolkit/builds/215958771

all3fox:
Yay! Finally, MacOS builds work too. However, I had to ditch the

```
env:
    matrix:
        TEST_SUITE = ""
```

structure because it was not properly picked up by the explicitly included `osx` builds. Ready to discuss/answer questions. Pretty much every change of the `.travis.yml` file has some specific reason behind it, I tried to capture that in the commit message.

Also, question: can current `scripts` section be reduced to:

```
pytest -n 2 -s --cov=src
./etc/release/release.sh
```

all3fox:
And Tracis CI passes! By the way, this is "threading the needle" for time on CI:

IIRC 50 mins for MaxOS on Travis, 60 mins for Linux on Travis and, looks like, 60 mins on Appveyor.

all3fox:
Hm, just noticed that `./etc/release/release.sh` cleans the virtualenv before `coveralls` is run.

That is bad, moving this `.sh` down to `deploy` section to keep the venv. Marking WIP again.

pombredanne:
> Also, question: can current scripts section be reduced to:
```
pytest -n 2 -s --cov=src
./etc/release/release.sh
```
Sure. I had tried to break the build across a matrix in the hope this would speed things up. But both travis and appeyor "free" plans run one task at a time mostly. It was speeding things up tad AFAIRC.

pombredanne:
>  just noticed that ./etc/release/release.sh cleans the virtualenv before coveralls is run.

The purpose of the release script is to build a release and do some simple smoke tests.  Hence why it it cleaning the venv and was running in its own job. But I am quite fine with having it otherwise

pombredanne:
@all3fox note that even though there 10K tests running it is quite telling to see how poor the coverage of some code is (and in some case highlighting some dead code too). This is great!

pombredanne:
Also it is quite interesting to see how both the runtime went up and elapsed time went down. With the caveat that it is tougher to get quicker feedback on a some tests for a given module. 
eg. 
- with the current `develop` branch: https://travis-ci.org/nexB/scancode-toolkit/builds/215617797
    -  Ran for 18 min 1 sec
    - Total time 1 hr 7 min 51 sec 

- with you new config: https://travis-ci.org/nexB/scancode-toolkit/builds/216037027
    - Ran for 46 min 35 sec
    - Total time 1 hr 58 min 52 sec

sschuberth:
> One possibility would be to drop Travis for Mac and consider Circle? @sschuberth configured it for spdx alright recently.

@all3fox If you're interested in my CircleCI script, have a look [here](https://github.com/spdx/tools-python/blob/master/circle.yml). Note that CircleCI builds default to Linux, you have to opt-in by emailing their support to get a free macOS plan for OSS projects. In the [tools-python](https://github.com/spdx/tools-python) project we use all of CircleCI, Travis CI and AppVeyor to build / test on macOS, Linux and Windows respectively in parallel as much as possible. The choice was to use CircleCI for macOS and Travis CI for Linux (and not vice versa) as I found the Circle CI macOS nodes to be available much quicker.

.travis.yml
```diff
@@ -1,35 +1,44 @@
...
-        - 
+  irc:
+    channels:
+      - "chat.freenode.net#aboutcode"
+  on_success: change
+  on_failure: always
+  use_notice: true
+  skip_join: true
+  template:
+    - "%{repository_slug}#%{build_number} (%{branch}-%{commit}:%{author})-%{message}- %{build_url}"
```
sschuberth: BTW, even if not part of this change, is the space before `%{build_url}` intended here?

pombredanne: AFAIRC not having a space made the url less linkable in IRC chat logs. But I may be wrong

pombredanne:
@sschuberth Thanks. Do you reckon I should enable circle on this repo then?

sschuberth:
> Do you reckon I should enable circle on this repo then?

Maybe not just now. Let's do one thing at a time and get coverage merged, and in the meantime continue looking at how well CircleCI works for us over at spdx/tools-python. Did you see any of the CI integrations not properly reporting back their status to PRs in there recently, or does that issue seem to be gone for good?

pombredanne:
@sschuberth you wrote:
>  Did you see any of the CI integrations not properly reporting back their status to PRs in there recently

Nope. Sounds good. And you are right to go there one step at a time !

.travis.yml
```diff
@@ -1,35 +1,44 @@
...
-        - TEST_SUITE="bin/py.test -n 2 -vvs --ignore=tests/scancode --ignore=tests/extractcode --ignore=tests/licensedcode --ignore=tests/cluecode --ignore=tests/packagedcode"
-        - TEST_SUITE="bin/py.test -n 2 -vvs tests/scancode"
-        - TEST_SUITE="bin/py.test -n 2 -vvs tests/extractcode"
-        - TEST_SUITE="bin/py.test -n 2 -s tests/licensedcode"
-        - TEST_SUITE="bin/py.test -n 2 -s tests/cluecode"
-        - TEST_SUITE="bin/py.test -n 2 -s tests/packagedcode"
-        - TEST_SUITE="./etc/release/release.sh"
+python:
+  - 2.7.12
+  - 2.7.13
```
pombredanne: Just curious: why two version of Python would be needed there? It may explain why things are taking twice the runtime?

sschuberth: Good point. Esp. as they only differ in the patch-level version. It probably makes sense to use this feature for testing with Python 2.x and 3.x, once we support 3.x officially ;-)

all3fox: Made two versions just because I could. That is indeed making the build twice as long, will reduce to `2.7.13`

all3fox:
> Sure. I had tried to break the build across a matrix in the hope this would speed things up. But both travis and appeyor "free" plans run one task at a time mostly. It was speeding things up tad AFAIRC.

Another option is to fail fast when the build contains errors:

```
matrix:
  fast_finish: true
```

However, if the build has no errors, then from my experience free plans are allocated 1-8 machines probably based on availability/load. So yes, my currently proposed build matrix does not take advantage of that. I tried keeping this idea of making a matrix out of `TEST_SUIT` variables but then they are not properly picked up by the `includ`ed parts of the matrix (the osx), only the last `TEST_SUIT` was essentially being picked up by osx.

all3fox:
> To properly support OSX I had to turn off that and use language: cpp
One possibility would be to drop Travis for Mac and consider Circle? @sschuberth configured it for spdx alright recently.

This time I found this `language: generic` trick that manually overrides the `language: python` in `include` section of the matrix. That is not pretty and is a hack but I can now at least finish configuring Travis (nearly there).

Also, sschuberth mentions "One thing at a time", so will do just that 😸

all3fox:
Finally, I stumbled upon a project that contains many decent examples of setting up travis/appveyor/codecov/others, might be useful for layout/script ideas at least.

https://github.com/SethMichaelLarson/python_project_template

all3fox:
Left one python version and one osx version (the travis default one). Also compressed `pytest` call into one line and moved the `release.sh` call to `after_success` section and after `coveralls` (for some reason `deploy:` section did not work even though [it is documented][1])

[1]: https://docs.travis-ci.com/user/customizing-the-build/#The-Build-Lifecycle

all3fox:
The Travis build now *really* works 😉 and the coverage is reported too!

The biggest problem is appveyor taking too long and failing. I would suggest:

1. Review this PR and merge it without bundling `pytest-cov`, `coveralls` and co with `scancode-toolkit`.

The bundling will not be easy (I suspect) and I would much rather do it in a separate PR. The issue #577 can be updated to reflect that coverage is there but a transition to something like a ci profile for configure (probably) is required.

Other issues with this PR that will come up during review I am happy to solve. 

2. Create an issue about speeding up appveyor build.

I have taken a look at apveyor in my toy project: https://github.com/all3fox/algos-py/ and it looks like appveyor is not actually slow at all. So need to figure out why it is taking so long and fix that.

3. Create an issue about possibly migrating to Circle CI, or not creating it, idk.

pombredanne:
@all3fox this is great. Please go ahead and create these other tickets. 

1. The bundling will be a mostly easy thing. I use https://github.com/pombreda/thirdparty and https://github.com/pombreda/thirdparty-manylinux/ to build wheels including native wheels on mac/linux/win alright. And they are then pushed to http://dl.bintray.com/pombreda/thirdparty/ for consumption.


2. Appevyor always feels faster than Travis here. Even though ScanCode runs faster on Linux than on Windows (because Python is typically faster there)


3. we can add circle and not remove Travis

4. I am open to anything on linting. We use pep8 but with relaxed line length in some cases so I expect this to blow up nicely as it should! btw, may be coala needs 3.4+ but can lint 2.7 code?

.travis.yml
```diff
@@ -1,35 +1,36 @@
...
+    - os: osx
+      language: generic
+      python:
+      osx_image: xcode7.3
 
 install:
-    - ./configure
+  - ./configure
+  - source bin/activate
+  - pip install pytest-cov coveralls
```
pombredanne: you could also eschew this activate call and use `bin/pip` instead for the same effect. Same for `bin/py.test` and `bin/coveralls` below too. Virtualenv-generated bin scripts have the logic to activate the virtualenv they belong to.

all3fox: Pushed an update, used `./bin` prefix everywhere. Added `./` because `./configure` and `./etc/release/release.sh` use it and I went consistent.

.travis.yml
```diff
@@ -1,35 +1,44 @@
...
+  - 2.7.12
+  - 2.7.13
 
+matrix:
+  include:
+    - os: osx
+      language: generic
+      python:
+      osx_image: xcode7.3
+    - os: osx
```
all3fox: Yes, again included two just because I could (the default one and the latest one). Left just the default one now.

all3fox:
Oh, wow.

First, I ditched the explicit `source bin/activate`, please review.

Second, I lurked around some issues here and saw that a common topic is clean commit messages. So, I went ahead and fixed mine: typos, grammar, etc. And then I got bitten hard: vanishing header lines! It took me a while to realise that starting commit message with `#564` actually turns it into a comment and removes it.

Luckily I heard of `git config core.commentchar ";"` before (this command uses `;` instead of `#` as your commentchar) but still had to spent some time searching for exact config.

I dont mean to say "let's stop doing that", just pointing out that it is surpisingly confusing (especially at 1 a.m. 😺 but granted, many things become confusing that late). 

Additionally, when I use `Refs: full/commit/url` and then scroll the results of `git log` in my terminal, I can Ctrl-click links and that takes me directly to page in browser. Obviously, no such effect with a simple `#564`. But I actually haven't seen many projects doing that. Even the ones who cared a lot about their commit messages.

pombredanne:
You bring a good point wrt to using #-prefixed issues ids. This can be a tad confusing with the default # being the git default for comment. And indeed this is not a clear URL.... Let me pounder all this. Eneter a ticket too so we can discuss this there.

PR #583, opened 2017-03-31
Edit README to correct commands for Debian and Ubuntu distros

johnmhoran:
This commit updates several of the README's installation commands for Ubuntu and Debian distributions.


JonoYang:
Looks good, @johnmhoran!

Issue #514, opened 2017-02-23
Add all the CC licenses to the ScanCode detectable licenses


pombredanne:
See https://github.com/creativecommons/creativecommons.org/tree/45420471049bbf7a6420a12c737376b6bf3fc9dd/docroot/legalcode
With all the translations and variants, this would be a good addition even if some are rather exotic...

pombredanne:
Another source is https://github.com/creativecommons/license.rdf/tree/3b47cda59c2ef60ad713ca14e111b9867a1be997/cc/licenserdf/licenses

pombredanne:
And this contains some code to handle them: https://github.com/warpr/licensedb.git

singh1114:
What is needed to be done to solve this issue?

pombredanne:
well, this is about adding new licenses and rules and detection tests as needed by checking that each and every of the CC licenses, including all the translations exist as ScanCode licenses in src/licensedcode/data/licenses

pombredanne:
the best way to do this would be to use some simple script to automate the basics

pombredanne:
You should also check recent PRs related to adding new licenses, rules and tests for examples.

singh1114:
Any particular one that you want to point out?

pombredanne:
Check the opened and closed PR. There are plenty of these and some are linked in the wiki doc on licenses

singh1114:
So what you want to say is that I need to create a script that will check the file names and compares the file names in the directory and the link?

singh1114:
Please see if the procedure that I am going to is right or wrong.

Let's say that I am adding the license for this one: https://github.com/creativecommons/creativecommons.org/blob/45420471049bbf7a6420a12c737376b6bf3fc9dd/docroot/legalcode/by-nc-sa_3.0_es_gl.html

First of all, I will check if the license already exist or not. As this is an HTML file I need open it in the browser. After that, I copy the content and save it in src/licensedcode/data/license, with a file name ending with .LICENSE and corresponding .yml file. And then I am done.

pombredanne:
@singh1114 let me come back with more details tomorrow. I have actually a script for something similar that I am working for SPDX licenses as part #41 .... Once I push this you could use it as a template and/or update this to support CC licenses.

aviaryan:
Hi @pombredanne .
I understand what is to be done in this issue. Should I try to solve it? 
Here is what I am thinking as of now - 
[...]

PR #587, opened 2017-04-04
#514 Add some CC licenses

aviaryan:
As discussed at https://github.com/nexB/scancode-toolkit/issues/514#issuecomment-291488367, this PR adds some CC licenses to the ScanCode database. 
These files were auto-generated using a script which can be found here (ugly code, no docs as of now). https://github.com/aviaryan/cc-licenses-parser-scancode

**Note** - I wasn't able to parse the `category` from the licenses. Therefore, they are fixed now at `Copyleft Limited`. @pombredanne Can you guide me a little bit about how these categories are decided?


aviaryan:
@pombredanne Can you please help me interpret the build logs. I can't figure out how the diff in this PR causes the build to fail.

pombredanne:
@aviaryan Thanks! Please check https://github.com/nexB/aboutcode/wiki/Writing-good-commit-messages
As for the CI build logs they are exactly the same as a local test run

In your case if you check this [job](https://ci.appveyor.com/project/nexB/scancode-toolkit/build/761/job/5h87w5fjjoe80bal) on windows for the license tests you see a failure in that test ` TestMatchAccuracyWithFullIndex.test_match_has_correct_line_positions_in_automake_perl_file` that will likely fail too if you run the test suite locally.
And since we do not run tests verbosely on the CI (otherwise we may exceed their max log size) you can rerun this test locally and see it fail. (You will need to check the py.test help for these options):
`py.test -vvs -k test_match_has_correct_line_positions_in_automake_perl_file` will likely do it.

aviaryan:
Hi @pombredanne 
Thanks for replying. 
So I ran the test locally and this is what I get. As you can see, none of the licenses mentioned here belong to the set that I have changed in this PR. So I don't get what I have done wrong. Sorry for bugging you again but I'm almost lost here. How is the build failing showing only the un-changed license files.
PS - I ran the same test as the [job](https://ci.appveyor.com/project/nexB/scancode-toolkit/build/761/job/5h87w5fjjoe80bal) you mentioned above. (`py.test -vvs -k test_match_has_correct_line_positions_in_automake_perl_file`)

```
E       AssertionError: assert [('gpl-2.0-pl...(1291, 1314))] == [('gpl-2.0-plu...(1292, 1315))]
E         At index 1 diff: (u'fsf-mit', (231, 238), Span(950, 1014)) != (u'fsf-mit', (231, 238), Span(951, 1015))
E         Full diff:
E         [(u'gpl-2.0-plus', (12, 25), Span(48, 159)),
E         -  (u'fsf-mit', (231, 238), Span(950, 1014)),
E         ?                                  ^     ^
E         +  (u'fsf-mit', (231, 238), Span(951, 1015)),
E         ?                                  ^     ^
E         -  (u'free-unknown', (306, 307), Span(1291, 1314))]
E         ?                                        ^     ^
E         +  (u'free-unknown', (306, 307), Span(1292, 1315))]
E         ?                                        ^     ^
```

src/licensedcode/data/licenses/cc-GPL-2.0-pt.yml
```diff
@@ -0,0 +1,11 @@
+key: cc-GPL-2.0-pt
+short_name: CC-GPL-2.0-PT
+name: >-
```
pombredanne: Please keep the names on one line

src/licensedcode/data/licenses/cc-GPL-2.0-pt.yml
```diff
@@ -0,0 +1,11 @@
+key: cc-GPL-2.0-pt
```
pombredanne: Use all lowercase keys

src/licensedcode/data/licenses/cc-GPL-2.0-pt.yml
```diff
@@ -0,0 +1,11 @@
...
+short_name: CC-GPL-2.0-PT
+name: >-
+    Creative Commons Licença Pública Geral do GNU (GPL) [General Public License]
+
+category: Copyleft Limited
+owner: Creative Commons
+homepage_url: https://creativecommons.org/licenses/GPL/2.0/
+spdx_license_key: CC-GPL-2.0-PT
```
pombredanne: Are you sure this is a valid SPDX  license identifier at spdx.org?

pombredanne:
What is likely happening here that by adding new licenses especially the ones in Portuguese and German you introduced a new batch of words aka. tokens that were never seen in the current set of licenses and rules. The way matching works is based on this set of tokens and whether they are known or not. This may subtly affect the way words are found and change the offsets where they are found (as these may only consider known words). Or this highlighted a bug somewhere in the code.

The fix is easy: e.g. update the expected spans in the test with the new observed values. Yet this is likely the sign of some bug elsewhere: so the fix should not be applied for now.

What would be great is to manually check in develop and in your branch what the matched words are exactly on the query side by running a scan of the test files in that test with `--license --diag --license-text` and verify if and then why the captured license texts may be different (or the same).

pombredanne:
Thanks for the quick commit!

aviaryan:
@pombredanne After rebase, build is still failing in the same manner. So I tried your suggestion of testing how captured licenses are differing in develop and my branch but it seems that the results come out identical. (See https://www.diffchecker.com/FiIaJdna)
Is there a scancode param to let us know at which position was a license matched ? (not line number, but absolute char position). 
As it seems from the error message I posted above, licenses are being detected correctly but they are being detected at a different position from what is specified in the test. 
I will now modify the test to see if this assumption is right. 

Thanks.

pombredanne:
@aviaryan can you post the results of running `scancode --license --diag --license-text --format json-pp` on the same file as used in the failing test and in both develop and your branch? 

that should tell us where the quirks is. It smells like a bug for now. Not in your code. But induced by the new data. So we need to fully investigate this

aviaryan:
> @aviaryan can you post the results of running scancode --license --diag --license-text --format json-pp on the same file as used in the failing test and in both develop and your branch?

Both outputs are identical. Please see https://www.diffchecker.com/hf6g9ZHZ for the logs.
I ran the following command. 

```
./scancode tests/licensedcode/data/positions/automake.pl  --license --diag --license-text --format json-pp
```

---

Also, I would like to repeat the question that I asked in my previous comment. 

> Is there a scancode param to let us know at which position was a license matched ? (not line number, but absolute char position).

Thank you @pombredanne for your patience.
Regards.

pombredanne:
> Is there a scancode param to let us know at which position was a license matched ? (not line number, but absolute char position).

not in the CLI. You could call the `Index.match()` and get `LicenseMatch` objects back with [this shortcut](https://github.com/nexB/scancode-toolkit/blob/0b6a589282133319e68da87d44c362ccbbf06550/src/licensedcode/index.py#L108). Note that the offset in a `Span` is never a character offset but instead a token position e.g. this is based on the tokenized "words" stream and not the character stream.

PR #590, opened 2017-04-11
URL unreachable if not 200 HTTP status

marcreyesph:
Signed-off-by: Marc Reyes <marcreyesmedia@gmail.com>

@pombredanne, I added something to fetch.py (which is stated as TO DO in the said file) where the URL to be fetched is in 200 HTTP status or not. Returns True if the HTTP status is 200 and False if not.

### DIFF
--- a/src/commoncode/fetch.py
+++ b/src/commoncode/fetch.py
@@ -77,18 +77,16 @@ def ping_url(url):
     """
     import urllib2
     # If there is no 200 HTTP status, then the URL may not be reachable.
-    http_response_error_list = [400, 401, 402, 403, 404,
-                                405, 406, 407, 408, 409,
-                                500, 501, 502, 503, 504,
-                                505, 506, 507, 508, 509]
     request = urllib2.Request(url)
     try:
         response = urllib2.urlopen(request)
-    except urllib2.HTTPError as e:
-        if e.code not in http_response_error_list:
-            return True
+    except urllib2.HTTPError as e:  # HTTP error like 404, 501, etc.
+        msg = 'HTTP error response code is %d' % e.code
+        print (msg)
         return False
     except urllib2.URLError as e:  # Not an HTTP-specific error (e.g. connection refused)
+        msg = 'Connection refused or no connection available'
+        print (msg)
         return False
     else:
         return True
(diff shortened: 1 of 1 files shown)


src/commoncode/fetch.py
```diff
@@ -76,11 +76,17 @@ def ping_url(url):
     Returns True is `url` is reachable.
     """
     import urllib2
-
-    # FIXME: if there is no 200 HTTP status, then the ULR may not be reachable.
+    # If there is no 200 HTTP status, then the URL may not be reachable.
+    http_url_error_list = [400,401,403,404,408,500,501,502,503,504,505]
```
pombredanne: Can you make sure this follow PEP8?

src/commoncode/fetch.py
```diff
@@ -76,11 +76,17 @@ def ping_url(url):
...
+    request = urllib2.Request(url)
     try:
-        urllib2.urlopen(url)
-    except Exception:
+        response = urllib2.urlopen(request)
+    except urllib2.HTTPError as e:
+        if e.code == any(http_url_error_list):
```
pombredanne: any would return True or False.... How can this possibly work? Also, you would be better off with a set of return code IMHO.

marcreyesph:
Thanks for the comments @pombredanne! I will apply necessary changes and commits to this pull request.

src/commoncode/fetch.py
marcreyesph: @pombredanne, what I did in the changes is for this to only have a single return code and print the HTTP response code as it returns a value of False.

src/commoncode/fetch.py
```diff
@@ -39,6 +39,11 @@
 # logging.basicConfig(level=logging.DEBUG, stream=sys.stdout)
 # logger.setLevel(logging.DEBUG)
 
+def test_download(url):
```
pombredanne: Would you mind call this check_download? Also what's the value to download and discard?

src/commoncode/fetch.py
```diff
@@ -39,6 +39,11 @@
...
+    if ping_url(url) == True:
+        download_url(url, file_name=None, verify=True, timeout=10)
+    else:
+        print('Failed to start download for', url)
```
pombredanne: You should raise an exception here instead

marcreyesph:
Thanks, @pombredanne but I'm closing this pull request. I don't believe it has a purpose at this time. Thanks for the helping me in my proposal. Thank you very much!

PR #594, opened 2017-04-13
add --fullroot option

armijnhemel:
Add an option for printing the full path of a file.


pombredanne:
Thanks! let me review. We likely need a test for this  and this also makes some test fail.

armijnhemel:
Plus likely make sure that --strip-root and --full-root cannot be used at the same time ;-)

pombredanne:
This has been merged with some tests. Thanks!

PR #597, opened 2017-04-14
#524 Remove automatic prepending of `/code` to resource paths

JonoYang:
Now that the ScanCode scan output includes the full resource path, `json2csv.py` has been modified to not automatically prepend `/code` to the resource path. The prepending of `/code` has now been made an option, `--prefix_path`. Tests have been modified to reflect this change and a new test has been created to test the `--prefix_path` option.


pombredanne:
Thanks! let me review

Issue #603, opened 2017-04-21
Lock timeout....


pombredanne:
```
 File "/home/scancode/scancode-toolkit/src/licensedcode/index.py", line 135, in get_index
    _LICENSES_INDEX = get_or_build_index_from_cache()
  File "/home/scancode/scancode-toolkit/src/licensedcode/cache.py", line 83, in get_or_build_index_from_cache
    with yg.lockfile.FileLock(index_lock_file, timeout=LICENSE_INDEX_LOCK_TIMEOUT):
  File "/home/scancode/scancode-toolkit/local/lib/python2.7/site-packages/yg/lockfile/__init__.py", line 43, in __enter__
    self.acquire()
  File "/home/scancode/scancode-toolkit/local/lib/python2.7/site-packages/yg/lockfile/__init__.py", line 99, in acquire
    raise FileLockTimeout()
yg.lockfile.FileLockTimeout
```

This came after this commit https://github.com/nexB/scancode-toolkit/commit/9be63735625d19cdc2486a654b701217144a3a49

It it happens with this setup at https://github.com/nexB/scancode-toolkit/tree/0cd685915b9cf3b13864ba8a172d11d9b68cdfb1

* 144 cores Intel, 256GB, 16T SSD running Debian 64
* `scancode -clp --verbose --license-text -f spdx-tv -n 4 --strip-root pathtoscan/ spdxfile`
* ... spawned in 40 parellel runs, with a loop to scan a whole Debian pool.

It ran OK with 16685bd94a7dd091f53ccfae8ec6f7d30f8c15bc

pombredanne:
This is likely introduced weirdly enough by https://github.com/zopefoundation/zc.lockfile/pull/8

The main diffs are:
 - yg.lockfile https://github.com/yougov/yg.lockfile/compare/2.1.1...2.2.2
 - zc.lockfile https://github.com/zopefoundation/zc.lockfile/compare/1.1.0...1.2.1

pombredanne:
@jaraco @jimfulton I am pinging you there for reference. 

You may have an idea why this could timeout at times.

This is a license scan for a full Debian packages pool. Overall since the debian pool is about 24K packages, this will run about 24,000 x 4 = 96,000 processes total, ~160 concurrent at anytime, e.g acquiring and releasing 96K lock file locks over ~ 11 hours.

[...]

PR #604, opened 2017-04-21
Bypass license index cache rebuild and consistency check by default

pombredanne:
* by default, the license index cache is no longer checked for
   consistency in all cases. Instead, the cache is created on the first
   run of ./confirgure or ./scancode (with a proper file lock) and then
   reused as-is without extra check
 * if the SCANCODE_DEV_MODE environment variable is set for development
   usage, then the cache consistency is checked in all cases.

Link: https://github.com/nexB/scancode-toolkit/issues/603
Reported-by: Benedikt Spranger <bene@linutronix.de>
Reported-by: Thomas Gleixner <tglx@linutronix.de>


pombredanne:
This look stable so far The thing for a "production" run is to run once ./configure or ./scancode -h before launching the massive scan. And re-run ./configure if working from a clone on each update

pombredanne:
@sschuberth you opinion is welcomed... since you run from a checkout, can you tell me what the procedure on your side?

This PR basically deactivates the license cache checks entirely when in "production" mode. e.g. once the license index has been created, it is not checked or rebuilt again even if some license or code files have changed. 
The "production" mode is when you use instead a release tarball or run `./configure etc/conf` or run a plain `./scancode` without a configure.

The dev mode is when you have a full clone or tarball and run `./configure` or `./configure etc/conf/dev`

For development there is an env var called `SCANCODE_DEV_MODE` that if present keeps the same behavior as before, and does a cache consistency check at each run of scancode (or rather at each run of something that requires the license index like the license detection proper)

Eventually I want to find a more automated way though.

pombredanne:
@sschuberth actually ignore all this... I am switching to use simpler approach with a tag file rather than an environment variable. The "dev mode" will be discovered with the presence of a "SCANCODE_DEV_MODE" file at the root of the repo which is simpler than a variable and exist regadless of the process and launch mode

pombredanne:
Merging this now!

sschuberth:
Sorry for chiming in late, @pombredanne. Here're some thoughts, even if this is merged now:

I was wondering whether you'd simply want to query Git to determine whether you're in "development mode". So instead of calculating checksums yourself, you could rely on Git's tree / object hashes, or even simply the output of `git ls-files -m -o` (or similar). That implies that you could only trigger "development mode" if you work from a Git checkout, but that's probably fine.

pombredanne:
@sschuberth I actually thought about using git for this, but there are cases where a checkout is used for production use and other cases where a tarball is used for development mode, hence the presence of git metadata is not a sure sign. Now in earnest this should not be needed at all, but this cache check did trigger some weird heisenbug-like errors at a certain scale (where 60 scans were running in // each with -`n4` on a 144 processor machine) ... so this is hard to reproduce BUT bypassing the cache checks removed the contention and the random bugs. And this was a non-development usage from a git clone.

Therefore, a tag file is simple enough and its presence is good enough IMHO, would you agree?

sschuberth:
>[...] hence the presence of git metadata is not a sure sign.

Not the presence alone, but the state of the files Git reports (modified, untracked etc.).

> Therefore, a tag file is simple enough and its presence is good enough IMHO, would you agree?

I agree it's good enough as a band-aid, and avoiding the Heisenbug is probably more important than an elegant solution that re-uses knowledge from Git (if present) about the state of the files, esp. if you need a fall-back to use development mode for a tarball (instead of Git) anyway.

Issue #611, opened 2017-05-03
Add size and origin_date fields to Package model in packagedcode


linexb:
This is required since PythonPackage should have it, and we should make it generic in Package model

pombredanne:
@linexb on second thought, using `release_date` is much better than `origin_date` and this is really what this is. I am merging and updating the code at the same time.

pombredanne:
Fixed.

PR #612, opened 2017-05-04
#611 add size origindate fields in Package Model

linexb:
add size and  origin_date fields in Package Model


pombredanne:
This has been merged in develop with some edits

Issue #542, opened 2017-03-06
Paths in JSON output are URL-encoded


sschuberth:
This is a follow-up to [this discussion](https://github.com/nexB/scancode-toolkit/issues/390#issuecomment-267562291) to explicitly state that paths are now URL-encoded in JSON output. For example when scanning https://github.com/AFNetworking/AFNetworking the output contains

```
{"path":"UIKit%2BAFNetworking/AFAutoPurgingImageCache.h","type":"file","name":"AFAutoPurgingImageCache.h","extension":".h", ...
```

So `+` is encoded as `%2B`. I'm not sure whether that's still intended. If so, it should be clearly documented, since it means you cannot just use the path to open the file as you have to URL-decode the path first.

pombredanne:
@sschuberth Thank I think having percent/url encoded paths is **pure utter non-sense** that I introduced.

**This should NOT be needed at all.** So I need to get rid of it. 

I think the issue/need would go away entirely by using Unicode at the boundary, right when a path is first ingested in the scancode CLI function and then used this way throughout, % encoding must not be needed at all.

pombredanne:
@sschuberth so if you agree with me this % stuff is non-sense, I will work out to remove this encoding entirely.

The paths in a scan should be readily usage as paths on the file system, period. (with some possible future caveat when we implement #14 )

sschuberth:
Agreed!

ashutoshsaboo:
@pombredanne I guess, when we implement #14 , would you want the paths of the extracted archives to be also appended to the JSON, since wasn't there a discussion about deleting the extracted files after the scanning is complete, as it would also lead to a lot of clutter for the user? Should we just add the path of the archive to all the paths of it's extracted files? Does that seem fine?

I would like to work on resolving #14 completely, so your input would do a lot of good. I have started my work in this PR - #544 for resolving #14 .

pombredanne:
@ashutoshsaboo you wrote:
> Should we just add the path of the archive to all the paths of it's extracted files? Does that seem fine?

sure.

pombredanne:
This is fixed in develop @sschuberth feedback welcomed

sschuberth:
Confirmed to be fixed in `develop`, thanks!

PR #628, opened 2017-05-25
542 unicode paths and more

pombredanne:
This is a fix for #542 and other tickets. The support for unicode paths, non-URL encoded actually has rather wide ranging impact....


pombredanne:
It was rather involved, but ... all the tests now pass on Linux, Mac and Windows.... Merging!

Issue #633, opened 2017-06-12
spdx-rdf results in output error


abuhman:
I am running into this error for the spdx-rdf output type:
'TypeError: must be unicode, not string'

When I run this command, it scans as expected with no errors:
    'sudo scancode --format spdx-tv test.py output.SPDX'
However, this command results in the TypeError above:
    'sudo scancode --format spdx-rdf test.py output.SPDX'

The error appears to happen when ScanCode tries to write to the output file. In fact, if I delete 'output.SPDX' from the command above and allow it to instead send the rdf to standard output, I do not get this error.

pombredanne:
@abuhman Thank you for the report.... though I could not reproduce this bug. Can you tell me which file you used for this and which version/commit of ScanCode you used?

abuhman:
I downloaded ScanCode from this link:  https://github.com/nexB/scancode-toolkit/archive/v2.0.0.rc2.zip .  While I did originally have this issue when running ScanCode from a file, I also have the same problem when I tried again running the same command directly from the command line, so it shouldn't need the file in order to be reproduced.  

This is the file I ran it from when I first had the issue: https://github.com/spdx/spdx-github/blob/master/run_scan/download_repo_run_scan.py.  This morning I've tried the scancode command from the command line again after typing 'source bin/activate' to make sure the virtual environment is running.  It also results in this error after starting the virtual environment.

pombredanne:
@abuhman ok, and which file did you scan? I am trying to know if this is a generic problem or specific to a certain file.

abuhman:
Okay, this is the file I scanned:

https://github.com/spdx/spdx-github/blob/master/run_scan/unit_test.py

I am also able to duplicate the error by scanning a local directory that does not contain this file.

pombredanne:
@abuhman can you paste the exact error you see here?

pombredanne:
@abuhman also ideally, could you contribute a test that fails and reproduces your bug?

abuhman:
Traceback (most recent call last):
  File "/usr/local/bin/scancode", line 11, in <module>
    load_entry_point('scancode-toolkit===2.0.0.rc2.post322.a11ef2a', 'console_scripts', 'scancode')()
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/scancode/utils.py", line 76, in main
[...]

Issue #474, opened 2017-02-01
Error when saving results as spdx-rdf


majurg:
When running the command: `scancode -clip -f spdx-rdf samples/ ~/out.rdf`, I encounter the following error and stack trace:

```
Traceback (most recent call last):
  File "/Users/sesser/Documents/src/scancode-toolkit/bin/scancode", line 11, in <module>
    load_entry_point('scancode-toolkit', 'console_scripts', 'scancode')()
  File "/Users/sesser/Documents/src/scancode-toolkit/lib/python2.7/site-packages/click/core.py", line 664, in __call__
    return self.main(*args, **kwargs)
  File "/Users/sesser/Documents/src/scancode-toolkit/src/scancode/utils.py", line 64, in main
    standalone_mode=standalone_mode, **extra)
  File "/Users/sesser/Documents/src/scancode-toolkit/lib/python2.7/site-packages/click/core.py", line 644, in main
    rv = self.invoke(ctx)
  File "/Users/sesser/Documents/src/scancode-toolkit/lib/python2.7/site-packages/click/core.py", line 837, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/sesser/Documents/src/scancode-toolkit/lib/python2.7/site-packages/click/core.py", line 464, in invoke
    return callback(*args, **kwargs)
  File "/Users/sesser/Documents/src/scancode-toolkit/src/scancode/cli.py", line 289, in scancode
    save_results(files_count, results, format, input, output_file)
  File "/Users/sesser/Documents/src/scancode-toolkit/src/scancode/cli.py", line 735, in save_results
    write_document(doc, str_buffer)
  File "/Users/sesser/Documents/src/scancode-toolkit/lib/python2.7/site-packages/spdx/writers/rdf.py", line 513, in write_document
    writer.write()
  File "/Users/sesser/Documents/src/scancode-toolkit/lib/python2.7/site-packages/spdx/writers/rdf.py", line 491, in write
    files = self.files()
  File "/Users/sesser/Documents/src/scancode-toolkit/lib/python2.7/site-packages/spdx/writers/rdf.py", line 237, in files
    return map(self.create_file_node, self.document.files)
  File "/Users/sesser/Documents/src/scancode-toolkit/lib/python2.7/site-packages/spdx/writers/rdf.py", line 211, in create_file_node
    license_info_nodes = map(self.license_or_special, doc_file.licenses_in_file)
[...]

Issue #588, opened 2017-04-05
Add actual license texts and captured notices to SPDX output


pombredanne:
Non-SPDX licenses are represented as `LicenseRef-<scancode license key>` in the SPDX output.
The actual corresponding text should be included in the results using typically these tags:
```
LicenseID:
LicenseName:
ExtractedText:
```
etc

The actual extracted text should be included too and is collected as part of #260

pombredanne:
@sschuberth in order to close in on 2.0, and also for #504 my plan for now would be to add a link to the scancode license and work out the proper solution immediately post 2.0? would this be OK as temp approach?

sschuberth:
Would you have an example of how exactly such a link would look like? Like, would it be easy to retrieve the license text from the link in a post-processing step on our side? In any case, I guess that's OK as a temporary solution.

pombredanne:
@sschuberth This is merged in develop. 
For now this is the "matched_text" if --license-text was used, otherwise this is a url to the license in the scancode repo (as a text is mandatory in SPDX)

#504 is still open to include the proper full text post 2.0

PR #637, opened 2017-06-14
Add tests for SPDX and fix crash 

pombredanne:
This is for #633 and #474 and this adds the missing tests for #338
@abuhman you should give a trry to this branch for #633


pombredanne:
@sschuberth this contains fixes to include the extracted license texts .... e.g. more or less about #588 but not yet full there. The logic is this:
If you used the `--license-text` option, then the texts are added to the SPDX RDF and TV formats
Otherwise for licenseref-like licenses, a link to the ScanCode .yml file for that license is added (this is because the extracted text for a licenseref is mandatory so we put something)

sschuberth:
Thanks, these certainly are good changes. In our case, however, it probably doesn't help us as we've switched to using JSON output and convert that to our slightly customized flavor of SPDX tag / value in a post-processing step. I.e. we'd require the JSON output to already contain the license text as described in #504.

pombredanne:
@sschuberth Thanks! And I will work out the rest of #504 for sure.
Out of curiosity, do you use the `spdx-tools` library then?

sschuberth:
> Out of curiosity, do you use the `spdx-tools` library then?

No, because it does not allow us add custom tags, or comments to standards tags. Esp. the later is a feature we needed.

Issue #635, opened 2017-06-14
Copyright holder not assigned for single word company names


edbratt:
Copyright "holder" for copyright statement:
  Copyright 1997-2012 Oracle and/or its affiliates. All rights reserved.
is not assigned. Additional words in Company name (e.g. Oracle America, Inc.) do result in assigned copyright "holder."
Would expect Copyright holder in example statement to be 'Oracle.'
(This results in Copyright Summary (html-app) to be confusing (Large number of parsed files, but only a small fraction of files shown with copyright attribution.)
Sample source: https://github.com/javaee/javamail/archive/JAVAMAIL-1_5_6.zip
Scancode toolkit version(s) 2.0.0rc2 and 1.6.0 behave similarly.
Python version: 2.7.10
Windows 7 SP 1 64 bit

pombredanne:
@edbratt Thank you for this report. Getting holders right is sometimes a pesky task. Let me check this.

pombredanne:
@edbratt btw, in the fix I pushed this is reported as `Oracle and/or its affiliates` .... This is not not what you want though and I could refine this alright.

edbratt:
Oracle and/or its affiliates is what we use so this is perfect. (And I merged this into my local copy and I can confirm this works for me. Looking forward to the update after it's passed the tests. Thank you very much!

pombredanne:
All set and merged. Closing.

PR #638, opened 2017-06-14
Improve detection of Oracle as a copyright holder

pombredanne:
This deals with the "Oracle and/or its affiliates" case better as reported in #635
@edbratt you can give it a spin, but I am confident that the fix is sound.
I will merge once all the CI tests are passing
Thanks again for the report!


pombredanne:
@edbratt Thanks for testing... Merging.

Issue #552, opened 2017-03-10
Plugins


pombredanne:
We need a plugin architecture such that it is easier to add new scanners, new formats and other features in a decoupled way.

At a high level, this is similar to Firefox, Chromium or Eclipse plugins.

The goal is to create a decoupled architecture such that output formatters and scanners (and in the future deduction tools) can:

1. be their own project/library and be plugged in ScanCode at installation time

2. for scanners and scan "pipeline processors" , they can possibly be coded in any language beside Python with Python being a special privileged language in any case. And accept an ABC data doc in and spit an ABC data out.

See for ABCD:
https://github.com/nexB/aboutcode/blob/a84e207bbe579dd9b419cc52bbfd00239f5beeb9/aboutcode-data/README.rst

For instance for packages scanners, a plugin approach such that new package manifests parsers are implemented as external libraries, eventually using different languages than Python to reuse the many parsers natively available (Ruby, Js, Java, etc) would make a lot of sense. .

Formatter/output should be the first to be implemented as plugins for a start, then eventually all scanners and more should be implemented as **plugins.** See #381 for format plugins. 

There stages where we can plugin things are, in this order of implementation priority:
- [x] output/formatters for JSON, HTML, etc . Tracked in #381 
- [x] post-scans hooks at the end of the scan, for instance for summarization/deduction for #377. Tracked in #699  
- [x] pre scan hook, before a scan (e.g. for instance archive extraction) #14
- [x] scan plugins "proper" that run a given scan. Tracked in #698. These may be specific to some file types or attributes and  may have dependencies on other scans. 

Other points:
- A plugin should be able to add new command line options even if minimally.

- The ability to have plugins in other languages than Python would be great: they would be spawned in their process and receive a path to scan and/or an ABC data packet and return an ABC data packet after having done their stuff.

- Some Python tools on Pypi such as pluggy, yapsy,  stevedore and the likes as well as possible click extensions may be candidate for this.

pombredanne:
with #639 we now have support for pluggable output formats. This ticket should be used to track the high level feature completion. @yashdsaraf  please create separate ticket for each pluggable features (e.g. pre-scan, scan, post-scan)

pombredanne:
With the latest merge of #885 this concludes the creation of plugins! Closing!

PR #639, opened 2017-06-15
Plugin architecture

pombredanne:
This is for #552
@yashdsaraf I am creating this PR so it is easier for me to review your code


setup.py
```diff
@@ -125,6 +125,9 @@ def read(*names, **kwargs):
         'bz2file >= 0.98',
         'text-unidecode >= 1.0, < 2.0',
 
+        # extractcode and scancode
```
pombredanne: This is for plugins at large not for ` extractcode and scancode` ... Also you might want to put an upper bound on the pluggy version (e.g. 0.5.0)

src/_pluggy/__init__.py
```diff
@@ -0,0 +1,26 @@
+#
```
pombredanne: Why would the module be named `_pluggy`?  something like `pluggedcode` `plugincode` would be better IMHO

src/_pluggy/hookspec.py
```diff
@@ -0,0 +1,17 @@
+import pluggy
```
pombredanne: You are missing a standard header and the `__future__` import including the unicode ones

src/scancode/cli.py
```diff
@@ -321,6 +323,10 @@ def scancode(ctx,
     Error and progress is printed to stderr.
     """
 
+    pm = pluggy.PluginManager('scan_proper')
```
pombredanne: is scan_proper the best name?

src/_pluggy/__init__.py
yashdsaraf: I was going to name it `pluggy` but that's the name of the library so I went with the PEP8 naming convention. Anyway, your suggestions make more sense.

yashdsaraf:
@pombredanne I'm currently fixing a bug, the plugins load multiple times. I'll start with tests as soon as I'm done with that. 👍

src/scancode/cli.py
yashdsaraf: I've used `pre_scan` `post_scan` markers, for the respective scan levels. Pertaining to those `scan_proper` made sense, I'd be happy to change it if you like. Here are some of my suggestions `proper_scan` `in_scan` or `scan_cli`.


_Sent from my Xiaomi MI 4W using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

src/formattedcode/writers.py
```diff
@@ -31,16 +31,20 @@
...
 from formattedcode.format import HtmlAppAssetCopyWarning
 from formattedcode.format import HtmlAppAssetCopyError
 
+from plugincode import hookspec
+
 
 def write_formatted_output(
         scanners, files_count, version, notice, scanned_files,
-        format, options, input, output_file, _echo):
+        format, options, input, output_file, _echo, plugin_formats, pm):
```
pombredanne: What is `pm` there?

src/formattedcode/writers.py
```diff
@@ -59,6 +63,13 @@ def write_formatted_output(
 
     elif format in ('spdx-tv', 'spdx-rdf'):
         write_spdx(version, notice, scanned_files, format, input, output_file)
+
+    elif format in plugin_formats:
+        pm.register(plugin_formats[format])
+        pm.hook.print_output(format=format, files_count=files_count, version=version,
```
pombredanne: Check out the alignment of the lines below

src/formattedcode/writers.py
```diff
@@ -31,16 +31,20 @@
 import os
 from os.path import abspath
 
+import pluggy
```
pombredanne: Could pluggy be encapsulated in plugincode instead rather than being imported here? Or actually I am not sure it is used at all  here and could be removed?

src/scancode/cli.py
```diff
@@ -223,20 +232,39 @@ def reindex_licenses(ctx, param, value):
...
 class ScanCommand(BaseCommand):
     short_usage_help = '''
 Try 'scancode --help' for help on options and arguments.'''
 
+    options = [ op for op in scan_proper_pm.hook.add_cmdline_option() if isinstance(op, click.Option) ]
+
+    def __init__(self, name, context_settings=None, callback=None,
+                 params=None, help=None, epilog=None, short_help=None,
+                 options_metavar='[OPTIONS]', add_help_option=True):
+        BaseCommand.__init__(self, name, context_settings, callback,
```
pombredanne: Would a super() call be better?

src/scancode/cli.py
```diff
@@ -223,20 +232,39 @@ def reindex_licenses(ctx, param, value):
...
+                 params, help, epilog, short_help,
+                 options_metavar, add_help_option)
+        plugins_ret_list = scan_proper_pm.hook.add_format()
```
pombredanne: what is the meaning of this variable name?

src/scancode/cli.py
```diff
@@ -223,20 +232,39 @@ def reindex_licenses(ctx, param, value):
...
+        for plugin_ret in plugins_ret_list:
```
pombredanne: same here: what is the meaning of this plugin_ret variable name?

src/scancode/cli.py
```diff
@@ -223,20 +232,39 @@ def reindex_licenses(ctx, param, value):
...
+            if (len(plugin_ret) == 2 and plugin_ret[0] not in plugin_formats):
```
pombredanne: this line is obscure: what is plugin_ret[0] ? why do you test its length?

src/scancode/cli.py
```diff
@@ -223,20 +232,39 @@ def reindex_licenses(ctx, param, value):
...
+                plugin_formats[plugin_ret[0]] = plugin_ret[1]
+        if plugin_formats:
+            for param in self.params:
+                if param.human_readable_name == 'format':
+                    param.help = param.help.replace(
```
pombredanne: This looks weird to have a whart for custom formats: I reckon you are trying to keep the same CLI UX, but it might be cleaner to update the CLI have a separate option for `--format-custom <path to format template` and remove this kludge

src/scancode/cli.py
```diff
@@ -223,20 +232,39 @@ def reindex_licenses(ctx, param, value):
 number of files processed. Use --verbose to display file-by-file progress.
 '''
 
+formats = ('json', 'json-pp', 'html', 'html-app', 'spdx-tv', 'spdx-rdf')
```
pombredanne: May be we have a disconnect here: to me each of `('json', 'json-pp', 'html', 'html-app', 'spdx-tv', 'spdx-rdf')` should become a format plugin... e.g. there should not be both formats from plugins and formats that are built in.

src/formattedcode/writers.py
yashdsaraf: It can be removed, it's left over from an older commit where the post scan plugin manager was instantiated in this file. I'll remove it now.

src/formattedcode/writers.py
yashdsaraf: It's the `post_scan` pluginmanager instance passed here by `cli.py`, I do get your point though (I hope), I'll rename it to be same as it is in `cli.py` i.e `post_scan_pm`

src/scancode/cli.py
yashdsaraf: I was planning to eventually phase it out as I go on to implement existing formats as plugins, that's why I left it there for now.

src/scancode/cli.py
yashdsaraf: The plugins defining the `add_format` method are supposed to return a list of exactly 2 items.
1. The name of the format
2. The plugin to be used as a callback if user chooses that format
`plugins_ret_list` is the list of all such lists returned by the plugins.

src/scancode/cli.py
yashdsaraf: `plugin_ret` is the list of exactly 2 items (as I mentioned in the above comment).

src/scancode/cli.py
yashdsaraf: I'm checking whether the list has exactly 2 elements (format name and post scan plugin to be used as its callback) and whether or not the returned format already exists in the `plugin_formats` dict. i.e. Repeated format names are not loaded.

src/scancode/cli.py
yashdsaraf: I'm thinking, as all formats shift to plugins, dynamically supplying the `format` option might be a good idea. That will preserve the current CLI UX and explicit modification of `format` help text like above wouldn't be needed. :)

src/formattedcode/writers.py
pombredanne: what about a name such as `post_scan_plugins` ? the fact that it is a manager is not relevant here... Actually would it be instead a `format_plugins` ? post scan would be after all the scans and before the formatting of an output.

src/scancode/cli.py
pombredanne: fair enough

src/scancode/cli.py
pombredanne: are they returning a list or a tuple of two values?
IMHO either the plugin is well behaved and it returns two values or it is not and you should warp this in a try/catch and bail out loudly with a meaningful error if it does not 
Note that in the case of an output format plugin there can be ONLY one. e.g. we dump only one format out. So I am not sure why you would carry a list there. Also the name would be best as `plugins` rather than `plugins_ret_list`

src/scancode/cli.py
pombredanne: ok, use unpacking instead and error out if the plugin is not well behaved

src/scancode/cli.py
pombredanne: But to actually do anything meaningful you should start to add a test ASAP... otherwise you are adding things that do not have much usage for now.

src/scancode/cli.py
pombredanne: perfect and again the custom format can be refactored out of the `--format` option as its own. Can you submit a separate PR with tests on this?

src/formattedcode/writers.py
yashdsaraf: Do you mean adding a new marker for `format_plugins`? Currently the formatting plugins are handled by the `post_scan` marker, according to that `post_scan_plugins` make sense IMO.

src/scancode/cli.py
yashdsaraf: I can, but wouldn't this feature assume that all formats are already acting as plugins?

yashdsaraf:
@pombredanne I am still porting formats to plugins, so I haven't pushed changes specific to that feature yet. :)

src/plugincode/formats.py
```diff
@@ -0,0 +1,72 @@
...
+    Validate formats and template files. Raise a BadParameter on errors.
+    """
+    if check_if_template(value):
+        return value
+    else:
+        return value.lower()
+
+def check_if_template(format):
+    format_lower = format.lower()
+    if format_lower in plugin_formats or format_lower in formats:
```
pombredanne: You could write instead:
` if format.lower() in (plugin_formats, format):`

src/plugincode/formats.py
```diff
@@ -0,0 +1,72 @@
...
+def validate_formats(ctx, param, value):
+    """
+    Validate formats and template files. Raise a BadParameter on errors.
+    """
+    if check_if_template(value):
+        return value
+    else:
+        return value.lower()
+
+def check_if_template(format):
```
pombredanne: may you could name this `is_template()` ?

src/scancode/utils.py
```diff
@@ -59,13 +59,21 @@ class BaseCommand(click.Command):
     # "Try 'scancode --help' for help on options and arguments."
     short_usage_help = ''
 
+    options = []
```
pombredanne: Is there any possible issue with this being a class-level attribute, e.g. a class global?

yashdsaraf: The idea is since the `BaseCommand` class is extended by both `ScanCommand` and `ExtractCommand`, the `get_params` method utilizes the `self.options` attribute for both of them. It's the same reason why `short_usage_help` isn't a class level attribute I believe.

setup.py
```diff
@@ -198,5 +200,13 @@ def read(*names, **kwargs):
             'scancode = scancode.cli:scancode',
             'extractcode = scancode.extract_cli:extractcode',
         ],
+        'scancode_plugins': [
+            'formats = plugincode.formats',
```
pombredanne: what is this "formats" exactly (compared to the other ones)?

src/plugincode/format_csv.py
```diff
@@ -0,0 +1,232 @@
+#
```
pombredanne: This file and all the format plugins (not the format plugin specs) should be in `formattedcode` .

src/plugincode/formats.py
```diff
@@ -0,0 +1,84 @@
...
+
+from functools import partial
+import os
+
+from pluggy import HookimplMarker
+import click
+
+
+echo_stderr = partial(click.secho, err=True)
+hookimpl = HookimplMarker('scan_proper')
```
pombredanne: I am confused about why this is called `scan_proper` rather than `scan_ouput` or `scan_output_format`

src/plugincode/formats.py
```diff
@@ -0,0 +1,84 @@
...
+    return option
+
+def validate_plugins(plugins):
+    try:
+        for formats, plugin in plugins:
+                for format in formats:
+                    if format not in plugin_formats:
+                        plugin_formats[format] = plugin
+                    else:
+                        raise Exception('Invalid plugin found: Duplicate format passed')
```
pombredanne: You need to provide which value is the root of the problem.
Also rather than if/else, test instead the error condition first: `if format in plugin_formats:` raise and you do not need an else

src/plugincode/formats.py
```diff
@@ -0,0 +1,84 @@
...
+    option = click.Option(('-f', '--format'), is_flag=False, default='json', show_default=True, metavar='<style>',
+              help=('Set <output_file> format <style> to one of the standard formats: %s '
+                    'or the path to a custom template' % ' or '.join(plugin_formats.keys())),
+              callback=validate_formats)
+    return option
+
+def validate_plugins(plugins):
+    try:
+        for formats, plugin in plugins:
+                for format in formats:
```
pombredanne: This indentation is incorrect, 8 spaces

src/plugincode/formats.py
```diff
@@ -0,0 +1,84 @@
...
+                    if format not in plugin_formats:
+                        plugin_formats[format] = plugin
+                    else:
+                        raise Exception('Invalid plugin found: Duplicate format passed')
+    except ValueError:
+        raise ValueError('Invalid plugin found: add_format returned too many values')
```
pombredanne: How is this related to `add_format`? is there a test for this?

src/plugincode/formats.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    Validate formats and template files. Raise a BadParameter on errors.
+    """
+    if is_template(value):
+        return value
+    else:
+        return value.lower()
+
+def is_template(format):
+    format_lower = format.lower()
```
pombredanne: Add a docstring

src/plugincode/formats.py
```diff
@@ -0,0 +1,84 @@
...
+    if format_lower in plugin_formats:
```
pombredanne: I do not like the global for plugin_formats.

src/plugincode/hookspec.py
```diff
@@ -0,0 +1,57 @@
...
+@scan_proper
+def add_cmdline_option(scan_output_plugins):
+    """
+    Return a click.Option instance which will be added to scancode.cli.ScanCommand
+    """
+    pass
+
+@scan_output
+def write_output(format, files_count, version, notice, scanned_files, options, input, output_file, _echo):
+    pass
```
pombredanne: You need to document this in docstring as this is the key help for a plugin writer

src/plugincode/hookspec.py
```diff
@@ -0,0 +1,57 @@
...
+
+@scan_output
+def add_format():
+    """
+    Return a unique format name and a plugin to act as a callback for that format
+    """
```
pombredanne: You need to document this better as this is the key help for a plugin writer. What is the a format name? what is a plugin? etc.

src/scancode/cli.py
```diff
@@ -109,6 +113,13 @@
 
 notice = acknowledgment_text.strip().replace('  ', '')
 
+scan_proper_plugins = PluginManager('scan_proper')
+scan_output_plugins = PluginManager('scan_output')
+scan_proper_plugins.add_hookspecs(hookspec)
+scan_output_plugins.add_hookspecs(hookspec)
+scan_proper_plugins.load_setuptools_entrypoints('scancode_plugins')
```
pombredanne: Why is this needed here now? the current plugins are only about the output formatting at this stage.

src/scancode/cli.py
```diff
@@ -109,6 +113,13 @@
 
 notice = acknowledgment_text.strip().replace('  ', '')
 
+scan_proper_plugins = PluginManager('scan_proper')
+scan_output_plugins = PluginManager('scan_output')
```
pombredanne: My hunch is that the PluginManager should get you all you need to validate the formats are unique and more? e.g. I feel this is what you should use and not the `plugin_formats` global custom mapping

src/scancode/cli.py
```diff
@@ -109,6 +113,13 @@
...
+scan_proper_plugins.add_hookspecs(hookspec)
+scan_output_plugins.add_hookspecs(hookspec)
```
pombredanne: Why add this twice?

src/scancode/cli.py
```diff
@@ -109,6 +113,13 @@
...
+scan_proper_plugins.load_setuptools_entrypoints('scancode_plugins')
+scan_output_plugins.load_setuptools_entrypoints('scancode_plugins')
```
pombredanne: I think that a format plugin is not a general "scancode_plugins" but rather a "scancode_format_plugin" of sorts (may be rather singular than plural?)

src/scancode/cli.py
```diff
@@ -223,25 +234,11 @@ def reindex_licenses(ctx, param, value):
...
-    Validate formats and template files. Raise a BadParameter on errors.
-    """
-    value_lower = value.lower()
-    if value_lower in formats:
-        return value_lower
-    # render using a user-provided custom format template
-    if not os.path.isfile(value):
-        raise click.BadParameter('Invalid template file: "%(value)s" does not exist or is not readable.' % locals())
-    return value
+    options = [ op for op in scan_proper_plugins.hook.add_cmdline_option(scan_output_plugins=scan_output_plugins) if isinstance(op, click.Option) ]
```
pombredanne: there is an extra space after `[` . is there a possibility that `op` is not a `isinstance(op, click.Option)` here?

pombredanne: Is this the best place to add the options as a class variable of the ScanCommand? what is the alternative?

src/scancode/cli.py
```diff
@@ -785,4 +776,4 @@ def save_results(scanners, only_findings, files_count, results, format, options,
                     raise e
         return
 
-    write_formatted_output(scanners, files_count, version, notice, results, format, options, input, output_file, echo_stderr)
+    write_formatted_output(files_count, version, notice, results, format, options, input, output_file, echo_stderr, scan_output_plugins)
```
pombredanne: I do not like that you pass a global here to an inner function call.... what exactly is held in `scan_output_plugins` ?

src/scancode/utils.py
pombredanne: I think this would be best only in ScanCommand if this is needed in anyway. Let's leave the extraction out of this entirely for now.

thirdparty/prod/pluggy.ABOUT
```diff
@@ -0,0 +1,18 @@
+about_resource:
+    - pluggy-0.4.0-py2.py3-none-any.whl
+    - pluggy-0.4.0.zip
```
pombredanne: remove this line, since this is not a file you have in thridparty.

tests/formattedcode/test_writers.py
```diff
@@ -412,3 +412,11 @@ def test_spdx_rdf_tree():
     result = run_scan_click(['--format', 'spdx-rdf', test_dir, result_file])
     assert result.exit_code == 0
     check_rdf_scan(expected_file, result_file)
+
+def test_csv():
```
pombredanne: I think there are many other tests to carry over from https://github.com/nexB/scancode-toolkit/blob/4a1d1d3df5b2121b6a42d88c5796d48cc59f602e/etc/scripts/test_json2csv.py right?

pombredanne:
What if each format was its own CLI option? (`--json`, `--json-pp`, etc?) rather than `--format json-pp` ....
@sschuberth what's would be your take with such a CLI UX change?

setup.py
yashdsaraf: It's the `scan_proper` plugin which calls all other `scan_output` formats are returns the `--format` option

src/plugincode/format_csv.py
yashdsaraf: Okay

src/plugincode/formats.py
yashdsaraf: `scan_proper` handles adding the command line options, whereas `scan_output` handles the formats.

src/plugincode/formats.py
yashdsaraf: Ah got it. 👍

src/plugincode/formats.py
yashdsaraf: It's returned when a plugin's `add_format` returns more than two values, say `(('json', 'json-pp)), 'format_json', 'extra')` would be invalid. I've written tests for the `validate_plugins` method which tests this scenario

src/scancode/cli.py
yashdsaraf: I've mentioned all the hook specifications for all scan stages in `hookspec.py` so only that needs to be added in each `PluginManager` instance.

src/scancode/cli.py
yashdsaraf: To add the command line options, the `add_cmdline_option()` method would call the `formats` plugin which in turn will return `--format` option with all the plugin formats attached.

src/scancode/cli.py
yashdsaraf: I've added `options` as a class variable so in future when `extractcode` needs to add command line options, it need not define its own `get_params` method but rather use the existing one defined its super class. This is similar to what happens with `short_usage_help`

src/scancode/cli.py
yashdsaraf: It's the `PluginManager` instance initialized with the `scan_output` marker in `cli.py`, it is used to call the appropriate format plugin. I think I can get rid of it by supplying the appropriate plugin directly as an argument, is that a good idea?

tests/formattedcode/test_writers.py
yashdsaraf: Oops, I forgot to check out that file, I'll add them now.

yashdsaraf:
So currently there are 2 ways to go,
1. Preserve the CLI UI and add a separate entrypoint for format plugins (this would give us something similar to the `plugin_formats` global
2. Modify the CLI UI as each plugin will return it's respective cmdline option

src/scancode/cli.py
pombredanne: To the contrary... I think the whole purpose of a PluginManager is to avoid the type of `plugin_formats` global you have been using

yashdsaraf: Yes, but that would require adding a separate entrypoint, which can again be totally avoided if we choose to modify the CLI UI. I think we need to discuss it more before heading in either direction. :)

sschuberth:
> What if each format was its own CLI option? (--json, --json-pp, etc?) rather than --format json-pp ....
@sschuberth what's would be your take with such a CLI UX change?

I think I'd prefer "not to pollute the top-level namespace" in terms of command line options and stick with `--format=`. If ScanCode ever has to deal with other input / output / config file formats it would otherwise be unclear to which if these e.g. a `--json` option would refer to.

pombredanne:
@sschuberth Thanks. @yashdsaraf case closed then. Each format plugin will add to the format option. Now I would quite OK for the sake of simplification to make the "custom format" a special case which would be dealt with separately with its own and separate `--custom-format <path to a template>`  option and could therefore not be a plugin but a built-in. This would simplify the format plugin interface and avoid the wart that this custom thing can be.

src/scancode/cli.py
pombredanne: Having a separate entry point for format is not what you have now?

yashdsaraf:
Well, `--custom-format` option would still be functioning the same as dynamically adding `--format` does (I mean everything supplied to `--format` option is *plugged in*), I think I should try the entrypoint approach, I'll get back to you with the results.

pombredanne:
there is a separate issue that can be dealt with later. ATM the --format comes at the bottom of the list of options and not in the right spot. We need to have a way to ensure that here the option is inserted at the right spot. This may require avoiding the decorators entirely and create the options programatically instead?

yashdsaraf:
A shorter way IMO would be to instead specify (just for `formats`) where in the `options` list it should be added, I think this can be done in the `ScanCommand` class.

pombredanne:
@yashdsaraf you wrote:

> A shorter way IMO would be to instead specify (just for formats) where in the options list it should be added, I think this can be done in the ScanCommand class.

That is ok as a very short term hack. 
But your `BaseCommand` class-level `options` attribute is not cleanly plugged in click with this approach. 

The right way might be to instead what I described here: https://github.com/pallets/click/issues/604#issuecomment-314497575 and in the case of plugins contributed options here and in general for other contributed options beside formats, this likely means not using the `@click` decorators anymore but instead crafting a `Command` and its `Option`s using plain non-decorator code.

src/plugincode/format_csv.py
yashdsaraf: @pombredanne What about `formats.py`? The plugin which returns the `--format` option, should it be moved to `formattedcode` as well?

src/plugincode/formats.py
pombredanne: IMHO for the case of formats, the options should be handled in one place not two.

src/plugincode/format_csv.py
pombredanne: IMHO there is a distinction here in `formattedcode` that provides plugins  and `formats.py` that provides the plugin mechanism and is core.
Eventually `formattedcode` could become a separate projects. Actual plugins contributed as a seperate packages could contribute, but in all cases the definition of the extension mechanism stays in the core. So it should not move to `formattedcode`

src/plugincode/formats.py
pombredanne: ok, since we agreed to use only a single format per plugin, this is not a possibility anymore, right?

src/scancode/cli.py
pombredanne: ok, as commented above, there should be only one hook spec for a format plugin, not two. The spec defines what and where in the scan pipeline a plugin contributes. Not how this is done or separating adding options from adding actual feature.

src/scancode/cli.py
pombredanne: extractcode will not need to add command line options in the foreseeable future.

src/plugincode/formats.py
yashdsaraf: The flow here is that in `scan_proper` stage `formats.py` plugin is called which in turn calls the `scan_output` formats and returns the option. Perhaps I should rename `formats.py` to make its purpose more clear.

src/plugincode/formats.py
yashdsaraf: It is, at least till we decide how  to manage the format_name:plugin_name map (plugin_formats global that is).

src/scancode/cli.py
yashdsaraf: I'm sorry I didn't get you there. Do you mean hookspecs for different scan stages should be defined in separate files?

src/scancode/cli.py
yashdsaraf: Right now there's a single entrypoint for all scan stages. But I'll add a separate entry point for formats and see how that works out in place of `plugin_formats` global.

src/plugincode/formats.py
pombredanne: OK, but I think that scan_proper is not for formats at all. Only one hookspec is needed for the formats. The `scan_proper` will be used for plugins that do a scan (e.g. license, copyrights, etc). They should not be mixed up

src/scancode/cli.py
pombredanne: I mean that scan_proper should NOT be used at all for formats. Only one hookspec should be needed for formats that handles the the format code for the option and the callable that will return this format.

src/scancode/cli.py
pombredanne: The simplest at this stage might be to remove anything that is not a hookspec for formats. You can add this later alright and this will make your code cleaner

yashdsaraf:
@pombredanne you wrote:
> The right way might be to instead what I described here: pallets/click#604 (comment) and in the case of plugins contributed options here and in general for other contributed options beside formats, this likely means not using the `@click` decorators anymore but instead crafting a Command and its Options using plain non-decorator code.

Is this what's required in #643 as well?

pombredanne:
@yashdsaraf yes, the click approach I suggested is likely the way for #643 too

src/formattedcode/format_csv.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+hookimpl = HookimplMarker('scan_output')
+
+@hookimpl
+def write_output(format, files_count, version, notice, scanned_files, options, input, output_file, _echo):
+    """
+    Convert a ScanCode JSON scan file to a nexb-toolkit-like CSV.
+
+    JSON_INPUT is either a ScanCode json format scan or the data.json file from a ScanCode html-app format scan.
```
pombredanne: I think this comment is no longer relevant, right?

src/formattedcode/format_html.py
```diff
@@ -0,0 +1,46 @@
...
+from __future__ import print_function
+from __future__ import division
+from __future__ import unicode_literals
+
+from pluggy import HookimplMarker
+
+from formattedcode.format import as_template
+
+
+hookimpl = HookimplMarker('scan_output')
```
pombredanne: Why do you redefine this each time? I think it would be cleaner to have this var named `scan_output` and defined once globally in plugincode, and imported from there in a plugin rather than too leak the fact we are using pluggy to every format plugin?

src/formattedcode/format_html.py
```diff
@@ -0,0 +1,46 @@
...
+
+@hookimpl
+def write_output(format, files_count, version, notice, scanned_files, options, input, output_file, _echo):
```
pombredanne: Why is `format` an arg required in an implementation? It does not seem to be used?

src/formattedcode/format_html.py
```diff
@@ -0,0 +1,46 @@
...
+    for template_chunk in as_template(scanned_files):
+        try:
+            output_file.write(template_chunk)
+        except Exception as e:
```
pombredanne: That's not directly related to plugins, but we need to clean this exception handling: here we echo then raise, elsewhere we only echo....

src/formattedcode/format_json_pp.py
```diff
@@ -0,0 +1,46 @@
+#
+# Copyright (c) 2017 nexB Inc. and others. All rights reserved.
```
pombredanne: I have have though of both json and json-pp would be in one module (eventhough there are two scan_ouput implementations)... That would avoid plugin scripts inflation. Especially since they should share a common function since the only difference in the indent arg?

src/formattedcode/format_spdx_tv.py
```diff
@@ -0,0 +1,188 @@
+#
```
pombredanne: Same as for JSON. IMHO duplicating the code in two modules is not worth it and error prone.
There should be two hook implementations within one module rather and shared code would not be duplicated then.

tests/formattedcode/test_writers.py
```diff
@@ -30,13 +30,16 @@
 import codecs
 import os
 import re
+import json
```
pombredanne: The imports order is not sorted correctly

src/formattedcode/format_spdx_tv.py
yashdsaraf: Instead of two hook implementations, how about defining the same plugin for both the formats in the `scancode_formats` entrypoint? Then the plugin can differentiate between the formats using the `format` parameter you referred to in a previous comment.

pombredanne: Two hooks implementations is the way to go IMHO. Otherwise you would introduce coupling between two plugins at the api level.
Note that a format plugin implementation should not know which format was requested (and it should not receive it as an argument) which is something I commented above.

So two hooks implementation functions that call some shared function is fine, having these aware that a certain format has been requested is not.

src/formattedcode/format_html.py
yashdsaraf: Got it. I'll define `HookimplMarker`s in their respecitve hookspec files, i.e. `scan_proper_hookspec.py` and `scan_output_hookspec.py`

src/formattedcode/format_spdx_tv.py
yashdsaraf: Got it. 👍

src/formattedcode/writers.py
```diff
@@ -27,235 +27,17 @@
...
-    # As the spdx-tools package can only write the document to a
-    # "str" file but ScanCode provides a "unicode" file, write to a
-    # "str" buffer first and then manually write the value to a
-    # "unicode" file.
-    from StringIO import StringIO
-    str_buffer = StringIO()
-    write_document(doc, str_buffer, validate=True)
-    as_unicode = str_buffer.getvalue().decode('utf-8')
-    output_file.write(as_unicode)
+    output_plugin.write_output(
```
pombredanne: I think the hookimpl callable name should not be hardcoded like you are expecting it here.
Instead you should get the hook impl from the plugin manager

I think this is why you end up using classes and always the same function name for hookimpl
What makes a hookimpl a hookimpl is the marker decorator, not its name

src/scancode/cli.py
```diff
@@ -238,12 +253,11 @@ def validate_formats(ctx, param, value):
     value_lower = value.lower()
     if value_lower in formats:
         return value_lower
-    # render using a user-provided custom format template
+    # render using a userprovided custom format template
```
yashdsaraf: Oh, I must've changed it by mistake. :P

src/formattedcode/writers.py
yashdsaraf: I think I got that, working on this now.

yashdsaraf: @pombredanne I've been looking into `pluggy` source, it seems it does indeed register/hardcode the callable name i.e `write_output`, see: https://pluggy.readthedocs.io/en/latest/_modules/pluggy.html#PluginManager.parse_hookimpl_opts
Here in parameters, **plugin** is the module or class passed and **name** is that plugin's name.

yashdsaraf: I got how to get the `hookcaller` of a plugin, but without the callable name, how do we differentiate one caller from other?

pombredanne: If you have setuptools entry points that provide a module or class, you will get a module or class, and then you will need to know the method/function name of what to call next. If instead this is callable, then you do not need to know this.
For instance:
in setup.py use this instead: `'csv = formattedcode.format_csv:write_csv_output'
and rename the function in `formattedcode.format_csv` to `write_csv_output` and you will receive a function alright from the plugin manager.

pombredanne: so the problem is that you pass modules or class as entry points, instead of a function or callable.

yashdsaraf: Oh, wow, didn't think of that. Thanks, I got it now. :)

pombredanne:
Your tests are failing. But I am adding more commits, so do not work on this just now

pombredanne:
@yashdsaraf I pushed a bunch of cleanups. So do not force push over this unless you rebased on this first.One more is likely needed to relocate the plugin managers in plugincode.

yashdsaraf:
@pombredanne Understood, can I push the code with relocated plugin managers?

pombredanne:
@yashdsaraf you can sure push anything you want on top in this branch

pombredanne:
@yashdsaraf actually you will need to rebase this on develop fist, and you can force push alright, just do not waste the commits I made there :P

pombredanne:
@yashdsaraf  actually rather let me push a quick fix so this merges alright and the test pass on merge

pombredanne:
@yashdsaraf done. Allyours. The latest commit should build alright including on merge

src/plugincode/scan_proper_hooks.py
```diff
@@ -39,3 +42,6 @@ def add_cmdline_option(scan_output_plugins):
     Return a click.Option instance which will be added to scancode.cli.ScanCommand
     """
     pass
+
+scan_proper_plugins = PluginManager('scan_proper')
+scan_proper_plugins.add_hookspecs(sys.modules[__name__])
```
pombredanne: Why not be explicit about which hookspec to load?

yashdsaraf: So that more hookspec methods can be added without changing the `add_hookspecs` code.

pombredanne: explicit is better imho. This is hard to figure out and is not something that would be added everyday

pombredanne:
I somehow created a mess of some possibly some recursive import that confuses py.test

src/plugincode/scan_proper_hooks.py
yashdsaraf: @pombredanne Is this still needed before merging?

yashdsaraf:
@pombredanne Perhaps it's because the plugins are loaded from hookspec file itself?

src/plugincode/scan_proper_hooks.py
yashdsaraf: Actually, what did you mean by being explicit about the hookspec? I mean `add_hookspecs` method does not take functions so a module needs to be passed. 
And since the current module is the one containing hookspecs, it needs to be passed.

pombredanne:
@yashdsaraf great! thanks! I would like to avoid exposing too much of pluggy outside of  plugincode... So let we wrap and guard this loading in a function and call this from cli.py

PR #642, opened 2017-06-19
Add 'EPL 1.0 or LGPL 2.1' rule

jpopelka:
After bd6ae3a170c ScanCode no longer finds EPL 1.0 in any file in [qos-ch/logback](https://github.com/qos-ch/logback/tree/master/logback-core/src/main/java/ch/qos/logback/core)

Therefore this change adds a separate rule.


pombredanne:
@jpopelka  Thanks ++ let me merge this ASAP!

pombredanne:
@jpopelka Thank++! I added an extra test for the sake of it.

pombredanne:
@jpopelka actually if I had added a test in the first place, there would not have been any regression!

PR #646, opened 2017-06-22
Use the full browser window width for the tables in the HTML report

sschuberth:
This way content is more convenient to view, esp. for long paths. Fix some
minor CSS formatting issues along the way.


pombredanne:
Looking good. Thanks!

pombredanne:
For this change there is no need to wait for all the CI tests to complete

Issue #426, opened 2017-01-04
Proposal: high level files classification


pombredanne:
To support #377 and other scan-based deduction and related refinements, an important step is to "classify" the files in the codebase being scanned. This would mean defining a few high level buckets and heuristics to classify a file in a bucket.

With this, smarter results could be provided: for instance the license of documentation files or build scripts does not have the same impact as the license of the main code (and may often not be part of a build or redistributed software as used in a system or app).

I am opening this up for discussion to determine what would be the classes. I think there should be as few classes as possible. They could be part of a hierarchy, but flat is probably better and simpler.

Here is a first shot at what these classes could be:

* main code: would be all the code proper that is effectively built and used when a piece of code is used.
* build scripts: such as Makefiles, poms, CMake lists, etc
* test code: any code that is used for testing the main code, either unit or integration or else. In many cases this is stored in a tree separated from the main code and often this is not part of the build meant to be used, but instead invoked during a build step (make check, or similar, etc)
* doc: any code and documents that are documenting the code and often may not be part of the built code. This often includes generated documents for API docs.
* assets and media: such as images, video, sounds, fonts, etc. often used in GUI and web apps. They have often different licenses and origins from the main code.
* dev tools: these are scripts, binaries, packages, etc present in the codebase but meant to be used for development and not production. Frequently, their provenance and license may have little impact on the resulting licensing of the main built code.
* metadata/metafiles: such as package manifests, LICENSE or COPYING files, etc that are describing top-level information for a codebase or a subset of it.
* generated code: such as when using a parser generator such as Bison/lex, some ORM such as Hibernate or some WSDL or else: this may represent a large volume of code at times and may not have directly identified provenance which needs to be traced to the "descriptor" used to generate the code. This may contain or not injected code plugs under various license (such as a bison skeletton)

Note that a file may end up in more than one class... not sure this would be a good thing.

Beside this classification, determining if some file is `deployed` or `not deployed` as part of a production build and `built` vs. `not built` is another topic altogether which would not be covered explicitly here.

majurg:
I think pointing out or emphasizing metadata files like LICENSE or COPYING in scan results would be a great addition. When I am doing analysis of 3rd party stuff, these are the first things I look at and if a project takes the time to include these, they are almost always correct. 

If these files appeared somewhere near the top of scan results where ever they are being viewed (html app or AboutCode manager), that would really be helpful during analysis.

majurg:
[...]

PR #702, opened 2017-07-26
Don't mark SVG files as source

jpopelka:
SVG file is XML and the only (I guess) XML that we want to mark as source is maven's pom.xml

### DIFF
--- a/src/typecode/contenttype.py
+++ b/src/typecode/contenttype.py
@@ -459,9 +459,11 @@ class Type(object):
             return False
 
         ft = self.filetype_file.lower()
-        pt = self.filetype_pygment
+        pt = self.filetype_pygment.lower()
 
-        if not 'xml' in ft and (pt or self.is_script is True):
+        if 'xml' not in ft and \
+           ('xml' not in pt or self.location.endswith('pom.xml')) and \
+           (pt or self.is_script is True):
             return True
         else:
             return False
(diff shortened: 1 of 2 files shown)


pombredanne:
Thanks! 
Actually there may be also `.pom` and .Net `.xaml` and a good bunch ASP-related things and code from android .xml files... which could be considered to be source code .... this is unfortunately not straightforward!

The only sane way I know of to decide what a given xml file really is "in the wild" is to parse it and then assert some things based on its structure and possible schema (which is what used to/is done be done in Eclipse to recognize/describe what an XML file really is for instance here (https://github.com/eclipse/eclipse.platform/blob/3dce93196b299c5d1852d9a8fbe308b6fc42b32f/ant/org.eclipse.ant.core/src/org/eclipse/ant/internal/core/contentDescriber/AntBuildfileContentDescriber.java) )..

Now on the other hand large XML data files are legions and a curse. And tagging them as sources upfront is mostly counter productive.

Also I would likely consider a pom.xml not as source code, but instead as a package manifest and build script of sorts, the same way an RPM spec file or Debian control files would be manifests. 

So with all this said, I have no issue with your proposed change... but in the end this would need to be refined with the upcoming #426 for a higher level file classification (and then I would also likely include Github linguist or its Python port(https://github.com/douban/linguist/network) )
And this is a classification that could then be used to drive and configure what should be or not be scanned.

pombredanne:
Mreging... Thanks again!

jpopelka:
Thank you for the description and the merge.

Issue #699, opened 2017-07-22
Post scan plugins for plugin architecture


yashdsaraf:
#552, #377 
For now at a simpler level, the plugin is supposed to return at a directory level if all or the majority of a directory files are source code.
The goal is to set the `is_source` flag to True for a directory if and only if the majority (~90%) of the files under a directory have `is_source` set to True.
The flow would be that the plugin accepts `scanned_files` iterable, does its job and returns an iterable/yields updated results.
These results should then be updated in the cache (not by the plugin)
Each plugin for now should be defined by a single new option (one for each such plugin) and all such options should be grouped together in the CLI help (#643).

pombredanne:
Can you elaborate a bit on the your design here?

pombredanne:
This now complete in the post #885 merge

PR #704, opened 2017-07-27
#699 Post scan plugins

yashdsaraf:
#699, #552


setup.py
```diff
@@ -215,6 +215,16 @@ def read(*names, **kwargs):
...
             'csv = formattedcode.format_csv:write_csv',
+        ],
+
+        # scancode_post_scan is an entry point for post_scan_plugins.
+        # See plugincode.post_scan module for details and doc.
+        # note: the "name" of the entrypoint (e.g only-findings)
+        # becomes the ScanCode CLI boolean flag used to enable a
+        # given post_scan plugin
+        'scancode_post_scan': [
+        'only-findings = scancode.plugin_only_findings:process_only_findings',
```
pombredanne: Please indent these ;)

src/scancode/cli.py
```diff
@@ -239,6 +241,24 @@ class ScanCommand(BaseCommand):
...
+                 options_metavar='[OPTIONS]', add_help_option=True):
+        super(ScanCommand, self).__init__(name, context_settings, callback,
+                 params, help, epilog, short_help, options_metavar, add_help_option)
+        for name, callback in plugincode.post_scan.get_post_scan_plugins().items():
+            option = ScanOption(('--' + name,), is_flag=True, help=callback.__doc__.strip(), group=POST_SCAN)
+            self.options.append(option)
+
+    def get_params(self, ctx):
+        """
+        Overridden from click.Command to return plugged in options in addtion to the
```
pombredanne: Typo in addtion

pombredanne: Now is this override still needed? What is it used for?

src/scancode/plugin_mark_source.py
```diff
@@ -0,0 +1,52 @@
...
+def process_mark_source(scanners, results, options):
+    """
+    Set `is_source` to true for all packages with (>=90%) of source files.
+    """
+    if not options['--info']:
+        return
+    for scanned_file in results:
+        if is_dir(scanned_file['path']):
+            source_files_count = files_count = 0
+            for file in file_iter(scanned_file['path']):
```
pombredanne: Why are you re-walking the file system here?
All the data should be in in the results, right?
This brings a couple of food for thoughts:
- how to efficiently navigate up and down the results "tree" of files without re-walking the FS?
- what if a post-scan plugin needs specific scans to be able to operate? for instance say I make a post scan plugin that filters out scanned URLs and emails if these are contained in the detected text of a license scan.... It can only work if the required scans   (urls, emails, license) are there. I guess there are a few ways to handle this: 
  - one is to force adding these scan options even if not selected IFF the post-scan plugin in question was selected . This may be a tad rude.
  - one is to have the plugin check that its re-requisite are satisfied on its own and either return a failure or a warning
  - one is to have the plugin declare its re-requisites and let the plugin framework handle it (and either return a failure or a warning)

src/scancode/cli.py
yashdsaraf: Yes, it returns the plugged options as well, the `self.options` attribute stores the options created in `__init__` which this method returns on every call

yashdsaraf: Actually, I just realized, I can just append the new options to `self.params`.

src/scancode/plugin_mark_source.py
yashdsaraf: Currently, the plugin runs only if `--info` option was selected.

yashdsaraf: For handling the prerequisite scans, I think
> one is to force adding these scan options even if not selected IFF the post-scan plugin in question was selected . This may be a tad rude.

This approach seems to be the way to go, we can also add a note for this in plugin's CLI help text

yashdsaraf: > Why are you re-walking the file system here?
All the data should be in in the results, right?

> how to efficiently navigate up and down the results "tree" of files without re-walking the FS?

I'm looking into this now

pombredanne: Well given the fact that I received reports of folks have GB-size scan results, there is only one sane answer possible: a database ;)
The cache from which results are iterated should be an opaque structure that is strictly internal and whose details should not leak to plugins and the outside world. It was designed to stream from disk and avoid having everything loaded in memory.

So the right way would be some internal DB (sqlite) with some very simple schema and API on top.
Sqlite locking should not be an issue: by construction a post-scan plugin is kinda like the reduce phase of a map reduce and has exclusive access to all scan results in a single process.

We could use some dict-like wrapper on sqlite db, or a more featured ORM, such as Django (I have POC code for using Django in the `orm-models` branch for #275 )...

In any case this is not something trivial. I think this could be best left out as a separate ticket for future developments ass this is not strictly tied to the plugin architecture proper

Unless you have some brilliant idea that could handle things nicely.

yashdsaraf: Okay, my brain can't come up with anything that doesn't send post-scan plugin back to mapping phase (i.e either re-walking the FS or iterating the scanned_results over and over to find a directory's content files which I am afraid is even worse).

pombredanne: That's fine! Enter a separate ticket to deal with this. For now your example plugin shows what can be done as-is. Later we can add a DB storage ;) ... And we can focus on moving on to pre-scan plugins next!

yashdsaraf: Ok, I'll push the changes I've made. Should I add tests for the `mark_source` plugin?

tests/scancode/test_mark_source.py
```diff
@@ -0,0 +1,51 @@
...
+class TestMarkSource(FileBasedTesting):
+
+    test_data_dir = path.join(path.dirname(__file__), 'data')
+
+    def test_scan_mark_source_without_info(self):
+        test_dir = self.get_test_loc('mark_source/test')
+        expected_file = self.get_test_loc('mark_source/without_info.expected.json')
+        result_file = self.get_temp_file('.json')
+        result = run_scan_click(['--mark-source', test_dir, result_file])
+        assert open(expected_file).read() == open(result_file).read()
```
pombredanne: FYI, if you check other scan tests, there is a helper method to help with doing more robust asserts on scan results.... See test_clip.py

Now does --mark-source expects the --info option?

yashdsaraf: Yes it does. I also added a note in it's help text for same
> Set `is_source` to true for all packages with (>=90%) of source files. Has no effect unless --info is requested.

pombredanne: See https://github.com/nexB/scancode-toolkit/blob/22bf88df19768dd207973ed2db91249822a0c446/tests/scancode/test_cli.py#L275 for a test example.
The `check_json_scan` thing abstracts dates and keys order: https://github.com/nexB/scancode-toolkit/blob/develop/src/scancode/cli_test_utils.py#L45

src/plugincode/post_scan.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+@post_scan_spec
+def process(scanners, results, options, input):
+    """
+    Process the scanned files and yield the modified results.
+    Parameters:
+     - `scanners`: an ordered dict of all possible scans
+     - `results`: an iterable of scan results for each file
+    """
```
pombredanne: What are options and input there? Can you document?

src/scancode/cli.py
```diff
@@ -461,8 +468,14 @@ def scancode(ctx,
         if not quiet:
```
pombredanne: IMHO this message should come after the post scan plugin runs.
And we should also have a message when not quiet telling that we are running post scans plugins and which one

pombredanne: I do not get where and when the post scan plugins are invoked.... can you help me there?

src/scancode/cli.py
```diff
@@ -458,11 +465,19 @@ def scancode(ctx,
             full_root=full_root,
             ignore=user_ignore)
 
+        for option, process in plugincode.post_scan.get_post_scan_plugins().items():
+            if kwargs[option.replace('-', '_')]:
+                options['--' + option] = True
+                if not quiet:
+                    echo_stderr('Running post-scan plugin: %(option)s...' % locals(), fg='green')
+                results = process(scanners, results, options, input)
```
pombredanne: Are you calling process once for each option? I am not sure Get the calling logic here. Where are the post scan plugins called vs where is the main call the run the "scanners"?

yashdsaraf: `process` is called for a `post-scan` plugin only if user supplied its corresponding option,
`scanners` are called in the `scan` function [here](https://github.com/nexB/scancode-toolkit/blob/7f2fb29d21e42b394b58ab1680bd0bccf5a2f6f3/src/scancode/cli.py#L455)
where as post-scan plugins are called [here](https://github.com/nexB/scancode-toolkit/blob/7f2fb29d21e42b394b58ab1680bd0bccf5a2f6f3/src/scancode/cli.py#L473)

yashdsaraf: Also, I'm not sure what you mean by *calling* the `scanners`, I mean `scanners` is just passed as a param which might or might not be used by a `post-scan` plugin.
e.g `only-findings` uses it, whereas `mark-source` doesn't

src/scancode/cli.py
pombredanne: we discussed this alright.

src/scancode/cli.py
```diff
@@ -458,11 +465,30 @@ def scancode(ctx,
...
+        # Find all scans that are both enabled and have a valid function
+        # reference. This deliberately filters out the "info" scan
+        # (which always has a "None" function reference) as there is no
+        # dedicated "infos" key in the results that "plugin_only_findings.has_findings()"
+        # could check.
+        # FIXME: we should not use positional tings tuples for v[0], v[1] that are mysterious values for now
+        active_scans = [k for k, v in scanners.items() if v[0] and v[1]]
+
+        for option, post_scan_handler in plugincode.post_scan.get_post_scan_plugins().items():
+            if kwargs[option.replace('-', '_')]:
```
pombredanne: could you create a variable for this, something named like `is_requested`?

src/scancode/cli.py
```diff
@@ -458,11 +465,30 @@ def scancode(ctx,
...
+                options['--' + option] = True
+                if not quiet:
+                    echo_stderr('Running post-scan plugin: %(option)s...' % locals(), fg='green')
+                results = post_scan_handler(active_scans, results)
+
+        # FIXME: computing len needs a list and therefore needs loading it all ahead of time
```
pombredanne: This code block should be executed if and only if a post scan plugin was invoked. Otherwise in the general case this is not needed and make every scan pay for the extra memory usage. So you should first loop to get all enabled post scan plugins. Then run your loop as above on these. And only if this list is not empty, then recompute the files_count

src/scancode/plugin_mark_source.py
```diff
@@ -0,0 +1,64 @@
...
+def process_mark_source(active_scans, results):
+    """
+    Set `is_source` to true for all packages with (>=90%) of source files. Has no effect unless --info is requested.
+    """
+
+    # FIXME: this is forcing all the scan results to be loaded in memory
+    # and defeats lazy loading from cache
+    loaded_results = list(results)
+
+    if 'type' not in loaded_results[0]:
```
pombredanne: can you make this a variable with an explicit name? otherwise what is happening here is not crystal clear. Now it could be called `has_file_info` or something else.... or better since you are passing the list of active scans, you can test that arg instead?

src/scancode/plugin_mark_source.py
```diff
@@ -0,0 +1,64 @@
...
+        for scanned_file in loaded_results:
+            yield scanned_file
+        return
```
pombredanne: replace this block instead by either `return iter(for scanned_file in loaded_results)` or a generator expression such as `return (scanned_file for scanned_file in loaded_results)`

src/scancode/plugin_mark_source.py
```diff
@@ -0,0 +1,64 @@
...
+
+    for scanned_file in loaded_results:
+        if scanned_file['type'] == 'directory' and scanned_file['files_count'] > 0:
```
pombredanne: Are you sure that the order of iteration is top-down, breadth first? (e.g. that directories comes before their children) ... I am not sure this is specified in anyway. .... Ahh! now I see below that you are iterating twice the list.   OK, this is likely to break quickly at scale, but may be OK for now.

src/scancode/plugin_mark_source.py
yashdsaraf: The `active_scans` list can't be tested for `--info` scan as noted here, 
https://github.com/nexB/scancode-toolkit/blob/fe29b8c6279adc69f40680585a701cd769caff29/src/scancode/cli.py#L469
I'll add an explicit name instead.

src/scancode/plugin_mark_source.py
yashdsaraf: Wouldn't this lead to a `SyntaxError` since generators don't allow return statements with arguments?

src/scancode/cli.py
yashdsaraf: @pombredanne If I may, wouldn't it be a better approach to rather set a boolean variable which is set to `True` if any `post-scan` plugins are executed.
Then we can check that variable to decide whether or not to recompute `files_count`

src/scancode/plugin_mark_source.py
pombredanne: may be... give it a try

src/scancode/cli.py
pombredanne: sure! good idea

src/scancode/plugin_mark_source.py
pombredanne: hum... this is an ugly wart. Not your fault but a wart. Work around it and we can fix it up later

src/scancode/plugin_mark_source.py
yashdsaraf: It does give out `SyntaxError: 'return' with argument inside generator`.

src/scancode/cli.py
```diff
@@ -458,11 +465,35 @@ def scancode(ctx,
...
 
+        # Find all scans that are both enabled and have a valid function
+        # reference. This deliberately filters out the "info" scan
+        # (which always has a "None" function reference) as there is no
+        # dedicated "infos" key in the results that "plugin_only_findings.has_findings()"
+        # could check.
+        # FIXME: we should not use positional tings tuples for v[0], v[1] that are mysterious values for now
+        active_scans = [k for k, v in scanners.items() if v[0] and v[1]]
+
+        plugins_invoked = False
```
pombredanne: This is too generic as a variable name but I will merge this and rename when I push

yashdsaraf: Okay :)

pombredanne:
All merged! Thanks!

Issue #697, opened 2017-07-22
Pre scan plugins for plugin architecture


yashdsaraf:
#552, #14

pombredanne:
It might best to explain you design here than in #719 

And after #719 is merged we will need to create a sample new plugin that would skip files based on their size and filetype data (which comes from the info, and therefore makes things a tad more complicated than the simpler, path-based filter)

yashdsaraf:
Every pre-scan plugins has to extend the AbstractFactory class
In it the get_option_attrs method will be called to get the attributes for the plugin's cmdline option, so plugin doesn't handle the creation of click options directly.
The process_resources method will be called in resource_paths function for plugins which require to make changes before the paths are supplied for scanning (e.g like --ignore), for plugins which do not need this method can simply call the super method which returns the resources back.

I chose this approach pre-scan plugins can't be plugged in a single spot, so AbstractFactory can add more method signatures in the future which can be called at different stages in cli.py

yashdsaraf:
Also `get_option_attrs` is a staticmethod since it gets called before a plugin class is initialized,
later if user supplied the cmdline option corresponding to that plugin, then the option's input is passed to that plugin's `get_instance` which is a classmethod.

pombredanne:
ok, this is a tad complex for my taste. Could there be a simpler way?

yashdsaraf:
Well, we can't plug-in pre-scan plugins in a single spot, like post-scan plugins.
And we can't assume attributes for click options for these plugins.

yashdsaraf:
How about having different hooks for different functions?
e.g
- for processing the `resources` i.e function which returns all the absolute paths of files to be scanned we could have a separate plugin marker (this can handle the --ignore part)
- for adding new scans or deleting them a different marker

This is something like `output` and `post-scan` are different markers

yashdsaraf:
@pombredanne I'm now using a normal class `PreScanPlugin` which needs to be extended by all `pre-scan` plugins.
It has an attribute `option_attrs` which defined click option attributes for the corresponding plugin. Click option name is taken from the entry point and help text from its docstring.
For now, it has `process_resources` method which is called when files to be scanned are decided.
We can add more methods to this class which can represent different hooks, say to add or delete a scan we might have a method called `process_scan`.
If a plugin's option was passed by the user, then the plugin class is instantiated with the user input and its above mentioned methods are called.
This way a plugin need not override all methods.

pombredanne:
[...]

PR #719, opened 2017-08-09
#697 Pre scan plugins

yashdsaraf:
Signed-off-by: Yash D. Saraf <yashdsaraf@gmail.com>


pombredanne:
Can you elaborate on the deisgn? The use of abstract meta class seems like a big thing here.

src/plugincode/pre_scan.py
```diff
@@ -0,0 +1,93 @@
...
+from pluggy import HookimplMarker
+from pluggy import HookspecMarker
+from pluggy import PluginManager
+
+
+pre_scan_spec = HookspecMarker('pre_scan')
+pre_scan_impl = HookimplMarker('pre_scan')
+
+@pre_scan_spec
+class AbstractFactory:
```
pombredanne: AbstractFactory is a too generic name: tell in its name what this is in more specific terms

src/plugincode/pre_scan.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @abstractmethod
+    def process_resources(self, resources):
+        """
+        Yield the absolute paths after processing.
+         - `resources`: a generator with absolute paths of files to be scanned.
+        """
+        return resources
+
+    @staticmethod
```
pombredanne: A staticmethod is weird here. Why this? a classmethod would be best

src/plugincode/pre_scan.py
```diff
@@ -0,0 +1,93 @@
...
+        Returns a dict of attributes to be used while creating the option
+        for this plugin.
+        """
+        return {}
+
+    @classmethod
+    def get_instance(cls, user_input):
+        """
+        Return the current class object initialized with input supplied to
+        the respective plugin cmdline option.
```
pombredanne: cmdline? a tad terse for a doc

src/plugincode/pre_scan.py
```diff
@@ -0,0 +1,93 @@
...
+
+pre_scan_plugins = PluginManager('pre_scan')
+pre_scan_plugins.add_hookspecs(sys.modules[__name__])
+
+
+def initialize():
+    # NOTE: this defines the entry points for use in setup.py
+    pre_scan_plugins.load_setuptools_entrypoints('scancode_pre_scan')
+    for plugin in pre_scan_plugins.get_plugins():
+        AbstractFactory.register(plugin)
```
pombredanne: what is the register method doing?

src/scancode/cli.py
```diff
@@ -456,7 +469,7 @@ def scancode(ctx,
             scans_cache_class=scans_cache_class,
             strip_root=strip_root,
             full_root=full_root,
-            ignore=user_ignore)
+            plugins=plugins)
```
pombredanne: is this really plugins and not pre_scan_plugins?

src/scancode/cli.py
```diff
@@ -725,11 +737,11 @@ def resource_paths(base_path, user_ignores):
     base_path = os.path.abspath(os.path.normpath(os.path.expanduser(base_path)))
     base_is_dir = filetype.is_dir(base_path)
     len_base_path = len(base_path)
-    ignores = dict()
-    ignores.update(user_ignores)
-    ignores.update(ignore.ignores_VCS)
-    ignored = partial(ignore.is_ignored, ignores=ignores, unignores={})
+    ignored = partial(ignore.is_ignored, ignores=ignore.ignores_VCS, unignores={})
```
pombredanne: does this mean that a pre scan plugin is limited for now to ignore files only?

tests/scancode/test_ignore_files.py
```diff
@@ -68,49 +68,49 @@ def test_ignore_glob_file(self):
...
-            'user',
-            'user/ignore.doc',
-            'user/src',
-            'user/src/ignore.doc',
-            'user/src/test',
-            'user/src/test/sample.txt'
-        ]
-        test = [rel_path for abs_path, rel_path in resource_paths(test_dir,{'sample.doc': 'test ignore'})]
-        assert expected == sorted(test)
+    # def test_resource_paths_with_single_file(self):
```
pombredanne: Why is this commented out? You could instead keep a function somewhere that is testable? It is likely dangerous to remove this test

pombredanne: The same applies to the other tests.

yashdsaraf:
Every `pre-scan` plugins has to extend the [AbstractFactory](https://github.com/nexB/scancode-toolkit/pull/719/files#diff-5cb8c26dd5b585283916cdf8145b65d1R41) class
In it the `get_option_attrs` method will be called to get the attributes for the plugin's cmdline option, so plugin doesn't handle the creation of click options directly.
The `process_resources` method will be called in `resource_paths` function for plugins which require to make changes before the paths are supplied for scanning (e.g like --ignore), for plugins which do not need this method can simply call the super method which returns the resources back.

I chose this approach pre-scan plugins can't be plugged in a single spot, so `AbstractFactory` can add more method signatures in the future which can be called at different stages in `cli.py`

src/plugincode/pre_scan.py
yashdsaraf: Since `get_option_attrs` is called before the plugin class is initialized, and whatever input (if any) is supplied to the plugin's option, it is passed to `get_instance` which is a `classmethod`

src/plugincode/pre_scan.py
yashdsaraf: It makes sure all the plugins subclass `AbstractFactory` and override all `abstractmethod`s

tests/scancode/test_ignore_files.py
yashdsaraf: I need to add updated tests for `resource_paths` since that function's signature in modified. I was waiting for your go for the abstract class approach :P

tests/formattedcode/data/json/simple-expected.json
```diff
@@ -6,7 +6,6 @@
     "--package": true,
     "--info": true,
     "--license-score": 0,
-    "--ignore": [],
```
yashdsaraf: Yes

src/scancode/plugin_ignore.py
```diff
@@ -0,0 +1,50 @@
...
+    Ignore files matching <pattern>.
+    """
+    option_attrs = dict(multiple=True, metavar='<pattern>')
+
+    def __init__(self, user_input):
+        super(ProcessIgnore, self).__init__(user_input)
+
+    def process_resources(self, resources):
+        user_ignore = {patt: 'User ignore: Supplied by --ignore' for patt in self.user_input}
+        ignored = partial(is_ignored, ignores=user_ignore, unignores={})
```
yashdsaraf: No, these are applied after the resources are processed for ignore.ignores_VCS.
Here https://github.com/nexB/scancode-toolkit/blob/8a7b660821dd6cfe7c1c077ccc052225044bb9c2/src/scancode/cli.py#L776
The `resources` are passed to `process_resources` function of a plugin after all the default processing is done

yashdsaraf:
@pombredanne I might've granted `ignore` a bit too much leeway here, but since it is pretty much the only type of use case to applied on files before they're decided for scanning, this may be the way to go IMO as the heavy lifting here is ultimately done by commoncode.ignore and not the plugin itself.

pombredanne:
Thanks++

Issue #688, opened 2017-07-14
UnicodeDecodeError backtrace during scan


dengste:
Using scancode 2.0.1. I get an abort with a UnicodeDecodeError backtrace. I'm guessing this is due to same illegal encoding in some file, but I cannot figure out which file exactly is responsible. The backtrace does not tell me much because I cannot see the original backtrace that triggered this but only the one from the multiprocessing pool:

```
Traceback (most recent call last):
  File "/home/engsterd/scancode-toolkit-2.0.1/bin/scancode", line 11, in <module>
    load_entry_point('scancode-toolkit', 'console_scripts', 'scancode')()
  File "/home/engsterd/scancode-toolkit-2.0.1/local/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/home/engsterd/scancode-toolkit-2.0.1/src/scancode/utils.py", line 76, in main
    standalone_mode=standalone_mode, **extra)
  File "/home/engsterd/scancode-toolkit-2.0.1/local/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/engsterd/scancode-toolkit-2.0.1/local/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/engsterd/scancode-toolkit-2.0.1/local/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/engsterd/scancode-toolkit-2.0.1/local/lib/python2.7/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/engsterd/scancode-toolkit-2.0.1/src/scancode/cli.py", line 402, in scancode
    ignore=user_ignore)
  File "/home/engsterd/scancode-toolkit-2.0.1/src/scancode/cli.py", line 515, in scan
    result = scanned.next()
  File "/home/engsterd/scancode-toolkit-2.0.1/local/lib/python2.7/site-packages/click/_termui_impl.py", line 259, in next
    rv = next(self.iter)
  File "/home/engsterd/scancode-toolkit-2.0.1/src/scancode/pool.py", line 46, in wrap
    return func(self, timeout=timeout or 1e10)
  File "/usr/lib/python2.7/multiprocessing/pool.py", line 668, in next
    raise value
[...]

PR #723, opened 2017-08-15
Support non-utf8/unicode-decodable bytes-only paths on Linux

pombredanne:
This PR is for #688 and provides a way to walk byte-only, non-unicode decodable paths on Linux and keep unicode elsewhere (on Windowss and Mac)

The root cause of all this is explained here: https://github.com/nexB/scancode-toolkit/issues/688#issuecomment-322790224 and is a Python2 bug/inconsistency in os.listdir when getting unicode inputs.

The approach I used is brutal and not super elegant but works for now: 
- top, entry level paths are forced to bytes on Linux and unicode elsewhere
- anywhere a path is needed for OS-level operations, it is forced to bytes on Linux as a safety (using the surrogateescape handler for errors as needed), left as unicode on other OSes
- at the tail end of processing, and whenever output is needed (terminal, file, etc) on Linux the paths are converted back to unicode with decoding using the surrogateescape handler using the pure Python v2 version in https://github.com/PythonCharmers/python-future by @edschofield with code from @haypo

This could be made cleaner with a major refactoring with a Path-like object that mediates all paths and os-level path-related operations using for instance @mcmtroffaes backport of Python 3 pathlib https://github.com/mcmtroffaes/pathlib2 that has a similar behavior (e.g. it provides bytes on Linux irrespective of the input, unicode on unicode input elsewhere)... but we would still need though to handle the surrogateescape ourselves as long as we are on Python2. When on Python 3, the builtins os.walk/os.path/pathlib should work consistently all the way with unicode inputs and handle the surrogateescape encoding/decoding as needed thanks in big parts to @haypo the unicode guru! Though I am sure some adjustments will be needed there too.

Also for reference: http://lucumr.pocoo.org/2013/7/2/the-updated-guide-to-unicode/


pombredanne:
This looks good for me, so I am merging this

PR #737, opened 2017-08-22
Miscellaneous minor formatting and wording improvements

sschuberth:
Please have a look at the individual commit messages for the details.

### DIFF
--- a/etc/conf/base.py
+++ b/etc/conf/base.py
@@ -1,3 +1,7 @@
+"""
+This base config script gets automatically executed for all platforms via
+configure.
+"""
 
 import sys
 
--- a/scancode.bat
+++ b/scancode.bat
@@ -1,34 +1,32 @@
 @echo OFF
-@rem  Copyright (c) 2015 nexB Inc. http://www.nexb.com/ - All rights reserved.
-@rem  
-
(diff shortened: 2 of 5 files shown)


etc/conf/base.py
```diff
@@ -1,3 +1,7 @@
+"""
+This base config script gets automatically executed for all platforms
+via configure if no sub-configuration is specified.
```
pombredanne: Actually I think the behavior is that it gets automatically executed in all cases.

sschuberth: Done.

scancode.bat
```diff
@@ -1,34 +1,31 @@
 @echo OFF
-@rem  Copyright (c) 2015 nexB Inc. http://www.nexb.com/ - All rights reserved.
-@rem  
-
+@rem Copyright (c) 2015 nexB Inc. http://www.nexb.com/ - All rights reserved.
 
 @rem  A minimal shell wrapper to the CLI entry point
 
 set SCANCODE_ROOT_DIR=%~dp0
-set SCANCODE_CMD_LINE_ARGS=
```
pombredanne: I think that's part of the joy of Windows BAT scripting ;)

sschuberth: Very good point, thanks for noticing. Fixed.

pombredanne:
Thanks! LGTM, merging!

Issue #736, opened 2017-08-22
Rework cli.py to use a Resource class instead of iterating paths


yashdsaraf:
For now the goal is to create a `Resource` class which stores paths and infos for a resource.
Eventually this class should hold all the scanned data for a resource and encapsulate the caching logic entirely.

pombredanne:
This has been fixed in #885 !

PR #738, opened 2017-08-23
#736 Add a `Resource` class to hold all scanned info for a resource

yashdsaraf:
Signed-off-by: Yash D. Saraf <yashdsaraf@gmail.com>

### DIFF
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -574,7 +574,7 @@ def scan(input_path,
 
     pool = None
 
-    resources = resource_paths(input_path, diag, pre_scan_plugins=pre_scan_plugins)
+    resources = resource_paths(input_path, diag, scans_cache_class, pre_scan_plugins=pre_scan_plugins)
     paths_with_error = []
     files_count = 0
 
@@ -740,7 +740,6 @@ def _scanit(resource, scanners, scans_cache_class, diag, timeout=DEFAULT_TIMEOUT
     """
     success = True
     scans_cache = scans_cache_class()
-    is_cached = scans_cache.put_info(resource.rel_path, resource.infos)
 
     # note: "flag and function" expressions return the function if flag is True
     # note: the order of the scans matters to show things in logical order
@@ -756,7 +755,7 @@ def _scanit(resource, scanners, scans_cache_class, diag, timeout=DEFAULT_TIMEOUT
     if any(scanner_functions):
         # Skip other scans if already cached
         # FIXME: ENSURE we only do this for files not directories
-        if not is_cached:
+        if not resource.is_cached:
             # run the scan as an interruptiple task
             scans_runner = partial(scan_one, resource.abs_path, scanners, diag)
             success, scan_result = interrupter(scans_runner, timeout=timeout)
@@ -766,7 +765,7 @@ def _scanit(resource, scanners, scans_cache_class, diag, timeout=DEFAULT_TIMEOUT
                 # "scan" key is used for these errors
                 scan_result = {'scan_errors': [scan_result]}
 
-            scans_cache.put_scan(resource.rel_path, resource.infos, scan_result)
+            scans_cache.put_scan(resource.rel_path, resource.get_info(), scan_result)
 
             # do not report success if some other errors happened
             if scan_result.get('scan_errors'):
@@ -791,7 +790,7 @@ def build_ignorer(ignores, unignores):
     return partial(ignore.is_ignored, ignores=ignores, unignores=unignores)
 
 
-def resource_paths(base_path, diag, pre_scan_plugins=()):
+def resource_paths(base_path, diag, scans_cache_class, pre_scan_plugins=()):
     """
     Yield tuples of (absolute path, base_path-relative path) for all the files found
     at base_path (either a directory or file) given an absolute base_path. Only yield
@@ -821,9 +820,9 @@ def resource_paths(base_path, diag, pre_scan_plugins=()):
     resources = fileutils.resource_iter(base_path, ignored=ignorer)
 
     for abs_path in resources:
-        resource = Resource(abs_path, base_is_dir, len_base_path)
+        resource = Resource(scans_cache_class, abs_path, base_is_dir, len_base_path)
         # always fetch infos and cache.
-        resource.infos.update(scan_infos(abs_path, diag=diag))
+        resource.put_info(scan_infos(abs_path, diag=diag))
         yield resource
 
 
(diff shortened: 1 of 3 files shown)


pombredanne:
@yashdsaraf FYI, I am waiting to merge #723 first before looking at this which looks fine otherwise. We will need some rebase as #723 is rather invasive.

pombredanne:
@yashdsaraf I merged #723 in develop, so you PR now has a merge conflict. Do you think you can look into this? Thanks!

src/scancode/cli.py
```diff
@@ -825,11 +819,10 @@ def resource_paths(base_path, pre_scan_plugins=()):
...
 
     for abs_path in resources:
-        posix_path = fileutils.as_posixpath(abs_path)
-        # fix paths: keep the path as relative to the original
-        # base_path. This is always Unicode
-        rel_path = get_relative_path(posix_path, len_base_path, base_is_dir)
-        yield abs_path, rel_path
+        resource = Resource(abs_path, base_is_dir, len_base_path)
+        # always fetch infos and cache.
+        resource.infos.update(scan_infos(abs_path, diag=diag))
```
pombredanne: We would need to cache these to avoid recomputing them a second time if info are needed downstream? Or is this taken care of otherwise?

src/scancode/cli.py
```diff
@@ -725,26 +725,20 @@ def _resource_logger(logfile_fd, resources):
...
-    abs_path, rel_path = paths
-    # always fetch infos and cache.
-    infos = OrderedDict()
-    infos['path'] = rel_path
-    infos.update(scan_infos(abs_path, diag=diag))
-
     success = True
     scans_cache = scans_cache_class()
-    is_cached = scans_cache.put_info(rel_path, infos)
+    is_cached = scans_cache.put_info(resource.rel_path, resource.infos)
```
pombredanne: Here we put in the cache, but I think that at this stage, the `info` has already been fetched when iterating Resource before but was not cached then? IMHO we likely need to properly define this:
- when we first collect the `info` they should be cached then
- when we work on a file, check if has info or results cached and use these from the cache

I reckon that I am not 100% clear on what the current step-by-step process is!

yashdsaraf: @pombredanne  a couple of questions,
 - Can a whole Resource object be cached?
 - Does a `Resource` object need to be cached? I mean can't we pass around the object as is and all the scan info can be accessed from it directly?

pombredanne: One approach could be indeed to have the Resource object being something that is "cached" behind the scenes, or at least hides the caching logic
The caching logic seems simple but this is not entirely trivial to get something that is fast (and actually helping getting things to be faster) and required several iterations to get it right and working nicely on all OSes. For instance, an earlier version used a SQLite-backed cache and was a catastrophic choke point.

A Resource object (or at least its underlying data) needs to be cached if its attributes were computed from the actual file (e.g. checksums, attributes) otherwise these would need to be recomputed each time. For instance in this new approach, the "info" are collected as part of the resource iteration for the pre-scan plugins. This is done for now in a single process... AND once the pre-scan plugins have been processes in the stream, the actual scans are enqueued and handled asynchronously, sending their results back in the cache. 
The caching proper is keyed with the file "info" 

Resource cannot be all kept in memory as this would blow up size-wise (and is the main reason for the existence of a cache in the first place).

So all this is not 100% straightforward and I think we would best benefit from documenting the steps in the process and can then refactor this as needed.

yashdsaraf: Ok, what I get from that is:
We need to add the caching logic for the redundant data such as checksums, attributes, etc. in the `Resource` object itself, thus when the `Resource` object's methods are called, they lookup the cache and return the data from there or recompute the data and store it in cache if not found. Am I getting this right?

yashdsaraf: @pombredanne ping?

pombredanne: @yashdsaraf you wrote
> We need to add the caching logic for the redundant data such as checksums, attributes, etc. in the Resource object itself, thus when the Resource object's methods are called, they lookup the cache and return the data from there or recompute the data and store it in cache if not found. Am I getting this right?

That sounds like a decent approach.

out.json
```diff
@@ -0,0 +1 @@
...7ed60cd20295e64a8","files_count":null,"mime_type":"text/x-c","file_type":"C source, ASCII text","programming_language":"C","is_binary":false,"is_text":true,"is_archive":false,"is_media":false,"is_source":true,"is_script":false,"scan_errors":[]},{"path":"samples/zlib/iostream2/zstream.h","type":"file","name":"zstream.h","base_name":"zstream","extension":".h","date":"2017-05-25","size":9283,"sha1":"fca4540d490fff36bb90fd801cf9cd8fc695bb17","md5":"a980b61c1e8be68d5cdb1236ba6b43e7","files_count":null,"mime_type":"text/x-c++","file_type":"C++ source, ASCII text","programming_language":"C","is_binary":false,"is_text":true,"is_archive":false,"is_media":false,"is_source":true,"is_script":false,"scan_errors":[]},{"path":"samples/zlib/iostream2/zstream_test.cpp","type":"file","name":"zstream_test.cpp","base_name":"zstream_test","extension":".cpp","date":"2017-05-25","size":711,"sha1":"e18a6d55cbbd8b832f8d795530553467e5c74fcf","md5":"d32476bde4e6d5f889092fdff6f8cdb0","files_count":null,"mime_type":"text/x-c","file_type":"C source, ASCII text","programming_language":"C++","is_binary":false,"is_text":true,"is_archive":false,"is_media":false,"is_source":true,"is_script":false,"scan_errors":[]}]}
```
pombredanne: May be you did not mean to commit this file?

src/scancode/cli.py
```diff
@@ -795,7 +788,7 @@ def build_ignorer(ignores, unignores):
     return partial(ignore.is_ignored, ignores=ignores, unignores=unignores)
 
 
-def resource_paths(base_path, pre_scan_plugins=()):
+def resource_paths(base_path, diag, scans_cache_class, pre_scan_plugins=()):
```
pombredanne: You are now yielding Resource and the docstring needs to be updated I guess?

tests/scancode/test_ignore_files.py
```diff
@@ -29,13 +29,15 @@
...
 from commoncode.ignore import is_ignored
+from scancode.cache import get_scans_cache_class
 from scancode.cli import resource_paths
 from scancode.plugin_ignore import ProcessIgnore
 
 
 class TestIgnoreFiles(FileBasedTesting):
 
     test_data_dir = path.join(path.dirname(__file__), 'data')
+    scan_cache_class = get_scans_cache_class()
```
pombredanne: You might want to create a "mock" cache in a test dir to avoid polluting the main cache?

pombredanne:
@yashdsaraf Thanks for all the changes. I will make a few checks to ensure we do not keep all things in memory: I wonder if we could do such an automated test btw....

And then I will merge.

Issue #706, opened 2017-07-28
Windows Installation Error


scottctr:
I've tried installing 2.0.1 on Windows 10 twice -- ensuring I followed the directions. I've gotten the same error both times when running ScanCode the first time:

```
....snip....
  File "c:\scancode-toolkit-2.0.1\src\typecode\magic2.py", line 201, in load_lib
    new_path = os.pathsep.join([lib_dir, os.environ['PATH']])
UnicodeDecodeError: 'ascii' codec can't decode byte 0xf6 in position 1447: ordinal not in range(128)

Failed to execute command:
C:\scancode-toolkit-2.0.1\bin\python.exe "C:\scancode-toolkit-2.0.1\etc\conf\base.py". Aborting...
```

pombredanne:
@scottctr thank you for this report.
Can you tell which exact command you used for scanning?
And what were you scanning? is this public code?

scottctr:
This actually occurred right after decompressing the latest ScanCode files into a new directory and using `scancode -h`. Note that I've previously been using 2.0.0 rc2 and never had a problem.

pombredanne:
@scottctr Thanks. This is surprising... but there have been some changes in handling unicode paths... Anything special about your environment? is this is an English installation of Windows 10 or a some other language? 
Is there something special in your `PATH` variable? Would it contain paths with special, non-Ascii like characters may be? I am trying to find a way to reproduce this correctly. 
For instance what shows up if you run this command:
```
c:\python27\python.exe -c "import os; print repr(os.environ['PATH'])"
```

scottctr:
I should also add that I'm using the scancode-toolkit-2.0.1.zip download and the Windows 10 decompress tool. I'll try the scancode-toolkit-2.0.1.tar.bz2 and 7zip shortly.

pombredanne:
@scottctr using the windows built-in zip should be fine normally I doubt this would make a difference to use another archive type and unarchiver. The reason for the failure is more likely in the `PATH` and some bug on our side about dealing with this

scottctr:
This is a standard US/English Windows 10 machine, but I think you're onto something with the path. Notice the entry in the path is `C:\\Program Files (x86)\\V\xf6r Security\\DevAudit\\`, but should be `C:\\Program Files (x86)\\Vör Security\\DevAudit\\`. I'll test this shortly.

```
[...]

PR #741, opened 2017-08-29
Path environment var with unicode

pombredanne:
This is a fix for #706 
This PR comes on top the branch in #723 that adds support for proper FS encoding/decoding for weird non-utf byte paths.


pombredanne:
Merging.

Issue #740, opened 2017-08-28
TypeError: invalid length: 8 when scanning PDF from fop project


sschuberth:
Using ScanCode version 2.0.1.post135.94d8aaa8, I get

    ERROR: licenses: Traceback (most recent call last):
      File "/workspace/tools/scancode/src/scancode/cli.py", line 730, in scan_one
        scan_details = list(scan_details)
      File "/workspace/tools/scancode/src/scancode/api.py", line 121, in get_licenses
        for match in idx.match(location=location, min_score=min_score):
      File "/workspace/tools/scancode/src/licensedcode/index.py", line 437, in match
        qry = query.build_query(location, query_string, self)
      File "/workspace/tools/scancode/src/licensedcode/query.py", line 107, in build_query
        if not T.contains_text:
      File "/workspace/tools/scancode/src/typecode/contenttype.py", line 429, in contains_text
        elif self.is_pdf and not self.is_pdf_with_text:
      File "/workspace/tools/scancode/src/typecode/contenttype.py", line 413, in is_pdf_with_text
        doc = PDFDocument(parser)
      File "/workspace/tools/scancode/local/lib/python2.7/site-packages/pdfminer/pdfdocument.py", line 328, in __init__
        self.info.append(dict_value(trailer['Info']))
      File "/workspace/tools/scancode/local/lib/python2.7/site-packages/pdfminer/pdftypes.py", line 155, in dict_value
        x = resolve1(x)
      File "/workspace/tools/scancode/local/lib/python2.7/site-packages/pdfminer/pdftypes.py", line 75, in resolve1
        x = x.resolve(default=default)
      File "/workspace/tools/scancode/local/lib/python2.7/site-packages/pdfminer/pdftypes.py", line 62, in resolve
        return self.doc.getobj(self.objid)
      File "/workspace/tools/scancode/local/lib/python2.7/site-packages/pdfminer/pdfdocument.py", line 462, in getobj
        (strmid, index, genno) = xref.get_pos(objid)
      File "/workspace/tools/scancode/local/lib/python2.7/site-packages/pdfminer/pdfdocument.py", line 261, in get_pos
        f2 = nunpack(ent[self.fl1:self.fl1+self.fl2])
      File "/workspace/tools/scancode/local/lib/python2.7/site-packages/pdfminer/utils.py", line 175, in nunpack
        raise TypeError('invalid length: %d' % l)
    TypeError: invalid length: 8

[...]

PR #742, opened 2017-08-29
Alternate pdf handling

pombredanne:
This is a fix for #740 
The fix consists of

1. switching to pdfminer.six maintained by @goulu which is actually a maintained version of pdfminder that upstream seems to have abandoned otherwise.... Thanks to  @goulu and team !

2. use io.StringIO instead of plain StringIO to better cope with unicode issues


pombredanne:
@sschuberth yes! good catch

sschuberth:
@pombredanne, any objections to merge this?

pombredanne:
@sschuberth no objections, I just need to add the pre-built native wheels for pycyptodome on Linux and macOS.
I kicked a build at  https://github.com/pombreda/thirdparty-manylinux/ and https://github.com/pombreda/thirdparty

These should show up shortly at http://dl.bintray.com/pombreda/thirdparty/

pombredanne:
All set... Merging! thanks again for the report

Issue #735, opened 2017-08-22
ImportError: No module named pluggy


sschuberth:
Using ScanCode version 2.0.1.post135.94d8aaa8 I'm running into `ImportError: No module named pluggy` errors on CI, but not locally. Still investigating why. Stacktrace:

    Traceback (most recent call last):
    File "/workspace/tools/scancode/bin/scancode", line 11, in <module>
      load_entry_point('scancode-toolkit', 'console_scripts', 'scancode')()
    File "/workspace/tools/scancode/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 561, in load_entry_point
      return get_distribution(dist).load_entry_point(group, name)
    File "/workspace/tools/scancode/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2649, in load_entry_point
      return ep.load()
    File "/workspace/tools/scancode/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2303, in load
      return self.resolve()
    File "/workspace/tools/scancode/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2309, in resolve
      module = __import__(self.module_name, fromlist=['__name__'], level=0)
    File "/workspace/tools/scancode/src/scancode/cli.py", line 52, in <module>
      import plugincode.output
    File "/workspace/tools/scancode/src/plugincode/output.py", line 32, in <module>
      from pluggy import HookimplMarker

sschuberth:
I don't see

    Configuring ScanCode for first use...

in the CI logs. We're running ScanCode from a Git checkout that is being fetched to incrementally. Can it be that the "first use" check does not properly work if new dependencies like `pluggy` are being introduced?

sschuberth:
Interestingly, I also don't see the `SCANCODE_DEV_MODE` file being created. Shouldn't that be the case when running from a Git checkout?

pombredanne:
@sschuberth is this a fresh checkout?

pombredanne:
@sschuberth if this is a incremental checkout and this was already "configured" or run before, then you should run `./configure` on this after each incremental pull

sschuberth:
> is this a fresh checkout?

No, it's not.

> if this is a incremental checkout and this was already "configured" or run before, then you should run `./configure` on this after each incremental pull

Ah, then this is a misunderstanding on my side. I though the logic would be smart enough to know for which commit `./configure` was run in case of running from a checkout.

[...]

PR #744, opened 2017-08-30
Improvements to the configuration logic

sschuberth:
Most notably, automatically re-configure if ScanCode is run from a Git working tree and new commits are present. This is useful for running on CI which incrementally fetches ScanCode's code from Git (without applying any local changes).

Note that the new `SCANCODE_GIT_MODE` is orthogonal to the existing `SCANCODE_DEV_MODE`: In development mode the license cache is always checked for consistency, which is not needed when running from a clean Git working tree without any local changes.

This fixes issues like the one I've been running into as described at #735.

### DIFF
--- a/scancode
+++ b/scancode
@@ -105,10 +105,33 @@ _canonicalize_file_path() {
 SCANCODE_BIN="$( realpath "${BASH_SOURCE[0]}" )"
 SCANCODE_ROOT_DIR="$( cd "$( dirname "${SCANCODE_BIN}" )" && pwd )"
 
+SCANCODE_CONFIG_DIR="etc/conf"
+
+GIT_STATUS=$(git -C "$SCANCODE_ROOT_DIR" status --porcelain 2> /dev/null)
+if [ $? -eq 0 ]; then
+    # This is a Git working tree, check whether it is clean.
+    if [ -z "$GIT_STATUS" ]; then
+        # This is a clean working tree, check whether configure has been run
+        # before for this commit.
+        GIT_COMMIT_CONFIGURED=$(cat "$SCANCODE_ROOT_DIR/SCANCODE_GIT_MODE" 2> /dev/null)
+        GIT_COMMIT_HEAD=$(git -C "$SCANCODE_ROOT_DIR" rev-parse HEAD)
+        if [ "$GIT_COMMIT_CONFIGURED" != "$GIT_COMMIT_HEAD" ]; then
+            echo "* (Re-)Configuring ScanCode because Git commits have changed..."
+            CONFIGURE_QUIET=1 $SCANCODE_ROOT_DIR/configure $SCANCODE_CONFIG_DIR
+        fi
+        echo $GIT_COMMIT_HEAD > "$SCANCODE_ROOT_DIR/SCANCODE_GIT_MODE"
+    else
+        echo "* (Re-)Configuring ScanCode because Git working tree is not clean..."
+        CONFIGURE_QUIET=1 $SCANCODE_ROOT_DIR/configure $SCANCODE_CONFIG_DIR
+    fi
+else
+    rm -f "$SCANCODE_ROOT_DIR/SCANCODE_GIT_MODE"
+fi
+
 SCANCODE_CONFIGURED_PYTHON=$SCANCODE_ROOT_DIR/bin/python
 if [ ! -f "$SCANCODE_CONFIGURED_PYTHON" ]; then
     echo "* Configuring ScanCode for first use..."
-    CONFIGURE_QUIET=1 $SCANCODE_ROOT_DIR/configure etc/conf
+    CONFIGURE_QUIET=1 $SCANCODE_ROOT_DIR/configure $SCANCODE_CONFIG_DIR
 fi
 
 $SCANCODE_ROOT_DIR/bin/scancode "$@"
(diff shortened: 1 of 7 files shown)


scancode
```diff
@@ -105,10 +105,33 @@ _canonicalize_file_path() {
 SCANCODE_BIN="$( realpath "${BASH_SOURCE[0]}" )"
 SCANCODE_ROOT_DIR="$( cd "$( dirname "${SCANCODE_BIN}" )" && pwd )"
 
+SCANCODE_CONFIG_DIR="etc/conf"
+
+GIT_STATUS=$(git -C "$SCANCODE_ROOT_DIR" status --porcelain 2> /dev/null)
```
pombredanne: Does this also implicitly checks for Git being available locally? Just curious what the different with a plain `git status`?

scancode
```diff
@@ -105,10 +105,33 @@ _canonicalize_file_path() {
...
+if [ $? -eq 0 ]; then
+    # This is a Git working tree, check whether it is clean.
+    if [ -z "$GIT_STATUS" ]; then
+        # This is a clean working tree, check whether configure has been run
+        # before for this commit.
+        GIT_COMMIT_CONFIGURED=$(cat "$SCANCODE_ROOT_DIR/SCANCODE_GIT_MODE" 2> /dev/null)
+        GIT_COMMIT_HEAD=$(git -C "$SCANCODE_ROOT_DIR" rev-parse HEAD)
+        if [ "$GIT_COMMIT_CONFIGURED" != "$GIT_COMMIT_HEAD" ]; then
+            echo "* (Re-)Configuring ScanCode because Git commits have changed..."
+            CONFIGURE_QUIET=1 $SCANCODE_ROOT_DIR/configure $SCANCODE_CONFIG_DIR
```
pombredanne: IMHO to be really safe, you should first issue a `configure --clean` but let see first if this is enough.
There could be cases where files have been removed and there could be dangling .pyc files left over in some rare cases

sschuberth:
> Do you think this is worth having for Windows too?

I forgot to mention that I wanted to do this for Linux first to align on the logic, and they (maybe) later do it for Windows, too. But as CI is mostly running on Linux, I don't have any immediate plans to implement this for Windows.

scancode
sschuberth: Exactly. It implicitly checks for `$SCANCODE_ROOT_DIR` to be a Git working tree (`git status` returns code 128 otherwise), and also for the `git` executable to be present (in this case it's Bash that returns code 127).

scancode
sschuberth: FWIW, I tried to avoid a full clean by doing e.g. 773bffa.

pombredanne:
LGTM. Thanks!

pombredanne:
Somehow there a few things I am not entirely getting on the overall behavior here. I will dive in it a bit more and eventually submit a new PR/ticket about this

sschuberth:
Like, did it break something? Or is it "only" a matter of understanding how it works?

pombredanne:
@sschuberth I guess I am not sure how this works and interacts with the SCANCODE_DEV_MODE :P

sschuberth:
I was hoping to make that clear in the second paragraph of the top post in this PR :-)

pombredanne:
fair enough: let me reread all this first

pombredanne:
@sschuberth here are the issues that I can see:

If I have a git clone and I ran  plain  `./configure` explicitly, this will use the `etc/conf/dev` and create the `SCANCODE_DEV_MODE` tag file, which is then used by the license cache check code to force a recheck of cache consistency wrt. the current code and license rules/files.

Now with the changes  in this PR, a run of `./scancode` with a git tree that is not clean will force a new `./configure` run using `etc/conf` as a conf instead which as a side effect wipes the `SCANCODE_DEV_MODE` tag file, creates a `SCANCODE_GIT_MODE` and  reconfigures and then re-index licenses even though I did not want any reconfiguration to take place in this case.

It furthermore echoes messages such as `* (Re-)Configuring ScanCode because Git ....` which should not be echoed at all IMHO.

There are other possible issues: doing a ./configure after a pull may not be enough and in many cases a ./configure --clean is needed: for instance say a python .py file was renamed: without a clean there may be dangling and orphaned .pyc files left over from before the rename of the corresponding .py file and this can create weird issues 

I will come up with fixes for this

sschuberth:
> It furthermore echoes messages such as `* (Re-)Configuring ScanCode because Git ....` which should not be echoed at all IMHO.

Just a quick reply to this one: I disagree here. We already have [echo "* Configuring ScanCode for first use..."](https://github.com/heremaps/scancode-toolkit/blob/c95017a319b8b7ebe87c4a777bb425929a89ac39/scancode#L133), so I believe if we reconfigure not necessarily because it's the first use, but because there are Git changes, we should say so, too.

pombredanne:
@sschuberth 

> Just a quick reply to this one: I disagree here.

Fair point!

So let me come with a new ticket with a clear statement of what we want , can/could/should do here

pombredanne:
When I say here, I meant what to do wrt. to running and configuring scancode from different user type standpoints and what should be automated and what should not

pombredanne:
@sschuberth since merging is creating some confusion based on @armijnhemel feedback ... may be we could revert these changes in a new PR and apply them anew with some updates?

sschuberth:
As you prefer, I'm fine with that!

pombredanne:
@sschuberth I reverted the #744 PR commit in develop and pushed the original branch of this PR as https://github.com/nexB/scancode-toolkit/tree/735-769-git-reconfigure

Issue #750, opened 2017-09-03
Sort out classpath-related licenses issues


pombredanne:
We have a bit of a mess to sort out for the classpath exception-related licenses.

Historically we have these:

 - gpl-2.0-library: this is a notice with a GPL (version unspecificed)
    and most of the classpath-exception text except for the first paragraph.
    This should not be a license but a rule instead

 - gpl-2.0-independent-module-linking: this is a notice of gpl-2.0-plus with
    classpath-exception-2.0. This should not be a license but a rule instead
    This was used in https://github.com/icsharpcode/SharpZipLib which has since switched to MIT

 - gpl-2.0-oracle-openjdk:  this is a notice of gpl-2.0-plus with
    classpath-exception-2.0. This should not be a license but a rule instead

 - gpl-2.0-openjdk: this is a notice of gpl-2.0-plus with
    classpath-exception-2.0 and the assembly exception
    The combo of classpath with assembly exception is rather unique. I think we
    should create a new exception for the combo "classpath-exception-2.0 + 
    the assembly exception and call this the openjdk-exception

 - gpl-2.0-classpath: this is IMHO to be replaced by classpath-exception-2.0. 
    In particular the calsspath exception is not specific to a version of the GPL

Recently, SPDX added this:
 - classpath-exception-2.0

 
 So after rework we would end up with:
 1. several new rules and tests
 2. classpath-exception-2.0 as a license exception
 3. openjdk-exception as a new license exception that contains the classpath exception and the assembly exception texts http://openjdk.java.net/legal/assembly-exception.html

pombredanne:
Here is the todo: 

 - [ ] gpl-2.0-library: deprecated and transformed as rules for `gpl-2.0 with classpath-exception-2.0`
 - [ ] gpl-2.0-independent-module-linking:  deprecated and transformed as rules for `gpl-2.0 with classpath-exception-2.0`
 - [ ] gpl-2.0-oracle-openjdk:  deprecated and transformed as rules for `gpl-2.0-plus with classpath-exception-2.0`
 - [ ] gpl-2.0-openjdk: deprecated and transformed as rules for `gpl-2.0-plus with openjdk-exception` 
 - [ ] gpl-2.0-classpath: deprecated and transformed as rules for  `classpath-exception-2.0`. 
 - [ ] openjdk-exception: new creation

pombredanne:
This has long been addressed. Closing

PR #753, opened 2017-09-08
Classpath licenses improvements

pombredanne:
See #750 for details of these changes.

@waynebeaton this changes were made directly in relation to the discussions around your submission of the EPL 2.0 to SPDX.

The net effect is that we have now:
1. the classpath-exception-2.0 as before
2. a new openjdk-exception exception, which contains both the classpath-exception text AND the openjdk assembly exception. If SPDX goes this route too in the future, then there would be no need to support a 'with A with B' syntax or a `with A and B` syntax. My rationale here is that the assembly exception is quite unique and intricately tied to the classpath exception.

Fedback welcomed!


pombredanne:
Merging this: all tests pass.

Issue #776, opened 2017-09-28
Does scanning of PHP composer work properly?


scandinaf710:
Hi all,

I am using scancode 2.1.0 and according to the release notes scanning PHP composer is already there. However, even though scancode finds _composer.json_, I don't see any analysis of dependencies section even when I put only this file as input.

How to run this package check correctly?

pombredanne:
@scandinaf710 good catch. The package parser is there but somehow not active. I pushed a branch and PR to fix this

pombredanne:
@scandinaf710 A fix is waiting for you in the lucky #777 PR !
You can test it by downloading this https://github.com/nexB/scancode-toolkit/archive/776-composer-package-detection.tar.gz 
Thanks again for fining this.

scandinaf710:
@pombredanne Well, I downloaded tarball and ran this version of scancode asking to analyze just one single file composer.json with the following "require" section:

> "require": {
        "php": "7.1.*",
        "symfony/symfony": "3.2.*",
        "sensio/distribution-bundle": "^5.0",
        "doctrine/orm": "^2.5",
        "doctrine/doctrine-bundle": "^1.6",
        "doctrine/doctrine-cache-bundle": "^1.2",
        "symfony/swiftmailer-bundle": "^2.3.10",
        "symfony/monolog-bundle": "^3.0.2",
        "symfony/polyfill-apcu": "^1.0",
        "sensio/framework-extra-bundle": "^3.0.2",
        "incenteev/composer-parameter-handler": "^2.0",
        "twig/twig": "^1.0||^2.0",
        "doctrine/doctrine-migrations-bundle": "^1.0",
        "greenlion/php-sql-parser": "^4.1",
        "eightpoints/guzzle-bundle": "^5.3",
        "jms/serializer-bundle": "^1.1",
        "friendsofsymfony/rest-bundle": "^2.1",
        "nelmio/api-doc-bundle": "^2.13",
        "snc/redis-bundle": "^2.0",
        "predis/predis": "^1.0",
        "paragonie/halite": "^3.2",
        "knplabs/knp-paginator-bundle": "^2.5",
        "egeloen/base64-file-bundle": "^1.0",
        "liip/imagine-bundle": "^1.7",
        "sonata-project/admin-bundle": "^3.17",
        "sonata-project/doctrine-orm-admin-bundle": "^3.1",
        "doctrine/doctrine-fixtures-bundle": "^2.3",
[...]

PR #777, opened 2017-09-28
Ensure that PHP Composer packages are recognized

pombredanne:
This is a fix for #776 
@scandinaf710 if you can test this this would be great!


pombredanne:
@scandinaf710 the reason why your package was not detected at all was because a package without a description was skipped (based on https://getcomposer.org/doc/04-schema.md#description ) as it could not be published... but this is still a valid package. This is fixed now. I am merging this in develop where you can test it.

PR #792, opened 2017-10-06
[pre-scan] Rename process_resources() and use it.

### LINKED ISSUE #689: How to ignore binary files ?
I've been thinking about how to speed up scans of packages, which contain big binary files.
I [can](https://github.com/fabric8-analytics/fabric8-analytics-worker/blob/master/f8a_worker/workers/license.py#L20) pass patterns to `--ignore` if I know the extensions I want to ignore, but that's tedious, because there's so many of them.
For example [pypi/alfanous](https://pypi.python.org/pypi/alfanous) contains some index with huge binary files with extensions like `dcz`, `pst`, `tiz`, etc.

Also when scancode reaches timeout while scanning such file it returns non-zero exit code, which is not what I want.

So I've been thinking about something like `--ignore-binaries`, which would tell scancode to scan only text files. But I'm not sure if we can reliable detect binary/text files with Python, see.
https://stackoverflow.com/questions/898669/how-can-i-detect-if-a-file-is-binary-non-text-in-python

Ideas ?

jpopelka:
I've been looking at how the pre-scan plugins work in order to fix #689.

I've realized that the [process_resources()](https://github.com/nexB/scancode-toolkit/blob/develop/src/plugincode/pre_scan.py#L52) is actually never used,
so I'm adding a code which runs it.

The pre-scan plugin which does what I need would then look like (it's not part of this PR):

```python
@pre_scan_impl
class ProcessIgnoreBinaries(PreScanPlugin):
    """
    Ignore binaries, medias and files bigger than specified size.
    """
    option_attrs = dict(metavar='<size>', default=1000000)

    def __init__(self, user_input):
        super(ProcessIgnoreBinaries, self).__init__(user_input)

    def process_resource(self, resource):
        if any((resource.infos['is_binary'],
                resource.infos['is_media'],
                resource.infos['size'] > self.user_input)):
            return None
        return resource
```

### DIFF
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -536,7 +536,7 @@ def scan(input_path,
          scans_cache_class=None,
          strip_root=False,
          full_root=False,
-         pre_scan_plugins=()):
+         pre_scan_plugins=None):
     """
     Return a tuple of (files_count, scan_results, success) where
     scan_results is an iterable and success is a boolean.
@@ -794,7 +794,7 @@ def build_ignorer(ignores, unignores):
     return partial(ignore.is_ignored, ignores=ignores, unignores=unignores)
 
 
-def resource_paths(base_path, diag, scans_cache_class, pre_scan_plugins=()):
+def resource_paths(base_path, diag, scans_cache_class, pre_scan_plugins=None):
     """
     Yield `Resource` objects for all the files found at base_path
     (either a directory or file) given an absolute base_path. Only yield
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -827,7 +827,11 @@ def resource_paths(base_path, diag, scans_cache_class, pre_scan_plugins=()):
         resource = Resource(scans_cache_class, abs_path, base_is_dir, len_base_path)
         # always fetch infos and cache.
         resource.put_info(scan_infos(abs_path, diag=diag))
-        yield resource
+        if pre_scan_plugins:
+            for plugin in pre_scan_plugins:
+                resource = plugin.process_resource(resource)
+        if resource:
+            yield resource
 
 
 def scan_infos(input_file, diag=False):
--- a/src/plugincode/pre_scan.py
+++ b/src/plugincode/pre_scan.py
@@ -49,12 +49,13 @@ class PreScanPlugin(object):
     def __init__(self, user_input):
         self.user_input = user_input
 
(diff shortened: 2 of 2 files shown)


pombredanne:
@jpopelka good catch! :blush:  let me review and possibly add a test!

src/scancode/cli.py
```diff
@@ -536,7 +536,7 @@ def scan(input_path,
          scans_cache_class=None,
          strip_root=False,
          full_root=False,
-         pre_scan_plugins=()):
+         pre_scan_plugins=None):
```
yashdsaraf: Giving an empty tuple as a default parameter gives the reader an idea of what type of data to expect. Why did you change it?

src/scancode/cli.py
```diff
@@ -794,7 +794,7 @@ def build_ignorer(ignores, unignores):
     return partial(ignore.is_ignored, ignores=ignores, unignores=unignores)
 
 
-def resource_paths(base_path, diag, scans_cache_class, pre_scan_plugins=()):
+def resource_paths(base_path, diag, scans_cache_class, pre_scan_plugins=None):
```
yashdsaraf: See my previous comment

src/scancode/cli.py
jpopelka: Well, it gives a wrong idea to me since the `pre_scan_plugins` is actually a list.
I should have explain it in the commit message, sorry.

yashdsaraf: It's not a good idea to pass mutable objects as default parameters. Using a tuple gives an idea that a collection is passed here. :)
BTW I didn't know this myself till @pombredanne posted a similar comment in an earlier review.
I'll try to find that for you. :P

jpopelka: I know that. That's why I changed it to `None` and not `[]`.

pombredanne: I might prefer using an empty tuple as an hint that the expected value is an iterable ofsorts, but that's a weak indication for sure... @jpopelka your way works too.

pombredanne:
I am merging this as is... we can add tests later!
@jpopelka Thanks ++

src/scancode/cli.py
pombredanne: @jpopelka to your point, confusing is not a good thing, hence why I went ahead and merged it as-is :) Thanks again!

PR #810, opened 2017-10-20
Add new and improved licensesand rules

pombredanne:
This is a set of new minor rules and licenses collected left and right.


pombredanne:
LGTM. merging

PR #823, opened 2017-10-27
Streamline package manifests models

pombredanne:
This is a major simplification of the Package models for #275
This breaks any API that existed for packages.


pombredanne:
If there is no objection, I might merge this very soon. This is a major change in the data API, this will warrant likely a bump to a V3 OR, we could adopt separate versioning at the scanner/plugin level too.... though that feels overkill at this stage

pombredanne:
there has been no objection to this, so I am merging now.

PR #848, opened 2017-11-18
Include scan errors in the result when invoked with --only-findings

haikoschol:
Signed-off-by: Haiko Schol <ext-haiko.schol@here.com>

### DIFF
--- a/tests/scancode/test_has_findings.py
+++ b/tests/scancode/test_has_findings.py
@@ -25,32 +25,25 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
-from plugincode.post_scan import post_scan_impl
+from unittest import TestCase
+from scancode.plugin_only_findings import has_findings
 
 
-@post_scan_impl
-def process_only_findings(active_scans, results):
-    """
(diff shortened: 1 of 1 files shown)


src/scancode/plugin_only_findings.py
```diff
@@ -49,6 +49,8 @@ def process_only_findings(active_scans, results):
 def has_findings(active_scans, scanned_file):
     """
     Return True if the `scanned_file` has findings for any of the
-    `active_scans` names list (excluding basic file information).
+    `active_scans` names list (excluding basic file information)
+    or any errors occured when scanning the file.
     """
-    return any(scanned_file.get(scan_name) for scan_name in active_scans)
+    findings = active_scans + ['scan_errors']
```
pombredanne: That's a sleek and simple way to include this, even though this is not a "scan" :+1:

haikoschol: Thanks. First I looked into adding it to `active_scans`, but it doesn't really belong there and I did not want to make changes that affect other plugins.

PR #854, opened 2017-11-23
Deal with install path containing spaces

ajeans:
So I am trying to run this on Cygwin with an install path containing spaces.
I understand that Cygwin is not supported (as this fails inside python with a clear message).
However, there is no harm in being more robust in the bash scripts

Signed-off-by: Arnaud Jeansen <arnaud.jeansen@gmail.com>

### DIFF
--- a/configure
+++ b/configure
@@ -23,7 +23,7 @@ fi
 
 CONFIGURE_ROOT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-python2.7 $CONFIGURE_ROOT_DIR/etc/configure.py $CFG_CMD_LINE_ARGS
+python2.7 "$CONFIGURE_ROOT_DIR/etc/configure.py" $CFG_CMD_LINE_ARGS
 if [ -f "$CONFIGURE_ROOT_DIR/bin/activate" ]; then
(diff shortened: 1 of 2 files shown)


pombredanne:
Thank you ++!
Do you mind adding a `Signed-off-by:`  to your commit?
https://github.com/nexB/scancode-toolkit/blob/develop/CONTRIBUTING.rst !

pombredanne:
Or at least to the PR message above, so I can make this part of the merge commit.

ajeans:
Whoops, I didn't read CONTRIBUTING.rst and managed to miss almost all the rules, sorry about that.
I added my Sign-Off to the first comment, hopefully this is enough. I can rebase the branch and re-add it but that would be tonight at the earliest.

Nice project organization btw.

pombredanne:
Thanks! and also for the kind words. Looking good: merging.

About cygwin, there have been unpublished versions that I tried on cygwin.... but the semantics of cygwin being what they are this is half between POSIX and Windows and was a major PITA to make working with one codebase for Win/Lin/Mac.... so I gave up long ago. We received some reports that scancode runs in Windows 10 Linux subsystem FWIW, but I am not too inclined to add support for this.

PR #855, opened 2017-11-23
Refactor post scan plugins so they can have command line options

haikoschol:
Signed-off-by: Haiko Schol <ext-haiko.schol@here.com>


haikoschol:
@yashdsaraf Is this what you had in mind for post-scan plugins?

Passing `active_scans` is a hack. Ideally I think plugins should receive all command line options, so they can check that the scan included whatever feature they need to function.

src/plugincode/post_scan.py
```diff
@@ -38,14 +38,30 @@
...
+    def process_results(self, results, active_scans):
+        """
+        Process the scan results.
+        results - an iterable of resources
+        active_scans - iterable of scanners that were used to obtain the results (e.g. "copyrights", "licenses")
+        """
+        return results
+
+    @staticmethod
+    def get_click_options():
```
pombredanne: calling this just `get_options` might be enough IMHO

src/plugincode/post_scan.py
```diff
@@ -38,14 +38,30 @@
...
     """
-    pass
+
+    def __init__(self, option, user_input):
+        self.option = option
+        self.user_input = user_input
+
+    def process_results(self, results, active_scans):
+        """
+        Process the scan results.
```
pombredanne: May I would say:
Return an iterable of results (eventually transformed or filtered) based on the `results` iterable.

src/plugincode/post_scan.py
```diff
@@ -57,6 +73,9 @@ def initialize():
     NOTE: this defines the entry points for use in setup.py
     """
     post_scan_plugins.load_setuptools_entrypoints('scancode_post_scan')
+    for name, plugin in get_post_scan_plugins().items():
+        if not issubclass(plugin, PostScanPlugin):
+            raise Exception('Invalid post-scan plugin "%(name)s": does not extend "plugincode.post_scan.PostScanPlugin".' % locals())
```
pombredanne: I guess this is OK for now, though eventually we may not care that a plugin extends a certain class, ass long as it quacks like one.

src/scancode/plugin_mark_source.py
```diff
@@ -28,43 +28,51 @@
...
+                yield scanned_file
+            return
 
-    if not has_file_info:
-        # just yield results untouched
+        # FIXME: this is an nested loop, looping twice on results
+        # TODO: this may not recusrively roll up the is_source flag, as we
+        # may not iterate bottom up.
         for scanned_file in results:
+            if scanned_file['type'] == 'directory' and scanned_file['files_count'] > 0:
```
pombredanne: This makes me think that we should invert this test and `continue` rather than having the deeper nested if... But you just carried over the code. So nothing to change here for  now.

src/scancode/plugin_only_findings.py
```diff
@@ -25,25 +25,33 @@
...
-    # FIXME: this is forcing all the scan results to be loaded in memory
-    # and defeats lazy loading from cache. Only a different caching
-    # (e.g. DB) could work here.
-    # FIXME: We should instead use a generator or use a filter function
-    # that pass to the scan results loader iterator
-    for scanned_file in results:
-        if has_findings(active_scans, scanned_file):
-            yield scanned_file
+    def process_results(self, results, active_scans):
+        # FIXME: this is forcing all the scan results to be loaded in memory
```
pombredanne: FYI, my idea there is that eventually the cache will be wrapped such that you can navigate the results anyway you like without memory impact. Likely storing them in a local sqlite DB.

pombredanne:
@haikoschol in https://github.com/nexB/scancode-toolkit/pull/855#issuecomment-346698442 you wrote:

> Passing active_scans is a hack. Ideally I think plugins should receive all command line options, so they can check that the scan included whatever feature they need to function.

I could not agree more. We need to find a cleaner way to pass around the scan run "arguments" consistently.

src/plugincode/post_scan.py
haikoschol: Agreed. Should I also change it on `PreScanPlugin`?

pombredanne: Please do!

pombredanne: eventually the class template should be the same for all plugins. And a plugin becomes a pre/scan/post/output plugin because it is declared to be so :) not because of its "shape"

yashdsaraf:
@pombredanne Sorry for replying so late
@haikoschol Yes this was exactly what I had in mind, I'm taking a look why the tests are failing and the merge conflicts.

haikoschol:
@yashdsaraf The branch was getting stale, so I rebased onto develop. Before that the tests passed. I didn't see any merge conflicts when rebasing.

pombredanne:
@haikoschol I am a tad lost about the history of this PR right now. Especially since you have merge things from @yashdsaraf with some updates. I guess I can unwind this, but it is best to avoid mixing things up, especially other branches, as this makes my life miserable ;)

pombredanne:
Ok I get it this is to in this local 787 branch

haikoschol:
@pombredanne Sorry about that! I rebased the branch in my fork onto develop to avoid merge conflicts later on. Which ended up creating more work for you, not less. :-/

pombredanne:
@haikoschol not a problem, I get it now. Let's try to avoid this in the future :P

Issue #246, opened 2016-04-10
file information `is_archive` should use extractcode archive detection


pombredanne:
The current reporting of an archive is weakly based on the file type and extension. Instead we should use extractcode more accurate detection.

pombredanne:
@shreyateeza sorry if I never replied to you: I just saw that just now... a long time!
Now, this has been implemented in #869 by @Sidsharik 
So I am closing.

pombredanne:
@shreyateeza sorry if I never replied to you: I just saw that just now... a long time!
Now, this has been implemented in #869 by @Sidsharik 
So I am closing.

Again my apologies for this falling through the cracks.

PR #869, opened 2017-12-12
Improve archive type detection with extractcode #869

Sidsharik:
Added extract code more accurate detections.

### DIFF
--- a/src/typecode/contenttype.py
+++ b/src/typecode/contenttype.py
@@ -181,6 +181,7 @@ class Type(object):
         self._is_text = None
         self._is_binary = None
         self._is_data = None
+        self._is_archive = None
         self._contains_text = None
 
     def __repr__(self):
--- a/src/typecode/contenttype.py
+++ b/src/typecode/contenttype.py
@@ -44,8 +44,12 @@ from pdfminer.pdftypes import PDFException
 from commoncode import fileutils
 from commoncode import filetype
 
+from extractcode import archive
+from extractcode import extract
+
 from typecode import magic2
 from typecode import entropy
+
 
 """
 Utilities to detect and report the type of a file or path based on its name,
@@ -285,8 +289,13 @@ class Type(object):
         """
         Return True if the file is some kind of archive or compressed file.
         """
-        # FIXME: we should use extracode archive detection
+
         # TODO: also treat file systems as archives
+
+        y = self.can_extract(location)
+        if can_extract(location) == True:
+            return can_extract(location)
+
         ft = self.filetype_file.lower()
         if (not self.is_text
         and (self.is_compressed
--- a/src/typecode/contenttype.py
+++ b/src/typecode/contenttype.py
@@ -292,9 +292,9 @@ class Type(object):
 
         # TODO: also treat file systems as archives
 
-        y = self.can_extract(location)
-        if can_extract(location) == True:
-            return can_extract(location)
+        extractable = can_extract(location)
+        if extractable == True:
+            return extractable
 
         ft = self.filetype_file.lower()
         if (not self.is_text
--- a/src/typecode/contenttype.py
+++ b/src/typecode/contenttype.py
@@ -292,10 +292,9 @@ class Type(object):
 
         # TODO: also treat file systems as archives
 
-        extractable = can_extract(location)
-        if extractable == True:
-            return extractable
-
+        if self.is_archive(self) is None:
+            self.is_archive = can_extract(location)
+        return self.is_archive
         ft = self.filetype_file.lower()
         if (not self.is_text
         and (self.is_compressed


Sidsharik:
@pombredanne can you help me figure out what went wrong ?

thanks

pombredanne:
Thank you!
For the subject of your commit, please use the imperative style as if giving an order to the codebase to update thyself and add a trailing ticket number for nice cross-linking.

So instead of:

`Added extract code more accurate detections.`

Please use

`Use extractcode in is_archive type detection #246`
or
`Improved archive type detection with extractcode #246`

src/typecode/contenttype.py
```diff
@@ -46,7 +46,9 @@
 
 from typecode import magic2
 from typecode import entropy
-
+#importing the archive.py from extract code
```
pombredanne: Why this comment? I think that this is obvious what you are doing, comment not needed. Also we like imports to be sorted. So extractcode comes before typecode

src/typecode/contenttype.py
```diff
@@ -46,7 +46,9 @@
...
+from extractcode import archive
+from extractcode import extract
```
pombredanne: Always add two empty lines after the imports block

src/typecode/contenttype.py
```diff
@@ -287,6 +289,11 @@ def is_archive(self):
         """
         # FIXME: we should use extracode archive detection
         # TODO: also treat file systems as archives
+        #adding extract code detector
+        can_extract(location)
```
pombredanne: Since you are fixing things, remove the line
`# FIXME: we should use extracode archive detection`

Also the comment `#adding extract code detector` is not needed

And calling `can_extract(location)` without assigning to a variable does not make sense.

src/typecode/contenttype.py
```diff
@@ -287,6 +289,11 @@ def is_archive(self):
...
+        if can_extract(location) == True:
+            extract(location, kinds=extractcode.default_kinds, recurse=False)
```
pombredanne: Are you seriously going to extract all archives just when doing a file type check? Do not extract here!

pombredanne: So may be the only thing to do is:
`return can_extract(location)`

Sidsharik: Thanks alot.
Can you guide me about how to write new unit tests?

Sidsharik:
@pombredanne  did the required changes. Should I resubmit the pull request ?

pombredanne:
Nope, just add a commit to your branch

Sidsharik:
I added the changes and commited is there anything else i am missing?

pombredanne:
Please use imperative "Improve" and not the past "Improved"

Sidsharik:
ok. Done.

src/typecode/contenttype.py
```diff
@@ -285,8 +289,13 @@ def is_archive(self):
...
         Return True if the file is some kind of archive or compressed file.
         """
-        # FIXME: we should use extracode archive detection
+
         # TODO: also treat file systems as archives
+
+        y = self.can_extract(location)
+        if can_extract(location) == True:
+            return can_extract(location)
+
```
pombredanne: What is `y` here? Why not using it? and why such a meaningless name for a variable that you do not use? And what is `self.can_extract` ? it does not exists afaik.... Did you test your code? You also need to add new unit tests of course.
Something like this is likely all you need:

```
extractible = can_extract(location)
if extractible:
    return extractible
```

pombredanne: Or may be rather:
Add this cached attribute above at https://github.com/nexB/scancode-toolkit/pull/869/files#diff-c41ee25999eeb86203e67767ef5115c7R183
```
self._is_archive = None
```

Then this all you need is IMHO no more than this for the whole method:
```
    @property
    def is_archive(self):
        """
        Return True if the file is some kind of archive or compressed file.
        """
        # TODO: also treat file systems as archives
        if self._is_archive is None:
            self._is_archive = can_extract(location)
        return self._is_archive
```

And of course we need a lot of (possibly new) unit tests like this one: https://github.com/nexB/scancode-toolkit/blob/a9083191a04f62c05d588a22fa8f4839eeffc79d/tests/typecode/test_contenttype.py#L156

Sidsharik: where do I write the test cases? 
In the same file contenttype.py? 
I am not familiar with writing tests. Thanks.

pombredanne:
@Sidsharik before going into tests, please do this: https://github.com/nexB/scancode-toolkit/pull/869#discussion_r156369815

pombredanne:
@Sidsharik did you see my comment above?

> @Sidsharik before going into tests, please do this: https://github.com/nexB/scancode-toolkit/pull/869#discussion_r156369815 ?

pombredanne:
You can check the wiki for directions on testing. And if there is not enough doc there, more doc is welcomed!

pombredanne:
But before writing new tests, you need to ensure the current tests pass.

Sidsharik:
What went wrong this time ? It has failed again. Can you please look into it. 
Thanks .

pombredanne:
Well you need to check the Travis and Appveyor build logs.

pombredanne:
See for instance: https://ci.appveyor.com/project/nexB/scancode-toolkit/build/1501/job/inw6tkvtxrv820jc#L89

I gave you the code to use there exactly: https://github.com/nexB/scancode-toolkit/pull/869#discussion_r156369815

Can you use this and run the tests once locally?

pombredanne:
@Sidsharik ping, any update on this? you were almost there... we need some tests too of course

yash-nisar:
@Sidsharik Are you still working on this PR ?

pombredanne:
This has been merged (with quite a few fixes) in f52e1b43a184ab98bd76fbb94ed512bf266a7438
Thanks!

Issue #685, opened 2017-07-13
Install via pip tries to use cache directory in /usr/local/lib


dengste:
I installed 'scancode-toolkit' via 'pip' on a Debian 9 box. On startup I get the following backtrace because a normal user obviously does not have permission to create files in /usr/local/lib:

     Traceback (most recent call last):
       File "/usr/local/bin/scancode", line 11, in <module>
	 sys.exit(scancode())
       File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 722, in __call__
	 return self.main(*args, **kwargs)
       File "/usr/local/lib/python2.7/dist-packages/scancode/utils.py", line 76, in main
	 standalone_mode=standalone_mode, **extra)
       File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 697, in main
	 rv = self.invoke(ctx)
       File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 895, in invoke
	 return ctx.invoke(self.callback, **ctx.params)
       File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 535, in invoke
	 return callback(*args, **kwargs)
       File "/usr/local/lib/python2.7/dist-packages/click/decorators.py", line 17, in new_func
	 return f(get_current_context(), *args, **kwargs)
       File "/usr/local/lib/python2.7/dist-packages/scancode/cli.py", line 386, in scancode
	 scans_cache_class = get_scans_cache_class()
       File "/usr/local/lib/python2.7/dist-packages/scancode/cache.py", line 93, in get_scans_cache_class
	 cache_dir = fileutils.get_temp_dir(unicode(cache_dir), prefix=unicode(timeutils.time2tstamp()) + u'-')
       File "/usr/local/lib/python2.7/dist-packages/commoncode/fileutils.py", line 128, in get_temp_dir
	 return tempfile.mkdtemp(prefix=prefix, dir=base)
       File "/usr/lib/python2.7/tempfile.py", line 339, in mkdtemp
	 _os.mkdir(file, 0700)
     OSError: [Errno 13] Permission denied: '/usr/local/lib/python2.7/.cache/scan_results_caches/2017-07-13T121227

JonoYang:
Your can opt to install pip packages to your home directory using the `--user` option. If you wanted to install it systemwide, you'd need to run pip with root access.

pombredanne:
@JonoYang I think this has been installed system wide with sudo, right @dengste ?
[...]

Issue #357, opened 2016-11-07
Allow cache location configuration 


pombredanne:
Since this could be eventually large and does not play well with any network storage (NTFS shares or NFS) the location where the cache is stored on disk should be configurable, ideally with env variables

pombredanne:
Here is my proposal for this:

- For development usage (from a git checkout) we would detect the presence of .git directory and use a `.cache` directory (or may be a `.tmp` directory) at teh root of the checkout as today. 

- For non-development installations (e.g. with a .git directory) we would by default store all the temporary and cached files in a `.scancode` dir in the user home directory. Since there can be multiple installations and versions installed at the same time, we would use a unique key for each installation as a sub-directory in the `.scancode` directory. That key will be computed based on a hash of the code tree the same way it is computed for the license cache validity check today in here  https://github.com/nexB/scancode-toolkit/blob/c57dab57ff74723e783b7c0dfeb7032f9bb7e84e/src/licensedcode/cache.py#L61 . This would ensure that the sub directory is unique to an installation and that even if some files are updated manually in that installation, a new temp directory would be created.

- additionally if a `SCANCODE_TEMP_DIR` environment variable is defined pointing to a directory path, this would be used and override the location used for caching and temp files both in dev and regular mode.

Last we could also include some crude and quick stats on the amount temporary disk space used in the current temp dir.

In all cases, deleting the whole temp directory would have no impact (unless this is done while a scan is mid air of course)

Feedback welcomed!

haikoschol:
@pombredanne Your proposal makes sense to me and I'd like to give implementing it a try (in order to fix #685). Just two questions/suggestions:

- Why look for a `.git` directory to detect development mode instead of `SCANCODE_DEV_MODE`?
- Wouldn't it be easier to use the same directory structure in both cases for temporary files (i.e. always use `.cache/scancode/<unique_id>/...`)?

pombredanne:
@haikoschol excellent!

> Why look for a .git directory to detect development mode instead of SCANCODE_DEV_MODE?

well, either way works: your call SCANCODE_DEV_MODE is a bit of a wart I created IMHO . I am not very happy about this 

> Wouldn't it be easier to use the same directory structure in both cases for temporary files (i.e. always use .cache/scancode/<unique_id>/...)?

That works too. Having it everywhere the same is indeed better: either `.scancode/` or `.cache/scancode` everywhere is indeed best.

haikoschol:
[...]

PR #878, opened 2017-12-21
Allow cache location configuration

haikoschol:
This is a first attempt to address #357 and #685.

I tried to move around as few things as possible, but I'm still not totally happy with the result. IMHO The most problematic part are import side effects, especially the creation of directories.

Instead of the environment variable `SCANCODE_TEMP_DIR` the code looks for `SCANCODE_TMP`, because that was already in use in `commoncode/fileutils.py`.

### DIFF
--- a/src/commoncode/fileutils.py
+++ b/src/commoncode/fileutils.py
@@ -222,7 +222,7 @@ def build_cache_path(dir_name, dev_mode=False):
         if dev_mode:
             cache_root = dirname(dirname(dirname(abspath(__file__))))
         else:
-            cache_root = os.path.expanduser('~')
+            cache_root = abspath(os.path.expanduser('~'))
 
     return os.path.join(cache_root, '.cache', 'scancode', tree_checksum(), dir_name)
 
--- a/src/commoncode/fileutils.py
+++ b/src/commoncode/fileutils.py
@@ -224,7 +224,6 @@ def build_cache_path(dir_name, dev_mode=False):
         else:
             cache_root = os.path.expanduser('~')
 
-    # tree_hash = tree_checksum()
     return os.path.join(cache_root, '.cache', 'scancode', tree_checksum(), dir_name)
 
 
--- a/src/commoncode/fileutils.py
+++ b/src/commoncode/fileutils.py
@@ -23,8 +23,8 @@
 #  Visit https://github.com/nexB/scancode-toolkit/ for support and download.
 
 from __future__ import absolute_import
-from __future__ import unicode_literals
 from __future__ import print_function
+from __future__ import unicode_literals
 
 # Python 2 and 3 support
 try:
@@ -44,22 +44,22 @@ except ImportError:
 
 import codecs
 import errno
-import os
+from functools import partial
+from hashlib import md5
 import ntpath
+import os
+from os.path import abspath
+from os.path import dirname
 import posixpath
 import shutil
 import stat
 import sys
 import tempfile
-from functools import partial
-from hashlib import md5
-from os.path import abspath
-from os.path import dirname
 
 from commoncode import filetype
 from commoncode.filetype import is_rwx
-from commoncode import system
 from commoncode.system import on_linux
+from commoncode import system
 from commoncode import text
 
 # this exception is not available on posix
--- a/src/commoncode/fileutils.py
+++ b/src/commoncode/fileutils.py
@@ -51,7 +51,10 @@ import shutil
 import stat
 import sys
 import tempfile
-
+from functools import partial
+from hashlib import md5
+from os.path import abspath
+from os.path import dirname
 
 from commoncode import filetype
 from commoncode.filetype import is_rwx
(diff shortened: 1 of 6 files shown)


pombredanne:
@haikoschol Thank you!

src/commoncode/fileutils.py
```diff
@@ -51,7 +51,10 @@
 import stat
 import sys
 import tempfile
-
+from functools import partial
+from hashlib import md5
+from os.path import abspath
+from os.path import dirname
```
pombredanne: Do you mind keeping the imports sorted in a given "block" (e.g. here the stdlib imports)

src/licensedcode/__init__.py
```diff
@@ -27,25 +27,18 @@
...
 from os.path import dirname
 from os.path import abspath
-from os.path import getsize
-from os.path import getmtime
 from os.path import join
 from os.path import exists
 
-from commoncode import fileutils
-
-
```
pombredanne: I kinda like 2 empty lines after imports

src/commoncode/fileutils.py
```diff
@@ -164,6 +170,64 @@ def get_temp_dir(base_dir, prefix=''):
...
+    if cache_root and not os.path.exists(cache_root):
+        raise OSError('{} is set to non-existent directory: {}'.format(TMP_DIR_ENV, cache_root))
+
+    if not cache_root:
+        if dev_mode:
+            cache_root = dirname(dirname(dirname(abspath(__file__))))
+        else:
+            cache_root = os.path.expanduser('~')
+
+    # tree_hash = tree_checksum()
```
pombredanne: Remove this

src/licensedcode/cache.py
```diff
@@ -24,22 +24,17 @@
...
 from __future__ import absolute_import, print_function
 
-from functools import partial
-from hashlib import md5
 import os
 from os.path import exists
-from os.path import getmtime
-from os.path import getsize
 from os.path import join
+from functools import partial
```
pombredanne: Please keep imports sorted here and elsewhere

src/licensedcode/cache.py
```diff
@@ -48,45 +43,20 @@
...
-# If this file exists at the root, the cache is always checked for consistency
-DEV_MODE = os.path.exists(os.path.join(root_dir, 'SCANCODE_DEV_MODE'))
-
-
 def get_or_build_index_through_cache(
         check_consistency=DEV_MODE,
         return_index=True,
         # used for testing only
-        _tree_base_dir=src_dir,
+        _tree_base_dir=None,
```
pombredanne: Why this change?

src/scancode/cache.py
```diff
@@ -41,9 +41,7 @@
 from commoncode.fileutils import path_to_unicode
 from commoncode.system import on_linux
 from commoncode import timeutils
-
-from scancode import scans_cache_dir
-
+from licensedcode import DEV_MODE
```
pombredanne: I think this should be across all scrancode, not just licensedcode

src/commoncode/fileutils.py
```diff
@@ -164,6 +170,64 @@ def get_temp_dir(base_dir, prefix=''):
...
+    cache_root = os.environ.get(TMP_DIR_ENV)
+
+    if cache_root and not os.path.exists(cache_root):
+        raise OSError('{} is set to non-existent directory: {}'.format(TMP_DIR_ENV, cache_root))
+
+    if not cache_root:
+        if dev_mode:
+            cache_root = dirname(dirname(dirname(abspath(__file__))))
+        else:
+            cache_root = os.path.expanduser('~')
```
pombredanne: This may need an abspath too

src/commoncode/fileutils.py
```diff
@@ -164,6 +170,64 @@ def get_temp_dir(base_dir, prefix=''):
...
+    cache_path = build_cache_path(dir_name, dev_mode=dev_mode)
+    create_dir(cache_path)
+    return cache_path
+
+
+def build_cache_path(dir_name, dev_mode=False):
+    cache_root = os.environ.get(TMP_DIR_ENV)
+
+    if cache_root and not os.path.exists(cache_root):
+        raise OSError('{} is set to non-existent directory: {}'.format(TMP_DIR_ENV, cache_root))
```
pombredanne: Why raise an error there? The directory should be created instead IMHO

src/commoncode/fileutils.py
```diff
@@ -164,6 +170,64 @@ def get_temp_dir(base_dir, prefix=''):
...
+
+    if not cache_root:
+        if dev_mode:
+            cache_root = dirname(dirname(dirname(abspath(__file__))))
```
pombredanne: This is the root of scancode's code, right?

src/commoncode/fileutils.py
haikoschol: Setting the environment variable requires user intervention (unlike automatically picking cache dirs based on install method). Therefore I'd argue the user wants to be notified if there is a typo in the path, for example.
Creating a directory in the wrong place and proceeding with the scan might lead to filling up the wrong file system and could be time-consuming to diagnose.

src/licensedcode/cache.py
haikoschol: I wanted to minimize the need to import global variables. It tricky to avoid circular imports and with lower-case names, it is easy to get them mixed up with local variables.

src/commoncode/fileutils.py
haikoschol: Yes. I decided against using the global `src_code` to avoid to many imports across modules. But see my comment below regarding `DEV_MODE`.

src/scancode/cache.py
haikoschol: Agreed. I thought about what would be a better place for it, probably `commoncode`. But it seemed like moving a few of those globals would drag a whole lot of stuff with it, so I opted for keeping changes to a minimum.
I can try to move all globals to `commoncode/__init__.py` (and uppercase them while I'm at it). Should I?

sschuberth:
@pombredanne Can we please get an updated review from you on this? This is blocking to resolve https://github.com/nexB/scancode-toolkit/issues/685 for us.

pombredanne:
@sschuberth Looking at this now! Sorry for the delay

pombredanne:
@haikoschol I would like to rework this on top of the new code in #885 based on the detailed description in https://github.com/nexB/scancode-toolkit/issues/685#issuecomment-358929114

haikoschol:
As discussed offline, this has been superseded by #885.

Issue #469, opened 2017-01-29
De-vendor prebuilt binaries to ease packaging for Linux distros


pombredanne:
@pabs3 and @matthieucan suggested to me on the OFTC #debian-qa IRC channel that ScanCode could be packaged as a Debian Linux package.

@dvc94ch in #288 looks for help on packaging ScanCode for the Guix Linux distro.

Packaging in a distro should be a non-event but it would be a tad involved today. 
ScanCode is already buildable as a wheel which is a first good step towards this.

As a second step, any pre-built binaries that are vendored as a user convenience in the `src/*/bin` directories should be de-vendored from main source tree and provided instead either:
- as another Python package (which can be vendored again such that the current user experience to be able to just download or clone and run without any other install required still works fine)
- as distro-provided dependent packages such that a distro packaging of ScanCode can use proper deps on other distro packages in this case

This way a distro packaing would be vastly simplified

pabs3:
Generally, I suggest a good way to replace embedded code copies is to
create binary packages for Windows/macOS/etc and only embed deps in the
binary packages that can't be installed via binary package deps.

I think Python has some sort of system for binary packages too.

To satisfy the the folks who want to install from source with one
command you can make the setup.py install all dependencies, either via
their distro or via the relevant language package managers.

pombredanne:
@pabs3 wrote
> Generally, I suggest a good way to replace embedded code copies is to create binary packages for Windows/macOS/etc and only embed deps in the binary packages that can't be installed via binary package deps. 

I agree this is the way to go.

Python has a system for binary packages alright and I should be able to use that to include plain binaries for the few pure native exe and libs  (such as `libarchive`, `libmagic`, `p7zip` and a few more)  as extra packages.

And I can make these such that we have variants for Windows, Mac and Linux32 and 64 instead of the vendored directories at https://github.com/nexB/scancode-toolkit/tree/develop/src/typecode/bin or https://github.com/nexB/scancode-toolkit/tree/develop/src/extractcode/bin ...  

I still want to keep the experience to download and run without having anything other dep needed beside Python proper and that on all three OSes and this should work alright: I wish that dealing with installing dependent packages from sources or prebuilt binaries on Windows and Mac would be a well solved problem, but this is not the case.

So the way could be to break these binaries apart such that:
1. they would still be available but provided in some optional Python package such that the current experience of fetch or clone and run without `build-essential` like deps can be kept the same

[...]

Issue #443, opened 2017-01-10
extractcode fails on Fedora 24


akaihola:
On Fedora 24, I get this:
```
$ extractcode test/requests-2.11.1.tar.gz                        
Extracting archives...
  [------------------------------------]
Traceback (most recent call last):
  File "scancode-toolkit/bin/extractcode", line 9, in <module>
    load_entry_point('scancode-toolkit', 'console_scripts', 'extractcode')()
  File "scancode-toolkit/lib/python2.7/site-packages/click/core.py", line 664, in __call__
    return self.main(*args, **kwargs)
  File "scancode-toolkit/src/scancode/utils.py", line 64, in main
    standalone_mode=standalone_mode, **extra)
  File "scancode-toolkit/lib/python2.7/site-packages/click/core.py", line 644, in main
    rv = self.invoke(ctx)
  File "scancode-toolkit/lib/python2.7/site-packages/click/core.py", line 837, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "scancode-toolkit/lib/python2.7/site-packages/click/core.py", line 464, in invoke
    return callback(*args, **kwargs)
  File "scancode-toolkit/src/scancode/extract_cli.py", line 156, in extractcode
    for xev in extraction_events:
  File "scancode-toolkit/lib/python2.7/site-packages/click/_termui_impl.py", line 240, in next
    rv = next(self.iter)
  File "scancode-toolkit/src/scancode/api.py", line 43, in extract_archives
    from extractcode.extract import extract
  File "scancode-toolkit/src/extractcode/extract.py", line 37, in <module>
    from extractcode import archive
  File "scancode-toolkit/src/extractcode/archive.py", line 47, in <module>
    from extractcode import libarchive2
  File "scancode-toolkit/src/extractcode/libarchive2.py", line 91, in <module>
    libarchive = load_lib()
  File "scancode-toolkit/src/extractcode/libarchive2.py", line 79, in load_lib
    lib = ctypes.CDLL(libarchive)
  File "/usr/lib64/python2.7/ctypes/__init__.py", line 357, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: libbz2.so.1.0: cannot open shared object file: No such file or directory
```

It starts working correctly if I run this:

    sudo ln -s /usr/lib64/libbz2.so.1 /usr/lib64/libbz2.so.1.0

pombredanne:
[...]

PR #882, opened 2018-01-11
Docs: Added workaround for testing on RedHat OSs

nishakm:
Since Scancode binaries are built on Debian based machines, and the
prebuilt dependencies are named differently on RedHat based machines
py.test will not be able to find the libbz2.so library on those
machines.
See https://github.com/nexB/scancode-toolkit/issues/443

Added a note about how to work around this issue by symbolically
linking the existing libbz2.so on the filesystem to the expected
name.

This should be removed once
https://github.com/nexB/scancode-toolkit/issues/469 is solved

Signed-off-by: Nisha K <nishak@vmware.com>

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -119,6 +119,19 @@ To set up ScanCode for local development:
 
        py.test -n6
 
+   If you are running this on a RedHat based OS you may come across this
+   failure::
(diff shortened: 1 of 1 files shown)


nishakm:
Resubmitting as github went down while CI checks were running

PR #883, opened 2018-01-11
Docs: Added workaround for testing on RedHat OSs

nishakm:
Since Scancode binaries are built on Debian based machines, and the
prebuilt dependencies are named differently on RedHat based machines
py.test will not be able to find the libbz2.so library on those
machines.
See https://github.com/nexB/scancode-toolkit/issues/443

Added a note about how to work around this issue by symbolically
linking the existing libbz2.so on the filesystem to the expected
name.

This should be removed once
https://github.com/nexB/scancode-toolkit/issues/469 is solved

Signed-off-by: Nisha K <nishak@vmware.com>

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -119,6 +119,19 @@ To set up ScanCode for local development:
 
        py.test -n6
 
+   If you are running this on a RedHat based OS you may come across this
+   failure::
+   
+       OSError: libbz2.so.1.0: cannot open shared object file: No such file or directory
+ 
+   Try creating a symbolic link to libbz2.so.1.0 to solve this issue::
+
+       locate libbz2.so.1.0
+       cd <resulting libbz2.so directory>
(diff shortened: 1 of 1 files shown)


nishakm:
@pombredanne can you review this?

Questions I have about this PR:
1. Do I need to add my name to AUTHORS.rst?
2. Where do I put results from py.test? (This is just a change in docs so I am not sure if this is necessary for this PR)
3. Do I need to add something about this change in CHANGELOG.rst?

Thanks!

pombredanne:
Thanks. And sorry for the delay

>    Do I need to add my name to AUTHORS.rst?

Sure thing. This is a small PR though, so your call

>    Where do I put results from py.test? (This is just a change in docs so I am not sure if this is necessary for this PR)

They are run automatically as you can see above

>    Do I need to add something about this change in CHANGELOG.rst?

not for a small change like this

pombredanne:
LGTM. Merging!

pombredanne:
Thanks again.

PR #856, opened 2017-11-28
Plugin options

pombredanne:
This combines work for #787 by @yashdsaraf and @haikoschol

Issue #789, opened 2017-10-04
Support multiple outputs from a single scan


pombredanne:
It would be nice to have multiple outputs created at once. Today this requires re-running a scan as a scan has only a single output file.

sschuberth:
This probably should be supported via something like `--format json,spdx-tv`.

pombredanne:
@sschuberth How would you pass the path to the output file for each format then?
What about something like this? 
`--format-json  foo/bat/scan.json --format-spdx-tv  this/that/scan.spdx`

sschuberth:
Ah, sorry, I was too much thinking about how our own pre-processor takes command line arguments. We have a `--format` option that takes a comma-separated list of formats, and an `--output-dir` option that specifies the single output directory for all formats. The actual file base name is derived from the input directory / file.

If that's not flexible enough, I could also imagine `--format` to take a comma-separated list of file paths, and the output format is derived from the file extension, failing if it's not any of the known extensions.

I believe both suggestions above are more generic than having separate `--format-XXX` options for format `XXX`. Esp. if plugins may add new formats, is it even possible for plugins to add new command line options to the main program?

pombredanne:
@sschuberth you wrote

> We have a --format option that takes a comma-separated list of formats, and an --output-dir option that specifies the single output directory for all formats. The actual file base name is derived from the input directory / file.

I have thought of this approach, but that sounds to me as difficult to understand and removes control by the user of which files are created exactly. Passing a list of comma-separated file paths and inferring the format based on this feels also like a possible source of confusion.. 

I find it though to get the right CLI UI/UX here!
I wonder if there are other CLI tools that allow to specify multiple output and how they deal with it?

Or may be this is the wrong way to look at this entirely? What if instead we had ONLY a json output in ScanCode and a few additional small tools that could convert this output to another format (CSV, HTML, SPDX, etc)?

sschuberth:
> What if instead we had ONLY a json output in ScanCode and a few additional small tools that could convert this output to another format

That's what we do right now, after all, as we need your own "flavor" of SPDX-TV. While that approach certainly has its advantages, the one thing that I don't like about it is that you lose visibility over what output formats are supported. I simply like it when `scancode --help` tells me about that.

pombredanne:
@sschuberth so you want `scancode --help` to tell you about all the available formats, as it should.
Right now, the help list only brief format "code names" right after the `--format` option.
[...]

Issue #787, opened 2017-10-04
Allow a plugin to provide more than one option


pombredanne:
We should have a way for a plugin to provide not one, but several CLI options
Also a plugin should be able to provide fully configured click Options as options and not be constrained in what the options args are.

This is needed whenever more than one option is required or when more advanced settings need to be provided for a given option. This would be needed for scan plugins where several have multiple options.

@yashdsaraf ping :)

yashdsaraf:
@pombredanne Currently each `pre-scan` plugin works in a way that if an option such as `--ignore abc` is passed, any plugin in the `pre-scan` entrypoint with the entry name `ignore` is initialized with the user input `abc`.
If a plugin provides more than one option, what do you think would be a better approach?
* Initialize the plugin with different user inputs depending on the options?
for e.g `--ignore-glob '*txt' --ignore-regex '^\.'` will call the `ignore` plugin as
```python
plugin('*.txt')
plugin('^\.')
```
or
* Pass an iterable with all the user inputs and options to a single plugin object and leave it to the plugin to decide how to process that data.
for e.g `--ignore-glob '*txt' --ignore-regex '^\.'` will call the `ignore` plugin as
```python
plugin(dict('ignore-glob'='*.txt', 'ignore-regex'='^\.'))
```

yashdsaraf:
Also, should we add a common plugin prefix to the options? say if `ignore` plugin defines 2 options `glob` and `regex`, do we add the plugin name such as `--ignore-glob` and `--ignore-regex` or we let the plugins practice complete autonomy in the naming scheme?

pombredanne:
@yashdsaraf Thanks... here is my thinking on this:

1. We should generalize the notion of plugin to decouple the UI (e.g. the CLI option name) from the actual declaration (as an entry point). This would mean that every plugin would have the same structure and could contribute or more options as they please and define these options entirely.

2. With this generalization a plugin becomes something that receives an iterable of resources and returns an iterable of resources. We should therefore better specify this resource object. In the future a plugin could express interest only in some resource that match certain criteria too.

With this, the only thing that a plugin would declare in the entry point is its "name" and its plugin callable for a given stage of the scan (pre, scan, post, output). All options would be provided by the  plugin class itself. And all plugins would have the same interface: act on a stream of Resources.

This would mean reworking the way all the plugins work today to a single common way.
When they are invoked is the only thing that would be special to a given plugin. 
[...]

Issue #748, opened 2017-09-01
Post-scan plugins should be able to take in parameters


JonoYang:
It would be helpful if post-scan plugins has the ability to take in parameters. For example, you should be able to pass in an API url and API key into a post-scan plugin that uses scan results to perform lookups via an external API.

pombredanne:
This has been fixed with #885 and is in develop!

PR #885, opened 2018-01-17
New plugin architecture

pombredanne:
This is for #787 and a few more tickets.

- The core processing is centered now on a Codebase and Resource abstraction that represents the scanned filesystem in memory #717 #736
- Everything is a plugin including the scans #698 . 
- The CLI becomes therefore much simpler.
- A plugin can provide any option it needs and this support #787 and #748 
- We support multiple output for #789 and scans are also plugins for #698
- The interface for output plugins is the same as other plugins #715

There are a few changes on what is reported:
- scanning a single file or empty dir will always return a value even with strtip-root
- the root directory has a proper entry with its file info
- the counts are correct and do not need to load the whole set of scan results in memory, using instead the lighter weight Codebase and Resources tree
- the scan cache has been folded in the Codebase and Resource code. This will enable a simpler implementation for  #357 and #878 
 

I want to close #856 in favor of this heavily updated plugin and CLI processing architecture.

@yashdsaraf and @haikoschol ping!


JonoYang:
@pombredanne Running `scancode --json-pp --license --package  samples`, which from the help option's example section, does not show the scan results in stdout.

src/scancode/cli.py
```diff
@@ -217,724 +225,891 @@ def print_version(ctx, param, value):
...
+                    if all(to_keep(resource) for to_keep in filters):
+                        continue
+                    filtered_rids_add(rid)
+            finally:
+                msg = 'ERROR: failed to run output filter plugins'
+                echo_stderr(msg, fg='red')
+                echo_stderr(traceback.format_exc())
+                ctx.exit(2)
+
+        codebase.timings['output-filter'] = time() - post_scan_start
```
haikoschol: This probably should be `output_filter_start` instead of `post_scan_start`.

src/scancode/cli.py
```diff
@@ -217,724 +225,891 @@ def print_version(ctx, param, value):
...
+
+        scan_sorter = lambda s: (s.sort_order, s.name)
+        for scanner in sorted(scanner_plugins.values(), key=scan_sorter):
+            scanner_kwargs = scanner.get_own_command_options_kwargs()
+            func = scanner.get_scanner(**scanner_kwargs)
+            scanners.append(Scanner(key=scanner.name, function=func))
+
+        if TRACE_DEEP: logger_debug('scancode: scanners:', scanners)
+
+        if scanners:
```
haikoschol: Is it possible to run scancode without any actual scanners? Otherwise I'd suggest to change this to `if not scanners: errror...`. If it *is* possible, `scan_names` needs to be initialized outside of the `if`, since it is referenced at the end of the function when creating the summary.

src/scancode/cli.py
```diff
@@ -217,724 +225,891 @@ def print_version(ctx, param, value):
...
+            if hasattr(scans, '__enter__'):
+                scans.__enter__()
+
+        while True:
+            try:
+                if with_timing:
+                    location, rid, scan_errors, scan_result, scan_time, scan_result, scan_timings = scans.next()
+                else:
+                    location, rid, scan_errors, scan_time, scan_result = scans.next()
+                    if TRACE_DEEP: logger_debug('scan_codebase: results:', scan_result)
```
haikoschol: This should probably be on the same indentation level as `if with_timing:`.

src/scancode/cli.py
```diff
@@ -217,724 +225,891 @@ def print_version(ctx, param, value):
...
+                msg = 'ERROR: for scanner: ' + scanner.key + ':\n' + error
+                errors.append(msg)
+            if value:
+                # a scanner function MUST return a sequence
+                scanner_result.extend(value)
+        except Exception:
+            msg = 'ERROR: for scanner: ' + scanner.key + ':\n' + traceback.format_exc()
+            errors.append(msg)
+        finally:
+            scan_time = time() - scan_time
```
haikoschol: Shouldn't `scan_time` be calculated only once, outside of the `for` loop?

src/scancode/cli.py
```diff
@@ -217,724 +225,891 @@ def print_version(ctx, param, value):
...
+        if progress_manager:
+            scans = progress_manager(scans)
+            # hack to avoid using a context manager
+            if hasattr(scans, '__enter__'):
+                scans.__enter__()
+
+        while True:
+            try:
+                if with_timing:
+                    location, rid, scan_errors, scan_result, scan_time, scan_result, scan_timings = scans.next()
```
haikoschol: `scan_result` occurs twice. The first one needs to be removed.

src/scancode/cli.py
pombredanne: @haikoschol thanks. Indeed. This is fixed in the latest commit: I now use a single function to run a scan "stage" that is used for all stages except scanners.

src/scancode/cli.py
pombredanne: In the `develop` branch no scanners means that the `-clip` options will be used. Here I only enabled `--info` if no scans were selected. The scan names are therefore never empty at this stage. They are also created only inside the `if scanners:` at https://github.com/nexB/scancode-toolkit/blob/787-new-plugin-architecture/src/scancode/cli.py#L776
Thank you for your eagle eyes!

src/scancode/cli.py
pombredanne: Fixed in latest push. Thanks!

src/scancode/cli.py
haikoschol: Right, `scan_names` is only created inside the `if scanners:` block, but it is referenced after the block ends at https://github.com/nexB/scancode-toolkit/blob/787-new-plugin-architecture/src/scancode/cli.py#L857
Are you saying that `scanners` will never be empty? If so, why the `if`? Maybe this is more of a readability issue and not a bug. In that case, don't let me distract you from more important things. :)

src/scancode/cli.py
pombredanne: This has been updated in the latest push. Thanks!

src/scancode/cli.py
pombredanne: Thank you!
Fixed and I added tests that were missing for the `--timing` option: it was not working at all before that :(

pombredanne:
So I am still ironing a few Windows and Mac quirks. 
Beside this there are few extra things to clean up:
- do not scan for directories., only files
- make info a regular scanner plugin
- change and improve the way scan results are accessible on the Resource object. The current API is total crap and hard to use.

pombredanne:
@haikoschol @yashdsaraf @JonoYang we are back to green and this is ready to be merged in develop
It would be great if you can have a quick look before I merge in develop.
Otherwise I will merge and we can make further fixes from there

JonoYang:
@pombredanne Looks good to me!

src/scancode_config.py
```diff
@@ -0,0 +1,156 @@
...
+    WindowsError  # noqa
+except NameError:
+    WindowsError = None  # NOQA
+
+
+def _create_dir(location):
+    """
+    Create directory and all sub-directories recursively at `location`.
+    Raise Exceptions if it fails to create the directory.
+    NOTE: this is essentailly a copy of commoncode.fileutils.create_dir()
```
haikoschol: The function is slightly different from `commoncode.fileutils.create_dir()`. In particular, it does not call `fsencode()` for dealing with unicode issues on Linux. Is this intentional? If not, how about we get rid of the duplication by importing `_create_dir()` in `commoncode.fileutils`?

pombredanne: https://guides.github.com/features/mastering-markdown/You have eagle eyes! I did not want that module to depend on anything but the stdlib. Therefore I copied and duplicated the code from commoncode AND removed the dep on the fsencode from os.backports.... But now you are asking, why did I do so? there is probably no seriously good reason, except to make sure there would be no circular imports? Actually this is the reason: https://github.com/nexB/scancode-toolkit/blob/5784b20a5c4263a55b8efae0509767aca30caeea/src/commoncode/fileutils.py#L55 ... I guess I could add some comment about this weirdness.

pombredanne: because fileutils also imports scancode_config

pombredanne:
From  @haikoschol
> i ran the tests on a mac and got some failures related to character encoding of files and paths
```
self = <test_fileutils.TestFileUtilsWalk testMethod=test_os_walk_can_walk_non_utf8_path_from_unicode_path>

    def test_os_walk_can_walk_non_utf8_path_from_unicode_path(self):
>       test_dir = self.extract_test_tar_raw('fileutils/walk_non_utf8/non_unicode.tgz')

tests/commoncode/test_fileutils.py:387:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
src/commoncode/testcase.py:278: in extract_test_tar_raw
    return self.__extract(test_path, extract_tar_raw)
src/commoncode/testcase.py:265: in __extract
    verbatim=verbatim)
src/commoncode/testcase.py:294: in _extract_tar_raw
    tar.extractall(path=target_dir)
/usr/local/Cellar/python/2.7.14_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/tarfile.py:2081: in extractall
    self.extract(tarinfo, path)
/usr/local/Cellar/python/2.7.14_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/tarfile.py:2118: in extract
    self._extract_member(tarinfo, os.path.join(path, tarinfo.name))
/usr/local/Cellar/python/2.7.14_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/tarfile.py:2194: in _extract_member
    self.makefile(tarinfo, targetpath)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <tarfile.TarFile object at 0x10d8c9650>, tarinfo = <TarInfo 'non_unicode/non_utf8_decodable_\xff' at 0x10d8c98d0>
targetpath = '/Users/hs/code/scancode-toolkit/tmp/scancode-tests -EmjcWA/4zHUC_/non_unicode.tgz/non_unicode/non_utf8_decodable_\xff'

    def makefile(self, tarinfo, targetpath):
        """Make a file called targetpath.
            """
        source = self.extractfile(tarinfo)
        try:
>           with bltn_open(targetpath, "wb") as target:
E           IOError: [Errno 92] Illegal byte sequence: '/Users/hs/code/scancode-toolkit/tmp/scancode-tests -EmjcWA/4zHUC_/non_unicode.tgz/non_unicode/non_utf8_decodable_\xff'

/usr/local/Cellar/python/2.7.14_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/tarfile.py:2234: IOError
```

> filesystem on my disk is APFS. maybe the CI machine uses HFS+ and there's a difference


"IOError: [Errno 92] Illegal byte sequence:" is a super rare error message :grinning: There is only one hit in Google .... and that said, this test is really just for reference to test the standard OS walk.... here you choke on the tar itself.

If this is the only test that fails, I would rather leave it as failing as this may be a symptom of more issues to come up with APFS.

There may be a few other tests you had failiures with, but this is a core stdlib tar issue. (actually the test above is just there to be a canary on the behavior of os.walk/listdir that changes in Python 3 afairc)

pombredanne:
ok, so I am merging now. We can take it from develop for any other quirks.
@haikoschol @yashdsaraf thanks you for starting this in the first place. It took a while but I feel we are in a better place now!

PR #894, opened 2018-01-26
Limit number of urls and emails reported in a single file #384

inishchith:
PR for issue : #384 
Signed-off-by: Nishchith Shetty inishchith@gmail.com

Note : This fix can be extended later to a CLI option for the limit .  As of now the threshold is set to 50 .

### DIFF
--- a/src/scancode/api.py
+++ b/src/scancode/api.py
@@ -105,8 +105,9 @@ def get_emails(location):
     """
     Yield mappings of emails detected in the file at `location`.
     """
+    email_threshold  = 50
     from cluecode.finder import find_emails
-    for email, line_num  in find_emails(location):
+    for email_count, (email, line_num)  in enumerate(find_emails(location)):
         if not email:
             continue
         misc = OrderedDict()
@@ -114,14 +115,17 @@ def get_emails(location):
         misc['start_line'] = line_num
         misc['end_line'] = line_num
         yield misc
+        if email_count >= email_threshold:
+            break
 
 
 def get_urls(location):
     """
     Yield mappings of urls detected in the file at `location`.
     """
+    url_threshold  = 50 
     from cluecode.finder import find_urls
-    for urls, line_num  in find_urls(location):
+    for url_count, (urls, line_num)  in enumerate(find_urls(location)):
         if not urls:
             continue
         misc = OrderedDict()
@@ -129,6 +133,8 @@ def get_urls(location):
         misc['start_line'] = line_num
         misc['end_line'] = line_num
         yield misc
+        if url_count >= url_threshold:
+            break
 
 
 DEJACODE_LICENSE_URL = 'https://enterprise.dejacode.com/urn/urn:dje:license:{}'


inishchith:
@pombredanne need your review .

pombredanne:
Thanks! your commit is missing a sign off.

src/scancode/api.py
```diff
@@ -105,30 +105,42 @@ def get_emails(location):
     """
     Yield mappings of emails detected in the file at `location`.
     """
+    email_threshold  = 100 
+    email_count = 0
     from cluecode.finder import find_emails
     for email, line_num  in find_emails(location):
```
pombredanne: You could more simply enumerate and break when you reach the mit

src/scancode/api.py
```diff
@@ -105,30 +105,42 @@ def get_emails(location):
...
+        email_count += 1
         misc = OrderedDict()
         misc['email'] = email
         misc['start_line'] = line_num
         misc['end_line'] = line_num
-        yield misc
+        if email_count < 100:
+            yield misc
+    if email_count >= 100:
+        print("\nWarning : %d emails detected at %s but only %d were reported"%(email_count,location,email_threshold))
```
pombredanne: As said in the ticket no warning neeed

inishchith: So for now i'll just remove the warning and break while the hardcoded threshold is reached ?

pombredanne: yes, and use enumerate with the threshold as an arg with a default to 50

inishchith:
@pombredanne for now i've enumerated and used break on a default set 50 . Also once i get more acquainted with the codebase i think i can work on #787  .

src/scancode/api.py
```diff
@@ -105,30 +105,36 @@ def get_emails(location):
     """
     Yield mappings of emails detected in the file at `location`.
     """
+    email_threshold  = 50
```
pombredanne: please use a `threshold=50` function arg here and below

src/scancode/api.py
```diff
@@ -105,30 +105,36 @@ def get_emails(location):
...
+    for url_count, (urls, line_num)  in enumerate(find_urls(location)):
         if not urls:
             continue
         misc = OrderedDict()
         misc['url'] = urls
         misc['start_line'] = line_num
         misc['end_line'] = line_num
         yield misc
+        if url_count >= url_threshold:
+            break
```
pombredanne: Using `itertools.slice` would likely be best, in which case this would take place not here but in the corresponding url and email plugins in the #885 branch?
We could accept leave this API unchanged and in `get_scanner` https://github.com/nexB/scancode-toolkit/blob/877c03e5fd54e099102253e73d0100c1f7407dd7/src/scancode/plugin_url.py#L53 return a closure with `functools.partial` that would wrap its results in an `itertools.islice based on the received argument of a new option `--max-urls`  option added in  https://github.com/nexB/scancode-toolkit/blob/877c03e5fd54e099102253e73d0100c1f7407dd7/src/scancode/plugin_url.py 

What do you think?

inishchith: @pombredanne Yes , sounds good . it'd take some time for me to look at `how to add another option to the plugin` at #787  . I tried skimming the branch , couldn't figure out from where is `-u` sets `url` ( of kwargs ) to `true` I mean from where is URLScanner getting called  .

inishchith:
@pombredanne working on a different branch for #787 so i'll close this .

Issue #624, opened 2017-05-22
Add Git blob sha1 to checksums returned with info


pombredanne:
This can be handy for identification in a Git context

Chaitya62:
I would like to take this issue @pombredanne

Chaitya62:
@pombredanne  Please review the PR #902  and let me know the changes whenever you can.

D-lang14:
Hii @pombredanne sir, I am new to open source, I would like to solve this issue, can you please guide me through how can I help with this issue.

pombredanne:
@D-lang14 sorry for the late reply!
The checksum computation is in commncode.hash alright, therefore this would mean adding this in the scan for `--info` here https://github.com/nexB/scancode-toolkit/blob/cce8cbc64746ed58f2f0a9bf03d6c3500bd7cc1c/src/scancode/api.py#L322 ... plu fixing all the tests

sschuberth:
One more motivation to address this is that [Software Heritage actually uses Git-blob-style SHA1s](https://docs.softwareheritage.org/devel/swh-model/persistent-identifiers.html#git-compatibility), too, also to identify content that was *not* originally stored in Git!

pombredanne:
@AyanSinhaMahapatra we should schedule this in the next release

AyanSinhaMahapatra:
We already had this functionality in `commoncode.hash` with https://github.com/aboutcode-org/scancode-toolkit/commit/2c4f3cd7cfc74f93ec6813f161cb1c8e09b8130b, now this is being returned for all the files using `--info` plugin, along with other checksums we already returned. See https://github.com/aboutcode-org/scancode-toolkit/pull/4425

PR #902, opened 2018-01-27
Add sha1_git to file_infos #624

Chaitya62:
Signed-off-by: Chaitya Shah <chaitya.shah@somaiya.edu>


Chaitya62:
@pombredanne  I think I covered everything, ping me if something needs to be change.

pombredanne:
Thank you! I will merge this as soon as #885 is merged as this changes the architecture quite a bit, but the gist of the code should mergeable.

Chaitya62:
Great, ping me here if anything needs to be changed after the merge

pombredanne:
The merge took place... So the plugin that collects file info is the one to update. Thank you.

pombredanne:
@Chaitya62 do you think you can update this based on the latest develop? sorry for the extra work!

Chaitya62:
@pombredanne no problem, I'll update it.

Chaitya62:
@pombredanne  I have updated the branch please review it

tests/formattedcode/data/json/tree/expected.json
```diff
@@ -20,6 +20,11 @@
       "size": 91,
       "sha1": "3922760d8492eb8f853c10a627f5a73f9eaec6ff",
       "md5": "fc7f53659b7a9db8b6dff0638641778e",
+<<<<<<< HEAD
```
pombredanne: It sounds like you committed unresolved merge conflicts?

src/scancode/api.py
```diff
@@ -225,4 +226,4 @@ def extract_archives(location, recurse=True):
     from extractcode.extract import extract
     from extractcode import default_kinds
     for xevent in extract(location, kinds=default_kinds, recurse=recurse):
-        yield xevent
+        yield xevent
```
pombredanne: I think you removed the trailing line there.

src/scancode/api.py
```diff
@@ -196,9 +196,10 @@ def get_file_info(location, **kwargs):
     result['date'] = get_last_modified_date(location) or None
     result['size'] = getsize(location) or 0
 
-    sha1, md5 = multi_checksums(location, ('sha1', 'md5',)).values()
+    sha1, md5, sha1_git = multi_checksums(location, ('sha1', 'md5','sha1_git')).values()
```
pombredanne: neat and clean! that's the way to go!

tests/formattedcode/data/json/tree/expected.json
Chaitya62: oh sorry, I must have missed it

Chaitya62:
Yes, I'll take care of it this weekend

pombredanne:
I respun the build. It may just be some CI quirk

pombredanne:
So this is all good to go. But I have been thinking that may be this should not be part of the info after all.
May be there should be a plugin for checksums with this one and future ones (sha256, sha512, blake2, etc)... What do you think?

Chaitya62:
Then we will have a different flag for checksums as well right? if yes then, from user point of view won't I want to have all the checksums when I ask for file infos ?

pombredanne:
> Then we will have a different flag for checksums as well right? if yes then, from user point of view won't I want to have all the checksums when I ask for file infos ?

IMHO this would be a scanner plugin of its own. I would have a single option that could accept multiple values. Each values would be in a list of choices such as sha1-git, sha256, etc... and would add a `checksums` attribute to the scans.

something like 
```
"checksums" : [
  "sha256:12142512412121acfd",
  "sha512:12142512412121acfd",
  "sha1-git:12142512412121acfd",
]
```

pombredanne:
@Chaitya62 any feedback on the suggestion above?

pombredanne:
I will eventually merge this but reworked in a checksums plugin instead. Unless you care to tackle this?

pombredanne:
@Chaitya62 Thanks! let me reopen this as a reminder under you have it in.

majurg:
@Chaitya62 @pombredanne Ping. This has been hanging around for a while, unsure if we want to merge or not.

Chaitya62:
After, I wrote this the code was changed significantly ( architecture wise )

majurg:
@Chaitya62 Any plans to update this or should I close this?

Chaitya62:
you can close it

pombredanne:
@Chaitya62 actually I still want this... so we will start from your code to get a proper plugin in :)

pombredanne:
@Chaitya62 I will eventually reuse bit of this and I made a copy of your branch to https://github.com/nexB/scancode-toolkit/tree/Chaitya62/add-git-sha1-to-checksums

pombredanne:
So I am closing this for now. Thank you++

Issue #852, opened 2017-11-22
Cross check that all licenses on https://www.openhub.net/licenses are properly detected


pombredanne:
See https://www.openhub.net/licenses
Several are likely to be variants of BSD and MIT or existing license and will be detected as such.
Some oddities may require a new license addition

yudhik11:
can I take up this issue?

pombredanne:
@yudhik11 That would be very nice of you. Thanks!

yudhik11:
@pombredanne do I need to scrape those license online and check whether they are getting detected OR in the repo there are many licenses I should test them from there.And also the license data is present in src/licensedcode/data/licenses/ and tests/licensedcode/data/licenses/ and to me the thing that confused me was there are more licenses in tests then in src data set.
So with LICENSE data should I use.

pombredanne:
@yudhik11 so the stepos would be:
1. collect the texts online
2. make license tests from each of these in ` tests/licensedcode/data/licenses/` ....  
3. also test by hand the detection
4. if we do not have 100% detection then either:
4.1 add new rules
4.2 add a new license if this is a new license 


> And also the license data is present in src/licensedcode/data/licenses/ and tests/licensedcode/data/licenses/ and to me the thing that confused me was there are more licenses in tests then in src data set.

- `src/licensedcode/data/licenses/` contains the actual real license files used as master data and for detection
-  `tests/licensedcode/data/licenses/` contains test files that we use to verify that the detection works correctly and are used only when running the `py.test` test suite

yudhik11:
@pombredanne what do you meant by [[making license test from each of these in (Addr)]]
and also what is testing by hand the detection

pombredanne:
- testing by hand: ./scancode --license --diag --license-text --format json-pp <file to check>
- license tests: see the automated tests in tests/licensedcode/data/licenses/ where each pair of yml + other file with the same base name become a unit test dynamically created by this code https://github.com/nexB/scancode-toolkit/blob/develop/tests/licensedcode/test_detection_datadriven.py

yudhik11:
@pombredanne it will be nice if you can assign me a small work or any small issue because right now i am trying this from past two weeks and I keep on moving in circle and cannot proceed.

Anything will be much appreciated.

pombredanne:
@yudhik11 ok :) Let me break thing down in smaller issue that would be bite sized!

yudhik11:
[...]

PR #908, opened 2018-01-31
Add script to extract openhub licenses #852

yash-nisar:
Signed-off-by: Yash Nisar yash.nisar@somaiya.edu

A small part of https://github.com/nexB/scancode-toolkit/issues/852

### DIFF
--- a/etc/scripts/openhub_scraper.py
+++ b/etc/scripts/openhub_scraper.py
@@ -1,0 +1,96 @@
+#!/usr/bin/python2
+#
+# Copyright (c) 2017 nexB Inc. and others. All rights reserved.
+# http://nexb.com and https://github.com/nexB/scancode-toolkit/
+# The ScanCode software is licensed under the Apache License version 2.0.
+# Data generated with ScanCode require an acknowledgment.
+# ScanCode is a trademark of nexB Inc.
+#
+# You may not use this software except in compliance with the License.
+# You may obtain a copy of the License at: http://apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software distributed
+# under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+#
+# When you publish or redistribute any data created with ScanCode or any ScanCode
+# derivative work, you must accompany this data with the following acknowledgment:
+#
+#  Generated with ScanCode and provided on an "AS IS" BASIS, WITHOUT WARRANTIES
+#  OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+#  ScanCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  ScanCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/scancode-toolkit/ for support and download.
+
+import codecs
+import json
+import os
+import urllib2
+
+from bs4 import BeautifulSoup
+
+
+def write_license_info_to_file(license_list):
+    """
+    Neatly format all the scrapped license information in the json format and
+    write it to a file.
+    """
+    with codecs.open('openhub_licenses.json', 'w', 'utf-8') as f:
+        f.write(json.dumps(license_list, indent=4, ensure_ascii=False))
+
+
+def parse_local_file():
+    """
+    Parses the given file using 'BeautifulSoup' and returns html content of
+    that file.
+    """
+    local_test_file = os.path.join(os.path.dirname(__file__),
+                                   'testdata/openhub_scraper/openhub_html.html')
+    with open(local_test_file, 'r') as f:
+        parsed_page = BeautifulSoup(f.read(), 'html.parser')
+    return parsed_page
+
+
+def parse_webpage(url, page_no):
+    """
+    Parses the given webpage using 'BeautifulSoup' and returns html content of
+    that webpage.
+    """
+    page = urllib2.urlopen(url + page_no)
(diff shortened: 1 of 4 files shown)


yash-nisar:
Note, the linked issue is being solved in parts, and so the tests are bound to fail. We expect the tests to pass once everything is put up together.

Need your inputs on this @pombredanne :)

pombredanne:
Thanks! I respawned the build job as there is no reason that this fails.

etc/scripts/openhub_scraper.py
```diff
@@ -0,0 +1,28 @@
...
+total_pages = 17
+license_dict = {}
+license_list = []
+
+try:
+	for i in range(1, total_pages + 1):
+		url = 'https://www.openhub.net/licenses?page=' + str(i)
+		page = urllib2.urlopen(url)
+		parsed_page = BeautifulSoup(page, 'html.parser')
+		all_licenses = parsed_page.find(id = 'license').select("table.table-condensed.table-striped.table")
```
pombredanne: Please use only single quotes unless this is a docstring

etc/scripts/openhub_scraper.py
```diff
@@ -0,0 +1,28 @@
...
+import json
+import urllib2
+import requests
+from bs4 import BeautifulSoup
+
+total_pages = 17
+license_dict = {}
+license_list = []
+
+try:
```
pombredanne: Why wrap things in a try/except since you are only printing the error? Also it would be best if you do a proper function for this called in a `if __name__ == '__main__':` block

yash-nisar: @pombredanne I've added the exception `e` too for the stack trace. Should I keep the `try...except` block or remove it ?

etc/scripts/openhub_scraper.py
yash-nisar: Oh damn, how did I miss that ! I'll correct that asap :)

yash-nisar:
@pombredanne Updated the script to reflect the changes :)

pombredanne:
I think the data from open hub are collectively licensed... so what is the license of "openhub_licenses.json"?

etc/scripts/openhub_scraper.py
```diff
@@ -0,0 +1,36 @@
+# A script to extract openhub license names, their urls, and save it in a
```
pombredanne: Can you add a license header there?

etc/scripts/openhub_scraper.py
```diff
@@ -0,0 +1,36 @@
...
+# json file
+import json
+import urllib2
+from bs4 import BeautifulSoup
```
pombredanne: Please sort your imports in two blocks and add one blank line between stdlib and non-stdlib

etc/scripts/openhub_scraper.py
```diff
@@ -0,0 +1,36 @@
...
+
+total_pages = 17
+license_dict = {}
+license_list = []
+
+
+def extract_openhub_licenses():
+    try:
```
pombredanne: A docstring would be nice.

etc/scripts/openhub_scraper.py
```diff
@@ -0,0 +1,36 @@
...
+            for license in license_rows:
+                license_dict['openhub_url'] = license['href']
+                license_dict['name'] = license.get_text()
+                license_list.append(license_dict.copy())
+
+        with open('openhub_licenses.json', 'w') as f:
+            f.write(json.dumps(license_list, indent=4,
+                               ensure_ascii=False).encode('utf8'))
+
+    except Exception as e:
```
pombredanne: Just do not try/except here. Let exceptions bubble up: you are hiding the stack trace otherwise

pombredanne: dumps should return UTF alright. why encode?

pombredanne:
Add a unit test or two would go a long way!

etc/scripts/openhub_scraper.py
yash-nisar: Addressed !

etc/scripts/openhub_scraper.py
yash-nisar: Addressed !

etc/scripts/openhub_scraper.py
yash-nisar: Addressed !

etc/scripts/openhub_scraper.py
yash-nisar: Addressed !

yash-nisar:
@pombredanne Addressed your comments and added a few tests for the same ;)

etc/scripts/test_openhub_scraper.py
```diff
@@ -0,0 +1,76 @@
...
+#  OR CONDITIONS OF ANY KIND, either express or implied. No content created from
+#  ScanCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  ScanCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/scancode-toolkit/ for support and download.
+
+import openhub_scraper
+
+
+def test_scraping_ascii():
```
pombredanne: Looking good, but does this make a remote HTTP call? If so you need instead to use a local test file. We do not want tests that depend on the network.

pombredanne: So you need to need to separate code that fetches from code that parses

yash-nisar:
@pombredanne I've updated the script ! Need your review :)

etc/scripts/test_openhub_scraper.py
yash-nisar: Addressed both the comments :)

pombredanne:
Looking great now and the test are working with no live net connections!

Issue #802, opened 2017-10-11
Always return minimal type information with a scanned file


pombredanne:
AboutCode Manager and other tools need to know if a path is a file or directory at the minimum.
This information is only returned with the `--info` option atm. We should provide this all the times IMHO whether or not detailed file information was requested otherwise proper downstream processing of  scan results  may not be possible.

yash-nisar:
I'm taking this one !

pombredanne:
I have second thoughts on this: let me review and thank you for the commit!

yash-nisar:
@pombredanne What did you finally decide ?

pombredanne:
@yash-nisar I want this alright! so if you can update based on the latest develop I will merge

mjherzog:
Also time to reconsider naming file or directory fields as "resource" fields.

pombredanne:
@mjherzog sure thing, but that would be in another ticket

PR #914, opened 2018-02-04
Return minimal type information #802

yash-nisar:
AboutCode Manager and other tools need to know if a path is a file or
directory at the minimum. This information was previously returned
only with the `--info` option. We now always return minimal type
file information with a scanned file for proper downstream processing
of scan results by tools like the AboutCode Manager.

Signed-off-by: Yash Nisar <yash.nisar@somaiya.edu>

Closes https://github.com/nexB/scancode-toolkit/issues/802

### DIFF
--- a/src/scancode/plugin_info.py
+++ b/src/scancode/plugin_info.py
@@ -63,7 +63,7 @@ class InfoScanner(ScanPlugin):
     options = [
         CommandLineOption(('-i', '--info'),
             is_flag=True, default=False,
-            help='Scan <input> for file information (size, type, checksums, etc).',
+            help='Scan <input> for file information (size, checksums, etc).',
             help_group=OTHER_SCAN_GROUP, sort_order=10
             )
     ]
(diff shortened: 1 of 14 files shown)


yash-nisar:
@pombredanne Need your inputs for the patch :)

pombredanne:
You are really on a roll! Thanks ++ 👍 
I will have to reconcile this I guess with the changes in the #885 branch though!

yash-nisar:
Okay, so should I work & test my changes on the `787-new-plugin-architecture` branch and then create a PR ? @pombredanne

pombredanne:
Your implementation is good, but the underlying code has changed because of my switcharoo!
Do you mind checking the new develop code, rebase and redo you changes there? It should be simpler since `Resource.is_file` is now a flag available at all times on the resource objects. The likely place to fix may be only the `Resource.to_dict` method to always include the type?

yash-nisar:
Sure, I'll check the new architecture and then push my changes again. :)

yash-nisar:
@pombredanne I've updated the PR. Wanna have a look ?

pombredanne:
@yash-nisar Thanks: this is looking great and I am merging.

Issue #106, opened 2015-10-21
Filter out/ignore copyrights by a specific author/holder


nakami:
I want to ignore copyright statements by a specific author.
However, there are two sections in scancode-toolkit/src/cluecode/copyrights.py which might adress my problem.

there is the "patterns" array line 97-257, but also "junk" line 627-653.
which will do the job the best?

also, a bit more documentation on that matter would be great.

pombredanne:
@nakami sorry for not replying earlier to this ticket!
I think the best way to implement this would be to have some settings (possibly as a command line option and/or a `.scancode.yml` settings file where you would list exclusion patterns that if matched would exclude the matching detected copyright

So the copyright would be always detected, but filtered out if it matches an exclusion pattern 
As a refinement the patterns could be based on the copyright holder only, but that is likely too much for a start. 

It might much better than changing the copyright grammar.

Let me add some docs on the way the copyright detection works to the wiki

pombredanne:
vaguely related, @jdaguil pushed a new feature to a branch https://github.com/nexB/scancode-toolkit/tree/214-license-policy : this allows to define a policy on licenses. 
In a similar way I could think of a feature where a list of authors is "white listed" and either filtered out of considered a meeting a policy expectation.

sschuberth:
I'm also interested in this feature. Do you think it would make sense to build on the *214-license-policy* branch and allow to add regex patterns to `.scancode.yml` that filter out specific Copyright findings?

pombredanne:
@sschuberth you wrote:
> Do you think it would make sense to build on the 214-license-policy branch and allow to add regex patterns to .scancode.yml that filter out specific Copyright findings?

it would definitely make sense! and this could overtime evolve to offer provide other scan filtering and failure conditions capabilities.

haikoschol:
@pombredanne I've been looking into this issue and went down the rabbit hole of related issues, feature branches and proposals. Now I'm trying to figure out how I can address this specific issue while at the same time avoid
1. addressing all related issues and the architecture/refactoring work that entails
2. putting in a quick hack that makes 1. harder down the line

At the moment I'm thinking a plugin would be the simplest and cleanest way to filter the results. However, the question then would be how that plugin gets the exclusion patterns to apply. Either by adding a way to pass command line parameters to plugins (#748 and #787) or by reading them from a config file (#520 and the [214-license-policy branch](https://github.com/nexB/scancode-toolkit/tree/214-license-policy)).

[...]

PR #923, opened 2018-02-08
Add "ignore copyrights" plugin

haikoschol:
#106


src/scancode/plugin_ignore_copyrights.py
```diff
@@ -0,0 +1,90 @@
...
+
+@output_filter_impl
+class IgnoreCopyrights(OutputFilterPlugin):
+    """
+    Filter findings that match given copyright holder or author patterns.
+    Has no effect unless the --copyright scan is requested.
+    """
+
+    options = [
+        CommandLineOption(('--ignore-copyright-holders',),
```
pombredanne: I wonder if this should be singular as in `ignore-copyright-holder` and `ignore-author` since each option in the CLI accepts only one value, and multiple ones are allowed?

haikoschol: Good point. I went with plural since I was thinking of it as the "ignore copyrights" plugin. But I agree that singular is better. I'll change it.

pombredanne:
Looking great! Thanks++
Just one tiny comment about singular vs. plural and I merge.

src/scancode/__init__.py
```diff
@@ -187,12 +187,18 @@ def _validate_option_dependencies(ctx, param, value,
     if not other_option_names:
         return
 
-    def _is_set(_value, _default, typ):
-        if type in (BooleanType, BoolParamType):
```
pombredanne: Goo catch, that was a bug alright!

sschuberth:
> e.g. are you sure you want to filter out entirely a Resource based only on its copyrights/authors irrespective of any other scanner data?

That's actually a *good* question! The original motivation for this feature was to simply reduce the number of findings a Legal person would need to go though, and an easy way to do that is to ignore (proprietary) license / copyright statements from one's own company. *If* a file / resource would contain multiple license / copyright statements other than the ones to be ignored, they should still be listed as findings!

So:

> Or would it be instead deleting the copyrights with holders or authors that have a known value? And NOT filtering out the whole Resource?

Yes, that's how it should work.

But thinking about this again, I meanwhile tend to believe such filtering should not be done on the (raw) output at all, but in any separate tool that is visualizing the output. We're planning to do such a tool anyway as part of the [Open Source Review Toolkit](https://github.com/heremaps/oss-review-toolkit) which would implement various other ways of filtering and organizing the findings anyway.

But maybe it'd still good to have this plugin as an example on how such a post-processing would be implemented?

haikoschol:
If we merge it, we at least get the bugfix. Although it's trivial to cherry pick of course. I wonder whether @nakami still has a use for this plugin. :)

pombredanne:
Let me merge this for now and I will add a caveat in the help.

pombredanne:
done! Thank you all for your patience on this one!

pombredanne:
I rebased and merged!

Issue #919, opened 2018-02-07
Allow to load a codebase from an existing JSON scan


pombredanne:
This would be a way to reprocess a codebase though plugins (including outputs) but not through scanners.

vivonk:
@pombredanne is you are trying to say `backtracing`? or making a sample codebase from the json output of our report.

pombredanne:
@vivonk yes. @JonoYang also is onto something, and this may be a dupe of this ticket :P worth asking!

pombredanne:
This has been implemented and merged and is available in the latest tags. Thanks @JonoYang !

PR #937, opened 2018-02-16
Add the ability to load a codebase from a scan to ScanCode #919

pombredanne:
This PR is for #919 e.g. allow to reload a scancode JSON output back in scancode to reprocess it (excluding scanning since this is just data). It creates a virtual codebase from that scan and then it can be re-processed through plugins.

For instance, an SPDX output can be recreated from an original JSON output. Or any prescan or postcan filtering or summarization can be applied.


JonoYang:
@pombredanne I've added tests for the virtual codebase, cache usage with the virtual codebase, and comments. Travis CI and AppVeyor should pass. Please review when you have a moment :)

pombredanne:
Looking good. Thank you for all the tests!

Issue #925, opened 2018-02-10
add time stamp of scan to ScanCode output


armijnhemel:
when running scancode it would be good to record a timestamp of when the scan started, ended or perhaps both (duration?). This makes it easier to later on identify a particular run of scancode.

vivonk:
@pombredanne is it not implemented already? as I read something like this in the repository, in before.

armijnhemel:
I just verified, it is not in the scancode that I am running ( 2.2.1.post241.68ea2eac0 ) with default options.

pombredanne:
@vivonk nope, not yet implemented. This is now available in the Codebase object but not yet saved to JSON

pombredanne:
@armijnhemel Thanks for the ticket btw :)

pombredanne:
This has long been merged and available.
@techytushar Thanks++

PR #945, opened 2018-02-18
Add timestamp to scan summary #925

techytushar:
PR for issue #925 
Signed-off-by: Tushar Mittal <chiragmittal.mittal@gmail.com>

### DIFF
--- a/src/formattedcode/output_json.py
+++ b/src/formattedcode/output_json.py
@@ -64,6 +64,7 @@ class JsonCompactOutput(OutputPlugin):
                    files_count=files_count,
                    scancode_version=scancode_version,
                    scancode_notice=scancode_notice,
+                   scan_start = codebase.scan_start,
                    pretty_options=pretty_options,
                    pretty=False)
 
@@ -92,18 +93,21 @@ class JsonPrettyOutput(OutputPlugin):
                    files_count=files_count,
                    scancode_version=scancode_version,
                    scancode_notice=scancode_notice,
+                   scan_start = codebase.scan_start,
                    pretty_options=pretty_options,
                    pretty=True)
 
 
 def write_json(results, output_file, files_count,
                scancode_version, scancode_notice,
-               pretty_options, pretty=False):
+               scan_start, pretty_options, 
+               pretty=False):
 
     scan = OrderedDict([
         ('scancode_notice', scancode_notice),
         ('scancode_version', scancode_version),
         ('scancode_options', pretty_options),
+        ('scan_start', scan_start),
         ('files_count', files_count),
         ('files', results),
     ])
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -724,7 +724,7 @@ def scancode(ctx, input,  # NOQA
         # TODO: this is weird: may be the timings should NOt be stored on the
         # codebase, since they exist in abstract of it??
         codebase.timings.update(setup_timings)
-
+        codebase.scan_start = scan_start
         codebase.timings['inventory'] = time() - inventory_start
         files_count, dirs_count, size_count = codebase.compute_counts()
         codebase.summary['initial:files_count'] = files_count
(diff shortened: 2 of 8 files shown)


src/scancode/cli.py
```diff
@@ -47,7 +47,7 @@
 from scancode_config import scancode_temp_dir
 
 from commoncode.fileutils import PATH_TYPE
-from commoncode.timeutils import time2tstamp
+from commoncode.timeutils import time2tstamp,tstamp2time
```
pombredanne: Please keep each import on a separate line, but also based on my other comments `tstamp2time` is not likely needed at all.

src/scancode/cli.py
```diff
@@ -1202,7 +1202,14 @@ def display_summary(codebase, scan_names, processes, verbose):
                                 '%(final_size_count)s' % locals())
 
     echo_stderr('Timings:')
+
+    timestamp = codebase.timings.get('timestamp',None)
+    timestamp = tstamp2time(timestamp)
```
pombredanne: Why do you get a time back?

src/scancode/cli.py
```diff
@@ -1202,7 +1202,14 @@ def display_summary(codebase, scan_names, processes, verbose):
...
+    echo_stderr('  timestamp: %(timestamp)s'% locals())
+    
     for name, value, in codebase.timings.items():
+        if name == 'timestamp':
```
pombredanne: You should pop() above instead IMHO
e.g. not `timestamp = codebase.timings.get('timestamp',None)` but instead `timestamp = codebase.timings.pop('timestamp',None)`

src/scancode/cli.py
```diff
@@ -724,7 +724,7 @@ def scancode(ctx, input,  # NOQA
         # TODO: this is weird: may be the timings should NOt be stored on the
         # codebase, since they exist in abstract of it??
         codebase.timings.update(setup_timings)
-
+        codebase.timings['timestamp'] = scan_start
```
pombredanne: IMHO the scan_start should be its own attribute on the Codebase. This is not a "timing" proper.

src/scancode/cli.py
techytushar: IMHO `time2tstamp` returns date time like this `2018-02-19T145640.979020` which is quite hard to read and understand so i converted it to `2018-02-19 14:56:40.979020` using `tstamp2time`

pombredanne: IMHO It is pure luck if a datetime ends up being serialized alright.
Instead you should use this I just pushed https://github.com/nexB/scancode-toolkit/commit/179d983a222cb067a40cfdcf05de70da1ffa262f with `path_safe=False`

src/scancode/cli.py
pombredanne: But this should not be in timings... but a proper attribute, so this is moot.

techytushar:
@pombredanne please review, i have made the required changes.. :sweat_smile:

src/formattedcode/output_json.py
```diff
@@ -92,18 +93,21 @@ def process_codebase(self, codebase, output_json_pp, files_count,
...
                    scancode_notice=scancode_notice,
+                   timestamp = codebase.scan_start,
                    pretty_options=pretty_options,
                    pretty=True)
 
 
 def write_json(results, output_file, files_count,
                scancode_version, scancode_notice,
-               pretty_options, pretty=False):
+               timestamp,pretty_options,
```
pombredanne: your code is not properly formatted

techytushar:
@pombredanne please review, add the test

src/scancode/cli.py
```diff
@@ -1202,6 +1202,10 @@ def display_summary(codebase, scan_names, processes, verbose):
                                 '%(final_size_count)s' % locals())
 
     echo_stderr('Timings:')
+
+    timestamp = codebase.scan_start
+    echo_stderr('  timestamp: %(timestamp)s'% locals())
```
pombredanne: Could we call this Scan start instead here and in the JSON output? ... since this is what this is.

techytushar:
@pombredanne only one test is failing, but all tests pass on my local machine, please review..

pombredanne:
The travis failure is an heisenbug. I respun a build

pombredanne:
Merging now. Thanks!

PR #946, opened 2018-02-18
Add example text to external file

techytushar:
Signed-off-by: Tushar Mittal <chiragmittal.mittal@gmail.com>

### DIFF
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -82,6 +82,8 @@ from scancode.resource import Resource
 from scancode.utils import BaseCommand
 from scancode.utils import path_progress_message
 from scancode.utils import progressmanager
+from scancode.help import examples_text
+from scancode.help import epilog_text
 
 # Python 2 and 3 support
 try:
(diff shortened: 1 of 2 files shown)


pombredanne:
Thanks! You are on a roll!

pombredanne:
Why is it better to have this in an external file?

techytushar:
there was a comment in the code saying `FIXME: this should be pushed out in some external help or pushed down in plugins.`, besides IMHO this makes the `cli.py` file clean...

src/scancode/cli.py
```diff
@@ -82,6 +82,7 @@
 from scancode.utils import BaseCommand
 from scancode.utils import path_progress_message
 from scancode.utils import progressmanager
+from scancode.example_text import examples_text, epilog_text
```
pombredanne: Make each import on its own line

pombredanne:
I can buy that, but the file should be called help.py

pombredanne:
Thanks! If we try to go a tad further, what would be the way to let plugins provide their own sample or extended help instead?

pombredanne:
Thanks. Merging!

PR #970, opened 2018-03-09
Fix sample command

jamesward:
Running the provided sample command results in:
```
Error: Missing scan option(s): at least one scan option is required.
```

It seems the scan type is required.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -165,8 +165,8 @@ Run this command for a list of command line examples::
 
     ./scancode --examples
 
-To run a scan on sample data, first run this::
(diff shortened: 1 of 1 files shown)


pombredanne:
@jamesward You have eagle eyes. Thanks ++. Merging

pombredanne:
@jamesward actually this proves really useful as I realized how the doc needs a serious facelift to sync up with the latest additions we have in develop before we can make a proper release (the README and more are correct if you were to use the current stable master, but rather off with the tip of develop)

Issue #969, opened 2018-03-08
Multiple version of GPL detected in smartctl 6.5


JonoYang:
Reported by baskar.pammi@arris.com

ScanCode 2.2.1 detects multiple versions of the GPL in the binary `smartctl` built from [smartmontools 6.5](https://sourceforge.net/projects/smartmontools/files/smartmontools/6.5/). GPL 2.0 is the only license expected from `smartctl`.

![smartctl](https://user-images.githubusercontent.com/6044800/37172685-5cfe4bcc-22c6-11e8-8095-48484d16411e.png)

pombredanne:
@JonoYang Thanks. I have a few new license detection rules to fix this that I will commit shortly!

pombredanne:
This has been fixed and merged in develop. Review welcomed!

JonoYang:
The match to GPL 3.0 has been removed but there are still matches to GPL 1.0. This is due to ScanCode reporting the string "GPL" and "gpl" as the GPL 1.0 license. I'll submit a new PR for these cases.

JonoYang:
Fix merged into develop in #981

PR #975, opened 2018-03-10
smartmontools licenses

pombredanne:
Improve license detection for GPL and related notices found in smartmontools #969


pombredanne:
Looking good to me. Merging

PR #977, opened 2018-03-10
Improve copyright POS tagging

pombredanne:
* minor changes mostly from scanning several npms


pombredanne:
There is still a failure on Windows but this should be fixed in develop. I will merge this as it is.

PR #981, opened 2018-03-12
Additional negative rules for smartctl

JonoYang:
smartctl uses the abbreviation "GPL" for "General Purpose Logging". This causes ScanCode to falsely report that the GPL 1.0 license is detected. These rules should solve that issue.


pombredanne:
Thanks. You can ignore safely the timeout error in Windows tests. This has been fixed in develop: lately Appveyor runs slower than usual. Looking  good!

JonoYang:
Merging.

Issue #214, opened 2016-03-03
Include license policy using a configuration file


jdaguil:

pombredanne:
This ticket was originally an experiment but it makes sense to flesh out a bit the design and goals.
Here is what I think these could be:

**Context**

When running a scan:
- there are certain results that are useful to ignore such as a some copyright statements per #106, specific file paths to ignore per #36, some file types (such as PDF) that may not be super useful to scan per #119 
- there are also certain results that could be flagged as OK, not OK or requiring some review. This could be for instance a certain license or license category: e.g. if I am scanning my GPL-licensed project it may be problematic to include proprietary-licensed code in it. Or it could be known copyrights such as my own copyright, or it could even be some package, urls or emails. Or it could the absence of some data that could be a trigger: for instance the absence of a copyright or license. 
- the usage context or the type or classification of a scanned file may matter too: for instance it may be perfectly  OK for an Apache-licensed project to use a GPL-licensed build utility or some test code but that may not be OK when used in the main code.

**Approach**

Based on this a policy would be some rules that define:
- what to scan and not to scan
- with certain scan results, (e.g. a license, copyright, package, etc) whether this is OK, not OK or needed some review (e.g. a while list, black list and grey list or green, orange, red markers)

The rules could apply to a whole scan or several scans (and therefore should be stored in a config file that could be referenced in a command line option) and the rules could be also specific to certain portions of a code tree (either by having several config files scattered in a code tree or by having a central file that can reference sub tree elements)

The usage would be that once the scan is completed then the policy would be "applied" on the results and the final scan results enhanced with the matched policy elements. This could take the form of a new attribute attached to a file in the scan results and providing the results of "matching" the policy rules to the scan results. Specific outputs (such as the html-app) and GUI tools (such as the https://github.com/nexB/aboutcode-manager) could then use that data to provide feedback and navigation of the results based on the incorporated policy feedback. A scan summary could provide a summary of compliance with the defined policy listing things OK, not OK, and requiring review. We go even go as far as reporting an error return code from running a scan based on this.

@mjherzog you had some opinion on this. Input welcomed.
@nakami @sschuberth ditto, based on  ticket #106 and comments.

sschuberth:
Sounds good to me, but just one remark for clarity ...

> The usage would be that once the scan is completed then the policy would be "applied" on the results

[...]

PR #1008, opened 2018-03-26
License policy plugin

pombredanne:
This is the PR for #214


pombredanne:
@majurg the test failure is because some path is reported as absolute rather than relative.

majurg:
@pombredanne How do I prevent the plugin system from using absolute path?

src/scancode/plugin_license_policy.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+
+    attributes = dict(license_policy=attr.ib(default=attr.Factory(dict)))
+
+    sort_order = 9
+
+    options = [
+        CommandLineOption(('--license-policy',),
+            multiple=False,
+            metavar='<license_policy>',
```
pombredanne: Just curious: are we using metavars this way elsewhere? or rather a plain `FILE` ?

src/scancode/plugin_license_policy.py
```diff
@@ -0,0 +1,109 @@
...
+                resource_license_keys = [entry.get('key') for entry in resource.licenses]
+
+                for key in resource_license_keys:
+                    if key in license_policies.keys():
+                        # Apply the policy to the Resource
+                        resource.license_policy = license_policies[key]
+                        codebase.save_resource(resource)
+
+            except AttributeError:
+                # add license_policy regardless if there is --license info or not
```
pombredanne: I was about to say that another way could have been to require a license scan... But this is actually better as this always works!

src/scancode/plugin_license_policy.py
majurg: Ok I will make that change

src/scancode/plugin_license_policy.py
pombredanne: I think we are could have this both as a separate plugin and we should be able to run on a "virtual" codebase alright as it is

pombredanne:
@majurg the failure is based on @JonoYang pending fix in #1018 , so I am merging as is.
Thank you!

PR #1011, opened 2018-03-28
Do not attempt to write empty package info to SPDX

ThorstenHarter:
When you scan a component without any copyright or license info, ScanCode will report an error:
`spdx.writers.tagvalue.InvalidDocumentError: [u'Package must have at least one file.']`

Adding a check before writing spdx_output will fix this problem, the output file will only contain the string `""# No results for package [name]."`

Signed-off-by: Thorsten Harter <thorsten.harter@gmx.net>

### DIFF
--- a/src/formattedcode/output_spdx.py
+++ b/src/formattedcode/output_spdx.py
@@ -292,11 +292,10 @@ def write_spdx(output_file, results, scancode_version, scancode_notice,
     # Therefore in one case we do nothing (rdf) and in the other case we
     # encode to UTF8 bytes.
 
-    if len(package.files) > 0:
+    if package.files:
       from StringIO import StringIO
       spdx_output = StringIO()
-      if len(package.files) > 0:
-        write_document(doc, spdx_output, validate=True)
+      write_document(doc, spdx_output, validate=True)
       result = spdx_output.getvalue()
       if as_tagvalue:
           result = result.encode('utf-8')
--- a/src/formattedcode/output_spdx.py
+++ b/src/formattedcode/output_spdx.py
@@ -292,10 +292,12 @@ def write_spdx(output_file, results, scancode_version, scancode_notice,
     # Therefore in one case we do nothing (rdf) and in the other case we
     # encode to UTF8 bytes.
 
-    from StringIO import StringIO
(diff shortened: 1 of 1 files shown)


src/formattedcode/output_spdx.py
```diff
@@ -292,10 +292,12 @@ def write_spdx(output_file, results, scancode_version, scancode_notice,
...
     # encode to UTF8 bytes.
 
-    from StringIO import StringIO
-    spdx_output = StringIO()
-    write_document(doc, spdx_output, validate=True)
-    result = spdx_output.getvalue()
-    if as_tagvalue:
-        result = result.encode('utf-8')
-    output_file.write(result)
+    if len(package.files) > 0:
```
pombredanne: You could also write the simpler form: `if package.files:`

src/formattedcode/output_spdx.py
```diff
@@ -292,10 +292,12 @@ def write_spdx(output_file, results, scancode_version, scancode_notice,
...
+      from StringIO import StringIO
+      spdx_output = StringIO()
+      if len(package.files) > 0:
```
pombredanne: You likely do not need to repeat this test here since this is already `True` based on your test above.

ThorstenHarter:
@pombredanne : The output SPDX file is created, because of this code at line 250ff:
```
    if len(package.files) == 0:
        if as_tagvalue:
            output_file.write("# No results for package '{}'.\n".format(package.name))
        else:
            output_file.write("<!-- No results for package '{}'. -->\n".format(package.name))
```
For example: `# No results for package 'test'.`

pombredanne:
I respun the build on Appveyor that failed for some mysterious reason

pombredanne:
@ThorstenHarter This is merged with: https://github.com/nexB/scancode-toolkit/commit/71128ebeb2c6ddcb29863805cc4ce9103b6888c6
I ameneded your commits to add the signoff that you added to this PR message : `Signed-off-by: Thorsten Harter <thorsten.harter@gmx.net>` and pushed some tests too.

I also created a follow up #1011 as we should not produce a mostly empty document IMHO in these cases.

Thank you ++

Issue #1007, opened 2018-03-26
Crash when scanning


zoltanblenessygfi:
Hello, 

I am trying to scan a somewhat complex project, I use multiple --ignore switches and others. My command looks like this: 
sudo scancode --license --license-diag --output-html-app ossreport/index.html --timeout 7200 -n 10 . --ignore "*.gif" --ignore "*.pdf" --ignore "*.rtf" --ignore "*.idx" --ignore "*.csv" --ignore "**/cvs-rlog-full.txt" --ignore "**/ChangeLog" --ignore "/ossreport" --ignore "**/test" --ignore "**/tests" --ignore "**/doc" --ignore "protractor" --verbose

Scan always crashes with the following stack: 
Traceback (most recent call last):
  File "/usr/local/bin/scancode", line 11, in <module>
    sys.exit(scancode())
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/scancode/utils.py", line 79, in main
    standalone_mode=standalone_mode, **extra)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/scancode/cli.py", line 707, in scancode
    stage='scan', kwargs=kwargs)
  File "/usr/local/lib/python2.7/dist-packages/scancode/cli.py", line 852, in run_scanners
    with_timing=timing, progress_manager=progress_manager)
  File "/usr/local/lib/python2.7/dist-packages/scancode/cli.py", line 955, in scan_codebase
    codebase.save_resource(resource)
  File "/usr/local/lib/python2.7/dist-packages/scancode/resource.py", line 515, in save_resource
    raise UnknownResource('Not part of codebase: %(resource)r' % resource)
TypeError: format requires a mapping

OS: Ubuntu 16.04 with latest updates
Python: 2.7.12-1~16.04
[...]

PR #1013, opened 2018-04-02
Save parent Resource of the Resource being removed from the codebase

JonoYang:
This fixes the issue in #1007


pombredanne:
@JonoYang Thanks!

Issue #967, opened 2018-03-07
File counts in summary are completely wrong


pombredanne:
This are completely wrong:

Initial counts: 12452 resource(s): 11039 file(s) and 1413 directorie(s) 
Final counts:   12452 resource(s): 11039 file(s) and 1413 directorie(s) 

```
scancode -l -n5 linux-4.15.7/ --output-csv 417.csv 
Setup plugins...
Collect file inventory...
Scan files for: licenses with 5 process(es)...
[####################] 62061                                                           
Scanning done.
Summary:        licenses with 5 process(es)
Errors count:   0
Scan Speed:     9.22 files/sec. 
Initial counts: 12452 resource(s): 11039 file(s) and 1413 directorie(s) 
Final counts:   12452 resource(s): 11039 file(s) and 1413 directorie(s) 
Timings:
  setup_scan:licenses: 46.79s
  setup: 46.80s
  inventory: 39.83s
  scan: 1197.61s
  output:csv: 28.78s
  output: 28.78s
  total: 1389.84s
```

techytushar:
@pombredanne what should it be instead of this? i.e what is the correct output :confused:

JonoYang:
@pombredanne 

This looks to be an issue when using the disk cache. When everything is memory, the counts are correct. When we use the disk cache, the counts are off.

Everything in memory (`--max-in-memory 0`)

```
$ scancode -i -n4 ~/Desktop/linux-stable-4.15.7 --json ~/Desktop/kernel.json --max-in-memory 0
Setup plugins...
Collect file inventory...
Scan files for: info with 4 process(es)...
[####################] 62061                                                           
Scanning done.
Summary:        info with 4 process(es)
Errors count:   0
Scan Speed:     484.66 files/sec. 5.84 MB/sec.
Initial counts: 66438 resource(s): 62061 file(s) and 4377 directorie(s) 
Final counts:   66438 resource(s): 62061 file(s) and 4377 directorie(s) for 747.61 MB
Timings:
  inventory: 8.80s
  scan: 128.05s
  output:json: 9.31s
  output: 9.31s
  total: 147.48s
```

Default settings (`--max-in-memory 10000`)

```
$ scancode -i -n4 ~/Desktop/linux-stable-4.15.7 --json ~/Desktop/kernel.json
Setup plugins...
Collect file inventory...
Scan files for: info with 4 process(es)...
[####################] 62061                                                           
Scanning done.
Summary:        info with 4 process(es)
[...]

PR #1015, opened 2018-04-03
Save Resource after updating counts

JonoYang:
PR for #967


pombredanne:
Thanks! that is simple and neat. It also highlights some core issues in parts that compute counts and whether or not we should keep some bits of the codebase/resource objects in memory at all times or not

pombredanne:
I am rebasing first

Issue #1017, opened 2018-04-03
Failure with a division by zero


pombredanne:
See https://ci.appveyor.com/project/nexB/scancode-toolkit/build/job/76oibjmlrmqon3h5#L553

@JonoYang this was introduced after merging your latest code.

JonoYang:
```
       scan_time = codebase.timings.get('scan', 0.)

       ...

       if scan_size_count:
>           scan_size_speed = format_size(scan_size_count / scan_time)
E           ZeroDivisionError: float division by zero
```
It seems like `scan` isn't being added to `codebase.timings`, so `scan_time` defaults to `0`, or the `scan` is actually zero. In both cases, we divide by zero. I am not able to reproduce this on Ubuntu or on a Windows 10 VM.

pombredanne:
thx! It could very much be that we were not reaching this before and that the bug was masked. I cannot reproduce either on my laptop. But the Ci fails consistently

JonoYang:
I put an assertion on `scan_time` to see if it is `0`. `scan_time` is `0`, so we pass the assertion, and then attempt to divide by zero later on.  This only happens on Appveyor. The Travis CI build fails the assertion at this point on the same test because `scan_time` is a nonzero number. I'm not sure why Appveyor acts differently with regards to `scan_time`.

```
        scan_time = codebase.timings.get('scan', 0.)
        if TRACE:
            logger_debug('######## scan_time ########', scan_time)
            assert scan_time == 0
    
        scan_files_count = codebase.summary.get('scan:files_count', 0)
    
        if scan_time > 0:
            scan_file_speed = round(float(scan_files_count) / scan_time , 2)
        else:
            scan_file_speed = 0
    
        scan_size_count = codebase.summary.get('scan:size_count', 0)
    
        if scan_size_count:
>           scan_size_speed = format_size(scan_size_count / scan_time)
E           ZeroDivisionError: float division by zero
```

pombredanne:
ah, this may because something like the time precision on Windows? 
One thing you could do as a cheat?
`scan_time = codebase.timings.get('scan', 0.0000001)`

Or test if the time is 0 and the the speed would be zero.

JonoYang:
I'm trying out changing the default value of `scan_time` from `0` to `0.0000001` to see what happens, though I favor testing if the `scan_time` is `0`, then the `scan_size_speed` is `0` as a fix.

JonoYang:
[...]

PR #1018, opened 2018-04-03
Check if scan_time is zero before calculating scan_size_speed

JonoYang:
PR for #1017

### DIFF
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -1068,7 +1068,7 @@ def display_summary(codebase, scan_names, processes, verbose):
             prescan_scan_size_speed = ''
 
     ######################################################################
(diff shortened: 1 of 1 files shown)


src/scancode/cli.py
```diff
@@ -1080,7 +1080,11 @@ def display_summary(codebase, scan_names, processes, verbose):
     scan_size_count = codebase.summary.get('scan:size_count', 0)
 
     if scan_size_count:
-        scan_size_speed = format_size(scan_size_count / scan_time)
+        if scan_time > 0:
```
pombredanne: You could write just `if scan_time:`

pombredanne:
@JonoYang I am merging as-is as this is fixing a bug but we would still need some tests if possible.
Thank you.

Issue #817, opened 2017-10-25
Add new license matcher for SPDX-License-Identifer


pombredanne:
Do not use license rules for these but instead a specialised matcher

pombredanne:
This has been merged in devel

PR #1025, opened 2018-04-09
Add support for SPDX-License-Identifer tags

pombredanne:
This PR adds support for the detection of SPDX-License-Identifer for #817
Some caveats: this only detects expression composed of known SPDX license identifiers that are on a single-line and at the start of a line.
It can handle malformed or invalid expressions, including the Uboot list of licenses.


pombredanne:
Note that some of the code paves the way for #74 to report effectively license expressions for which is there is not much work left to do.

pombredanne:
All tests passing so I am merging. This is only incremental new features.

PR #1034, opened 2018-04-16
Change ignore plugin logic #1007

JonoYang:
The ignore plugin now works in two steps:

- The ignore plugin first walks the codebase top-down to collect the rid's of the resources that are to be ignored
- Then the plugin walks the codebase bottom up and removes resources if their rid is in the list of rids to remove


pombredanne:
Looking good! Merging.

PR #1061, opened 2018-05-10
fix a couple of spelling typos

paralax:


pombredanne:
@paralax Thank you: good catch! 
This is minor so I amreging this as-is. In the future, would you mind adding a proper  Signed-off-by to your commit or at least to the PR? that's our procedure.

Issue #1073, opened 2018-05-22
Add support for FreeBSD Ports and packages


pombredanne:
Ports are the packages for FreeBSD. They area built from sources and also available as pre-built binaries.

- The main site is at: https://www.freebsd.org/ports/
- The doc is at https://www.freebsd.org/doc/handbook/ports.html
- Source are in SVN at https://svnweb.freebsd.org/ports/head/
- And binaries at http://pkg.freebsd.org/

An example of a "port" would be this on the source side:
https://svnweb.freebsd.org/ports/head/sysutils/dmidecode/

The main source-side "metafile" is the Makefile:
https://svnweb.freebsd.org/ports/head/sysutils/dmidecode/Makefile?view=markup

The binaries are there (by architectures and releases):
https://pkg.freebsd.org/FreeBSD:10:amd64/release_1/All/

The index https://pkg.freebsd.org/FreeBSD:10:amd64/release_3/packagesite.txz may be of seom interest in the future and contains a packagesite.yaml which has one valid YAML doc in flowstyle per line, each being package data.

Note: the field "origin" contains a path back to the SVN ports tree
e.g. with origin="x11/xrefresh" we can go to  https://svnweb.freebsd.org/ports/head/x11/xrefresh/

For dmidecode as an example package, the tarball:

https://pkg.freebsd.org/FreeBSD:10:amd64/release_1/All/dmidecode-2.12.txz

would contain this +COMPACT_MANIFEST file as a YAML file (same strcuture as each line in the packagesite.yaml above):
```
    "name": "dmidecode", 
    "origin": "sysutils/dmidecode", 
    "version": "2.12", 
    "name": "dmidecode", 
    "origin": "sysutils/dmidecode", 
    "version": "2.12", 
    "arch": "freebsd:10:x86:64", 
    "maintainer": "anders@FreeBSD.org", 
    "prefix": "/usr/local", 
    "www": "http://www.nongnu.org/dmidecode/", 
    "flatsize": 188874, 
    "comment": "Tool for dumping DMI (SMBIOS) contents in human-readable format", 
    "licenselogic": "single", 
    "licenses": [
        "GPLv2"
    ], 
[...]

PR #1077, opened 2018-05-24
Add support for FreeBSD packages (ports) #1073

majurg:
* FreeBSD package support added for FreeBSD port's +COMPACT_MANIFEST
  file
* Creates scancode Package with the following fields filled out:
    * name
    * version
    * homepage_url
    * description
    * keywords
    * declared_licensing
    * parties
    * code_view_url
    * download_url


majurg:
@pombredanne The download URLs for the binary packages are not guaranteed to exist in all cases for an exact package name, version. This is due to the rolling-release nature of binary FreeBSD packages. If there was a version bump to any particular package, the download URLs pointing to the older version will no longer exist.

Even with the possibility of using the various `release_{{ }}` and `quarterly` repos, these download URLs are not not guaranteed to exist due to the same reasons as above.

That being said, I am unsure of weather or not we should include a `download_url` for these FreeBSD packages.

src/packagedcode/freebsd.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+def license_mapper(licenses, package):
+    """
+    Update package licensing and return package. Licensing structure for FreeBSD
+    packages is a list of (non-scancode) license keys.
+    """
+    if isinstance(licenses, list):
+        lics = [l.strip() for l in licenses if l and l.strip()]
+        lics = ', '.join(lics)
```
pombredanne: We would need to take into account the licenselogic for this too. This can be dealt with afterwards and I posted a comment
here https://github.com/nexB/scancode-toolkit/issues/1073#issuecomment-392041079

pombredanne:
@majurg thanks. this weird test failure in https://travis-ci.org/nexB/scancode-toolkit/jobs/385363822 is rather unexplained... or this is because the order of evaluation of packages has changed with  https://github.com/nexB/scancode-toolkit/pull/1077/files#diff-009913c41b348f948f7b14c8e061770bR50

Try may be to move your class just before this: 
https://github.com/nexB/scancode-toolkit/pull/1077/files#diff-009913c41b348f948f7b14c8e061770bR72
```
_________ TestRecognize.test_recognize_cpan_manifest_as_plain_package __________
[gw1] linux2 -- Python 2.7.13 /home/travis/build/nexB/scancode-toolkit/bin/python2.7
self = <test_recognize.TestRecognize testMethod=test_recognize_cpan_manifest_as_plain_package>
    def test_recognize_cpan_manifest_as_plain_package(self):
        test_file = self.get_test_loc('cpan/MANIFEST')
        package = recognize_package(test_file)
>       assert isinstance(package, packagedcode.models.CpanModule)
E       AssertionError: assert False
E        +  where False = isinstance(None, <class 'packagedcode.models.CpanModule'>)
E        +    where <class 'packagedcode.models.CpanModule'> = <module 'packagedcode.models' from '/home/travis/build/nexB/scancode-toolkit/src/packagedcode/models.py'>.CpanModule
E        +      where <module 'packagedcode.models' from '/home/travis/build/nexB/scancode-toolkit/src/packagedcode/models.py'> = packagedcode.models
tests/packagedcode/test_recognize.py:91: AssertionError
```

Issue #1088, opened 2018-06-01
Nugets not reported as packages (2.9.2)


tomeks666:
I used this:
extractcode c:\src\solution\main
then
scancode -l -p -c -i -n 2 --verbose c:\src\solution\main --json-pp solution.json

It took 8 hours to scan the folder. Unfortunately only zip,cab ans msi are detected as packages. Nugets are not and they are the main priority.

Is there something needed to be done like loading some additional module?

pombredanne:
@tomeks666 thank you for the report. Which was not detected?
A Nuget manifest will be detected if this is a "*.nuspec" file. This can surely be improved.
Is there some details on your code that you can share such to have a reproducible test case?

tomeks666:
I just ran scancode directly on a single package folder. Nuspec file is there.
scancode -l -p -c -i -n 2 --verbose  "C:\src\MeLEDES\main\packages\Castle.Core.4.2.1" --json-pp castlenuget.json
The result is attached. 

[castlereport.zip](https://github.com/nexB/scancode-toolkit/files/2063132/castlereport.zip)

pombredanne:
thanks!

pombredanne:
@tomeks666 do you mind testing again with the latest `develop` branch? this should be fixed.

On another note, I find a scan that runs for 8 hours a tad suspicious:
- is this a very large codebase? or a slow machine? or both?
- was there any timeout errors on some files, and if yes, that would be great if you can point me to these files assuming they would be public/open source or provide some indication of their type and size.

Thank you.

tomeks666:
The number of files is large like 30000+. I did not find any category I could exclude to make significant cut in this number. The file inventory alone takes many minutes. During scanning it makes around 1 file per second. The machine is not very fast - a laptop with i5-4300U 1.9/2.5 GHz with an SSD drive. The disk has full encryption and there is some corporate DLP/security services. But they do not seem to have a visible impact on other operations like solution rebuild. dir /s > dir.txt takes only 13 seconds. So maybe there is some other reason e.g. something wrong with python installation. Even when using --from-json option the file inventory takes minutes. On other hand importing to SQLite database in AboutCode-Manager.exe is normal - less that one minute.

pombredanne:
@tomeks666 sorry for the late reply. Some comments:
[...]

PR #1090, opened 2018-06-01
Detect Nuget .nuspec manifests properly #1088

pombredanne:
* this is a fix for #1088
 * use xmltodict to simplify XML handling. Remove xml utils
 * add repository tags
 * update and add new tests

Reported-by: @tomeks666


pombredanne:
The test failure has been fixed in develop and is not related to this PR. Merging

tomeks666:
I tested the PR and the nuget packages got detected as packages. Thank you.

One thing I could suggest that nuget packages have a license information detected. I can see detected licenses for files in the package, but not for the package as a whole. It would be great to be able to run a simple report with a list of packages used with licenses and files not being a part of packages that have some licenses associated (like libraries manually added to the solution).

pombredanne:
@tomeks666 good point and this is something that's in the plan.
There are actually several levels to this:

1. for package manifests, this is means running the license detection on the declared licensing which is in the case of nugets is rather poor and mostly a single URL... So part of doing a proper job will be to collect a list of every nuget `licenseUrl` values and possibly create a mapping to actual licenses/license expressions.

2. the licenses are detected alright for files beyond a manifest such as a .nuspec

3. as part of #377, the goal is to apply some smarts on top of these two facts such that top level, declared licensing found in a manifest and actual file-level licenses can be compared and we can verify if they match or not. And use this to infer the overall licensing of a package and what review of this licensing is needed. (based on consistency of information and coverage as the number of files with licensing info)

Beyond that, and in the context of .Net, there are also the cases of bare DLLs that may not be part of a nuget and that may be binary-only: these may also have structure metadata that can be detected with the code in https://github.com/nexB/scancode-toolkit/blob/35d0adb081bf46e8e2a97de35589ea7aaa4fb0e6/src/packagedcode/win_pe.py but that has not been wired in yet.

Issue #1070, opened 2018-05-18
corrupt license cache after running scancode -l


selmf:
Operating system: Arch Linux x64
Python version: Python 3.6.5,  2.7.15
Scancode version: 2.9.2 and latest checkout as of today

1) Run scancode -l on a code base.
2) Run scancode -l a second time on the same codebase

Expected behavior:

The second scan should run just as well as the first run.

Experienced behavior:

Scan fails with the following error message:

``Setup plugins...
ERROR: failed to setup plugin: scan:licenses:
Traceback (most recent call last):
  File "/home/herr_k/builds/scancode-toolkit/src/scancode/cli.py", line 568, in scancode
    plugin.setup(**kwargs)
  File "/home/herr_k/builds/scancode-toolkit/src/scancode/plugin_license.py", line 127, in setup
    return_value=False)
  File "/home/herr_k/builds/scancode-toolkit/src/licensedcode/cache.py", line 64, in get_index
    _LICENSES_BY_KEY_INDEX = get_cached_index(cache_dir, check_consistency)
  File "/home/herr_k/builds/scancode-toolkit/src/licensedcode/cache.py", line 175, in get_cached_index
    return load_index(cache_file)
  File "/home/herr_k/builds/scancode-toolkit/src/licensedcode/cache.py", line 229, in load_index
    return LicenseIndex.loads(ifc.read())
  File "/home/herr_k/builds/scancode-toolkit/src/licensedcode/index.py", line 782, in loads
    idx = cPickle.loads(saved)
ValueError: binary data truncated (2)
``

Deleting ~/.cache/scancode-tk temporarily solves this issue until the cache is regenerated and it fails again.

Please note that Arch Linux is one of the few distros shipping Python3 by default, i.e. /usr/bin/python points to Python3 instead of Python2.

pombredanne:
@selmf Thank you for the report. This definitely should not happen.

Do you mind to try the things below and report back here?

1. cd to your scancode install directory `/home/herr_k/builds/scancode-toolkit/`
2. run `source bin/actiavte`
3. run `python` and paste the results here.


Note that it is possible that this is hitting this pending bug  https://github.com/WojciechMula/pyahocorasick/issues/50 in pyahocorasick

Another thing to test specifically for this would be this:

1. cd to your scancode install directory `/home/herr_k/builds/scancode-toolkit/`
2. run `source bin/actiavte`
3. run `pip install --upgrade pyahocorasick`
[...]

PR #1096, opened 2018-06-05
warn if the license cache is corrupted

selmf:
Display a warning if loading the license cache fails. Tell the user to delete the cache file and retry, file a bug report if the problem persists. Reraise the original python exception as discussed in #1070 to avoid error masking.

Signed-off-by: Felix Kauselmann <licorn@gmail.com>

### DIFF
--- a/src/licensedcode/cache.py
+++ b/src/licensedcode/cache.py
@@ -226,15 +226,15 @@ def load_index(cache_file):
     from licensedcode.index import LicenseIndex
     with open(cache_file, 'rb') as ifc:
         # Note: weird but read() + loads() is much (twice++???) faster than load()
-	try:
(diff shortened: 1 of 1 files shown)


pombredanne:
@selmf Thanks ++ for this and following up from our discussion on #1070 !I will merge shortly... do you mind moving the click import inside the except block?

pombredanne:
Thanks for the update!
Also can you use four spaces indent and not tabs? that's the Python standard.

selmf:
Sorry about that. Usually my IDE does that for me, but this time I used a plain old text editor …

pombredanne:
@selmf Thanks ++ for all this. I pushed a few minor changes on top of your PR and added a test.

Issue #1039, opened 2018-04-20
Add license detection rule for "common cure rights commitment"


pombredanne:
The exact text or text variants are TBD. 
@richardfontana if you have some pointers of usage in the wild (beyond the actual press releases and web announcements) and the actual reference text(s) that would be quite useful.

pombredanne:
Some pointers:
- https://www.redhat.com/en/about/press-releases/technology-industry-leaders-join-forces-increase-predictability-open-source-licensing
- https://www.zdnet.com/article/big-four-linux-tech-firms-shift-open-source-licensing-policies/
- https://www.fsf.org/blogs/licensing/red-hat-leads-coalition-supporting-key-part-of-principles-of-community-oriented-gpl-enforcement

richardfontana:
@pombredanne belated reply: Here are a couple of usages in project repositories:
https://github.com/wildfly/wildfly/blob/master/COMMITMENT
http://git.gluster.org/cgit/glusterfs.git/tree/COMMITMENT

pombredanne:
@richardfontana See also this https://github.com/gplcc/gplcc ... I guess this is mostly the same but there is also an individual "variant" ?

pombredanne:
So what would be the license key: ccrc-1.0  or common-cure-1.0 ?

pombredanne:
This was first added as a `ccrc-1.0` and later updated to `gplcc-1.0` as parts of #1242 to sync with SPDX.
@richardfontana Closing this now... this is available in the latest develop for now.
Thanks!

PR #1100, opened 2018-06-07
Add ccrc license #1039

pombredanne:
reported by @richardfontana


pombredanne:
@richardfontana if you could review this PR that would be great. It adds a new "license" record as well as several rules to also detect with a high score the various variants seen in the wild.

pombredanne:
@richardfontana I guess you do not have comments. Merging now. We can always update this later as needed!

Issue #1129, opened 2018-07-15
scancode does not start on windows exec


jeffmcaffer:
I exec'ing "c:\install\scancode-toolkit-2.9.2\scancode" from node fails looking for  "c:\install\scancode-toolkit-2.9.2\bin\scancode" (note the "bin" segment. Seems that the bin folder is actually a symlink to Scripts. Using normal windows tools (i.e., drag and drop) to unzip the Release zip, you get a windows Junction for bin. That is fine in the UI and from the Command shell but when exec'ing from Node, it seems like the junction is not being followed (or is not available). I don't know enough about the internal workings of node and `exec` to know why this would be different.

Since the "Scripts" folder seems to include all the executable stuff, can it just called `bin` rather than using a symlink? 

as a workaround right now I just deleted the bin link, and copied Scripts to bin. Not sure if there has to be something called Scripts. (i.e., could I have just renamed?)

@pombredanne FWIW, this is for ClearlyDefined.

pombredanne:
@jeffmcaffer I do not know if node honours Windows junctions or not: it could be also that node may be less tested on Windows than POSIX

Now the reason why ScanCode generates a junction on first use (e.g. on the initial `configure`) is that the Python machinery that creates the `bin` and `Scripts` (e.g. `setuptools`) actually does generate a `bin` on POSIX and a `Scripts` dir on Windows. As a convenience, having a junction helps have the same paths referenced in help and documentation on every OS: e.g. `bin`.

These (the Scripts and the bin/junctions) are created either:
- on the first run of a new installation
- when calling the root `configure.bat` script (and this can be a combo of `configure --clean` then `configure` )

Can you point me to where in ClearlyDefined this happens?
Unless there is a bug in node I cannot fathom what happens exactly here.

Was this on the first run? or on other runs?
Was the 2.9.2 of scan a fresh extract in a clean directory? or was it overwritten on top of a previous installation?

jeffmcaffer:
Thanks @pombredanne. this is all runs of a freshly installed 2.9.2 (was also happening on 2.9.1). This happens in ClearlyDefined if you [setup the crawler](https://github.com/clearlydefined/crawler/blob/master/README.md). If you unzip scancode in c:\installs (default) and run, then queue a request by POSTing something like 

```
[...]

PR #1139, opened 2018-07-22
Use Scripts dir rather than bin junction #1129

pombredanne:
This has likely little or no impact but is harmless
It may help with #1129


pombredanne:
This does not harm. Merging.

Issue #979, opened 2018-03-11
Improve reporting of the importance of detected licenses 


pombredanne:
### Context

Today we report detected license with a score, length, relevance and coverage.
As documented:

        The score is an indication of the confidence that a match is
        good. It is computed from the number of matched tokens, the
        number of query tokens in the matched range (including
        unknowns and unmatched) and the matched rule relevance.

The definition of the various elements used to compute the score are:

- rule relevance: what is the relevance of a match to this rule text?
  a float between 0 and 100 where 100 means highly relevant and 0 means not relevant at all. For instance a match to the "gpl" or the "cpol" words have a fairly low relevance as they are a weak indication of an actual license and could be a false positive. In some cases, this may even be used to discard obvious false positive matches automatically. This is a value that can be either set manually on a given license detection rule or is compute if not set. In this case this is based on the length (as the number of words) of the rule: a rule with over 20 words is treated as 100% relevant and a rule with a single word as a relevance of 5%. Each additional word adds 5% of relevance up to 100%: 10 words means 50% relevance

- match magnitude: how big is the match from its starting to its ending positions, considering that the matched ranged may contain matched words, as well as unmatched known words or unknown words.

- query coverage: how much of the query/match side has been matched to a given rule text defined as the number of matched words divided by the match magnitude 

- rule coverage: how much of the rule side has been matched as the number of matched words divided by the length of matched rule (in words)

The score is finally computed with this formula:

- if the query and rule coverage are both 100%, the score is `rule coverage * relevance`
- otherwise, the score is `query coverage * rule coverage * relevance`

The gist of this formula is that when all words in a rule are matched without any words added or removed, the score is 100% or lower based on the rule relevance.

Otherwise, the score will degrade also based on the addition or removal of words wrt. the matched rule and its relevance.  

### Problem

The score has really two elements: how much of a detection rule was matched and how important/relevant a rule is. The matched rule relevance part is mostly computed and may in some case return weak scores for short but quite strong license statements such as `license under the Apache 2.0 license` or `released under the GPL 2.0 or later` that are both shorter than the 20 words threshold for a rule to reach 100% relevance. On the other hand some matches of the same length can indeed be a weak indication of licensing and should therefore receive a low score alright. This problem is amplified the shorter a rule gets: for instance `license: gplv2` only has two words but is eventually a strong license indication. The same applies to structured licenses expressions found in package manifests or `SPDX-License-Identifier`

### Solution

[...]

PR #1140, opened 2018-07-23
Add license rule importance flags

pombredanne:
This PR add a new set of flags on license rule to state the importance of a rule for #979. A rule is more important if its detection is the clearer mark of a license.
The importance flags are:
- `is_license_text`: this is a full license text
- `is_license_notice`: this is a license notice
- `is_license_reference`: this is a only a reference to a license and therefore may not be as strong
- `is_license_tag`: this is a a structured, coded license reference such as a SPDX-License-Identifier and is typically a strong license reference albeit a short one.
Licenses are also tagged using a first set of heuristics and then a review.

License importance is useful for improve summarization and license determination automation for #377 


pombredanne:
This is also renaming rule flaqs `negative` to `is_negative` and `false_positive` to `is_false_positive`
All tests passing ... merging now.

Issue #1141, opened 2018-07-23
ERROR: failed to run post-scan plugin: summary


tdruez:
Running a Scan on https://github.com/tytso/e2fsprogs/archive/v1.44.2.zip 
Using https://github.com/nexB/scancode-toolkit/archive/fd8e0750d5c293d29a540d20496c66e9fcd7aee9.tar.gz
```
Extracting archives...
Extracting done.
Setup plugins...
Collect file inventory...
Scan files for: info, licenses, copyrights, packages with 3 process(es)...
ERROR: failed to run post-scan plugin: summary:
Traceback (most recent call last):
  File "/opt/scancode-toolkit/src/scancode/cli.py", line 821, in run_plugins
    plugin.process_codebase(codebase, **kwargs)
  File "/opt/scancode-toolkit/src/summarycode/summarizer.py", line 129, in process_codebase
    summarize_codebase(codebase, keep_details=False, **kwargs)
  File "/opt/scancode-toolkit/src/summarycode/summarizer.py", line 185, in summarize_codebase
    _summary_data = summarizer(resource, children, keep_details=keep_details)
  File "/opt/scancode-toolkit/src/summarycode/summarizer.py", line 220, in license_summarizer
    for child_summary in child_summaries:
TypeError: 'NoneType' object is not iterable

Scanning done.
Some files failed to scan properly:
ERROR: failed to run post-scan plugin: summary:
Traceback (most recent call last):
  File "/opt/scancode-toolkit/src/scancode/cli.py", line 821, in run_plugins
    plugin.process_codebase(codebase, **kwargs)
  File "/opt/scancode-toolkit/src/summarycode/summarizer.py", line 129, in process_codebase
    summarize_codebase(codebase, keep_details=False, **kwargs)
  File "/opt/scancode-toolkit/src/summarycode/summarizer.py", line 185, in summarize_codebase
    _summary_data = summarizer(resource, children, keep_details=keep_details)
  File "/opt/scancode-toolkit/src/summarycode/summarizer.py", line 220, in license_summarizer
    for child_summary in child_summaries:
TypeError: 'NoneType' object is not iterable

Path: tmp5rzkvu1q/e2fsprogs-1.44.2.zip-extract/e2fsprogs-1.44.2/tests/i_qcow/image1024.orig.bz2-extract/image1024.orig.bz2-extract
[...]

PR #1143, opened 2018-07-25
Do not crash when computing a summary

pombredanne:
This is a fix for #1141  

Also improve copyright candidate selection: now using more selective hints.


pombredanne:
Looking good... merging!

PR #1156, opened 2018-08-20
Use attrs in packages

pombredanne:
This PR removes schematics and replaces this with attr instead.


majurg:
One step closer to python3?

pombredanne:
There is only one minor test failure that should be already fixed in develop. Merging this now.

PR #1160, opened 2018-08-29
Add openssl SPDX identifier

GiasemiSh:
OpenSSL SPDX identifier is listed in the SPDX License List: https://spdx.org/licenses/

### DIFF
--- a/src/licensedcode/data/licenses/openssl.yml
+++ b/src/licensedcode/data/licenses/openssl.yml
@@ -5,9 +5,8 @@ category: Permissive
 owner: OpenSSL
 homepage_url: http://openssl.org/source/license.html
 notes: |-
-    This is the OpenSSL license proper, without the SSLEay part. The SPDX
-    OpenSSL identifier does not apply here. Instead it matches the openssl-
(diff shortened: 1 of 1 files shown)


pombredanne:
@GiasemiSh I see you submitted then closed this... I reckon that the fact that the license key "openssl" does not match the "openssl" SPDX identifier may feel weird. This is historical: ScanCode licenses keys existed before SPDX.
- in ScanCode we track four openssl-related licenses: 
  - `openssl`  which is strictly the openssl part, 
  - `ssleay-windows`which is Eric Young bottom text with the windows-specific wording
  - `ssleay` which is Eric Young bottom text without the windows-specific wording, 
  - the combo   `openssl-ssleay` which is the full text provided as a tope level license by OpenSSL and is a combo of ScanCode's `openssl` and `ssleay-windows`. These licenses are used alone in some files hence why we track them. 
- SPDX only tracks the combo `openssl-ssleay` and calls it OpenSSL.

Also we try to avoid changing the keys in ScanCode: I hope this makes sense.

PR #1161, opened 2018-09-03
Add full name to license data in output

MartinPetkov:
As the title says, add the full name to the license data output. It would be useful for any applications that use scancode's output for reporting, or want to filter or match on the full name when doing post-processing.

The CHANGELOG.rst doesn't seem to have been updated in a while, but I added a line anyway. I can remove it if it's not necessary.

Signed-off-by: Martin Petkov <martin.p.petkov@gmail.com>


pombredanne:
@MartinPetkov Thank you ++ for this.

re

> The CHANGELOG.rst doesn't seem to have been updated in a while, but I added a line anyway. I can remove it if it's not necessary.

this is good, this is overdue and I reckon that we should update it as we go rather than just before a release...

Looking good... merging

Issue #1173, opened 2018-09-13
Scan results for ImageMagick-7.0.8-11.tar.gz are misleading


pombredanne:
from @DennisClark 
I initiated a Scan of  ImageMagick-7.0.8-11.tar.gz [with license and summary options]
https://github.com/ImageMagick/ImageMagick/archive/7.0.8-11.tar.gz

and it returned the scan results that are attached to this issue.  [ImageMagick-7.0.8-11.tar.gz_scan.json.zip](https://github.com/nexB/scancode-toolkit/files/2378451/ImageMagick-7.0.8-11.tar.gz_scan.json.zip) 

You'll notice that for Licenses it returns: 
mit: 437
apache-2.0: 77
gpl-2.0: 48
etc.

These results make it appear that the overall, or primary, license for ImageMagick is "mit"; however, if you take a look at the project README.md at 
https://raw.githubusercontent.com/ImageMagick/ImageMagick/7.0.8-11/README.md
you find the paragraph: 

```
ImageMagick is free software delivered as a ready-to-run binary distribution or as source code that you may use, copy, modify, and distribute in both open and proprietary applications. It is distributed under the Apache 2.0 [license](https://www.imagemagick.org/script/license.php).
```

So we have a classic case here of an prominent project file, in this case a README.md, providing an overall license of apache-2.0, but obviously the project includes many third-party subcomponents licensed under mit.   

It would be really nice if somehow ScanCode could detect the apache-2.0 notice in the README.md file, and because of the prominence of that file in the package, tell us that the primary license is indeed apache-2.0, even though the number of occurrences of an apache-2.0 notice in the entire project might be less than some other license.

pombredanne:
@DennisClark two things here:
1. the mit detection is a bug
2. the license is very close to an `apache-2.0` but  this is an `imagemagick` license (a modified apache license) which is also referenced at SPDX.

pombredanne:
This is fixed in develop now.

pombredanne:
@DennisClark FYI this is waht we get now as summaries:
```
  "scancode_version": "2.9.2.post329.9ac4f64d2.dirty.20180912123033",
  "scancode_options": {
    "input": "ImageMagick-7.0.8-11/",
    "--classify": true,
    "--info": true,
    "--json-pp": "im.json",
    "--license": true,
    "--processes": "6",
    "--summary": true,
    "--summary-key-files": true
  },
  "scan_start": "2018-09-13T185032.148740",
[...]

Issue #1171, opened 2018-09-12
incorrect license detection  for OpenSSL


pombredanne:
as reported privately by @DennisClark :

A scan of the files at https://github.com/openssl/openssl/archive/OpenSSL_1_1_1.tar.gz found a majority of `openssl` licenses and not a majority of `openssl-ssleay` licenses. The top level license of the project (https://github.com/openssl/openssl/blob/OpenSSL_1_1_1/LICENSE) clearly states that the project is under the combined license, which is `openssl-ssleay`.

pombredanne:
This is fixed with the commits in the #1172 PR

pombredanne:
This is fixed in develop

PR #1174, opened 2018-09-13
Add new license rules

pombredanne:
These new license rules are for #1171 and #1173 . There are also a few new rules for rarer BCM licenses


pombredanne:
All tests passing. Merging!

PR #1177, opened 2018-09-14
Return correct line numbers for long lines #1176

pombredanne:
* Long lines were broken in chunks without keeping the origianl line
   number. This fixes it.


pombredanne:
All tests passing... merging now.

Issue #1199, opened 2018-10-04
Multiple outputs with --html crashes with "TypeError: Object of type License is not JSON serializable"


MartinPetkov:
To reproduce:

```bash
$ git checkout v2.9.3
$ ./configure
$ source ./bin/activate
(scancode-toolkit) $ cd /tmp/
(scancode-toolkit) $ git clone https://github.com/madler/zlib && cd zlib/
(scancode-toolkit) $ scancode --verbose --license --json-pp ./scancode_results.json --html ./scancode_results.html .
```

This results in:

```bash
...
Save scan results...
 Save scan results as: html...
 Save scan results as: json-pp...
ERROR: failed to run output plugin: json-pp:
Traceback (most recent call last):
  File "/usr/local/google/home/mpetkov/development/scancode-dev/scancode-toolkit/src/scancode/cli.py", line 809, in run_plugins
    plugin.process_codebase(codebase, **kwargs)
  File "/usr/local/google/home/mpetkov/development/scancode-dev/scancode-toolkit/src/formattedcode/output_json.py", line 99, in process_codebase
    pretty=True)
  File "/usr/local/google/home/mpetkov/development/scancode-dev/scancode-toolkit/src/formattedcode/output_json.py", line 140, in write_json
    output_file.write(simplejson.dumps(scan, **kwargs))
  File "/usr/local/google/home/mpetkov/development/scancode-dev/scancode-toolkit/local/lib/python2.7/site-packages/simplejson/__init__.py", line 399, in dumps
    **kw).encode(obj)
  File "/usr/local/google/home/mpetkov/development/scancode-dev/scancode-toolkit/local/lib/python2.7/site-packages/simplejson/encoder.py", line 298, in encode
    chunks = list(chunks)
  File "/usr/local/google/home/mpetkov/development/scancode-dev/scancode-toolkit/local/lib/python2.7/site-packages/simplejson/encoder.py", line 696, in _iterencode
    for chunk in _iterencode_dict(o, _current_indent_level):
  File "/usr/local/google/home/mpetkov/development/scancode-dev/scancode-toolkit/local/lib/python2.7/site-packages/simplejson/encoder.py", line 652, in _iterencode_dict
    for chunk in chunks:
  File "/usr/local/google/home/mpetkov/development/scancode-dev/scancode-toolkit/local/lib/python2.7/site-packages/simplejson/encoder.py", line 708, in _iterencode
    for chunk in _iterencode_list(o, _current_indent_level):
[...]

PR #1200, opened 2018-10-08
Ensure that multiple outputs work with a license scan

pombredanne:
This is a fix for #1199
The key issue was that a data structure shared between scan output plugin was mutated by one of the plugins leading to some failures.

This improves the situation but does go (yet) as far as making a Resource object immutable.


pombredanne:
This is ready for merging now.

Issue #1194, opened 2018-09-28
Reporting only real programming languages as such


pombredanne:
We use the Pygments libraries to detect lexable text and report it as the programming language. But Pygments can lex many other types beyond programming langauges (e.g. INI files, etc).
We should instead:
1. return the Pygments lexer name as a new, extra filetype attribute
2. use a stricter, subset list of such types as real, actual programming languages AND confront that with the actual file extensions for improved accuracy.

pombredanne:
See also #778 #1109 #605

pombredanne:
This has been implemented mostly now... with #1201 merged and released.

PR #1201, opened 2018-10-08
Simplify programming language reporting

pombredanne:
This PR deals with  #1194 to only report real programming languages.


pombredanne:
This can be a tad controversial as it changes which programming languages and what file types are returned. Nevertheless this is a much welcomed simplification in practice as this was a rather annoying bug-like data issue until now: merging

Issue #1207, opened 2018-10-12
Allow "transient" scan data to be stored in Resource.extra_data


pombredanne:
Today scan results must be stored internally as returned attributes of the Resource class.
In some cases such as the "package_manifest" we do not want to keep these (see #1191) ... so we would want to store these under the extra_data attribute instead.

pombredanne:
This has been implemented as part of #1212

Issue #1191, opened 2018-09-28
Revert using only "packages" rather than "packages/package_manifest"


pombredanne:
A recent update introduced reporting `packages` and `package_manifest` when scanning for packages. The rationale was that `package_manifest` is where we put information about a single manifest such as a `pom.xml` or a `package.json`. And based on that, the actual package is often the whole parent directory of when the manifest is stored. And a given directory many have multiple package manifests.

While this is generally correct, this is rather verbose and I would like to find a way to avoid reporting the exact same data at both levels.

pombredanne:
In addition to report properly a single "packages" attribute we would need to still:
- report the package data at the right  level (as introduced by #1027)
- also add a new "manifest_path" attribute that point to the relative path of the package manifest of a package (which would be empty there is no distinction such as with an RPM)

pombredanne:
In order to deal with this cleanly, I suggest that we modify the API of a scanner plugin such that it can have a "process_codebase" method and that this would be called after the scans are done on the individual files (kinda like a post scan but done before all post scans)

pombredanne:
Done with #1212

PR #1212, opened 2018-10-16
Simplify package manifest data storage

pombredanne:
This PR updates the way package data is stored in the scan results for #1191 
This also contains a fix to allow a scan plugin to have a "process_codebase" method as well as support storing its results as extra data for #1207


src/packagedcode/models.py
```diff
@@ -516,7 +521,7 @@ def get_package_root(cls, manifest_resource, codebase):
         Resource object that represents a manifest in the `codebase` Codebase.
 
         Each package type and instance have different conventions on how a
-        package manifest realtes to the toor of a package.
+        package manifest relattes to the root of a package.
```
majurg: Typo here

pombredanne: Thanks!

pombredanne:
Appveyor seems to be clogged ... so merging without it.

Issue #1206, opened 2018-10-12
Simplify "related_packages"


pombredanne:
The way "related_packages" are defined with three attributes is too verbose in practice. Here is a current example:
```
from_purl: pkg:rpm/libproxy@0.3.0-4.el6_3?arch=src
relationship: source_of
to_purl: pkg:rpm/libproxy-bin@0.3.0-4.el6_3?arch=x86_64
```

The object is defined as:
```
@attr.s()
class PackageRelationship(BaseModel):
    """
    A directed relationship between two packages.

    This consiste of three attributes:
    - The "from" (or subject) package "purl" in the relationship,
    - the "to" (or object) package "purl" in the relationship,
    - and the "relationship" (or predicate) string that specifies the relationship.
    """

    from_purl = String(
        repr=True,
        label='"From" purl package URL in the relationship',
        help='A compact purl package URL.'
    )

    relationship = String(
        repr=True,
        label='Relationship between two packages.',
        help='Relationship between the from and to package '
             'URLs such as "source_of" when a package is the source '
             'code package for another package.'
    )

    to_purl = String(
        repr=True,
        label='"To" purl package URL in the relationship',
        help='A compact purl package URL.'
    )
```

In practice we only are interested by a single type of relationship: a package is the `source` for the current package (and we can only collect this from manifest)
Note: dependencies are handled alright and not concerned by this ticket 

I suggest that we replace this "related_packages" by a "source_packages" attribute that would be a list of Package URLs.

pombredanne:
Implemented in #1213

PR #1213, opened 2018-10-16
Simplify related packages

pombredanne:
This is a fix for #1206
Note that it builds on top of PR #1212


pombredanne:
Appevyor is stuck. Merging without Windows checks

Issue #1189, opened 2018-09-27
Revert packages to use direct attributes for checksums: md5, sha1, sha256, etc


pombredanne:
The Package model introduced a list of download_checksums with d403e9c3a9cca191fd4b2bf980c0cfea523c0af4 to replace md5/.sha1/etc...
This does not make much sense to have a list after all since we have a finite number of such checksums used anywhere. We should revert.

pombredanne:
This has been fixed in develop

Issue #1149, opened 2018-08-01
Add Package URL field to top-level package output


majurg:
Currently, `scancode-toolkit` returns all the needed details to construct a Package URL when running a scan with the `--package` option. However, a `purl` field is not present for top-level packages. 

This should be added, as all the necessary fields are present and we return `purl` values for package dependencies, so we should keep our top-level packages consistent with these dependencies.

pombredanne:
This has been fixed and merged in develop

PR #1215, opened 2018-10-18
Fix package checksums and other refinements

pombredanne:
This is a fix for #1189 and also #1149 and oher minor package data refinements


pombredanne:
All tests passing. Merging.

PR #1221, opened 2018-10-19
Place all third-party packages under thirdparty

JonoYang:
All third-party packages are now under the thirdparty directory. The configure script has been updated to reflect this change.


pombredanne:
Test runs are clogged because of GH webhook issues. But I verified that they pass on all OSes now... so I am merging.
@JonoYang Thanks!

PR #1241, opened 2018-10-31
Fix 'license detection tests' collection on Windows

### LINKED ISSUE #1182: License detection tests do not run on Windows/Appveyor
pytest does not collect them all with "py.test -n 2 -vvs tests/licensedcode --ignore=tests/licensedcode/test_zzzz_cache.py"
See the attached output from a collect-only.

nicoddemus:
The problem was that the `test_file_name` attribute generation was using a naive approach which only worked on Linux, resulting in all test methods having the same name (`test_detection`).

This PR fixes the problem by producing native path names using `os.path` functions:

```
$ py.test -vvs tests/licensedcode/test_detection_datadriven.py --collect-only
...
collecting ... collected 1762 items
```

Fix #1182
Fix pytest-dev/pytest#4006

### DIFF
--- a/tests/licensedcode/test_detection_datadriven.py
+++ b/tests/licensedcode/test_detection_datadriven.py
@@ -87,7 +87,7 @@ class LicenseTest(object):
         self.data_file = data_file
         self.test_file = test_file
         if self.test_file:
-            _, _, self.test_file_name = test_file.partition('licensedcode/data/')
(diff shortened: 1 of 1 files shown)


pombredanne:
@nicoddemus thanks you +++ that extremely gentle of you to chime in with a PR!

nicoddemus:
Sure, my pleasure! 👍

Issue #1242, opened 2018-11-01
Sync scancode-toolkit with the SPDX License List 3.3


DennisClark:
Sync scancode-toolkit with the SPDX License List 3.3.

See https://spdx.org/licenses/ 
and
https://github.com/spdx/license-list-XML/blob/master/RELEASE-NOTES.md

pombredanne:
@DennisClark sure thing. We need to have the JSON data generated though before I can do anything I created https://github.com/spdx/license-list-data/issues/37

pombredanne:
@DennisClark FYI the fix is in #1245 and this contains everything from the SPDX list 3.3

pombredanne:
This has been merged in develop.

PR #1245, opened 2018-11-02
Sync latest licenses

pombredanne:
This adds all the latest SPDX licenses and sync also with DejaCode latest. 
This is a fix for #1242


pombredanne:
There is a mysterious Windows-only extractcode test failure that is tracked with #1250 
All tests are passing otherwise.
Merging!

PR #1251, opened 2018-11-05
Improve generated code detection

pombredanne:
Add new clues, some comming from the OpenJDK


pombredanne:
All green. merging

Issue #1180, opened 2018-09-17
Implement a license clarity score


pombredanne:
For use here and in ClearlyDefined, we should compute a license clarity score. The specification is documented in this [ClearlyLicensed Metrics](https://docs.google.com/document/d/1o_VuVnP1PMoYOXZHSm0TyDX4CsqDXM3zqESAE0AoUnQ/edit#) doc.

pombredanne:
Here is the status so far:
The latest doc link for license scoring https://docs.google.com/document/d/1o_VuVnP1PMoYOXZHSm0TyDX4CsqDXM3zqESAE0AoUnQ/edit#

And I have a working scoring implementation https://github.com/nexB/scancode-toolkit/tree/license-scoring
```
  "license_score": {
    "score": 82,
    "has_top_level_declared_licenses": true,
    "file_level_license_and_copyright_coverage": 0.9,
    "has_consistent_key_and_file_level_license": true,
    "has_all_spdx_licenses": false,
    "has_all_license_texts": true
  },
```

that's using facets:
```
  "scancode_version": "2.9.2.post351.e23b1d79f.dirty.20180916153557",
  "scancode_options": {
    "input": "dnsmasq-2.45",
    "--classify": true,
    "--copyright": true,
    "--facet": [
      "dev=*/bld/*",
      "examples=*/contrib/*",
      "docs=*.html",
      "examples=*.example",
      "data=*/po/*",
      "data=*/dbus/*",
      "docs=*/man/*",
      "docs=CHANGELOG",
      "docs=FAQ"
    ],
    "--info": true,
    "--json-pp": "<stdout>",
    "--license": true,
    "--license-clarity-score": true,
    "--processes": "3",
    "--summary": true,
    "--summary-by-facet": true
  },

```

pombredanne:
This is completed and merged in develop

PR #1253, opened 2018-11-05
Improve summary and score

pombredanne:
This PR is a set of improvements for  #1180 

This add an is_key_file as a file flag and improves the computation.
This also improves generated code detection for #377


pombredanne:
All green. Merging

Issue #1259, opened 2018-11-06
Remove older 32 bits Linux pre-built binaries


pombredanne:
Since #297 we no longer officially support 32 bits Linux.
we should clean things up for 3.0 and drop it entirely

pombredanne:
Fixed and merged in develop

PR #1260, opened 2018-11-06
Remove Linux 32 bits and Linux narrow wheels #1259

pombredanne:
This is a welcome diet since 32 bits is
no longer officially supported


pombredanne:
All green...LGTM.... merging

Issue #1262, opened 2018-11-06
Configuring on Alpine Linux wants to build bitarray from source


sschuberth:
I'm trying to create a minimal Docker image based on Alpine Linux to run ScanCode. In particular, I do not want to build any native packages required by ScanCode from source (as the Docker base image does not contain `gcc`, and I do not want to use Docker multi-stage builds). Thus, I'm not installing ScanCode via `pip`, but from the release archive which comes with pre-built binaries for several platforms.

For some reason, running `./configure etc/conf` when creating the image does not seem to pick up the `bitarray` binaries but wants to build them from source:

    Installing collected packages: bitarray, intbitset, boolean.py, license-expression, pyahocorasick, Beautifulsoup, Beautifulsoup4, webencodings, html5lib, pycryptodome, pdfminer.six, binaryornot, pygments, pefile, lxml, idna, urllib3, requests, pymaven-patch, schematics-patched, packageurl-python, click, colorama, pluggy, attrs, typing, MarkupSafe, jinja2, simplejson, ply, SPARQLWrapper, pyparsing, isodate, rdflib, spdx-tools, unicodecsv, psutil, zc.lockfile, contextlib2, pytz, tempora, jaraco.timing, yg.lockfile, scancode-toolkit
      Running setup.py install for bitarray: started
        Running setup.py install for bitarray: finished with status 'error'
        Complete output from command /usr/local/scancode-toolkit-2.9.2/bin/python2.7 -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-9By3EB/bitarray/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-qQss3M-record/install-record.txt --single-version-externally-managed --compile --install-headers /usr/local/scancode-toolkit-2.9.2/include/site/python2.7/bitarray:
        running install
        running build
        running build_py
        creating build
        creating build/lib.linux-x86_64-2.7
        creating build/lib.linux-x86_64-2.7/bitarray
        copying bitarray/test_bitarray.py -> build/lib.linux-x86_64-2.7/bitarray
        copying bitarray/__init__.py -> build/lib.linux-x86_64-2.7/bitarray
        running build_ext
        building 'bitarray._bitarray' extension
        creating build/temp.linux-x86_64-2.7
        creating build/temp.linux-x86_64-2.7/bitarray
[...]

PR #1263, opened 2018-11-06
Use startswith('linux') to check for the Linux platform

sschuberth:
This is to support #1262

### DIFF
--- a/etc/conf/base.py
+++ b/etc/conf/base.py
@@ -23,7 +23,7 @@ else:
     arch = '32'
 
 sys_platform = str(sys.platform).lower()
-if 'linux' in sys_platform:
+if sys_platform.startswith('linux'):
     os = 'linux'
 elif 'win32' in sys_platform:
     os = 'win'
--- a/etc/configure.py
+++ b/etc/configure.py
@@ -67,7 +67,7 @@ import subprocess
 # platform-specific file base names
 sys_platform = str(sys.platform).lower()
 on_win = False
-if 'linux' in sys_platform:
+if sys_platform.startswith('linux'):
     platform_names = ('posix', 'linux',)
 elif 'win32' in sys_platform:
     platform_names = ('win',)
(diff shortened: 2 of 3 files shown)


pombredanne:
@sschuberth thank you. Do you mind adding you sign-off (either in each of the commit or at least in the body of this PR)?

etc/conf/base.py
```diff
@@ -23,11 +23,11 @@ def unsupported(platform):
     arch = '32'
 
 sys_platform = str(sys.platform).lower()
-if 'linux' in sys_platform:
+if sys_platform.startswith('linux'):
     os = 'linux'
-elif 'win32' in sys_platform:
+elif sys_platform == 'win32':
```
pombredanne: There may be some risk attached to changing the tests here... Do you mind leaving these as is?

etc/configure.py
```diff
@@ -67,12 +67,12 @@
 # platform-specific file base names
 sys_platform = str(sys.platform).lower()
 on_win = False
-if 'linux' in sys_platform:
+if sys_platform.startswith('linux'):
     platform_names = ('posix', 'linux',)
-elif 'win32' in sys_platform:
+elif sys_platform == 'win32':
```
pombredanne: Same as above... please only change the Linux test.

thirdparty/virtualenv.py
```diff
@@ -1052,10 +1052,10 @@ def copy_required_modules(dst_prefix, symlink):
             if f is not None:
                 f.close()
             # special-case custom readline.so on OS X, but not for pypy:
-            if modname == 'readline' and sys.platform == 'darwin' and not (
+            if modname == 'readline' and is_darwin and not (
```
pombredanne: You should not change anything in virtualenv... this is the official/standard code so it should not be modified .

pombredanne:
@sschuberth Thanks++, merging

Issue #1264, opened 2018-11-06
Duplicate spdx_license_keys


furuholm:
I have found two duplicated SPDX license keys in the license metadata:

### Nokia-Qt-exception-1.1
[nokia-qt-exception-1.1.yml](https://github.com/nexB/scancode-toolkit/blob/6ca3b5e8a74085f9390e7086b7e2a109b8afee52/src/licensedcode/data/licenses/nokia-qt-exception-1.1.yml) and [qt-company-exception-lgpl-2.1.yml](https://github.com/nexB/scancode-toolkit/blob/6ca3b5e8a74085f9390e7086b7e2a109b8afee52/src/licensedcode/data/licenses/qt-company-exception-lgpl-2.1.yml) both specifies spdx_license_key: `Nokia-Qt-exception-1.1`

Since spdx includes [`Qt-LGPL-exception-1.1`](
https://spdx.org/licenses/Qt-LGPL-exception-1.1.html) my guess is that simply changing the spdx_license_key to `Qt-LGPL-exception-1.1` should solve the problem for this one.

### GPL-2.0-with-bison-exception
[gpl-2.0-bison.yml](https://github.com/nexB/scancode-toolkit/blob/6ca3b5e8a74085f9390e7086b7e2a109b8afee52/src/licensedcode/data/licenses/gpl-2.0-bison.yml) and [gpl-2.0-bison-2.2.yml](
https://github.com/nexB/scancode-toolkit/blob/6ca3b5e8a74085f9390e7086b7e2a109b8afee52/src/licensedcode/data/licenses/gpl-2.0-bison-2.2.yml) both specifies spdx_license_key: `GPL-2.0-with-bison-exception`

The notes of [gpl-2.0-bison.yml](https://github.com/nexB/scancode-toolkit/blob/6ca3b5e8a74085f9390e7086b7e2a109b8afee52/src/licensedcode/data/licenses/gpl-2.0-bison.yml) says `replaced by bison-exception-2.0`, which is not an SPDX License exception. My guess is therefore that `spdx_license_key` should be removed from [gpl-2.0-bison.yml](https://github.com/nexB/scancode-toolkit/blob/6ca3b5e8a74085f9390e7086b7e2a109b8afee52/src/licensedcode/data/licenses/gpl-2.0-bison.yml)

Will create a PR with the suggested fixes.

furuholm:
[...]

PR #1265, opened 2018-11-06
Fix incorrect spdx_license_keys

furuholm:
Proposed fix for #1264


pombredanne:
@furuholm thank you++
Do you mind adding your "Signed-off-by" either in the main text of this PR or in each of the commit messages? See https://github.com/nexB/scancode-toolkit/blob/develop/CONTRIBUTING.rst#contributing

Also I will be checking why your fix makes several SPDX-related tests fail when the tests server are done,

furuholm:
@pombredanne I have added the signature.

pombredanne:
@furuholm thanks. I am pushing a few extra fixes:  this bug helped me fixed some areas that were less tested and found a few other bugs along the way! I also squashed your two commits in one. ... and I rebased your branch on the latest develop since you were a tad behind.

With all these and even though you ought to only load non-deprecated licenses, I have added an extra consistency test to ensure that SPDX ids are use uniquely even if a license is deprecated (this helps to keep things clean and consistent... and the data is cleaner (there was an another QT license that was a dupe). 

I wished QT/Digia/Nokia were not changing their licenses each time they change name and were more consistent... but at least we have a mostly clean slate for now thanks to you!

PR #1284, opened 2018-11-14
Base native plugins for Alpine linux

pombredanne:
This is to support  #1262 
For now these contain placeholder empty .so/.exe


sschuberth:
Hmm. Creating an `alpine` sub-directory below `plugins` does not seem to match the current conventions. I would have expected this to create directories like

    extractcode-7z-alpinelinux1_x86_64
    extractcode-libarchive-alpinelinux1_x86_64
    typecode-libmagic-alpinelinux1_x86_64

pombredanne:
@sschuberth since there is no support for now for linux distros in Wheels tags, this is a workaround: we will still be able to get pre-built for Alpine but they must use a name such as the one I use (note that this directory name does not matter much... what matter is the name inside setup.py AND the platform `plat-name` arg inside here https://github.com/nexB/scancode-toolkit/blob/88127230c2d12d8ede3f8e391e4b1157cf43ffb2/plugins/alpine/extractcode-7z-linux_x86_64/setup.cfg#L5 (which is wrong btw and should be instead `linux_x86_64` since this is the tag for Alpine based on your comments)

In anycase, if and when distro will be supported in `pip` the tag will most likely be in the style of: `linux_x86_64_alpine` ... and not `alpinelinux1_x86_64` as the distro would likely come last.

sschuberth:
Ok.

So by now it seems we have a bunch of not-strictly-Alpine-Linux-related cleanup commits in here. I don't have the background to judge all changes in-depth, but they look reasonable to me.

majurg:
@pombredanne Ping. Not sure if this is ready or not.

pombredanne:
Closing this for now.
1. this is stale
2. the whole set of native plugins would nee to be built from sources to tun on Alpine as there are dependencies on libc in these. https://github.com/nexB/scancode-plugins should help though

Issue #840, opened 2017-11-09
--only-findings should not omit scan_errors from JSON output


sschuberth:
When running ScanCode 2.2.1 with `--only-findings` and a scan error like exceeding a given `--timeout` occurs, that error is not written to the JSON output. Instead, the file is treated as if no findings occurred and is completely omitted from the output due to `--only-findings`.

I believe scan errors should be included in the JSON output despite `--only-findings` being specified because if there are errors we in fact do not know for sure if there were findings or not, hence it's unsafe to exclude those files when `--only-findings` is given.

sschuberth:
I believe having both would be beneficial: We need the per-file `scan_errors` so we know for *which* files errors occurred, but also having a top-level `error_summary` (?) which basically just is a de-duplicated set of all `scan_errors` for all files would be nice to have to get a quick overview of what *type* of errors occurred. For example, if that `error_summary` only contains `"ERROR: Processing interrupted: timeout after X seconds.` we might chose to ignore these errors since we're in control of `X` and have chosen its value as a compromise between scanning speed and scanning completeness.

pombredanne:
Excellent point again :+1:

sschuberth:
The main issue has been fixed via https://github.com/nexB/scancode-toolkit/pull/848, thanks @haikoschol! The additional idea that came up of having an `error_summary` (?) should be dealt with as part of #211. So I'm closing this.

PR #1285, opened 2018-11-14
Add new "headers" top level attribute

### LINKED ISSUE #211: Provide header level details for a scan: Enhance scancode to include a log or history with useful statistics
A run of scancode should generate a log file with meaningful statistics, including such things as: 
- the version number of scancode that was executed
- start date/time
- end date/time
- elapsed time
- name of the library that was scanned
- number of files scanned
- {{anything else that makes sense and is useful}}

pombredanne:
* This is a new data structure as designed in  https://github.com/nexB/aboutcode/issues/7
* This would fix #211 and address the errors summary discussed in #840
* For now, the old header-level data have been kept.
* The top level attribute is named `headers` as discussed https://github.com/nexB/aboutcode/issues/7 ?

Some questions:
- old header-level attributes such as `scancode_version` :  should we remove them entirely from the 3.0 output or deprecate them in doc and remove them only later?

@mjherzog @DennisClark @sschuberth @JonoYang @majurg @mnonnenmacher feedback welcomed!


pombredanne:
Note that this is the last feature-like ticket of the V3 milestone https://github.com/nexB/scancode-toolkit/milestone/8
After this only doc is left todo

sschuberth:
Please keep `scancode_version` (probably just renamed to "version" in new structure). We use that to determine whether scan results are directly comparable (different versions of ScanCode might detect different things).

And to be frank, I believe `history_log` to be a bad choice for a name. Neither "history" nor "log" (which also are somewhat redundant) capture the purpose properly, IMO. I'd probably generically call it `info`.

pombredanne:
@sschuberth re

> Please keep `scancode_version` (probably just renamed to "version" in new structure). We use that to determine whether scan results are directly comparable (different versions of ScanCode might detect different things).

This is kept and renamed to `tool_version` in the new structure such as here: https://github.com/nexB/scancode-toolkit/pull/1285/files#diff-a8e64273dc1cb3ca2a1ff8127a989c6cR136

Are you saying that you always need a top level `scancode_version`? or that you can get to the new version in `history_log.0.tool_version` alright?

> And to be frank, I believe `history_log` to be a bad choice for a name. Neither "history" nor "log" (which also are somewhat redundant) capture the purpose properly, IMO. I'd probably generically call it `info`.

Yes! `history_log` is a bad choice! 
... let's try something else: `info` is a candidate, this is a kind of "page info" that is offered in a browser, `header` is the name in the original ABC ticket.... let's find a few more candidate!

pombredanne:
let's move the top level name discussion to https://github.com/nexB/aboutcode/issues/7

sschuberth:
> Are you saying that you always need a top level `scancode_version`?

No, that's not a requirement for us, but we need to generally know from looking at the result file by which version of ScanCode it was created.

pombredanne:
@sschuberth re

> we need to generally know from looking at the result file by which version of ScanCode it was created.

that's going to be always available

pombredanne:
@sschuberth I renamed this attribute to `headers` alright.

pombredanne:
This looks all green now. Merging

Issue #1292, opened 2018-11-21
something wrong with output_html.py


arthur657834:
run command:  /opt/scancode-toolkit/scancode --license --copyright --package --html=scancode-2018-11-21.html --verbose ./test-master

and i get this:

16:04:13 Scanning done.
16:04:13 Some files failed to scan properly:
16:04:13 ERROR: failed to run output plugin: html:
16:04:13 Traceback (most recent call last):
16:04:13   File "/opt/scancode-toolkit/src/scancode/cli.py", line 864, in run_codebase_plugins
16:04:13     plugin.process_codebase(codebase, **kwargs)
16:04:13   File "/opt/scancode-toolkit/src/formattedcode/output_html.py", line 86, in process_codebase
16:04:13     write_templated(html, results, version, template_or_format='html')
16:04:13   File "/opt/scancode-toolkit/src/formattedcode/output_html.py", line 165, in write_templated
16:04:13     raise Exception(msg)
16:04:13 Exception: ERROR: Failed to write output for: u'\n            <tr>\n              <td>RD-front-master-extend/app/static/imgs/\u6570\u636e.svg</td>\n              <td>file</td>\n              <td></td>\n              <td></td>\n              <td></td>\n              <td></td>\n              <td></td>\n              <td></td>\n              <td></td>\n              <td></td>\n              <td></td>\n              <td></td>\n              <td></td>\n              <td></td>\n              <td></td>\n              <td></td>\n              <td></td>\n              <td></td>\n            </tr>\n        '
16:04:13 Traceback (most recent call last):
16:04:13   File "/opt/scancode-toolkit/src/formattedcode/output_html.py", line 160, in write_templated
16:04:13     output_file.write(template_chunk)
16:04:13 UnicodeEncodeError: 'ascii' codec can't encode characters in position 75-76: ordinal not in range(128)

pombredanne:
@arthur657834 Thanks for the report and sorry for that. This looks like the file and directory names you scanned where using special character encodings... and  that not should be a problem.
I was able to reproduce this this way:
```
$ mkdir han
$ touch  han/据.svg
$ $ scancode -i --html h.html han/
Setup plugins...
Collect file inventory...
Scan files for: info with 1 process(es)...
[####################] 1              
ERROR: failed to run output plugin: html:
Traceback (most recent call last):
  File "/tmp/scancode-toolkit/src/scancode/cli.py", line 864, in run_codebase_plugins
[...]

PR #1296, opened 2018-11-27
Support unicode path in HTML outputs #1292

pombredanne:
This PR is a fix for #1292


pombredanne:
@arthur657834  ping... if you could test this branch that would be great!

Issue #532, opened 2017-03-03
Define a "namespace" for SPDX LicenseRef- for ScanCode licenses that are NOT known to SPDX (yet)


pombredanne:
ScanCode tracks many more licenses that SPDX does. As a result `LicenseRef-xxx` needs to be created (e.g. e37fec9d4ee3eeab0a5481386d81f24fcfb44f10 )

For sanity and simplicity and clarity we should define a namespace for these non-SPDX refs.
For instance `LicenseRef-scancode-<license-key>` ... I am not sure there is a convention that exist for this in SPDX though we could suggest it.

@sschuberth input welcome!

sschuberth:
I was just wondering whether there are cases where it's uncontroversial how `<license-key>` should look like, e.g. because such a key / id is already provided by the creator of the license, and if for such cases we deliberately should *not* use a namespace to create a de-facto standard how the `LicenseRef` should look like for easier interoperability with other tools?

mjherzog:
@pombredanne I recommend contacting Gary and/or Mark G. about the namespace idea.  We have had many discussions and good ideas on this topic in the SPDX Tech group, but I cannot remember the details and life is too short to reread the emails.

pombredanne:
@goneall @MarkGisi ping!

goneall:
Currently, license-refs are local to a document.  We have discussed different license lists with different namespaces in the past, but so far have not standardized an approach.  The current convention is to use the license name field and comment field.  Probably something we should revisit in the legal and tech teams

pombredanne:
@goneall thanks. So for now I am inclined to namespace all the non-SPDX scancode licenses with `LicenseRef-scancode-<license-key>` ... This would help adopters and provide a clean reminder that the license is available and where.  that would still keep the ref logically local to a doc, but conveys better info (and certainly better than `LicenseRef-08` ;) )

pombredanne:
@sschuberth you wrote:

> I was just wondering whether there are cases where it's uncontroversial how `<license-key>` should look like, e.g. because such a key / id is already provided by the creator of the license, and if for such cases we deliberately should *not* use a namespace to create a de-facto standard how the `LicenseRef` should look like for easier interoperability with other tools?

That's a possibility, but would you have any such example? 
One thing that could be done in such cases if any would be to pre-assign a license with an `spdx_license_key: LicenseRef-XYZ` this way it would not be pre-computed from the scancode key ?
[...]

PR #1307, opened 2018-12-07
Prefix non SPDX licenses "LicenseRef-scancode-"

furuholm:
Fix for #532

ScanCode can detetect many more licenses than those listed in the
SPDX standard. Such licenses are considered external and are
named LicenseRef-xxx. This change adds a 'LicenseRef-scancode-'
prefix to such licenses to clarify that the license refers to the
license as defined by ScanCode and avoid name collisions with
output from other license scanners.

Signed-off-by: Tobias Furuholm <tobiasfuruholm@gmail.com>


pombredanne:
@furuholm thank you! Looking good and merging,

PR #1325, opened 2019-01-04
Simple, non-tested plugin to ignore binaries #689

pombredanne:
This is to support #689 


pombredanne:
@sschuberth there you go

Issue #1314, opened 2018-12-11
build_package function throws the error: ValueError: Invalid purl: name is a required argument.


linexb:
When build_package with this input json: https://registry.npmjs.org/angular2-autosize

The split_scoped_package_name function can return None,None result which is namespace and package name value, and when forming with None into a package url object, it will cause the exception since package url doesn't allow a None name, so the error happens:
```
package = build_package(data)
File "/tmp/lib/python2.7/site-packages/packagedcode/npm.py", line 258, in build_package
func(value, package)   

```

We should fix the code at: 
This is https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/npm.py#L597

pombredanne:
The actual culprit seems to be this https://raw.githubusercontent.com/stevepapa/angular2-autosize/a9f9d99143aaeb1b1108bea94336c121d7471dd8/package.json
it is not scanned properly at all

pombredanne:
And the root cause is indeed a faulty dependency:
```
  "devDependencies": {
    "@angular": "2.0.0-beta.0",
```  
```
>>> from packagedcode.npm import *
>>> split_scoped_package_name('@angular')
(u'@angular', None)

```

this return no name...  Thanks for the report!

pombredanne:
This has been fixed and merged. @linexb thanks for the report and @majurg thanks for the fix!

linexb:
@majurg  I have verified in the minecode side and and it works fine.  

This is a minor issue, please double check if it's the parser problem in scancode.

The vcs_url will have an extra "git+".
> "vcs_url":"git+git+https://github.com/stevepapa/angular2-autosize.git@f8ecfbb315d37c7354bb3fadbc50e29c42835226",

linexb:
The URL I tried to access is https://replicate.npmjs.com/registry/_changes?include_docs=true&limit=10&since=7333426, the first element in the json can reproduce the issue.

majurg:
@pombredanne and @linexb 

Looking closer at this issue, we have a number of npm packages and test cases that treat this vcs_url pattern (`git+git+https://blahblah`) as correct.

@pombredanne just to be sure, this is an actual issue we want to fix?

majurg:
#1333 Merged. This issue is now fixed.

linexb:
@majurg  There is another little issue. 

This is the result of latest scancode toolkit
    "vcs_url":"git://github.com/jprichardson/npm-research.git",

[...]

PR #1333, opened 2019-01-18
Fix duplicate vcs_tool string in vcs_url for npm packages #1314

majurg:


pombredanne:
Looking good.Thanks!

PR #1338, opened 2019-01-25
Fix-up the license text inside the `bsl-*.LICENSE` files

fviernau:
The license texts for the `Business Source License` version 1.0 and 1.1
were accidentally interchanged. Swap the contents of the two corresponding
files in order to have the right text in the right files.

See also:

https://enterprise.dejacode.com/licenses/public/bsl-1.0/#license-text
https://enterprise.dejacode.com/licenses/public/bsl-1.1/#license-text

### DIFF
--- a/src/licensedcode/data/licenses/bsl-1.0.LICENSE
+++ b/src/licensedcode/data/licenses/bsl-1.0.LICENSE
@@ -1,80 +1,31 @@
-Business Source License 1.1
+Business Source License 1.0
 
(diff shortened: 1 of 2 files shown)


fviernau:
As I missed signing-off my commit message I'll sign it off with this comment (as suggested above):

Signed-off-by: Frank Viernau <frank.viernau@here.com>

pombredanne:
@fviernau Thank you!

Issue #1351, opened 2019-02-03
Consider new category for modified FOSS licenses


pombredanne:
There are times when an existing FOSS license is modified it a little (see some Apache modifications for instance). It could make sense to report this as a new category

LeChasseur:
Marking them as "variant" or "aberrant version" would be helpful.

pombredanne:
Some examples of franken-apache-2.0 licenses
- http://www.sql-workbench.net/manual/license.html
- https://github.com/palantir/blueprint/blob/develop/LICENSE

PR #1357, opened 2019-02-09
Add new and improved license and rules

### LINKED ISSUE #1352: License classification for fraunhofer-fdk-aac-codec
The classification as "Copyleft Limited" for fraunhofer-fdk-aac-codec seems wrong for two reasons:

1. It is not an open source license since it explicitly excludes a patent license ("3.    NO PATENT LICENSE", "Patent licenses for necessary patent claims for the FDK AAC Codec (including those of Fraunhofer)
may be obtained through Via Licensing (www.vialicensing.com) or through the respective patent owners
individually for the purpose of encoding or decoding bit streams in products that are compliant with
the ISO/IEC MPEG audio standards."

2. There is no limitation for the requirement to provide the source code of modified versions, in particular the copyleft is not restricted to libraries ("You must make available free of charge copies of the complete source code of the FDK AAC Codec and your modifications thereto to recipients of copies in binary form." - "modification" could be interpreted as in GPL-3.0).

I suggest to classify the license as "Proprietary" (which seems to be called "commercial" - which is misleading since all FOSS can be commercially used)

pombredanne:
Beside a few new licenses and several new license rules, this also add support for quoted SPDX expressions and and these comment marker prefixes.
This also contains a fix #1352 and an update to #1351


pombredanne:
All green. Merging!

PR #1360, opened 2019-02-12
licensedcode: Remove the spdx_license_key from here-proprietary

sschuberth:
This is not a "core" SPDX license. While one could form an SPDX license
idstring out of any license by prefixing it with "LicenseRef-", we
currently don't do that for any of the other "non-core" SPDX /
proprietary licenses. Also, users might actually use the presence of
"spdx_license_key" to determine whether a license is a "core" SPDX
license or not. Thus, remove "spdx_license_key" from here again.

Signed-off-by: Sebastian Schuberth <sschuberth@gmail.com>

### DIFF
--- a/src/licensedcode/data/licenses/here-proprietary.yml
+++ b/src/licensedcode/data/licenses/here-proprietary.yml
@@ -3,5 +3,3 @@ name: HERE Proprietary License
 short_name: HERE Proprietary
 category: Commercial
 owner: HERE Global B.V.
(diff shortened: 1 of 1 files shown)


pombredanne:
@sschuberth thank you: that's your license, so your call! As you are aware we are discussing namespacing with SPDX, so that would be a natural way forward. I am fine either way.

sschuberth:
Thanks!

Issue #1373, opened 2019-02-15
parse function of packagedcode.maven module has exception


linexb:
When the location is None and text is not None, the _get_maven_pom will have exception:

> py", line 188, in is_posixpath
>     has_slashes = POSIX_PATH_SEP in location
> TypeError: argument of type 'NoneType' is not iterable
> 

The error should be thrown by parent_directory function below.
```
def _get_maven_pom(location=None, text=None, check_is_pom=False, extra_properties=None):
   """
    Return a MavenPom object from a POM file at `location` or provided as a
    `text` string.
    """
    if location and check_is_pom and not is_pom(location):
        return
    pom = MavenPom(location, text)
    if not extra_properties:
        extra_properties = {}
    # do we have a pom.properties file side-by-side?
    parent = fileutils.parent_directory(location)
    pom_properties = os.path.join(parent, 'pom.properties')
if os.path.exists(pom_properties):
```

pombredanne:
Thanks. And your fix has been merged and released in 3.0.2

PR #1375, opened 2019-02-15
Fix the location None error, #1373

linexb:
Quick fix to avoid the None error when getting parent folder of the passing an empty location.


pombredanne:
@linexb thanks!

pombredanne:
Thanks!

Issue #1397, opened 2019-02-26
Use case: Scanning files contained in pull request


nicobucher:
For projects using Git: "git diff --name-only master" shows a list of files which changed when comparing the current branch to master.

It would be great if the scanner could only scan those files in order to provide feedback about foss and licenses contained in the pull request. In bash-like environments, this task could be achieved by a one-liner (E.g. "git diff --name-only master | ./scancode --licenses --json-pp") if the scanner could take a list of files as input.

pombredanne:
@nicobucher thank you ++.
So the output of a `git diff --name-only` is something like:
```
$ git diff --name-only
src/formattedcode/output_csv.py
src/packagedcode/models.py
```
And we have one path per line and each path is relative to the root of the git repository.

Here is an idea for an implementation:
-  ScanCode has already an option to `--ignore` some files with this plugin https://github.com/nexB/scancode-toolkit/blob/develop/src/scancode/plugin_ignore.py
- there is an open ticket #875 to also have the opposite e.g a way to include file patterns
- the underlying fileset module used for --ignore already supports both inclusions and exclusions https://github.com/nexB/scancode-toolkit/blob/develop/src/commoncode/fileset.py

Therefore, an approach could be:
- the current single input directory argument would be kept as it is today
- a new option to include paths that would match a pattern would be added to the ignore plugin , and the logic would be to use this in conjunction with ignore 
- the logic when working with combos of  excludes/ignores and includes would be: 1. keep only included paths, 2. further exclude from these the ignored paths (this is the way of the `fileset` module)

This first step would not support yet something such as`git diff --name-only master | xargs scancode -i --json-pp -`  (note the use of xargs and the trailing dash to print the scan results) .... but a step towards this.

In fact the command may look more like `git diff --name-only master | xargs scancode -i --json-pp -` btw (note the use of xargs and the trailing dash to print the scan results) 



To get there, the process could be:
1. allow multiple args in the scancode function CLI definition https://github.com/nexB/scancode-toolkit/blob/a7de3052d540b2f005960521635e7816f50485a6/src/scancode/cli.py#L263 by adding `nargs= -1`

2. handle the case where there is only a single path provided as usual, processing the `input[0]` item.

3. if more than one paths are provided, we need to ensure first that all the paths share a common root (this would otherwise require quite a few changes to the codebase/ virtual filesystem model used internally by ScanCode)

- first case: the provided input paths share a common first path segment. In this case we would transform the list of input path arguments as:
[...]

Issue #875, opened 2017-12-21
Need specific file type pattern option and License Version


gnchalla:
1. I have many file patterns in my scan area, as of now the tool has an option to ignore those.

--ignore README --ignore "*.jar" --ignore "*.msb" --ignore "*.pl" --ignore "*.py" --ignore "*.log" --ignore "*.loc" --ignore "*.1" --ignore "*.xml" --ignore "*.ref" --ignore "*.rsf" --ignore "*.ini" --ignore "*.pom" --ignore "*.rar" --ignore "*.html" --ignore "*.css" --ignore "*.png" --ignore "*.xml" --ignore "*.nlb" --ignore "*.htm" --ignore "*.image" --ignore "*.gif" --ignore "*.po" --ignore "*.mo" --ignore "*.mak" --ignore "*.sh" --ignore "*.wsdl" --ignore "*.pm" --ignore "*.h" --ignore "*.pod" --ignore "*.map" --ignore "*.pls" --ignore "*.sql" --ignore "*.xsd" --ignore "*.cmd" --ignore "*.jpg" --ignore "*.js" --ignore "*.LAYOUT" --ignore "*.msg" --ignore "*.bat" --ignore "*.txt" --ignore "*.ctl" --ignore "*.cfg" --ignore "*.lst" --ignore "*.ttf" --ignore "*.jsp" --ignore "*.ini" --ignore "*.tif" --ignore "*.info" --ignore "*.zip" --ignore "*.ear" --ignore "*.bmp" --ignore "*.LIC" --ignore "*.ico" --ignore "*.keep" --ignore "*.db" --ignore "*.doc" --ignore "*.lnk" --ignore "*.cfs" --ignore "*.mkelem" --ignore "*.bnd" --ignore "*..SAVF" --ignore "*.xmit" --ignore "*.m" --ignore "*.xliff" --ignore "*.nls" --ignore "*.xlf" --ignore "*.properties"

Instead of ignore option, do we have an option to scan only particular pattern files?
for example want to scan only *.c

2. Do we have an option to get the license version of particular scanned file along with license name,score etc?

Please provide the info.

pombredanne:
Thanks for the report!

> Instead of ignore option, do we have an option to scan only particular pattern files? for example want to scan only *.c


[...]

PR #1399, opened 2019-02-26
Support multiple input path in CLI #875 #1397

pombredanne:
This adds support for #875 #1397 aka. multiple inputs and a new `--include` option. 
* you can now enter multiple input paths with the constraint that they all must be relative path and share a common root directory.
* also add a new enabling option for `--include` that works in tandem with the `--ignore` and has the same semantics. Includes are processed first then ignores are applied on top.
* it is now possible to run this:
`find samples/ | grep zlib |grep -v ada | grep -v iostrea | xargs scancode  --include="*JGroup*" --ignore "*.S" --json-pp -` 
or this:
`git diff --name-only master | xargs scancode -i --json-pp -`

Reported-by: Nico Bucher @nicobucher

### DIFF
--- a/etc/scripts/genlicspdx.py
+++ b/etc/scripts/genlicspdx.py
@@ -1,0 +1,113 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2019 nexB Inc. and others. All rights reserved.
+# http://nexb.com and https://github.com/nexB/scancode-toolkit/
+# The ScanCode software is licensed under the Apache License version 2.0.
+# Data generated with ScanCode require an acknowledgment.
+# ScanCode is a trademark of nexB Inc.
+#
+# You may not use this software except in compliance with the License.
+# You may obtain a copy of the License at: http://apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software distributed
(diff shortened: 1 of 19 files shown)


pombredanne:
@nicobucher ping... were you able to test this branch a bit? feedback is mucho welcomed!

nicobucher:
Hi @pombredanne, I was able to quickly try this out.
It seems to do exactly what I was looking for!

I will play around with this branch in the next days. Perhaps I can give more detailed feedback then.

pombredanne:
@nicobucher thank you for the feedback. Note that the current implementation is subpar performance wise because the logic goes this way:
1. find what is the root, shared ancestor directory of all the files arguments (and they must all be relative paths)
2. walk the whole tree for this
3. keep only the ones that match the requested paths.

If you have a large codebase (say a Linux kernel) and you modified only one file, this would walk 70K files first. So this can be optimized later if needed.

pombredanne:
@johnmhoran I moved the script to create one SPDX doc per license to this branch... https://github.com/nexB/scancode-toolkit/pull/1399/files#diff-09a4d8eb1415f0acf8c0e24ec59fc3d3

johnmhoran:
Thanks for the heads-up @pombredanne .

pombredanne:
All green ... merging now.

Issue #1403, opened 2019-02-28
Improve storage of package declared licenses


pombredanne:
### Description
The declared license of a package is what we find in a package manifest license-related fields.
Today we take the values of fields and use a simple line separation.
Instead we should keep the field names and store that as JSON too, keeping some of the original structure such that we can a better determination/normalization of the license when we populate the `license_expression` field.

For instance this is a Maven POM snippet:
```
<licenses>
    <license>
        <name>BSD</name>
        <url>http://www.jcraft.com/jzlib/LICENSE.txt</url>
    </license>
</licenses>
```
and an npm package.json snippet (rather weird as it has both a legacy and new style of license tags:
```
{
  "license": "Apache-2.0",
...
  "licenses": [
    {
      "type": "Apache 2.0",
      "url": "https://github.com/spenceralger/grunt-esvm/blob/master/LICENSE.md"
    }
  ],
...
}

```
We should in these cases store and report something such as:
- maven: `"name": "BSD, "url": http://www.jcraft.com/jzlib/LICENSE.txt"`
- npm: `"license": "Apache-2.0", "type": "Apache 2.0", "url": "https://github.com/spenceralger/grunt-esvm/blob/master/LICENSE.md"
`

vaibhav-k3:
Hello , i am new to Scancode Repo and i would like to work on this issue to start to learn about Scancode

pombredanne:
@vaibhav-k3 thanks! so each Package subclass is fed a `declared_license` [string](https://github.com/nexB/scancode-toolkit/blob/c6d1de1d97da1b6ba6d76edae95d30dbe5aa1206/src/packagedcode/models.py#L446)  by a manifest parser. This happens [here](https://github.com/nexB/scancode-toolkit/blob/c6d1de1d97da1b6ba6d76edae95d30dbe5aa1206/src/packagedcode/maven.py#L976) for Maven and [here](https://github.com/nexB/scancode-toolkit/blob/c6d1de1d97da1b6ba6d76edae95d30dbe5aa1206/src/packagedcode/npm.py#L266) for npm for instance ... there are more places of course as each maifest parser does this eventually.

vaibhav-k3:
@pombredanne , please review PR #1405

linexb:
We have done maven, pypi and npm packages now, and we will develop further more for other types.

pombredanne:
This has been merged. Thanks!

PR #1405, opened 2019-03-01
 Improve storage of package declared licenses #1403

vaibhav-k3:
Hello , i tried to work on the issue #1403 . as far as i understood , we want to store the related field(keys) and value both in json file , but currently the program stores only values. i made the changes accordingly to dictionaries and ran the tests script, but it did not pass all the tests. Please review it and suggests corrections , Thank you !

### DIFF
--- a/src/packagedcode/maven.py
+++ b/src/packagedcode/maven.py
@@ -976,7 +976,7 @@ def parse(location=None, text=None, check_is_pom=True, extra_properties=None):
     # TODO: we join all data in a single text: this may not be right
     declared_license = []
     for lic in pom.licenses:
-        lt = (l for l in [lic['name'], lic['url'], lic['comments']] if l)
+        lt = (l for l in ['name:'+lic['name'], 'url:'+lic['url'], 'comments:'+lic['comments']] if l)
         declared_license.extend(lt)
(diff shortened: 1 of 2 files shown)


pombredanne:
@vaibhav-k3 thank you for this.
A couple things:

1. You cannot create strings there. Instead you want to create a proper data structure with a mapping or a list of mappings and that will be serialized accordingly. For now you are trying to craft some hand rolled JSON string which cannot work properly at scale and is likely not parsable afterwards in many cases.

2. There are thousand of tests for a good reason: they help make sure things work: you will need to update the tests that fail accordingly by either:
2.1 update the tests
2.3 OR fix your code

3. Can you check https://github.com/nexB/scancode-toolkit/blob/develop/CONTRIBUTING.rst and the DCO sign off requirements in your commit messages and update them accordingly? Also `made some changes` is not a great commit message. See https://github.com/nexB/aboutcode/wiki/Writing-good-commit-messages

pombredanne:
@vaibhav-k3 I am closing this PR now but I appreciate the efforts you had put in it. This is a tad more engaged than what you envisioned as a solution: please see the pr of @linexb #1420 for the solution that I will merge.

Issue #1363, opened 2019-02-13
VirtualCodebase should be able to import a scan where not every resource has a parent


JonoYang:
When we create a VirtualCodebase from a scan, we assume every Resource has a parent and that is not always the case. Some Resources are pruned from a scan using the `--ignore` option or are removed for privacy purposes. ScanCode should create the parents for these Resources so the scan can be loaded.

JonoYang:
@pombredanne 

Currently, to create a VirtualCodebase from a scan, the Resources of the scan must follow these conditions:

- The first Resource in the scan needs to be the root
- Each subsequent Resource needs to be in top-down order (where the paths start from the root directory down)
- No parent Resource can be missing from the scan

I have created a function that returns a parent resource for a given path if the parent Resource exists. If it doesn't exist, it will recursively create the parent resource for a given path and return it.

This approach works for lightly damaged scans, where:

- The first Resource is the root
- Each Resource's path is base off the root path

This approach doesn't work in the cases where we are missing the root resource in the scan. For example, we cannot import a scan where the `--only-findings` options was used because `--only-findings` only reports the files that have findings and removes all other Resources that do not (the root Resource is removed since directories do not contain findings). 

Are we expected to be able to import scans that do not contain a root Resource or where the root Resource is not the first Resource in the scan? How damaged can a scan be before we refuse to create a VirtualCodebase from it?

pombredanne:
Good points. The `--only-findings` is really the evil there I guess. How complex would it be to have something along these lines, ignoring for now how this would be plugged in the CLI?

We could have a codebase "repair" function that would take a damaged "files" list then add missing parent directories and ensure that all resources have all the same attributes (e.g. from looking at all "files" to collect each attribute (name, type). And if there is no common root, then a synthetic root named "root" would be created. Finally the list would be sorted in the proper expected and stable order and then fed to create a codebase.

An alternative design would be to evolve the whole Codebase code to support doing this on the fly... but IMHO that would be detrimental to the actual physical non-virtual codebase performance and will likely make the code messy.

So having something that takes a list or iterable of "files" and returns the same but usable for VC usage seems the simplest and most testable approach at first.  It could also be used for possible future codebase "type" conversion (such as getting for instance some iterable from a CSV or from an SPDX document or from a `find . -ls` or some other representation of a filesystem and "repair" these such that they can be used as the input to a VC)

[...]

PR #1406, opened 2019-03-02
Create missing parent nodes #1363

JonoYang:
I've updated the VirtualCodebase creation process to be more robust when importing damaged scans. as a fix for #1363

VirtualCodebase can now accept scans that do not have a root Resource and has missing parent Resources. 

VirtualCodebase will first create an empty root resource by extrapolating the root directory from the path of the first Resource in the scan. If the root is encountered during the codebase loading loop, it will be updated.

VirtualCodebase will recursively create parent Resources for a Resource if its parents do not exist in the scan. If a parent is encountered later during the codebase loading loop, its fields will be updated from the scan data.


src/scancode/resource.py
```diff
@@ -1189,7 +1190,7 @@ def ancestors(self, codebase):
 
     def descendants(self, codebase):
         """
-        Return a sequence of descendant Resource objects 
+        Return a sequence of descendant Resource objects
         (doe NOT include self).
```
pombredanne: nit picking: While you are at it may be you can fix the "doe" typo?

src/scancode/resource.py
```diff
@@ -1497,26 +1497,75 @@ def get_resource_basic_attributes(_resource_data):
...
             return _name, _path, _is_file
 
-        # Create root resource
-        root_name, root_path, root_is_file = get_resource_basic_attributes(root_data)
+        def create_empty_resource_data():
+            # Get fields from the base Resource class and the ScannedResource class
+            base_fields = attr.fields(Resource)
+            resource_fields = attr.fields(self.resource_class)
+            # Create dict of {field: field_default_value} for the dynamically created fields
+            resource_data = {}
```
pombredanne: IMHO you want this to be an OrderedDict? Or does order not matter here?

src/scancode/resource.py
```diff
@@ -1497,26 +1497,75 @@ def get_resource_basic_attributes(_resource_data):
...
+                    # For fields that have Factories as values, we set their values to be an
+                    # instance of whatever type the factory makes
+                    value = value.factory()
+                resource_data[field.name] = value
+            return resource_data
+
+        # Create root resource without setting root data just yet. If we run into the root data
+        # while we iterate through `resources_data`, we fill in the data then.
+        sample_resource_path = sample_resource_data.get('path')
+        root_path = sample_resource_path.split('/')[0]
```
pombredanne: A naive split does not work in all cases. You should strip the paths from slashes first as a precaution
```
>>> '/this/is/'.split('/')[0]
''
```

This would happen if we used `--full-root` option

src/scancode/resource.py
```diff
@@ -1497,26 +1497,75 @@ def get_resource_basic_attributes(_resource_data):
...
+                if isinstance(value, attr.Factory):
+                    # For fields that have Factories as values, we set their values to be an
+                    # instance of whatever type the factory makes
+                    value = value.factory()
+                resource_data[field.name] = value
+            return resource_data
+
+        # Create root resource without setting root data just yet. If we run into the root data
+        # while we iterate through `resources_data`, we fill in the data then.
+        sample_resource_path = sample_resource_data.get('path')
```
pombredanne: If this return None, the next line will crash. If this should crash if path is not there (which is a correct assumption IMHO), use instead `sample_resource_data['path']`

src/scancode/resource.py
```diff
@@ -1444,11 +1445,10 @@ def _populate(self, scan_data):
             self.has_single_resource = True
         if not resources_data:
             raise Exception('Input has no file-level scan results.')
-        resources_data = iter(resources_data)
 
-        # The root MUST be the first resource. We use it as a template for
-        # actual Resource attributes
-        root_data = resources_data.next()
+        # We collect the first Resource so we can see what attributes it has and determine
```
pombredanne: What is the empty `def _get_top_level_attributes(self, scan_data):` above?
Also IMHO we are starting to abuse inner nested functions in this function and this starts to make reading hard. It would be good to refactor that soon enough (and that's OK to do this with this PR)

src/scancode/resource.py
```diff
@@ -1497,26 +1497,75 @@ def get_resource_basic_attributes(_resource_data):
...
+            """
+            Return a parent resource for a given `path` from `parent_by_path`.
+
+            If a parent resource for a `path` does not exist in `parent_by_path`, it is created recursively.
+
+            Note: the root path and root Resource must already be in `parent_by_path` or else this
+            function does not work.
+            """
+            parent_path = parent_directory(path).rstrip('/')
+            if parent_path in parent_by_path:
```
pombredanne: Since this a test that will be done often, it might be best and faster to do instead this where only one dict lookup occurs:
```
existing_parent = parent_by_path.get(parent_path)
if existing_parent:
    return existing_parent
```

src/scancode/resource.py
```diff
@@ -1497,26 +1497,75 @@ def get_resource_basic_attributes(_resource_data):
...
-            parent_path = parent_directory(path).rstrip('/')
-            # this must succeed since we must be in the proper sort order
-            parent = parent_by_path[parent_path]
-            resource = self._create_resource(name, parent, is_file, path, resource_data)
-
-            # Files are not parents (for now), so we do not need to add this
-            # to the parent_by_path mapping
-            if not is_file:
-                parent_by_path[path] = resource
+            if path in parent_by_path:
```
pombredanne: Same comment as above, parent_by_path.get would be better to avoid extra lookup without loosing any readability IMHO

tests/scancode/test_resource.py
```diff
@@ -1168,4 +1175,41 @@ def test_VirtualCodebase_can_be_created_from_dict(self):
...
         expected = ['bar.svg', 'han']
-        assert expected == results
\ No newline at end of file
+        assert expected == results
+
+    def test_VirtualCodebase_create_from_scan_with_no_root_and_missing_parents(self):
+        test_file = self.get_test_loc('resource/virtual_codebase/samples-only-findings.json')
+        codebase = VirtualCodebase(test_file)
+        results = sorted(r.name for r in codebase.walk())
+        expected = [
```
pombredanne: This expectation is not really easy to make sense off... It would be best to either use a file to expect the whole Resource tree

pombredanne:
@JonoYang Thanks! there was an heisen failure of one test suite on Travis. I respun a build.

src/scancode/resource.py
JonoYang: I don't think it matters there. We're just collecting the dict of field and values so we can add them in as keyword arguments when we create a Resource.

src/scancode/resource.py
JonoYang: I didn't know about this option. I'll make the change.

src/scancode/resource.py
JonoYang: I left that method in on accident. I'll take a look at refactoring the code.

JonoYang:
@pombredanne I've made the changes you requested, with the exception of refactoring the `_populate()` method (as to make sure everything works before doing so).

JonoYang:
@pombredanne I moved the in-line functions out of the `_populate()` method and folded the logic from `get_resource_basic_attributes()` into the Resource creation loop as we are just using that function there. 

Is this good enough for readability or are more changes necessary?

src/scancode/resource.py
JonoYang: I moved the in-line functions to be class methods and folded the logic of `get_resource_basic_attributes()` into the main Resource creation loop.

src/scancode/resource.py
pombredanne: that works now. An alternative could have bee to use the commoncode.fileutils/paths utility functions too

Issue #295, opened 2016-08-20
Upgrade to support also Python 3.x


pombredanne:
The time is now!
 Some important reads:
- the official guide: https://docs.python.org/3/howto/pyporting.html
- a great extensive guide from @fedora-python  https://portingguide.readthedocs.io/en/latest/process.html

pombredanne:
And this likely means only Python 3.5 and up

sschuberth:
Here's one more vote to support Python 3 as that would make [debugging multiprocessing exceptions](http://stackoverflow.com/questions/29080650/debugging-errors-in-python-multiprocessing) like issue #413 a bit easier.

pombredanne:
At a high level the key issues I can think of are:
- unicode vs. bytes, since a good many functions handle low level bytes and not unicode and Py3 is unicode by default.
- native libraries using C/C++
- other third-party libraries
- doctests (with `u''` in their expectations) or other similar tests

And in all cases, I want to support 2 and 3 at the same time with a single codebase which is not a big deal.


FWIW, most libraries should be OK except some natives. Upcoming ones (e.g. https://github.com/nexB/license-expression/ ) are also ok (I just ported that to Python 3)

pombredanne:
For the 2 to 3 update, **the first step** will be effectively to gather wheels for every ScanCode-bundled dependencies

I do not know which one would be OK or not: this has to be tested

And eventually a Python 3 wheel --if not available-- can be built for all OSes/archs using https://github.com/pombreda/thirdparty-manylinux/ https://github.com/pombreda/thirdparty that are Travis and Appveyor build loops that create wheels and upload them to http://dl.bintray.com/pombreda/thirdparty/

Note that I have already started collecting Python 36 wheels there

And once all deps are available, **the second step** is to ensure that all the tests pass both on Py2 and Py3 and start making fixes, one at a time which will be the guide to make the port work

And BTW the minimum Python3 version we will support in ScanCode is 3.6.1.
Nothing before that :wink:

vivonk:
According to me, most helpful is [this](http://python-future.org/compatible_idioms.html) so why we are collecting python wheels on version dependent. Are you trying say that packages of python3 should be installed as well like it do in python 2 case`(bin/python2.7)` for virtualenv purposes?

sschuberth:
More reasons to upgrade to Python 3.x, and version 3.7 in particular: https://docs.python.org/3/whatsnew/3.7.html#optimizations

[...]

PR #1407, opened 2019-03-02
used argparser instead of raw processing sign off: seeni<seeni0424@gm…

Seenivasanseeni:
I am using argparser . The current code just reads command line arguments and then processing it. Using this library will greatly enhance readability and feature rich.

### DIFF
--- a/etc/configure.py
+++ b/etc/configure.py
@@ -344,18 +344,15 @@ if __name__ == '__main__':
     etc_dir = os.path.abspath(os.path.dirname(__file__))
     root_dir = os.path.dirname(etc_dir)
 
-    args = sys.argv[1:]
-    if args:
-        arg0 = args[0]
-        if arg0 == '--clean':
-            clean(root_dir)
-            sys.exit(0)
-        elif arg0.startswith('-'):
-            print()
-            print('ERROR: unknown option: %(arg0)s' % locals())
-            print(usage)
-            sys.exit(1)
 
+    import argparse
+    argParser = argparse.ArgumentParser()
+    argParser.add_argument("--clean",help="Clean the build directories",action="store_true")
+    args = argParser.parse_args()
+    if args.clean:
+        clean(root_dir)
+        sys.exit(0)
+    
     sys.path.insert(0, root_dir)
     bin_dir = os.path.join(root_dir, 'bin')
     standard_python = sys.executable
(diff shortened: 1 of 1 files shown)


pombredanne:
@Seenivasanseeni Thank you for this!
A couple things for your review:
1. I appreciate your efforts but the configure script is meant to have as few moving parts as possible and very few options if any. What do I gain to use argparse there? there is no plan to add more options here. And when designed `configure` was also running on Python 2.6 (and is eventually still used on 2.6) and there is no argparse builtin on 2.6. So based on that I am not inclined to use argparse at all.

2. please check out https://github.com/nexB/scancode-toolkit/blob/develop/CONTRIBUTING.rst and especially the Sign-off-by format for your commit messages... this must be exactly a DCO signoff.

If you have other ideas this is great and let's discuss them first on the chat channel.
Thank you again!

QuLogic:
>     1. And when designed `configure` was also running on Python 2.6 (and is eventually still used on 2.6) and there is no argparse builtin on 2.6. So based on that I am not inclined to use argparse at all.

I don't care about this change too much in particular, but wasn't the result of #295 to change to Python 3 _only_ after scancode 3.x (which is out), so support of 2.6 is kind of irrelevant, isn't it?

pombredanne:
@QuLogic True, 2.6 is kind of irrelevant (I mentioned it to provide some historical background) ... but so is making a change to use argparse in this specific context where we have so little argument parsing involved AND this is a script that is used likely many thousand++ times  as every new install of scancode runs this script. This is NOT an area I want to change likely and introduce a new dependency. argparse even if standard lib IS a new dependency.

Also as explained in https://github.com/nexB/scancode-toolkit/issues/295#issuecomment-469076254 by @dankegel the port from 2 to 3 has to be a gradual process where small, incremental and consistent changes that work on both Python versions are merged on a regular basis. I will NOT accept some kind of massive PR that upgrades for 2 to 3 at once, that would be a complete waste of time. 
Also note that work on #295 is not even started yet. At least I have not seen any code yet.

I hope this makes sense :smile_cat:

dankegel:
Configure scripts are used while bootstrapping, as it were, and it may be wise to do raw processing rather than pull in libraries.

What problem are you trying to solve?  Try filing an issue for that first.  If you're working on #295, 
it might be better to ignore the configure script and thirdparty for now, and focus on a local editable pip install as I described in my long comment there.

PR #1410, opened 2019-03-03
fixed typo

Saichethan:
there is no pom.license element which is creating error while normalizing

### DIFF
--- a/src/packagedcode/maven.py
+++ b/src/packagedcode/maven.py
@@ -975,7 +975,7 @@ def parse(location=None, text=None, check_is_pom=True, extra_properties=None):
 
     # TODO: we join all data in a single text: this may not be right
     declared_license = []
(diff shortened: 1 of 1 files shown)


pombredanne:
@Saichethan Thank you! 

Also I am a bit puzzled by your proposed change:
1. there is not much explanations beyond the subject. and no tests... making it hard to understand
2. there is a `licenses` (plural) tag in a Maven POM and this is not a typo
3. your change does break everything and the tests are failing massively with over 500+ tests that fail.

Did you run the tests or any test?

So I cannot do much with your PR at this stage: it does not solve any problem but creates new one without explanations.

Also can you check https://github.com/nexB/scancode-toolkit/blob/develop/CONTRIBUTING.rst and the DCO sign off requirements in your commit messages and update them accordingly in the future?

PR #1420, opened 2019-03-06
Improve declared licenses storage #1403

linexb:
Update the declared license storage in maven module, the new change is to use the list of order dict instead of using a single string by append each license with a line break.


src/packagedcode/maven.py
```diff
@@ -27,12 +27,13 @@
 from __future__ import unicode_literals
 
 from collections import OrderedDict
-import io
-import logging
-import os.path
 from os.path import dirname
 from os.path import join
 from pprint import pformat
+import io
```
pombredanne: Could you sort your imports?

src/packagedcode/maven.py
```diff
@@ -145,6 +146,9 @@ def api_data_url(self, baseurl=default_api_baseurl):
             filename=filename,
             baseurl=baseurl)
 
+    def compute_normalized_license(self):
+        declared_license_str = json.dumps(self.declared_license)
```
pombredanne: I am not sure we want to use a straight dump. Instead the whole idea is to detect on the filed values and not the whole thing at once which is the problem we have right now. Did you add some tests of problematic POMs?

src/packagedcode/maven.py
```diff
@@ -1026,6 +1025,21 @@ def parse(location=None, text=None, check_is_pom=True, extra_properties=None):
...
+    Return a list of order dictionary to store license by paring the passing maven licenses
+    """
+    declared_license = []
+    if licenses:
+        for each in licenses:
+            declared_license.append(OrderedDict([
+                ('name', each.get('name')),
+                ('url', each.get('url')),
+                ('comments', each.get('comments')),
+                ('distribution', each.get('comments')),
```
pombredanne: Do we care about the distribution? And this is unlikely the comments, right? What are the typical values?

tests/packagedcode/data/m2/bytebuddy/pom.xml.package.json
```diff
@@ -30,7 +30,14 @@
   "vcs_url": "git+https://github.com/raphw/byte-buddy.git", 
   "copyright": null, 
   "license_expression": null, 
-  "declared_license": "The Apache Software License, Version 2.0\nhttp://www.apache.org/licenses/LICENSE-2.0.txt\nA business-friendly OSS license", 
+  "declared_license": [
+    {
+      "name": "The Apache Software License, Version 2.0", 
+      "url": "http://www.apache.org/licenses/LICENSE-2.0.txt", 
+      "comments": "A business-friendly OSS license", 
+      "distribution": "A business-friendly OSS license"
```
pombredanne: Distribution should be repo (but I am not sure we care for that field at all)
Here is the original XML POM snippet:
```
    <licenses>
        <license>
            <name>The Apache Software License, Version 2.0</name>
            <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
            <distribution>repo</distribution>
            <comments>A business-friendly OSS license</comments>
        </license>
</licenses>
```

tests/packagedcode/data/m2/com/jcraft/jsch/0.1.23/jsch-0.1.23.pom.package.json
```diff
@@ -30,7 +30,14 @@
   "vcs_url": null, 
   "copyright": null, 
   "license_expression": null, 
-  "declared_license": "BSD\nhttp://www.jcraft.com/jsch/LICENSE.txt\nLicense information from http://www.jcraft.com/jsch", 
+  "declared_license": [
+    {
+      "name": "BSD", 
+      "url": "http://www.jcraft.com/jsch/LICENSE.txt", 
+      "comments": "License information from http://www.jcraft.com/jsch", 
+      "distribution": "License information from http://www.jcraft.com/jsch"
```
pombredanne: Same as above this is not a proper repo value.

src/packagedcode/maven.py
linexb: Good idea.  We didn't carry the original data in this code, I would use declared_license to carry the original data and then it's easy to pass the normalized_license as the original data. That would be better for the license detection.

This is the our discussion,  2 or 3 is the best solution.

1. keep it as an unstructured multiline strings as we have today
2. store it as JSON carrying the structure that was in the original manifest
3. store it as a string exactly as it was in the original manifest: an XML snippet for a POM, a JSON snippet for an npm, etc

linexb: Thanks for the comment, I got the idea and have some ideas, I will update and let  you review again.

pombredanne: > Good idea. We didn't carry the original data in this code, I would use declared_license to carry the original data and then it's easy to pass the normalized_license as the original data. That would be better for the license detection.
> 
> This is the our discussion, 2 or 3 is the best solution.
> 
>     1. keep it as an unstructured multiline strings as we have today
> 
>     2. store it as JSON carrying the structure that was in the original manifest
> 
>     3. store it as a string exactly as it was in the original manifest: an XML snippet for a POM, a JSON snippet for an npm, etc

we need 2. alright ... but the whole point is that we want to detect these in sequence for the maven case:
1. the name
2. the url. If url is not properly detected ignore url. If fully detected but with a different license than name, combine the results of the two in an expression
3. the comment, with a process similar to the url.

We need to do this for all the <licenses>

Later for npms, the idea would be the same using `license` and `licenses.type` and `licenses.url` instead

src/packagedcode/maven.py
linexb: Done.Thanks.

tests/packagedcode/data/m2/bytebuddy/pom.xml.package.json
linexb: Add a 'text" key in ordered dictionary to keep the original license text info.

tests/packagedcode/data/m2/com/jcraft/jsch/0.1.23/jsch-0.1.23.pom.package.json
linexb: Yep. Fixed.

src/packagedcode/maven.py
linexb: Yep, agree, removed it.``

linexb:
@pombredanne  Fixed according your latest comments. Please review.

linexb:
Discussion with Philippe:

so the detection should likely be something such as
1. iterate over each `license`
2. try detection on the value of name if not empty and keep this
3. try detection on the value of url  if not empty and keep this
4. try detection on the value of comment  if not empty and keep this
5. if the three detection are for the same license, this becomes the kept license for that one licenses item
6. if not, the name should have precedence and any unknowns in url and comment should be ignored.
7. if there are some not unknowns in url and comment then they should be added in a combined, merged expression with an AND (license_expressions knows how to combine expressions)
8. finally merge with AND all detections from all the `licenses.license`

Finally we may need to add some score/confidence level for the detection that led to the license_expression determination... but that's for later (

linexb:
@pombredanne  Fixed according the latest discussion between you and me. :)

src/packagedcode/maven.py
```diff
@@ -145,6 +146,33 @@ def api_data_url(self, baseurl=default_api_baseurl):
...
+                # in url and comment should be ignored.
+                detected_items = [models.compute_normalized_license(item) for item in (detected_from_name, detected_from_url, detected_from_comments) if item and item != 'unknown']
+                if not detected_items:
+                    # all detected value from name, url and comments are
+                    # unknown, we should return unknown
+                    license_detected = 'unknown'
+                else:
+                    # use and expression to combine detected values from name,
+                    # url and comment
+                    license_detected = models.compute_normalized_license(' AND '.join(detected_items))
```
pombredanne: this is not how you combine tow license expressions. You need to create a Licensing() and then parse the two expressions and create a new And(exp1, exp2) OR ... `' AND '.join(detected_items)` as you did here and parse it with a Licensing(). You cannot invoke `compute_normalized_license` on them

src/packagedcode/maven.py
```diff
@@ -145,6 +146,33 @@ def api_data_url(self, baseurl=default_api_baseurl):
...
+                detected_licenses.append(license_detected)
+            return models.compute_normalized_license(' AND '.join(detected_licenses))
```
pombredanne: Sam as above. This will not work.

src/packagedcode/maven.py
```diff
@@ -973,12 +1001,7 @@ def parse(location=None, text=None, check_is_pom=True, extra_properties=None):
             # complex defeinition in Maven
             qualifiers['type'] = extension
 
-    # TODO: we join all data in a single text: this may not be right
-    declared_license = []
-    for lic in pom.licenses:
-        lt = (l for l in [lic['name'], lic['url'], lic['comments']] if l)
-        declared_license.extend(lt)
-    declared_license = '\n'.join(declared_license)
+    declared_license = get_maven_licenses(pom.licenses)
```
pombredanne: What do you gain with this vs. using the plain and simple `declared_license = pom.licenses` ?

src/packagedcode/maven.py
```diff
@@ -1026,6 +1049,20 @@ def parse(location=None, text=None, check_is_pom=True, extra_properties=None):
     return package
 
 
+def get_maven_licenses(licenses):
```
pombredanne: See my comment above. I think this function is not needed at all.

src/packagedcode/maven.py
```diff
@@ -145,6 +146,17 @@ def api_data_url(self, baseurl=default_api_baseurl):
...
 
+    def compute_normalized_license(self):
+        """
+        Get the license expression by combing the pure text from the declared license. 
+        The  value is original xml format
+        """
+        if self.declared_license:
+            declared_license_str = []
+            for license_declaration in self.declared_license:
+                if license_declaration.get('text'):
```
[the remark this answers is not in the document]
linexb: The code was refined. Due to some reason, it shows outdated code. :)

src/packagedcode/maven.py
```diff
@@ -557,6 +569,10 @@ def _find_licenses(self):
                 ('comments', self._get_attribute('comments', lic)),
                 # arcane and seldom used
                 ('distribution', self._get_attribute('distribution', lic)),
+                # Add the whole text of the licenses
```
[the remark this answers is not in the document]
linexb: This was fixed, it's a system problem that shows outdated code. :)

src/packagedcode/maven.py
```diff
@@ -1026,6 +1037,21 @@ def parse(location=None, text=None, check_is_pom=True, extra_properties=None):
...
+    Return a list of order dictionary to store license by paring the passing maven licenses
+    """
+    declared_license = []
+    if licenses:
+        for each in licenses:
+            declared_license.append(OrderedDict([
+                ('name', each.get('name')),
+                ('url', each.get('url')),
+                ('comments', each.get('comments')),
+                ('text', each.get('text')),
```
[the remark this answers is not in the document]
linexb: This was fixed, it's a system problem that shows outdated code. :)

linexb:
@pombredanne  I have refined with latest comments. Please review.  :)

src/packagedcode/maven.py
linexb: Yep, I agree. We already have pom.licenses which is a list of ordered dict.

src/packagedcode/maven.py
linexb: Yep, I agree. We already have pom.licenses which is a list of ordered dict.

src/packagedcode/maven.py
```diff
@@ -145,6 +146,69 @@ def api_data_url(self, baseurl=default_api_baseurl):
             filename=filename,
             baseurl=baseurl)
 
+    def compute_normalized_license(self):
+        """
+        Get the license expression by combining the expression with AND of all licenses, and in each license, combining the expression with AND from detected value from name, url and comments.
```
pombredanne: this docstring is too complicated IMHO. It would be best removed entirely

tests/packagedcode/test_maven.py
```diff
@@ -354,6 +354,89 @@ def test_parse_will_load_extra_pom_properties_if_file_present(self):
         assert 'org.activiti' == pom.namespace
 
 
+class TestMavenComputeNormalizedLicense(testcase.FileBasedTesting):
+    test_data_dir = os.path.join(os.path.dirname(__file__), 'data')
+
+    def test_compute_normalized_license_two_names_only(self):
+        list_license_dictionary = [{'name': 'apache-2.0'},
```
pombredanne: nit picking:
Use instead a formatting like this:
```
        declared_license = [
            {'name': 'apache-2.0'},
            {'name': 'mit'}
        ]

**Comment by pombredanne** [comment]
@linexb sounds like this went stale with the latest merges... Do you mind to rebase or merge and resolve conflicts so that I can merge alright? Thanks!

File: src/packagedcode/npm.py [hunk 15]
```diff
@@ -339,22 +398,20 @@ def split_scoped_package_name(name):
...
 
-    declared_license = None
+    declared_licenses = []
     if isinstance(license_object, string_types):
         # current form
         # TODO: handle "SEE LICENSE IN <filename>"
         # TODO: handle UNLICENSED
         # TODO: parse expression with license_expression library
-        declared_license = license_object
+        declared_licenses.append({'type': license_object})
```
pombredanne: Why do you create fake type there?

src/packagedcode/npm.py
```diff
@@ -374,22 +431,16 @@ def get_licensing(license_object):
         or
         "licenses": ["MIT"],
         """
-        lics = []
         for lic in license_object:
             if not lic:
                 continue
             if isinstance(lic, string_types):
-                lics.append(lic)
+                declared_licenses.append({'type': lic})
```
pombredanne: Again why having a fake type?

src/packagedcode/npm.py
```diff
@@ -374,22 +431,16 @@ def get_licensing(license_object):
...
-                lics.append(lics_val)
+                declared_licenses.append(lic)
             else:
-                lics.append(repr(lic))
-        declared_license = u'\n'.join(lics)
-
-    else:
-        declared_license = repr(license_object)
-    return declared_license
+                declared_licenses.append({'type': repr(lic)})
```
pombredanne: Same here: I think I get now why you are using a fake type... but the whole purpose of the declared license is to keep the original structure. So by introducing a fake "type" attribute in a mapping here and above you are not faithfully carrying over the license data structure. Furthermore, "type" is the legacy and obsolete way to document npm licenses... so ceating fake ones is not a solution instead we should have this a list (like you have now) with items being either:
- a string, for the cases where we have a single string in `license` or a legacy list of strings in `licenses`
- a mapping as it was e.g. the type/url map

And that's the job of license normalization to normalize this. Normalization should be done before when crafting the declared license field.

pombredanne:
@linexb I merged the maven part of your branch then rebased this for the npm part. You will need to use this update branch for the npm remainder

tests/packagedcode/data/npm/legacy_license_dict/package.json.expected
```diff
@@ -36,8 +36,13 @@
   "code_view_url": null,
   "vcs_url": "git://github.com/sartak/grunt-inline-imgsize.git",
   "copyright": null,
-  "license_expression": null,
-  "declared_license": "MIT\nhttps://github.com/sartak/grunt-inline-imgsize/blob/master/LICENSE-MIT",
+  "license_expression": "mit AND (mit AND unknown)",
```
pombredanne: I wonder if the results should not rather be a plain MIT in this case?

linexb:
@pombredanne I have fixed NPM changes according to your latest review ideas. :) Please review.

src/packagedcode/npm.py
```diff
@@ -407,15 +392,10 @@ def licenses_mapper(license, licenses, package):  # NOQA
...
-    licensing1 = get_licensing(license)
-    licensing2 = get_licensing(licenses)
-    declared_license = '\n'.join([v for v in (licensing1, licensing2) if v])
-
-    if declared_license:
-        if package.declared_license:
-            package.declared_license = '\n'.join([package.declared_license, declared_license])
-        else:
-            package.declared_license = declared_license
+    declared_license = get_declared_licenses(license)
```
pombredanne: What do we do if we have both a `license` and a `licenses` attribute? This is a case that can happen. See https://github.com/nexB/scancode-toolkit/pull/1420/files#diff-44d0ea49f77a17ae876679244da9e490R37

In this case we need to return both as part of the declared license.

src/packagedcode/npm.py
```diff
@@ -339,22 +407,20 @@ def split_scoped_package_name(name):
     return ns, name
 
 
-def get_licensing(license_object):
+def get_declared_licenses(license_object):
     """
-    Return actual declared licensing info from a license_data object that can be
-    a string, list, dict or dict list and come from the license or licenses
-    package.json tag.
+    Return the declared licenses, which is a list of order dictionary of licenses
```
pombredanne: Is this really true `... which is a list of order dictionary of licenses`?
This need to be updated with the changes below. Effectively this `Return a list of declared license items that can be strings or mappings`

src/packagedcode/npm.py
```diff
@@ -339,22 +407,20 @@ def split_scoped_package_name(name):
...
     """
     if not license_object:
-        return
+        return []
```
pombredanne: We need to be clear about what is the declared license of an npm. IMHO we can have all these cases:
we can have both "license" and "licenses"
- `license` should only be a single string when present. But we still need to test if this is a list of strings, dict or list of dicts.
- `licenses` can be a list of strings, a dict or a list of dict.

And we can have weird cases we have both `license` and `licenses`


So we could:
1. always return a list of things that contains either strings or dicts
2. return a string if we have a single `license` and no `licenses` (default common cases) or a list otherwise as in 1.

Below we are implementing 2.  but above we are returning an empty list which would be more consistent with 1. None would be better  for 2.

But after all I think it would be simpler in this function to always return a list.
So the line below e.g. "return license_object" should be changed to `return [license_object]`

src/packagedcode/npm.py
```diff
@@ -407,18 +459,23 @@ def licenses_mapper(license, licenses, package):  # NOQA
     - (Deprecated) an array or a list of arrays of type, url.
     -  "license": "UNLICENSED" means this is proprietary
     """
-    licensing1 = get_licensing(license)
-    licensing2 = get_licensing(licenses)
-    declared_license = '\n'.join([v for v in (licensing1, licensing2) if v])
+    # Value by parsing 'license' element
+    declared_license_via_license = get_declared_licenses(license)
+    # Value by parsing 'license' element
```
pombredanne: remove this comment and the one above: this one is wrong we are parsing "licenses" (plural) and not "license"

src/packagedcode/npm.py
```diff
@@ -407,18 +459,23 @@ def licenses_mapper(license, licenses, package):  # NOQA
...
+    declared_license_via_licenses = get_declared_licenses(licenses)
```
pombredanne: With the change above to always have `get_declared_licenses()` return a string, then the code below can be vastly simplified...

src/packagedcode/npm.py
```diff
@@ -407,18 +459,23 @@ def licenses_mapper(license, licenses, package):  # NOQA
...
+    if declared_license_via_license and declared_license_via_licenses:
+        declared_licenses = []
+        for declared_license_via in [declared_license_via_license, declared_license_via_licenses]:
+            if isinstance(declared_license_via, list):
+                declared_licenses.extend(declared_license_via)
+            else:
+                declared_licenses.append(declared_license_via)
+        package.declared_license = declared_licenses
+    else:
+        package.declared_license = declared_license_via_license or declared_license_via_licenses
```
pombredanne: ... this whole method body can be IMHO replaced by these few line:
```
def licenses_mapper(license, licenses, package):
"""....
...
"""
    declared_license = get_declared_licenses(license) or []
    declared_license.extend(get_declared_licenses(licenses)  or [])
    package.declared_license = declared_licenses
    return package
```

src/packagedcode/utils.py
```diff
@@ -28,6 +28,10 @@
 
 from six import string_types
 
+from license_expression import Licensing
+
+from packagedcode import models
+
```
pombredanne: These tow imports are not used anymore and need to be removed

src/packagedcode/npm.py
```diff
@@ -95,8 +96,75 @@ def api_data_url(self, baseurl=default_api_baseurl):
...
+    
+            # 2. try detection on the value of url  if not empty and keep this
+            url = license_declaration.get('url')
+            via_url = models.compute_normalized_license(url)
+    
+    
+            if via_type:
+                # The type should have precedence and any unknowns
+                # in url should be ignored.
+                if via_url == 'unknown':
```
pombredanne: add `TODO: find a better way to detect unknown licenses`

pombredanne:
@linexb thanks. I might make a few cosmetics updates but this is good for merging!

PR #1421, opened 2019-03-06
Add new and improved licenses and licenses rules

pombredanne:
This is an "omnibus" PR for miscellaneous new licenses and license detection rules.


etc/scripts/genlicspdx.py
```diff
@@ -0,0 +1,102 @@
+# -*- coding: utf-8 -*-
```
pombredanne: @johnmhoran you mention some interest for the SPDX script... this is it.

johnmhoran: Thank you @pombredanne .  👍

pombredanne:
All greenish :green_apple:

Issue #1424, opened 2019-03-07
Does not detect all author names in Cargo.toml


ritiek:
### Description
Hi, I ran scancode to detect copyright information but it does not correctly detect authors from `Cargo.toml` file (which get used in Rust crates). For example:
```jq
$ wget https://raw.githubusercontent.com/rust-lang/rust-clippy/master/Cargo.toml
$ scancode Cargo.toml --copyright --json-pp cargo.json
$ cat cargo.json
{
  "headers": [
    {
      "tool_name": "scancode-toolkit",
      "tool_version": "3.0.2.post331.45bbafd17",
      "options": {
        "input": "Cargo.toml",
        "--copyright": true,
        "--json-pp": "cargo.json"
      },
      "notice": "Generated with ScanCode and provided on an \"AS IS\" BASIS, WITHOUT WARRANTIES\nOR CONDITIONS OF ANY KIND, either express or implied. No content created from\nScanCode should be considered or used as legal advice. Consult an Attorney\nfor any legal advice.\nScanCode is a free software code scanning tool from nexB Inc. and others.\nVisit https://github.com/nexB/scancode-toolkit/ for support and download.",
      "start_timestamp": "2019-03-07T064943.463630",
      "end_timestamp": "2019-03-07T064943.891931",
      "message": null,
      "errors": [],
      "extra_data": {
        "files_count": 1
      }
    }
  ],
  "files": [
    {
      "path": "Cargo.toml",
      "type": "file",
      "holders": [],
      "copyrights": [],
      "authors": [
        {
          "value": "Manish Goregaokar <manishsmail@gmail.com> , Andre Bogus <bogusandre@gmail.com>",
          "start_line": 4,
          "end_line": 6
        }
      ],
      "scan_errors": []
    }
  ]
}
```

Here it displays only 2 authors:
```
"Manish Goregaokar <manishsmail@gmail.com> , Andre Bogus <bogusandre@gmail.com>"
```
where as there are 5 authors mentioned in the acutal [`Cargo.toml`](https://raw.githubusercontent.com/rust-lang/rust-clippy/master/Cargo.toml):
```
	"Manish Goregaokar <manishsmail@gmail.com>",
	"Andre Bogus <bogusandre@gmail.com>",
	"Georg Brandl <georg@python.org>",
	"Martin Carton <cartonmartin@gmail.com>",
	"Oliver Schneider <clippy-iethah7aipeen8neex1a@oli-obk.de>"
```

### System configuration

[...]

PR #1426, opened 2019-03-08
Basic Packagedcode module for handling Rust's cargo crates

ritiek:
A very basic implementation at the moment. This is still missing tests which I'll add in shortly.

Otherwise, this is ready for review!


ritiek:
Also see #1424.

ritiek:
For those looking to quickly test this out. Clone any Rust repository and run scancode on it:
```
$ git clone https://github.com/jvns/snake.rs
$ scancode --package snake.rs/ --json-pp -
```

ritiek:
I've added some tests as well.

This should be ready to merge if everything looks good. I think we could extend the subclass for handling more `Cargo.toml` attributes in an another PR. As it would be a big task in itself.

tests/packagedcode/data/rubygems/gemspec/cat.gemspec.expected.json
```diff
@@ -32785,6 +32785,9 @@
     "thirdparty/text_unidecode-1.2-py2.py3-none-any.whl", 
     "thirdparty/text_unidecode-1.2-py2.py3-none-any.whl.ABOUT", 
     "thirdparty/text_unidecode-1.2-py2.py3-none-any.whl.NOTICE", 
+    "thirdparty/toml-0.10.0-py2.py3-none-any.whl",
```
pombredanne: That's not your making but this is a weird test that I will need to fix.

PR #1427, opened 2019-03-08
Adding support for ipaddress module for python3.

Abhishek-Dev09:


setup.py
```diff
@@ -127,7 +127,7 @@ def read(*names, **kwargs):
         # Some nltk version ranges are buggy
         'nltk >= 3.2, < 4.0',
         'publicsuffix2',
-        'py2-ipaddress >= 2.0, <3.5',
+        'ipaddress',
```
pombredanne: Thnnks! Is this really the correct way since `ipadress` is built in Pytho 3.3 and up?
Would the fix rather be to have `py2-ipaddress` only required when we are on Python 2 and not when on Python3?

Abhishek-Dev09: Since ,ipaddress supports python 2 and 3 both.
https://pypi.org/project/ipaddress/

PR #1428, opened 2019-03-09
Add support python 3 for plugins

Abhishek-Dev09:
Previous version of these plugins is not compatible with python 3.And it also not available on pypi. So, i have build wheels for these plugins that compatible with python 2/3.Also these are universal wheels and platform independent.If possible i can upload these on thirdparty repo.But its not,so it is available in my repo.As three months ago lib has been uploaded by JonoYang except these plugins that support python3.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -121,7 +121,11 @@ setup(
         'open source', 'scan', 'license', 'package', 'dependency',
         'copyright', 'filetype', 'author', 'extract', 'licensing',
     ],
-    python_requires='>=2.7,<3',    
+    python_requires='>=2.7,<4',    
+    dependency_links = ['https://github.com/Abhishek-Dev09/libarchive/raw/master/extractcode-libarchive-3.1.2.tar.gz#egg=extractcode-libarchive-3.1.2',
+                       'https://github.com/Abhishek-Dev09/typecode-libmagic/raw/master/typecode-libmagic-5.22.tar.gz#egg=typecode-libmagic-5.22',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -121,7 +121,11 @@ def read(*names, **kwargs):
         'open source', 'scan', 'license', 'package', 'dependency',
         'copyright', 'filetype', 'author', 'extract', 'licensing',
     ],
-    python_requires='>=2.7,<3',    
+    python_requires='>=2.7,<4',    
+    dependency_links = ['https://github.com/Abhishek-Dev09/libarchive/raw/master/extractcode-libarchive-3.1.2.tar.gz#egg=extractcode-libarchive-3.1.2',
```
pombredanne: I cannot merge things that would work with a dependency link that is off your repository. If you modified somehow these plugins... why not make this a PR? the code is under `/plugins`

setup.py
```diff
@@ -121,7 +121,11 @@ def read(*names, **kwargs):
         'open source', 'scan', 'license', 'package', 'dependency',
         'copyright', 'filetype', 'author', 'extract', 'licensing',
     ],
-    python_requires='>=2.7,<3',    
+    python_requires='>=2.7,<4',
```
pombredanne: This is a small but important change... so you should make this its own commit

setup.py
Abhishek-Dev09: I do not have push access to your repository.Thats why i have uploaded on my repo.
No need to worry i have built from /plugins only.

PR #1430, opened 2019-03-09
Add python 3 version support to Raising exceptions

Abhishek-Dev09:
To make compatible with python 2/3 compatible.

### DIFF
--- a/src/commoncode/fileutils.py
+++ b/src/commoncode/fileutils.py
@@ -25,7 +25,7 @@
 from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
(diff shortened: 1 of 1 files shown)


src/commoncode/fileutils.py
```diff
@@ -25,6 +25,7 @@
 from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
+from shutil import Error
```
pombredanne: please separate the stdlib imports from the future imports by a blank line.

Abhishek-Dev09: Yes,I can do after testing.

PR #1432, opened 2019-03-09
To create universal wheel for extractcode-7z

Abhishek-Dev09:
Currently it support python 2 only.By making universal it can be installed anywhere by pip and support Python 2 and 3 .
After it can be build by python **setup.py bdist_wheel --universal** in this repo.
We can also create source distribution by **python3 setup.py sdist bdist_wheel**
Reference-https://packaging.python.org/guides/distributing-packages-using-setuptools/#universal-wheels

### DIFF
--- a/plugins/extractcode-7z-manylinux1_x86_64/setup.cfg
+++ b/plugins/extractcode-7z-manylinux1_x86_64/setup.cfg
@@ -1,5 +1,15 @@
 [metadata]
+# This includes the license file(s) in the wheel.
+# https://wheel.readthedocs.io/en/stable/user_guide.html#including-license-files-in-the-generated-wheel-file
 license_file = LICENSE.txt
-
+[bdist_wheel]
+# This flag says to generate wheels that support both Python 2 and Python
+# 3. If your code will not run unchanged on both Python 2 and 3, you will
+# need to generate separate wheels for each Python version that you
+# support. Removing this line (or setting universal to 0) will prevent
+# bdist_wheel from trying to make a universal wheel. For more see:
+# https://packaging.python.org/guides/distributing-packages-using-setuptools/#wheels
+universal=1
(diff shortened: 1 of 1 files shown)


plugins/extractcode-7z-manylinux1_x86_64/setup.cfg
```diff
@@ -1,5 +1,15 @@
...
 license_file = LICENSE.txt
-
+[bdist_wheel]
+# This flag says to generate wheels that support both Python 2 and Python
+# 3. If your code will not run unchanged on both Python 2 and 3, you will
+# need to generate separate wheels for each Python version that you
+# support. Removing this line (or setting universal to 0) will prevent
+# bdist_wheel from trying to make a universal wheel. For more see:
+# https://packaging.python.org/guides/distributing-packages-using-setuptools/#wheels
+universal=1
```
pombredanne: nit: there should be a space around the equal sign

plugins/extractcode-7z-manylinux1_x86_64/setup.cfg
```diff
@@ -1,5 +1,15 @@
...
 [aliases]
-release = clean --all bdist_wheel --plat-name manylinux1_x86_64
+release = clean --all bdist_wheel
```
pombredanne: Why would you  remove the --plat-name manylinux1_x86_64 ? AFAIK this is essential for these wheels to be built properly for each OS (e.g. Windows, Linux and macOS for instance here: https://pypi.org/project/extractcode-7z/#files )

plugins/extractcode-7z-manylinux1_x86_64/setup.cfg
Abhishek-Dev09: Ok, I will fix it now.

PR #1440, opened 2019-03-12
Add Support python3 to ipaddress module

Abhishek-Dev09:
Since py2-ipaddress does not support python3.So we add PyPI dependency which is compatible with both python2/3. 

### DIFF
--- a/plugins/extractcode-7z-macosx_10_9_intel/setup.cfg
+++ b/plugins/extractcode-7z-macosx_10_9_intel/setup.cfg
@@ -7,6 +7,6 @@ license_file = LICENSE.txt
 # support. Removing this line (or setting universal to 0) will prevent
 # bdist_wheel from trying to make a universal wheel. For more see:
 # https://packaging.python.org/guides/distributing-packages-using-setuptools/#wheels
-universal=1
+universal = 1
 [aliases]
 release = clean --all  bdist_wheel --plat-name macosx_10_9_intel
--- a/plugins/extractcode-7z-macosx_10_9_intel/setup.cfg
+++ b/plugins/extractcode-7z-macosx_10_9_intel/setup.cfg
@@ -1,5 +1,12 @@
 [metadata]
 license_file = LICENSE.txt
-
+[bdist_wheel]
+# This flag says to generate wheels that support both Python 2 and Python
+# 3. If your code will not run unchanged on both Python 2 and 3, you will
+# need to generate separate wheels for each Python version that you
+# support. Removing this line (or setting universal to 0) will prevent
+# bdist_wheel from trying to make a universal wheel. For more see:
+# https://packaging.python.org/guides/distributing-packages-using-setuptools/#wheels
+universal=1
 [aliases]
 release = clean --all  bdist_wheel --plat-name macosx_10_9_intel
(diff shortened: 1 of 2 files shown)


plugins/extractcode-7z-macosx_10_9_intel/setup.cfg
```diff
@@ -1,5 +1,12 @@
...
 license_file = LICENSE.txt
-
+[bdist_wheel]
+# This flag says to generate wheels that support both Python 2 and Python
+# 3. If your code will not run unchanged on both Python 2 and 3, you will
+# need to generate separate wheels for each Python version that you
+# support. Removing this line (or setting universal to 0) will prevent
+# bdist_wheel from trying to make a universal wheel. For more see:
+# https://packaging.python.org/guides/distributing-packages-using-setuptools/#wheels
+universal=1
```
pombredanne: you need a space around the =

pombredanne:
@Abhishek-Dev09 most of the comments I did on previous PR still apply: you are still missing a sign off on your commits and please use explicit commit titles and messages: you need to tell a story when you write them down.

plugins/extractcode-7z-macosx_10_9_intel/setup.cfg
Abhishek-Dev09: Ok,I will do after test

plugins/extractcode-7z-macosx_10_9_intel/setup.cfg
```diff
@@ -1,5 +1,12 @@
 [metadata]
 license_file = LICENSE.txt
-
+[bdist_wheel]
```
pombredanne: This is not needed for now if you run a `pip install -e ....` instead

setup.py
```diff
@@ -121,13 +121,13 @@ def read(*names, **kwargs):
         'open source', 'scan', 'license', 'package', 'dependency',
         'copyright', 'filetype', 'author', 'extract', 'licensing',
     ],
-    python_requires='>=2.7,<3',    
+    python_requires='>=2.7,<4',
```
pombredanne: We are not supporting Python 3 yet, so you need to keep this chnage in a branch of yours and I cannot merge this in develop yet

setup.py
```diff
@@ -121,13 +121,13 @@ def read(*names, **kwargs):
...
     ],
-    python_requires='>=2.7,<3',    
+    python_requires='>=2.7,<4',    
     install_requires=[
         # cluecode
         # Some nltk version ranges are buggy
         'nltk >= 3.2, < 4.0',
         'publicsuffix2',
-        'py2-ipaddress >= 2.0, <3.5',
+        'ipaddress',
```
pombredanne: If you want this merged in develop, I will need to have a pre-built wheel created and contributed. Also there is may be a better way since ipaddress is builtin 3.6.x: you can keep the dep on `py2-ipaddress` and make it conditional on being on Python2 only. That would be easier to merge.

pombredanne:
Also can you please squash these in a single rebased commit?

plugins/extractcode-7z-macosx_10_9_intel/setup.cfg
Abhishek-Dev09: pip install -e ... is not working on python3.I have tried.It saying that "Collecting extractcode-libarchive (from scancode-toolkit===3.0.2)
Could not find a version that satisfies the requirement extractcode-libarchive (from scancode-toolkit===3.0.2) (from versions: )"

PR #1442, opened 2019-03-12
Minor changes in comments

PyAgni:


src/commoncode/codec.py
```diff
@@ -62,7 +62,7 @@ def to_base_n(num, base):
 def to_base85(num):
     """
     Convert `num` number to a string representing this number in base 85,
-    padded as needed.
+    padding is done as needed.
```
pombredanne: Are you sure this is a valuable comment update?

PyAgni: Now realizing that this isn't of much value. Reverting this to as it was.

PyAgni:
> Thanks for the patch! No patch is too small... though can you check the guidelines in https://github.com/nexB/scancode-toolkit/blob/develop/CONTRIBUTING.rst > I would need a sign off on your commit.
> See also my feeback inline

Added the sign off on the commit. I amended the commit and force pushed it, but some errors occurred in the checks.

src/commoncode/codec.py
PyAgni: > @PyAgni Thank you for the update. Do you mind to squash these two commits in a single one?

Done. Hope everything seems good now.

pombredanne:
Thanks your commit comments looks not right now:
```
Made minor changes in comments

Update codec.py

Squashed the commits.
```

Instead it should be more like this:
```
Update to_base10() docstring in codec.py

Add proper markup to argument names.

```

Please read this https://github.com/nexB/aboutcode/wiki/Writing-good-commit-messages

PyAgni:
Changed the commit comments. Please have a look.

pombredanne:
Thanks!

Issue #1356, opened 2019-02-07
Introduce stop words to license detection


pombredanne:
There are a few rare words that should be entirely excluded at all times from the license detection.
These are mostly comment markers such as `dnl` (in autotools .m4 files) and `rem` (in DOS .bat files) and possibly `LIST` used in some ASCII firmware dumps as a comment marker. The same could be used for a limited set of HTML or XML tags (`<em>`, `<strong>`, `<href>`, `<a`, etc.) that are otherwise not helping with detection. (And this would be in contrast with stripping markup)

pombredanne:
As chatted with @majurg and @johnmhoran , other candidates for stopwords would be common HTML/XML character entities such as `&apos;`

pombredanne:
This is now included in develop and in #1552 . Also as a related thing, the list of frequent words has been replaced by a list of "legalese" words which is roughly the complement of what the junk words list was, over the set of license texts and rules. Another way to think of it would be the intersection of ranked common words vs. ranked common words in licenses and where these do not agree.
Closing now.

pombredanne:
This has been added alright. Closing now.

PR #1443, opened 2019-03-12
Add new and improved license rules

pombredanne:
This is an "omnibus" PR for various new and improved licenses and license detection rules.
It also implements stopwords #1356 which explains why there are several rules and tests impacted.


pombredanne:
Deprecated in favor of #1456

PR #1446, opened 2019-03-13
Applying condition on ipaddress module.

Abhishek-Dev09:
Since py2-ipaddress does not support python3.So we add ipaddress module which supports both python2/3.And "py2-ipaddress" will be installed only if python version<3(say2.7).
I have also updated python requires(version<4) field so that it can be needed after porting of scancode-toolkit.


setup.py
```diff
@@ -121,13 +121,14 @@ def read(*names, **kwargs):
...
-    python_requires='>=2.7,<3',    
+    python_requires='>=2.7,<4',    
     install_requires=[
         # cluecode
         # Some nltk version ranges are buggy
         'nltk >= 3.2, < 4.0',
         'publicsuffix2',
-        'py2-ipaddress >= 2.0, <3.5',
+        'py2-ipaddress >= 2.0, <3.5;python_version<"3"',
+        'ipaddress',
```
pombredanne: You do not need this line anymore now.

setup.py
```diff
@@ -121,13 +121,14 @@ def read(*names, **kwargs):
         'open source', 'scan', 'license', 'package', 'dependency',
         'copyright', 'filetype', 'author', 'extract', 'licensing',
     ],
-    python_requires='>=2.7,<3',    
+    python_requires='>=2.7,<4',
```
pombredanne: As I said I cannot merge this for now as develop does not support Python3 yet. So do not send a PR with this for now. It will have to either stay local or it will need to be defined conditionally based on the current Python version (a simple sys check)

setup.py
Abhishek-Dev09: I have removed this line.

pombredanne:
@Abhishek-Dev09 why did you close this? you only needed to squash and force push alright.

Abhishek-Dev09:
Because,this pr become very dirty.i will open fresh.

PR #1448, opened 2019-03-14
Do not use py2-ipaddress on Python3 #295

Abhishek-Dev09:
The py2-ipaddress library is a backport of the Python 3 ipadress built-in module. Therefore it is not needed on Python 3

Signed-off-by: Abhishek-Dev09 <abhishek.kasyap09@gmail.com>


Abhishek-Dev09:
Whats wrong with test cases, in previous PR it was passed, but now it doesn't.

PR #1449, opened 2019-03-15
Update to get_file_mtime() in date.py

PyAgni:
Fixed typo error and changed variable name 'iso' to 'ISO' for better
code readability.

### DIFF
--- a/src/commoncode/date.py
+++ b/src/commoncode/date.py
@@ -38,7 +38,7 @@ def isoformat(utc_date):
 def get_file_mtime(location, iso=True):
     """
     Return a string containing the last modified date of a file formatted
-    as an ISO time stamp is ISO is True or a as raw number since epoch.
+    as an ISO time stamp if ISO is True or as a raw number since epoch.
     """
     date = ''
     # FIXME: use file types


src/commoncode/date.py
```diff
@@ -35,16 +35,16 @@ def isoformat(utc_date):
     return datetime.isoformat(utc_date).replace('T', ' ')
 
 
-def get_file_mtime(location, iso=True):
+def get_file_mtime(location, ISO=True):
```
pombredanne: We do not use uppercase for variable names.

PyAgni: Will keep that in mind. Reverted the variable name changes.

PyAgni:
@pombredanne 
In the comment what exactly do you mean by using file types? Could you please elaborate this?
[https://github.com/nexB/scancode-toolkit/blob/9e04eb55cd4a049409b86dac0b76445d87474c3a/src/commoncode/date.py#L44](url)

pombredanne:
You wrote:

> In the comment what exactly do you mean by using file types? Could you please elaborate this?
> https://github.com/nexB/scancode-toolkit/blob/9e04eb55cd4a049409b86dac0b76445d87474c3a/src/commoncode/date.py#L44

I do not think I recall what this was for. This is a stale comment.

pombredanne:
Thanks!

PyAgni:
> Thanks!

You are welcome! Also, should I consult with you on gitter before making any pull requests? That would probably relieve you of requesting changes so many times. :)

pombredanne:
@PyAgni re:

> You are welcome! Also, should I consult with you on gitter before making any pull requests? That would probably relieve you of requesting changes so many times. :)

Your call. It is definitely best to discuss first any larger thing. Tickets work well too.

PR #1452, opened 2019-03-15
Add support python3.X on module #295

Abhishek-Dev09:
Since module used on python 2.X. Therefore in order to port scancode-toolkit to python 3.X. It need to be ported.

Signed-off-by: Abhishek-Dev09 <abhishek.kasyap09@gmail.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -127,7 +127,7 @@ setup(
         # Some nltk version ranges are buggy
         'nltk >= 3.2, < 4.0',
         'publicsuffix2',
-        'py2-ipaddress >= 2.0, <3.5',
+        'py2-ipaddress >= 2.0, <3.5;python_version<"3"',
         'url >= 0.1.4, < 0.1.6',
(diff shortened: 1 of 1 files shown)


src/commoncode/urn.py
```diff
@@ -25,6 +25,8 @@
 from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
+from future.standard_library import install_aliases
```
pombredanne: What does this do?

pombredanne:
@Abhishek-Dev09 thanks... See my comment. Also how could this be tested automatically?

src/commoncode/urn.py
Abhishek-Dev09: This module has been used to import urlib module in python 2/3. Refer to this documentation of future module.
Reference: https://python-future.org/compatible_idioms.html#urllib-module

PR #1453, opened 2019-03-15
Add support for python3 in urn.py

PyAgni:
Used urllib.parse python3 module and made changes in encode() and
decode()

### DIFF
--- a/src/commoncode/urn.py
+++ b/src/commoncode/urn.py
@@ -85,8 +85,12 @@ Examples:
 The product object type syntax is the same as the component syntax.
 """
 
-from urllib import quote_plus
-from urllib import unquote_plus
+try:
+    from urllib.parse import quote_plus
+    from urllib.parse import unquote_plus
+except ImportError:
+    from urllib import quote_plus
+    from urllib import unquote_plus
 
 
 class URNValidationError(Exception):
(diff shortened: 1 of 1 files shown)


src/commoncode/urn.py
```diff
@@ -141,13 +144,13 @@ def decode(urn):
 
     # object type is always lowercase
     object_type = segments[2].lower()
-    if object_type not in URN_SCHEMAS.keys():
+    if object_type not in list(URN_SCHEMAS.keys()):
```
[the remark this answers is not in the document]
PyAgni: > Thank you for this. There are a few stylistic comments ... but more important is that this urn.py code is not use anywhere so IMHO it should be instead deleted with its tests and not patched for use on Python3.

Okay! So I'm making the requested changes now and whether urn.py should be kept or deleted can be decided after the whole porting process is done.

src/commoncode/urn.py
```diff
@@ -85,9 +85,12 @@
 The product object type syntax is the same as the component syntax.
 """
 
-from urllib import quote_plus
-from urllib import unquote_plus
+#The future package makes the Python 3.x APIs available on Python 2.x
+from future.standard_library import install_aliases
+install_aliases()
```
pombredanne: Could use something that would not depend on `future`: I would rather have a simple try/except on imports.
Also check if this `urllib` is in use anywhere else that would need fixing and apply that one related fix everywhere.

src/commoncode/urn.py
```diff
@@ -141,7 +145,7 @@ def decode(urn):
 
     # object type is always lowercase
     object_type = segments[2].lower()
-    if object_type not in URN_SCHEMAS.keys():
+    if object_type not in URN_SCHEMAS:
```
pombredanne: This would have been best left out as this is unrelated to the port.

src/commoncode/urn.py
```diff
@@ -116,13 +120,13 @@ def encode(object_type, **fields):
 
     # case is not significant for the object type
     object_type = object_type.strip().lower()
-    urn_prefix = u'urn:dje:{0}:'.format(quote_plus(object_type))
+    urn_prefix = 'urn:dje:{0}:'.format(quote_plus(object_type))
```
pombredanne: Keeping u prefixes is harmless and supported in Python 3.

PR #1457, opened 2019-03-19
Use "exception as e" on Python 3 #295

Abhishek-Dev09:
Since 'Exception, e' is used on python 2.X .Therefore it is not needed on Python 3.X .

### DIFF
--- a/src/commoncode/fileutils.py
+++ b/src/commoncode/fileutils.py
@@ -127,14 +127,14 @@ def create_dir(location):
 
         # avoid multi-process TOCTOU conditions when creating dirs
         # the directory may have been created since the exist check
-        except WindowsError, e:
+        except WindowsError as e:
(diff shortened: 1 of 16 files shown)


pombredanne:
Thanks! I will review tomorrow. two things:
- please use your real name for the sign-off (not Abhishek-Dev09)
- are you sure that this cover all the places everywhere that need to be updated with this idiom?
- please do not add other changes when you do such change (even and including small formatting changes that are not bad per se such as adding an empty line). There is going to be a lot of changes, so these have to be done quite strictly

pombredanne:
Thanks!
A better message could have been:
```
Use "except Exception as e" idiom everywhere #295

This works on both Python 2 and 3. 
The idiom "except Exception, e" does not work on Python 3. 
```

but I will merge as is for now. Make you sure you work on crafting descriptive message next time.

PR #1460, opened 2019-03-21
Raising exception on Python 3.X #295

Abhishek-Dev09:
Since 'raise ValueError, str' is used on Python 2.X. Therefore, to use raise exception on python 3.X ,we use it as 'raise ValueError(str)'.

### DIFF
--- a/src/commoncode/fileutils.py
+++ b/src/commoncode/fileutils.py
@@ -499,7 +499,7 @@ def copytree(src, dst):
             errors.append((srcname, dstname, str(why)))
 
     if errors:
-        raise shutil.Error, errors
+        raise shutil.Error(errors)
 
 
 def copyfile(src, dst):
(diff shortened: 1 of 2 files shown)


src/commoncode/fileutils.py
```diff
@@ -540,7 +540,7 @@ def copytime(src, dst):
     if hasattr(os, 'utime'):
         try:
             os.utime(dst, (st.st_atime, st.st_mtime))
-        except OSError, why:
+        except OSError as why:
```
Abhishek-Dev09: Since, one exception has left in #1457 .

Abhishek-Dev09:
If anyone find more suitable title and msg for this commit.Suggest it

tests/extractcode/test_archive.py
```diff
@@ -379,7 +379,7 @@ def assertRaisesInstance(self, excInstance, callableObj,
         excClass = excInstance.__class__
         try:
             callableObj(*args, **kwargs)
-        except excClass, e:
+        except excClass as e:
```
Abhishek-Dev09: exception left in #1457

src/commoncode/fileutils.py
pombredanne: :+1:

PR #1465, opened 2019-03-22
Import "types" & "urllib2" on python 3.X  #295

Abhishek-Dev09:
The 'types' module contains a variety of constants to help you determine the type of an object. In Python 3, these constants have been eliminated;just use the primitive type name instead.

The old urllib module in Python 2 had a variety of functions, including urlopen() for fetching data; and splittype(), splithost(), and splituser() forsplitting a URL into its constituent parts. These functions have been reorganized more logically within the new urllib package.
The old urllib2 module in Python 2 has been folded into the urllib package in Python 3. All your urllib2 favorites–the build_opener() method, Request objects, and HTTPBasicAuthHandler and friends–are still available. 

### DIFF
--- a/src/commoncode/functional.py
+++ b/src/commoncode/functional.py
@@ -25,7 +25,7 @@
 from __future__ import absolute_import, print_function
 
 import functools
-from types import ListType, TupleType, GeneratorType
+from types import *
 from array import array
 
 
--- a/src/commoncode/functional.py
+++ b/src/commoncode/functional.py
@@ -25,7 +25,7 @@
 from __future__ import absolute_import, print_function
 
 import functools
-from types import *
+from types import GeneratorType
 from array import array
 
 
@@ -53,7 +53,7 @@ def flatten(seq):
     """
     r = []
     for x in seq:
-        if isinstance(x, (ListType, TupleType)):
+        if isinstance(x, (list, tuple)):
             r.extend(flatten(x))
         elif isinstance(x, GeneratorType):
             r.extend(flatten(list(x)))
@@ -102,7 +102,7 @@ def memoize(fun):
         if kwargs:
             return fun(*args, **kwargs)
         # convert any list args to a tuple
-        args = tuple(tuple(arg) if isinstance(arg, (ListType, tuple, array)) else arg
+        args = tuple(tuple(arg) if isinstance(arg, (list, tuple, array)) else arg
                      for arg in args)
         try:
             return memos[args]
--- a/src/scancode/__init__.py
+++ b/src/scancode/__init__.py
@@ -35,7 +35,6 @@ from os.path import getsize
 from os.path import getmtime
 from os.path import join
 from os.path import exists
-from types import *
 
 import click
 from click.types import BoolParamType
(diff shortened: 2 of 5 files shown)


Abhishek-Dev09:
If anyone find more suitable title and msg for this commit.Suggest it

src/commoncode/functional.py
```diff
@@ -25,7 +25,7 @@
 from __future__ import absolute_import, print_function
 
 import functools
-from types import ListType, TupleType, GeneratorType
+from types import *
```
pombredanne: Please never ever use `import *` but always list what you import explicitly

src/scancode/__init__.py
```diff
@@ -35,7 +35,7 @@
 from os.path import getmtime
 from os.path import join
 from os.path import exists
-from types import BooleanType
+from types import *
```
pombredanne: No * please

tests/licensedcode/data/slic-tests/identification/repo
```diff
@@ -123,7 +123,10 @@ if sys.version_info[0] == 3:
   import urllib.error
 else:
   import imp
-  import urllib2
+  try:
```
pombredanne: This is a test data file that is NOT part of the code we run, just something we test license detection on... so this file should not be modified at all.

src/commoncode/functional.py
```diff
@@ -25,7 +25,15 @@
...
-from types import ListType, TupleType, GeneratorType
+
+# Python 2 and 3 support
+try:
+    # Python 2
+    from types import ListType, TupleType, GeneratorType
+except ImportError, NameError:
+    from types import GeneratorType
+    ListType = list
+    TupleType = tuple
```
pombredanne: that's clean... but could we instead just use list.tuple everywhere and only import GeneratorType on both py2 and 3?

src/commoncode/functional.py
```diff
@@ -53,7 +61,7 @@ def flatten(seq):
     """
     r = []
     for x in seq:
-        if isinstance(x, (ListType, TupleType)):
+        if isinstance(x, (ListType, TupleType, array)):
```
pombredanne: Why did you add array to this? Aldo with my other comment above, using `(list, tuple,)` would be enough IMHO

src/scancode/__init__.py
```diff
@@ -35,7 +35,14 @@
...
 from os.path import exists
-from types import BooleanType
+
+# Python 2 and 3 support
+try:
+    #Python2
+    from types import BooleanType
+except NameError:
+    # Python 3
+    BooleanType = bool
```
pombredanne: I think it is simpler to use `bool` on both versions.: no import needed anymore

src/commoncode/functional.py
```diff
@@ -25,7 +25,15 @@
 from __future__ import absolute_import, print_function
 
 import functools
-from types import ListType, TupleType, GeneratorType
+
+# Python 2 and 3 support
```
singh1114: No need for this comment.

src/commoncode/functional.py
singh1114: Why have you included array as well?

src/scancode/__init__.py
```diff
@@ -35,7 +35,14 @@
 from os.path import getmtime
 from os.path import join
 from os.path import exists
-from types import BooleanType
+
+# Python 2 and 3 support
```
singh1114: Again no need for such comments.

PR #1467, opened 2019-03-22
setup.py: add two dependencies to make older systems happy; fixes #1463

### LINKED ISSUE #1463: pip install fails on fresh ubuntu 16.04 instance
While pip is not the main install method preferred by the scancode authors,
it's still an important mainstream way to distribute python code, so we should
keep it working.

Trying it on a fresh ubuntu 16.04 instance fails at the moment, e.g.

```
$ lxc launch ubuntu:16.04 ubu1604-demo
$ lxc exec ubu1604-demo bash
# apt update
# apt install python-pip
# su ubuntu
$ cd
$ pip install --user scancode-toolkit
...
Collecting pyicu>=1.9.3 (from normality>=0.4.0->fingerprints==0.5.4->scancode-toolkit)
...
    RuntimeError:
    Please set the ICU_VERSION environment variable to the version of
    ICU you have installed.
```
i.e. there's a missing native package that can't be installed by pip.  That's easy to work around:
```
$ sudo apt install libicu-dev
$ pip install --user scancode-toolkit
...
Collecting more-itertools (from jaraco.functools>=1.20->tempora->scancode-toolkit)
  Using cached https://files.pythonhosted.org/packages/dd/f4/cdfbb6f07f767e0cd8a11b16adfecc735b6eb87dbebda026ce8898e1cc22/more-itertools-6.0.0.tar.gz
...
```
That succeeds, but note:
```
$ pip list | grep itertools
more-itertools (6.0.0)
```
That's not going to end well, since more-itertools dropped support for python 2.7,
and indeed:
```
$ .local/bin/scancode --json-pp - --license --package .
Setup plugins...
ERROR: failed to setup plugin: scan:licenses:
...
  File "/home/ubuntu/.local/lib/python2.7/site-packages/jaraco/functools.py", line 27, in <module>
    import more_itertools.recipes
  File "/home/ubuntu/.local/lib/python2.7/site-packages/more_itertools/__init__.py", line 1, in <module>
    from more_itertools.more import *  # noqa
  File "/home/ubuntu/.local/lib/python2.7/site-packages/more_itertools/more.py", line 329
    def _collate(*iterables, key=lambda a: a, reverse=False):
                               ^
SyntaxError: invalid syntax
```

Filed https://github.com/jaraco/jaraco.functools/issues/10 and
https://github.com/jaraco/jaraco.functools/pull/11 for that.

That's also not too hard to work around:
```
$ pip uninstall more-itertools
$ pip install 'more-itertools<6.0.0'
$ .local/bin/scancode --json-pp - --license --package .
```
And that goes swimmingly, using up 1.3 gigabytes of RAM as expected :-)

So, the proper action is probably to wait a while for jaraco.functools to push a fixed version to pypi.  If they don't, we could paper over the problem by adding a dependency spec in scancode's setup.{cfg,py}.

Oh, and maybe we could have a line of code to hint to users they need to install native libicu-dev,  written generally enough to apply regardless of distro.

dankegel:
Python purists may scratch their heads over this commit, but it should make life easier for people trying to support ubuntu 16.04 and 18.04.

See https://github.com/nexB/scancode-toolkit/issues/1463

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -123,6 +123,11 @@ setup(
     ],
     python_requires='>=2.7,<3',    
     install_requires=[
+        # Hack to support pip 8 (for those poor sods forced to use ubuntu 16.04's system pip)
+        # See https://github.com/nexB/scancode-toolkit/issues/1463
+        'more_itertools <  6.0.0; python_version == "2.7"',
+        # end hack
+
         # cluecode
         # Some nltk version ranges are buggy
         'nltk >= 3.2, < 4.0',
@@ -140,6 +145,11 @@ setup(
         'extractcode-7z',
 
         # commoncode
+        # Hack to avoid broken backports.os on systems with old configparser
+        # See https://github.com/PiDelport/backports.os/issues/11
+        # and https://github.com/jaraco/configparser/issues/17
+        'configparser >= 0.6.0',
+        # end hack
         'backports.os == 0.1.1',
         'future == 0.16.0',
         'text-unidecode >= 1.0, < 2.0',


pombredanne:
Thanks! Looks good to me I will merge as soon as the CI tests are finished though the CI is likely to fail since it uses exclusively bundled thirdparty libraries for now. So no configparser yet

setup.py
```diff
@@ -123,6 +123,11 @@ def read(*names, **kwargs):
     ],
     python_requires='>=2.7,<3',    
     install_requires=[
+        # Hack to support pip 8 (for those poor sods forced to use ubuntu 16.04's system pip)
+        # See https://github.com/nexB/scancode-toolkit/issues/1463
+        'more_itertools <  6.0.0; python_version == "2.7"',
```
Abhishek-Dev09: "more_itertools " is written as "more-itertools"

dankegel: Curses!  :-)  Thanks.

Abhishek-Dev09: > Curses! :-) Thanks.

Is i am wrong?

dankegel: You're probably right, and I should learn to test my code better.

setup.py
```diff
@@ -140,6 +145,11 @@ def read(*names, **kwargs):
         'extractcode-7z',
 
         # commoncode
+        # Hack to avoid broken backports.os on systems with old configparser
+        # See https://github.com/PiDelport/backports.os/issues/11
+        # and https://github.com/jaraco/configparser/issues/17
+        'configparser >= 0.6.0',
```
Abhishek-Dev09: I see the min version of configparser is 3.2

dankegel: Yes, that should be something like 3.5.1, not 0.6.0.

dankegel:
Closing in favor of a future clean MR that does not futz with configparser.

PR #1478, opened 2019-03-28
Use new Python3 syntax for octal #295

Abhishek-Dev09:
Octal literals must now be specified with a leading "0o" or "0O" instead of "0" .
Reference: http://www.python.org/dev/peps/pep-3127/


pombredanne:
Thank you. Neat and clean. Merging!

PR #1483, opened 2019-03-30
Use 'maxsize' instead of 'maxint' on Python 3 #295

Abhishek-Dev09:
The sys.maxint constant was removed, since there is no longer a limit to the value of integers. However, sys.maxsize can be used as an integer larger than any practical list or string index. It conforms to the implementation’s “natural” integer size and is typically the same as sys.maxint in previous releases on the same platform (assuming the same build options).                                                                             
              Reference = http://docs.python.org/3.1/whatsnew/3.0.html#integers

### DIFF
--- a/src/commoncode/testcase.py
+++ b/src/commoncode/testcase.py
@@ -50,7 +50,7 @@ from commoncode.system import on_windows
 test_run_temp_dir = None
 
 # set to 1 to see the slow tests
-timing_threshold = sys.maxint
+timing_threshold = sys.maxsize
 
 POSIX_PATH_SEP = b'/' if on_linux else '/'
 WIN_PATH_SEP = b'\\' if on_linux else '\\'


tests/cluecode/data/ics/chromium-net-tools-testserver/chromiumsync.py
```diff
@@ -6,4 +6,4 @@
     # to its nickname.
     self.clients = {}
     self.client_name_generator = ('+' * times + chr(c)
-        for times in xrange(0, sys.maxint) for c in xrange(ord('A'),ord('Z')))
+        for times in xrange(0, sys.maxsize) for c in xrange(ord('A'),ord('Z')))
```
pombredanne: This is test data and not our own code. Please do not change this

Abhishek-Dev09: ok,I have done.

Issue #1480, opened 2019-03-29
 Improve declared licenses storage for Rubygems


linexb:
### Description
This is a sub ticket of #1403, the idea is let's keep the original data format of the declared license and use the class based function "compute_normalized_license"  to populate the license_expression field.

pombredanne:
FYI I entered this https://github.com/rubygems/rubygems/issues/2713

pombredanne:
This has been merged. Thanks

PR #1490, opened 2019-04-02
Improve Rubygems declared license storage #1480

linexb:
See #1480 
Store the declared license as a list of strings and compute a license expression from these.

### DIFF
--- a/src/packagedcode/maven.py
+++ b/src/packagedcode/maven.py
@@ -47,6 +47,7 @@ from commoncode import filetype
 from commoncode import fileutils
 from packagedcode import models
 from packagedcode.models import Package
+from packagedcode.utils import combine_expressions
 from packagedcode.utils import normalize_vcs_url
 from packagedcode.utils import VCS_URLS
 from textcode import analysis
@@ -202,27 +203,16 @@ def compute_normalized_license(declared_license):
                 detections = via_name, via_url, via_comments
                 detections = [l for l in detections if l]
                 if detections:
-                    if len(detections) == 1:
-                        combined_expression = detections[0]
-                    else:
-                        expressions = [
-                            licensing.parse(le, simple=True) for le in detections]
-                        combined_expression = str(licensing.AND(*expressions))
-                    detected_licenses.append(combined_expression)
-
+                    combined_expression =  combine_expressions(detections)
+                    if combined_expression:
+                        detected_licenses.append(combined_expression)
         elif via_url:
             detected_licenses.append(via_url)
         elif via_comments:
             detected_licenses.append(via_comments)
 
-    if len(detected_licenses) == 1:
-        return detected_licenses[0]
-
     if detected_licenses:
-        # Combine if pom contains more than one licenses declarations.
-        expressions = [licensing.parse(le, simple=True) for le in detected_licenses]
-        combined_expression = licensing.AND(*expressions)
-        return str(combined_expression)
+        return combine_expressions(detected_licenses)
 
 
 def build_url(group_id, artifact_id, version, filename, baseurl='http://repo1.maven.org/maven2'):
--- a/src/packagedcode/rubygems.py
+++ b/src/packagedcode/rubygems.py
@@ -291,9 +291,7 @@ def build_rubygem_package(gem_data, download_url=None, package_url=None):
     descriptions = [d for d in (short_desc, long_desc) if d and d.strip()]
     description = '\n'.join(descriptions)
 
-    # Since the gem spec doc is not clear https://guides.rubygems.org
-    # /specification-reference/#licenseo, we will treat a list of licenses and a
-    # conjunction for now (e.g. AND)
+
     declared_licenses = []
     licenses = gem_data.get('licenses') or []
     for lic in licenses:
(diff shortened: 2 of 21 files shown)


src/packagedcode/rubygems.py
```diff
@@ -385,6 +398,22 @@ def build_rubygem_package(gem_data, download_url=None, package_url=None):
...
+    `licenses` values found in a package.
+    """
+    declared_licenses = []
+    if license:
+        declared_licenses.append(str(license).strip())
+    if licenses:
+        for lic in licenses:
+            if lic and lic.strip():
+                declared_licenses.append(lic.strip())
+    package.declared_license = declared_licenses
```
pombredanne: Instead of updating the package, please return the list of declared licenses.

src/packagedcode/rubygems.py
```diff
@@ -314,6 +320,13 @@ def build_rubygem_package(gem_data, download_url=None, package_url=None):
         declared_license=declared_license,
     )
 
+    # Since the gem spec doc is not clear https://guides.rubygems.org
+    # /specification-reference/#licenseo, we will treat a list of licenses and a
+    # conjunction for now (e.g. AND)
+    license = gem_data.get('license')
+    licenses = gem_data.get('licenses')
+    package = licenses_mapper(license, licenses, package)
```
pombredanne: do not pass the package around.... instead get a declared_license list back and add it to the package creation above.... so this can and should be moved before the package creation

src/packagedcode/rubygems.py
```diff
@@ -306,6 +311,7 @@ def build_rubygem_package(gem_data, download_url=None, package_url=None):
 
     declared_license = '\n'.join(declared_licenses) or None
 
+
```
pombredanne: `declared_license = '\n'.join(declared_licenses) or None` 

.... this no longer makes sense. We want to explicitly keep a list of such licenses and no longer a string

src/packagedcode/utils.py
```diff
@@ -146,3 +147,22 @@ def build_description(summary, description):
...
 
     return description
+
+def combine_expressions(expressions, licensing=Licensing()):
+    """
+    Return a combined license expression string with AND, given a list of
+    license expressions.
+
+    For example:
+    >>> a = 'mit'
```
pombredanne: Can you add a test for what happens when expressions is None, an empty list or a string or a string with a single character?

src/packagedcode/utils.py
```diff
@@ -146,3 +147,22 @@ def build_description(summary, description):
...
+    >>> b = 'gpl'
+    >>> combine_expressions([a, b])
+    'mit AND gpl'
+    """
+    if not expressions:
+        return
+    if len(expressions) == 1:
```
pombredanne: what happens if expressions is `q` ? this tests alone does not make sense to me.

src/packagedcode/maven.py
```diff
@@ -202,27 +203,16 @@ def compute_normalized_license(declared_license):
...
                 if detections:
-                    if len(detections) == 1:
-                        combined_expression = detections[0]
-                    else:
-                        expressions = [
-                            licensing.parse(le, simple=True) for le in detections]
-                        combined_expression = str(licensing.AND(*expressions))
-                    detected_licenses.append(combined_expression)
-
+                    combined_expression =  combine_expressions(detections)
```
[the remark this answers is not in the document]
pombredanne: Actually my bad: there is no change t the logic here! :+1:

src/packagedcode/rubygems.py
linexb: Fixed.

src/packagedcode/rubygems.py
linexb: Fixed

src/packagedcode/rubygems.py
```diff
@@ -167,6 +150,30 @@ def extra_root_dirs(cls):
...
+    """
+    if not declared_license:
+        return
+
+    detected_licenses = []
+
+    for declared in declared_license:
+        #TODO:  scancode convention is to put one license per line when there are
+        # multiple licenses as a list in the manifest.
+        # lines = [l for l in self.declared_license.splitlines(False) if l and l.strip()]
```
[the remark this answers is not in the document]
linexb: Fixed.

src/packagedcode/rubygems.py
linexb: Fixed.

src/packagedcode/rubygems.py
```diff
@@ -167,6 +150,30 @@ def extra_root_dirs(cls):
...
+        if declared:
```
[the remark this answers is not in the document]
linexb: Agree, the licenses_mapper already checked the empty case, so it's impossible to be empty here. Fixed.

src/packagedcode/utils.py
linexb: Yep, I have added the tests on  the pypi branch, https://github.com/nexB/scancode-toolkit/commit/801bb5d92525fbc036e8fdab829ee41724f57f65

src/packagedcode/utils.py
linexb: @pombredanne Can you please explain more about q? Do you mean non list case? :)

pombredanne: what if this expression is only a single character string and not a list?

pombredanne:
Thanks. All merged

PR #1494, opened 2019-04-02
Avoid explicitly specifying str type as Unicode #295

chashiv:
In Python 3.x implicit **str** type is **Unicode**.
Signing off : Shivam Chauhan <chauhanshivam999@gmail.com>

### DIFF
--- a/src/cluecode/finder_data.py
+++ b/src/cluecode/finder_data.py
@@ -31,7 +31,7 @@ def set_from_text(text):
     return set(u.lower().strip('/') for u in text.split())
 
 
-JUNK_EMAILS = set_from_text(u'''
+JUNK_HOSTS_AND_DOMAINS = set_from_text('''
     test@test.com
     exmaple.com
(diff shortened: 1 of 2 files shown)


pombredanne:
Thank you but we need to keep Python2 compatibility and a `u` prefix is  OK in Python3. Also if this was a desirable change (which it is not) there would be many other places where this would be needed.
Furthermore this change makes all tests break. So I cannot accept this because it breaks everything... but I appreciate the effort and the PR nonetheless.

chashiv:
> Thank you but we need to keep Python2 compatibility and a `u` prefix is OK in Python3. Also if this was a desirable change (which it is not) there would be many other places where this would be needed.
> Furthermore this change makes all tests break. So I cannot accept this because it breaks everything... but I appreciate the effort and the PR nonetheless.

@pombredanne I do understand now . Thank you for your valuable review/feedback over this PR.

PR #1495, opened 2019-04-02
Found an unclosed group, closing it

chashiv:
There lies an **unclosed group**, closing it.
Signing off : Shivam Chauhan <Chauhanshivam999@gmail.com>

### DIFF
--- a/src/textcode/strings.py
+++ b/src/textcode/strings.py
@@ -199,7 +199,7 @@ def is_relative_path(s):
     Return True if s looks like a relative posix path.
     Example: usr/lib/librt.so.1 or ../usr/lib
     """
-    relative = re.compile('^(?:([^/]|\.\.)[\w_\-]+/.*$', re.IGNORECASE).match
+    relative = re.compile('^(?:([^/]|\.\.)[\w_\-]+/.*$)', re.IGNORECASE).match
     return relative(s)
 
 
--- a/src/textcode/strings.py
+++ b/src/textcode/strings.py
@@ -27,6 +27,7 @@ from __future__ import print_function
 
 import re
 import string
+import os
 
 from commoncode.text import toascii
 
@@ -34,11 +35,8 @@ from commoncode.text import toascii
 Extract raw ASCII strings from (possibly) binary strings.
 Both plain ASCII and UTF-16-LE-encoded (aka. wide) strings are extracted.
 The later is found typically in some Windows PEs.
-
 This is more or less similar to what GNU Binutils strings does.
-
 Does not recognize and extract non-ASCII characters
-
 Some alternative and references:
 https://github.com/fireeye/flare-floss (also included)
 http://stackoverflow.com/questions/10637055/how-do-i-extract-unicode-character-sequences-from-an-mz-executable-file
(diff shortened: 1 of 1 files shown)


chashiv:
> Thank you! Good catch. I think you DCO signature is not correct. Also this is code that is not yet used but a test would be much welcomed, possibly a doctest?

You're Welcome

chashiv:
Updated requested changes with correct DCO and put up a doctest . Also the **is_relative_path** was not returning **bool** values, updated it.

pombredanne:
Thnk you for the updates. Your doctest is failing for now https://travis-ci.org/nexB/scancode-toolkit/jobs/514873900#L1831

chashiv:
> Thnk you for the updates. Your doctest is failing for now https://travis-ci.org/nexB/scancode-toolkit/jobs/514873900#L1831

@pombredanne I will surely do the required changes so as to pass all the cases.

src/textcode/strings.py
```diff
@@ -27,18 +27,16 @@
 
 import re
 import string
+import os
```
pombredanne: Please sort your imports

src/textcode/strings.py
```diff
@@ -198,9 +196,14 @@ def is_relative_path(s):
...
+    >>> is_relative_path("../usr/lib")
+    True
+    >>> is_relative_path("../usr")
+    True
+    >>> is_relative_path("..")
+    True
     """
-    relative = re.compile('^(?:([^/]|\.\.)[\w_\-]+/.*$', re.IGNORECASE).match
-    return relative(s)
+    return not os.path.isabs(s)
```
pombredanne: IMHO this cannot work as the paths could be POSIX or Windows and you may be running on either one OSes.

src/textcode/strings.py
```diff
@@ -299,4 +302,4 @@ def filter_strings(strs, nglen=4):
     import sys
     location = sys.argv[1]
     for s in strings_from_file(location):
-        print(s)
+        print(s)
```
pombredanne: Why did you remove the final LF?

src/textcode/strings.py
pombredanne: os.path is either ntpath or posixpath depending on the OS. So this cannot work here since you do not know:
1. which OS will be running scancode
2. for which OS the path found is
You need a cross OS solution. And `os.path` is not it

chashiv: Should I detect which os scancode will be running on and implement separate regular expression pattern for each case?

src/textcode/strings.py
```diff
@@ -198,9 +196,14 @@ def is_relative_path(s):
     """
     Return True if s looks like a relative posix path.
     Example: usr/lib/librt.so.1 or ../usr/lib
+    >>> is_relative_path("../usr/lib")
+    True
+    >>> is_relative_path("../usr")
+    True
+    >>> is_relative_path("..")
+    True
```
singh1114: A test for false will help understand this thing better.

src/textcode/strings.py
pombredanne: Of course not. The whole point is that scancode should be able to detect a path in a binary that came from any OS and that while running on any OS.

majurg:
@chashiv @pombredanne There have been no updates to this for a long time. I will close this shortly if no response from @chashiv

pombredanne:
@chashiv this has gone stale so I am closing this at this time. Please reopen if you want to finish this!
Thank you ++

PR #1500, opened 2019-04-03
Minor changes to determine supported os

chashiv:
@pombredanne Please provide your feedback to this PR. Just a few changes so as to avoid extraneous **searching** and functions like **startswith** .
Signed-off-by: Shivam Chauhan <chauhanshivam999@gmail.com>

### DIFF
--- a/etc/conf/base.py
+++ b/etc/conf/base.py
@@ -4,7 +4,7 @@ configure.
 """
 
 import sys
-
+import platform
 """
 Check that we run a supported OS and architecture.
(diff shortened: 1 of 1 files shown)


pombredanne:
Thanks... but what are the benefits of this? I have no confidence that your change is guaranteed work in practice... and there is no script to tests this for now. So I cannot accept a change here unless super robust and tested everywhere and frankly changing this script in a mostly cosmetic way is no desirable.

PR #1501, opened 2019-04-03
Use proper str or 'Unicode' for path Sep on Python 3 #295

Abhishek-Dev09:
Python 2 and Python 3 behave quite differently there.Eventually we will be able to use unicode throughout when working fully on Python 3.But until then, we are left with using the os.fsencode and fs.decode backports and dealing only with bytes on Linux and unicode elsewhere.


Abhishek-Dev09:
If anyone has better title than this.Please suggest it.

pombredanne:
@Abhishek-Dev09 thanks. I think the title is OK

pombredanne:
thanks!

PR #1504, opened 2019-04-04
Minor changes to .travis.yml to improve efficiency

chashiv:
- Travis was testing only on **python 2.7.13**, added **python 3.7.2** as well.

- Why build all the branches, should not we build only **develop** branch?

Signed-off-by: Shivam Chauhan <chauhanshivam999@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: python
 
 python:
   - 2.7.13
+  - 3.7.2
 
 # Use a matrix of TEST_SUITE variables to distribute tests to several machines
 # Use a global PYTEST variable to run all of the tests in a uniform fashion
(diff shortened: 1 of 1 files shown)


pombredanne:
I guess you closed this as this cannot work out:

>    Travis was testing only on python 2.7.13, added python 3.7.2 as well.

this cannot work until we port ScanCode to Python 3

>    Why build all the branches, should not we build only develop branch?

The branch we release from is master so develop would never be enough ... and the branches we code from are feature branches from which PR are crafted. We want to make sure a PR is tested/built. This is to catch PR like this one that will massively fail to pass tests. Also this is important to distinguish failures from merge failures at times.

chashiv:
> I guess you closed this as this cannot work out:
> 
> > Travis was testing only on python 2.7.13, added python 3.7.2 as well.
> 
> this cannot work until we port ScanCode to Python 3
> 
> > Why build all the branches, should not we build only develop branch?
> 
> The branch we release from is master so develop would never be enough ... and the branches we code from are feature branches from which PR are crafted. We want to make sure a PR is tested/built. This is to catch PR like this one that will massively fail to pass tests. Also this is important to distinguish failures from merge failures at times.

Got it ! thanks @pombredanne .

PR #1509, opened 2019-04-07
Update copyright year

chashiv:
Updated copyright year in various licenses.
Signed-off-by: "Shivam Chauhan" <"chauhanshivam999@gmail.com">

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,7 +1,7 @@
 Software license
 ================
 
-Copyright (c) 2017 nexB Inc. and others. All rights reserved.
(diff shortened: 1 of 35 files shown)


Abhishek-Dev09:
I think this PRs make no change and no needed.It will be will not be accepted until you use git.

pombredanne:
Thank you for this, but we do not update years unless really needed and this is typically only done by the primary author or modifier of a given module. Otherwise this creates changes that bring no value at all beyond polluting the history so I cannot accept this.

Issue #1502, opened 2019-04-04
extractcode fails to extract docx file - skips


a-tinsmith:
### Description
Steps to reproduce:
* `./extractcode --verbose sample.docx`
* Nothing is extracted

I tried to debug the problem and found that:
* `OfficeDocHandler` is initialised with the properties:
`strict=True`
`filetypes=('zip archive',)`
* During the examination of the file is was determined as `Microsoft Word 2007+`
* Because of the `strict=True` the handler was **not** picked to extract the file.

### System configuration
For bug reports, it really helps us to know:
* What OS are you running on? (Windows/MacOS/Linux)
MacOS 10.14.4
* What version of scancode-toolkit was used to generate the scan file?
`extractcode-libarchive        3.1.2`
* What installation method was used to install/run scancode? (pip/source download/other)
```bash
$ brew install icu4c
$ export PATH="/usr/local/opt/icu4c/bin:$PATH"
$ export PATH="/usr/local/opt/icu4c/sbin:$PATH"
$ pipenv install extractcode-libarchive
```

pombredanne:
@a-tinsmith great catch: I think that this "strict=True" was originally set this way to avoid too much noise around non office documents with an office document extension. That said, if "Microsoft Word 2007+" is  reported as a type, the extraction should occur alright.
Do you have a `scancode -i` for that document so that we can see the various types detected?

a-tinsmith:
@pombredanne , no problem. Here is the full JSON output of `scancode -i <file> --json-pp output.json`:

```json
{
  "headers": [
    {
      "tool_name": "scancode-toolkit",
      "tool_version": "3.0.2",
      "options": {
        "input": "sample-2007.docx",
        "--info": true,
        "--json-pp": "output.json"
      },
      "notice": "Generated with ScanCode and provided on an \"AS IS\" BASIS, WITHOUT WARRANTIES\nOR CONDITIONS OF ANY KIND, either express or implied. No content created from\nScanCode should be considered or used as legal advice. Consult an Attorney\nfor any legal advice.\nScanCode is a free software code scanning tool from nexB Inc. and others.\nVisit https://github.com/nexB/scancode-toolkit/ for support and download.",
      "start_timestamp": "2019-04-04T064438.402889",
      "end_timestamp": "2019-04-04T064438.517834",
      "message": null,
      "errors": [],
      "extra_data": {
        "files_count": 1
      }
    }
  ],
  "files": [
    {
[...]

PR #1518, opened 2019-04-09
Extract MS Office .docx files as archives #1502

a-tinsmith:
Changes:
- `OfficeDocHandler` now recognises MS Office file types as specified in libmagic: Microsoft Word 2007+, Microsoft Excel 2007+, Microsoft Powerpoint 2007+. For reference see https://github.com/file/file/blob/master/magic/Magdir/msooxml
- New unit test for MS Office extractcode
- Updated unit test for typecode
- Fixed an extra space typo in the temporary directory name prefix: 'scancode-tk-tests-'


src/commoncode/testcase.py
```diff
@@ -183,7 +183,7 @@ def get_temp_dir(self, sub_dir_path=None):
             test_tmp_root_dir = os.path.join(scancode_root_dir, 'tmp')
             # not we add a space in the path for testing path with spaces
             test_run_temp_dir = fileutils.get_temp_dir(
-                base_dir=test_tmp_root_dir, prefix='scancode-tk-tests -')
+                base_dir=test_tmp_root_dir, prefix='scancode-tk-tests-')
```
pombredanne: This is a small space that is there for a reason: to ensure that we run tests on paths that contain at least one space, so do you mind reverting this?

a-tinsmith: Updated. I didn't notice the comment one line above!

.gitignore
```diff
@@ -44,6 +44,7 @@ htmlcov
 .idea
 org.eclipse.core.resources.prefs
 .vscode
+.vs
```
pombredanne: Just curious: what would this file be?

a-tinsmith: I was using Visual Studio and it created this folder. This is VS internal bookkeeping of the project.

pombredanne: thx

pombredanne:
Thank you ++

Issue #1521, opened 2019-04-10
scancode installed via pip fails if PyYaml 5.1 is installed


dankegel:
Now that saneyaml is updated to not depend on a particular version of PyYAML, scancode's explicit dependency on a particular PyYAML is getting in the way.

To reproduce (on clean, updated Ubuntu 16.04, with https://launchpad.net/~dank/+archive/ubuntu/python-fixes installed):
```
$ rm -rf ~/.local
$ pip install scancode-toolkit
$ scancode --version
ScanCode version 3.0.2
$ pip list | grep -i pyyaml
PyYAML (3.13)
$ pip install PyYaml==5.1
$ scancode --version
pluggy.manager.PluginValidationError: Plugin 'json-pp' could not be loaded: (PyYAML 5.1 (/home/buildbot/.local/lib/python2.7/site-packages), Requirement.parse('PyYAML<=3.13,>=3.11'))!
```

See also https://github.com/nexB/saneyaml/pull/3#issuecomment-481779602

dankegel:
See https://github.com/nexB/scancode-toolkit/pull/1522

pombredanne:
All merged at last.

PR #1522, opened 2019-04-10
setup.py: Remove explicit dependency on PyYAML 3.1x #1521

dankegel:
Is it this simple?

Untested.  Leaning on the github test machinery.


pombredanne:
> Is it this simple?

I think it is! Thank you: If that works we just then need to update the bundled saneyaml and its version requirement to `>= 0.4`

dankegel:
Verified that 'pip install -e .' after this commit does The Right Thing for me, so from my point of view, this works.

PR #1526, opened 2019-04-12
Improve declared_license storage #1403

linexb:
Improve storage of declared_license for FreeBSD & phpcomposer packages.
Now using correct data structures instead of plain string

Signed-off-by: Li <li@nexb.com>


src/packagedcode/freebsd.py
```diff
@@ -148,20 +181,21 @@ def license_mapper(package_data, package):
     if not licenses:
         return
 
+    declared_license = OrderedDict()
```
pombredanne: I think that we should stay true to the original data structure:
This should be something such as:
`{ 'licenselogic': 'dual', 'licenses': [a list of license strings, ... ]}`

This was we can do the detection much better... e.g. we should NOT try to craft an a pseudo expression in the `declared_license`

pombredanne:
LGTM ! thanks and merging now.

PR #1529, opened 2019-04-12
Fixed string formatting

TechnicallyMay:
There was a broken string format, it is now fixed 
Signed-off-by: Mason Wells <mwells@tellus.us>

### DIFF
--- a/src/scancode/resource.py
+++ b/src/scancode/resource.py
@@ -389,7 +389,7 @@ class Codebase(object):
         def err(_error):
             """os.walk error handler"""
             self.errors.append(
-                ('ERROR: cannot populate codebase: %(_error)r\n' % _error)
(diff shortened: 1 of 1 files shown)


pombredanne:
Thank you ++... good catch! Do you mind adding a Signed-off-by to you commit message? or at least paste one as a comment in this PR and I will add it to the merge commit?
Also do you know if we could craft a unit test for that?

TechnicallyMay:
> Thank you ++... good catch! Do you mind adding a Signed-off-by to you commit message? or at least paste one as a comment in this PR and I will add it to the merge commit?
> Also do you know if we could craft a unit test for that?

Sure we could! I would do it by intentionally causing that error, and asserting that the error is added to `self.errors`. I made this change to get the project working, but if I get some free time I'll take a look at doing this.

Issue #1519, opened 2019-04-09
Make tests smarter to run tests faster


linexb:
### Description
Some ideas form Philippe:
1. split scancode in its libraries commoncode, licensedcode, etc (rather than the big mono repo it is today)
2. be smarter on what is tested when
- only retest things that are affected by a commit in general
- restest all in some cases such as a tag or master or develop merge

There are some pytest plugins that might help for this.

pombredanne:
Chat log with py.test developers on #pylib on Freenode:
```
<pombreda> hum, I am still confused by y.test test collection https://gist.github.com/pombredanne/b268ac258ca7d773c4b35dc9b431dde6
<pombreda> that's a minimal example, the real one is https://github.com/nexB/scancode-toolkit/pull/1531
* pombreda is likely confused by the firstresult pluggy thing OR there is a bug
<pombreda> s/y.test/py.test
<pombreda> basically I have a pytest_ignore_collect plugin that ignores EVERYTHING
<pombreda> but things are still collected 
<cdunklau> pombreda: something else is returning non-None before your function
<pombreda> cdunklau: so that would have to be the standard code for collection then?
<pombreda> in particular this in main?
<pombreda> https://github.com/pytest-dev/pytest/blob/49d690d137be1edc79bfffce6720d1186c3eabf2/src/_pytest/main.py#L295
<pombreda> firstresult smells to me a tad like a wart, from a distance 
<pombreda> not that I would know how  wart smells :P
<cdunklau> pombreda: so i have tests/conftest.py and tests/test_blarg.py. When i run `pytest tests`, the hook is invoked, and nothing is collected
<cdunklau> pombreda: when i run `pytest tests/test_blarg.py`, the hook is not invoked, and the tests run
<pombreda> cdunklau, hum.... thx, that should give enough to dive and fix this. I want the combo of passing a list of paths AND selectively further filtering things from these paths
<pombreda> the weird thing is that the main.py hook is properly invoke in this case
<pombreda> no, my bad
<pombreda> shrikes
<pombreda> so for now I cannot do both 
<pombreda> even "py.test foo.py --ignore foo.py" does not ignore foo.py at all
<blueyed> pombreda: re "basically I have a pytest_ignore_collect plugin that ignores EVERYTHING" I would say this sounds like a bug/missing feature, but this hook might just not be called for args.
<blueyed> Unfortunately collection code is duplicated in main and python, and messy in general.. mostly since namespace packages were added there IIRC.
[...]

PR #1531, opened 2019-04-16
Do not run all the tests all the times #1598 #1519

pombredanne:
This is a pytest conftest plugin that bring new options:

- To support the Python 3 port as part of #295 it adds a new test marker `scanpy3` to mark a ScanCode test as supported on Python 3. 
- To support running fewer tests faster for #1598 it adds two new tests markers: `scanslow` to mark a ScanCode test as slow, long running and `scansmoke`  to mark a test as a quick "smoke" test.   

The markers full names are:
```
@pytest.mark.scanpy3: Mark a ScanCode test as supported on Python 3.
@pytest.mark.scansmoke: Mark a ScanCode test as a smoke test.
@pytest.mark.scanslow: Mark a ScanCode test as a slow, long running test.
```

... and then can be used to tag a whole test module, or a test class or a single test function or method.

There also are two new options that work with these markers:
- `--force-py3` will force the run of all tests on Python 3 even if they have not been tagged as supported with the `scanpy3` marker. By default only `scanpy3`-marked tests run on Python3 for now during the port #295
```
  --force-py3           Run all the scancode tests on Python 3, irrespective
                        of their markers. If not selected, only the tests
                        marked as compatible with Python 3 will run on Python
                        3. Use the @pytest.mark.scanpy3 marker to mark test
                        functions, classes or modules that are supported on
                        Python3. This marker is there to help with the Python
                        3 port and will be removed in the future.
```

- `--test-suite={smoke,standard,all}`  can select which tests to run. "standard" is the default and excludes `scanslow`-marked tests.

```
  --test-suite={smoke,standard,all}
                        Select which test suite to run: "smoke" is a minimal
                        integration test suite; "standard" is the standard
                        test suite that is designed to run reasonably fast; it
                        does not include "smoke" tests and "slow" tests."all"
                        runs all the tests including "slow" (long running),
                        "smoke" and "standard" tests. Use the
                        @pytest.mark.scansmoke marker to mark a test as a
                        "smoke" test.Use the @pytest.mark.scanslow marker to
                        mark a test as "slow" test that will only run when
                        running "all" tests.

```

In addition you can optionally only rerun the tests roughly impacted by code changes for #1519:
```
  --changed-only        Run only the subset of tests impacted by your
                        changes.If selected, you can provide an optional
                        --base-branch and the changes are checked against that
                        branch. Otherwise, a git diff is made against the
                        current branch.
  --base-branch=BASE_BRANCH
                        Optional branch name to use if --changed-only is
                        selected.
```


Abhishek-Dev09:
@pombredanne : This PR is about tests that you have commented in #295 at [here](https://github.com/nexB/scancode-toolkit/issues/295#issuecomment-480854808).

pombredanne:
@Abhishek-Dev09 

> This PR is about tests that you have commented in #295 at [here](https://github.com/nexB/scancode-toolkit/issues/295#issuecomment-480854808).

yes, I am experimenting with using some py.test local plugins (conftest.py) to eventually use this as a way to selectively activate or deactivate running some tests  on Python 3 as a prep for the port.

Abhishek-Dev09:
What about this PR. Is there any any still work left?

pombredanne:
@Abhishek-Dev09 yes. The code does not work as it is.
The thing to implement is what is recommended in https://github.com/nexB/scancode-toolkit/issues/1519#issuecomment-486156268

pombredanne:
All tests passing where they should. Merging!

PR #1544, opened 2019-04-23
Chef metadata parser

JonoYang:
This is a new Package scanning parser for Chef `metadata.json` files and its associated tests. This does not handle Chef `metadata.rb` files.


JonoYang:
I've updated the Chef metadata parser to be able to parse metadata.rb files. It does this by using Pygments to parse `metadata.rb` and a custom formatter to format the contents to JSON, where each variable name in `metadata.rb` is a key and the variable value is the value. I then pass the data to `build_package()` as normal.

pombredanne:
I had rebased this and all tests pass now. The Windows error was because of #1555

PR #1557, opened 2019-05-08
Improved Example

sk9331657:
- [:heavy_check_mark:]  Added Example for reading copyright in directory while ignoring header files and saving result in Json file in Pretty Printed Format .

### DIFF
--- a/src/scancode/help.py
+++ b/src/scancode/help.py
@@ -24,6 +24,11 @@ Print scan results to stdout as pretty formatted JSON.
 
     scancode -lc --package --ignore README --processes 4 --json-pp - samples/
 
+Scan a file for copyright and packages. Print scan results to file as pretty
+formatted JSON.
+
+    scancode --copyright --package  samples/zlib/zlib.h --json-pp - > scan.json
+
 Scan a directory while ignoring all files with .txt extension.
 Print scan results to stdout as pretty formatted JSON.
 It is recommended to use quotes around glob patterns to prevent pattern
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -16,3 +16,4 @@ The following organizations or individuals have contributed to ScanCode:
 - Tushar Mittal @techytushar
 - Martin Petkov @MartinPetkov
 - Mrinal Paliwal @mriiinal
+- Shubham Kumar @sk9331657


sk9331657:
@pombredanne  Can you please review??

src/scancode/help.py
```diff
@@ -24,6 +24,11 @@
 
     scancode -lc --package --ignore README --processes 4 --json-pp - samples/
 
+Scan a directory while ignoring header(.h) files . Scan for copyright and license.
+Print scan results to file as pretty formatted JSON.
```
pombredanne: You use a redirect from the standard output to a file which is a more advanced feature. Could your example use a file directly?

sk9331657: > Thank you! See my small comment

Ok Will do that Sir.

sk9331657:
@pombredanne  Please review

majurg:
@sk9331657 This is not quite what was requested by @pombredanne. Your example should show how to exclude ALL header files, not just the single specific one you have in your example.

src/scancode/help.py
```diff
@@ -24,6 +24,11 @@
 
     scancode -lc --package --ignore README --processes 4 --json-pp - samples/
 
+Scan a file for copyright and packages. Print scan results to file as pretty
+formatted JSON.
+
+    scancode --copyright --package  samples/zlib/zlib.h --json-pp - > scan.json
```
pombredanne: Why would you promote the the use of a stdout output with a redirection to a file?
Also this kind of example might be best in the doc than in the cli... @AyanSinhaMahapatra what do you think?

AyanSinhaMahapatra: IMHO if the scan target is only a file or two then using stdout output makes sense, otherwise, file output works best(your opinions?). Are you asking about having examples only in the doc and not CLI(minimal examples), if so then yes that would be best. Do you suggest any changes to the current doc then?

pombredanne: > IMHO if the scan target is only a file or two then using stdout output makes sense, otherwise, file output works best(your opinions?). Are you asking about having examples only in the doc and not CLI(minimal examples), if so then yes that would be best. Do you suggest any changes to the current doc then?

@AyanSinhaMahapatra I am not suggesting any change. Your call if you want to integrate something like that in the doc. Otherwise I would like this PR closed if we do not merge it.


as for the stdout... yes output to `-` is fine what I find dubious would be to recommend to use `>` a redirection to a file as a standard example (this is always possible but not the recommended default way IMHO)

pombredanne:
@sk9331657 I am closing this for now as the feedback was not applied fully. Please feel free to reopen and/or resubmit!

PR #1582, opened 2019-06-05
Find python packages #253

linexb:


linexb:
@pombredanne  A new request is ready. :)

PR #1589, opened 2019-06-08
Improve license detection rules and detection

pombredanne:
This is an "omnibus" PR to add new and improved license detection rules and other license-related refinements.


pombredanne:
All green... merging!

PR #1591, opened 2019-06-09
Add Initial Debian Packaging

aj4ayushjain:
These are files generated using the `dh_make` command on upstream source tarball dir and after some alterations are made to them which throws a lot of errors on building via
`sudo dpkg-buildpackage -b`

Signed-off-by: aj4ayushjain <aj4ayushjain@gmail.com>

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,6 +1,6 @@
 trigger:
   branches:
-    include: ['pipeline']
+    include: ['*']
   tags:
     include: ['*']
 


etc/debian/control
```diff
@@ -0,0 +1,36 @@
+Source: scancode-toolkit
+Section: unknown
+Priority: optional
+Maintainer: root <root@40825ee16090>
+Build-Depends: debhelper (>= 11), dh-python, python, python-setuptools
+Standards-Version: 4.1.3
+Homepage: <insert the upstream URL, if relevant>
+X-Python-Version: >= 2.6
```
pombredanne: For now this has to be 2.7 exactly until the port for Python 3 is complete

etc/debian/control
```diff
@@ -0,0 +1,36 @@
...
+X-Python3-Version: >= 3.2
```
pombredanne: This is not supported

etc/debian/control
```diff
@@ -0,0 +1,36 @@
...
+
+Package: python-scancode-toolkit
```
pombredanne: Why would this be prefixed with `python` ?

etc/debian/control
aj4ayushjain: Ok will take down this one in next iteration.

etc/debian/control
aj4ayushjain: Correctly pointed out ! 
Is scancode-toolkit been used as library(python-module) apart from application.
In debian there is a general convention to name python module package as python-module*.

aj4ayushjain:
Will make a PR soon with updated files and adding .ex files to .gitignore

PR #1592, opened 2019-06-11
Define base os strings for each OS #295

Abhishek-Dev09:
The os.sep string is now unicode so we adapt accordingly.


pombredanne:
All good.Thanks!

PR #1593, opened 2019-06-11
Remove references to funname and func_name #295

Abhishek-Dev09:
These are weird and either changed (func_name is now __name__) or
incorrect (funname was a typo and never really existed/was not used).


pombredanne:
Thanks. :+1:

PR #1594, opened 2019-06-11
Use unicode for ENV var names #295

Abhishek-Dev09:
Environment variable names are now unicode and no longer bytes in Python 3.


pombredanne:
There is on test failing on Windows but not related. Merging then.

PR #1599, opened 2019-06-13
Add new compatibility layer for Python 2 and 3 #295

Abhishek-Dev09:
This module used in condition to apply on Python 2 and Python 3


pombredanne:
I am merging this, but I would need to understand what is your approach for this: will you be using six or this kind of local compat layer?

Issue #1447, opened 2019-03-14
Add Conda meta.yml manifest parser


pombredanne:
### Description
See for example https://github.com/bioconda/bioconda-recipes/blob/master/recipes/abeona/meta.yaml#L42

This is the format used by  Conda and bioconda

shubhscoder:
@pombredanne I am interested in taking this ! Can you guide me on this ? Thank you so much!

shubhscoder:
@pombredanne is  (scancode-toolkit/src/packagedcode/jar_manifest.py) the right file to look?

PyAgni:
@pombredanne I'm willing to do this if anyone already hasn't been assigned. I have a few questions:

- We have to use PyYaml for this. Right? Or this has to be done by any other method?
- Once parsed, the collected metadata from the manifest has to be stored in a variable(say 'package'). And make an object that returns that package.
- Do we also need to make unittests for the parser?

And since I'm already working on #295 I'll try to make this python 2/3 compatible.

shubhscoder:
Hi Agni, I am trying to complete this as my first issue. Would you mind
helping me in this ?

PyAgni:
Hello @shubhscoder . Although I'm new here, I would love to help you. I don't have prior experience with yml files but the format seems pretty straight forward if you have worked JSON files or something similar. I think we have to write a script to parse all the contents of the meta.yml file like package details,requirements,about etc. Then we would also probably need to create unittests for that script. (See tests folder) 
Apart from this @pombredanne  will be able to clarify more about it.

shubhscoder:
Hello @PyAgni we can work together on this. Thank you!

pombredanne:
@PyAgni @shubhscoder you can see the many parsers available in `src/packagedcode`. The main purpose is to normalize the data in our Package model objects. A recent and simple one was contributed by @ritiek in #1426 for Rust Cargo for instance. 

About  YAML: we do not use PyYAML directly.  We use instead the `saneyaml` parser. Search for "saneyaml" for details.
And we always write automated tests of course.

pombredanne:
I would be good to start looking into this. @linexb do you mind taking a look?

epruesse:
FYI: While the files in the recipes are called `meta.yaml`, they are not actually valid yaml. Example:

```
{% set name = "package-name" %}
package:
  name: {{ name }}
  version: 0.10
source:
  url: https://somewhere   # [py2k and osx]
[...]

PR #1605, opened 2019-06-18
Conda support #1447 

linexb:
Add Conda package parser


linexb:
@pombredanne  The Conda parser request is ready, please review. :)

src/packagedcode/conda.py
```diff
@@ -0,0 +1,215 @@
...
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from collections import OrderedDict
+import io
+import logging
+import sys
+
+import attr
+from  saneyaml import load as yamlload
```
pombredanne: that's minor but you could have imported only saneyaml instead

src/packagedcode/conda.py
```diff
@@ -0,0 +1,215 @@
...
+    @classmethod
+    def recognize(cls, location):
+        return parse(location)
+
+    def compute_normalized_license(self):
+        return models.compute_normalized_license(self.declared_license)
+
+
+def is_conda_yaml(location):
+    return (filetype.is_file(location) and fileutils.file_name(location).lower() == 'meta.yaml')
```
pombredanne: are there other manifest files than a meta.yml?

src/packagedcode/conda.py
```diff
@@ -0,0 +1,215 @@
...
+    default_download_baseurl = None
+    default_api_baseurl = None
+
+    @classmethod
+    def recognize(cls, location):
+        return parse(location)
+
+    def compute_normalized_license(self):
+        return models.compute_normalized_license(self.declared_license)
+
```
pombredanne: Is a meta.yml at the root of a package or elsewhere? IF not at the root, we need to override the method t get the root.

src/packagedcode/conda.py
```diff
@@ -0,0 +1,215 @@
...
+    logger.setLevel(logging.DEBUG)
+
+    def logger_debug(*args):
+        return logger.debug(' '.join(isinstance(a, string_types) and a or repr(a) for a in args))
+
+
+@attr.s()
+class CondaPackage(models.Package):
+    metafiles = ('meta.yaml',)
+    filetypes = ('.yaml',)
```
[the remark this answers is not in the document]
linexb: Yes, I agree. I think "ASCII text" would be better and plus the extension ".yaml" here.

src/packagedcode/conda.py
```diff
@@ -0,0 +1,215 @@
...
+    mimetypes = ('application/json',)
```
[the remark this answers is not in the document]
linexb: "text/plain" would be better here.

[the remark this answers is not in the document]
linexb: .tar.bz2 file types is also possible here...

src/packagedcode/conda.py
```diff
@@ -0,0 +1,215 @@
...
+from six import string_types
+
+from commoncode import filetype
+from commoncode import fileutils
+from packagedcode import models
+from packageurl import PackageURL
+
+
+"""
+Parse Conda manifests, see https://getcomposer.org/ and
```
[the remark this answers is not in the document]
linexb: Fixed. :)

src/packagedcode/conda.py
linexb: It can be under info/recipe.tar/recipe/meta.yaml too. Thanks for the kind reminder, I tried to add the get_package_root function to overwrite it.

src/packagedcode/conda.py
linexb: Agree, add more generic match condition for "*.yml" or "*.yaml" files

src/packagedcode/conda.py
linexb: Yep, fixed by using saneyaml package from commoncode.

src/packagedcode/conda.py
```diff
@@ -0,0 +1,215 @@
...
+
+@attr.s()
+class CondaPackage(models.Package):
+    metafiles = ('meta.yaml',)
+    filetypes = ('.yaml',)
+    mimetypes = ('application/json',)
+    default_type = 'conda'
+    default_web_baseurl = None
+    default_download_baseurl = None
+    default_api_baseurl = None
```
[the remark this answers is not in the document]
linexb: This a little hard to form a download  url right now. But I think we can keep this as a default_download_baseurl and we can use it to form a real download url for the resource in the future. ;)

linexb:
@pombredanne  I have fixed according to your latest comments, and it's ready again for review. Thanks.

pombredanne:
@linexb Thank you!

pombredanne:
@linexb you have committed all the files of  conda package for testing but we only need a yaml file instead, correct?

src/packagedcode/conda.py
```diff
@@ -0,0 +1,234 @@
...
+    logger.setLevel(logging.DEBUG)
+
+    def logger_debug(*args):
+        return logger.debug(' '.join(isinstance(a, string_types) and a or repr(a) for a in args))
+
+
+@attr.s()
+class CondaPackage(models.Package):
+    metafiles = ('meta.yaml',  'META.yml',)
+    filetypes = ('.bz', '.bz2', 'bzip2',)
```
pombredanne: I am not sure we want to cast so wide a net. ... Let's chat about this

pombredanne: @linexb ping?

linexb: Thanks, I have removed extra test files. :)

src/packagedcode/conda.py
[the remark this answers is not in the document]
linexb: @pombredanne  Agree. As we have discussed in slack, I will remove the filetypes and minetypes limit right now.

src/packagedcode/conda.py
linexb: @pombredanne  Yep, I will remove "filetypes and minetypes" limit  :)

linexb:
@pombredanne  I have refined according to the latest comments and our discussion, please review.  :)

pombredanne:
There are likely still things to collect with https://github.com/nexB/scancode-toolkit/issues/1447#issuecomment-513074714 ... but this is a great start and I am merging this now. Thanks!

PR #1607, opened 2019-06-18
 Use 'str' or compat instead of 'basestring' on Python 3 #295

Abhishek-Dev09:
Python 2 had two string types: Unicode and non-Unicode. But there was also another type: basestring.It was an abstract type, a superclass for both the str and unicode types. It couldn’t be called or instantiated directly, but you could pass it to the global isinstance() function to check whether an object was a Unicode or non-Unicode string. In Python 3, there is only one string type, so basestring has no reason to exist.

### DIFF
--- a/src/commoncode/fileutils.py
+++ b/src/commoncode/fileutils.py
@@ -26,14 +26,7 @@ from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
 
-# Python 2 and 3 support
-try:
-    # Python 2
-    unicode
-    str = unicode  # NOQA
-except NameError:
-    # Python 3
-    unicode = str  # NOQA
+from commoncode import compat
 
 try:
     from os import fsencode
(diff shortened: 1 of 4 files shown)


src/commoncode/fileutils.py
```diff
@@ -83,7 +83,7 @@ def logger_debug(*args):
     logger.setLevel(logging.DEBUG)
 
     def logger_debug(*args):
-        return logger.debug(' '.join(isinstance(a, basestring) and a or repr(a) for a in args))
+        return logger.debug(' '.join(isinstance(a, str) and a or repr(a) for a in args))
```
pombredanne: Why do not you use compat there too?

Abhishek-Dev09: This can also be used.If you want i can make changes to compat but i have already close that related PR.

Abhishek-Dev09: As both can be used because it passed on both version of python 2 and 3.

pombredanne: We want to be consistent and always refer and use `compat` when we tests for strings. So please use compat here.

Abhishek-Dev09: Ok, i am doing.

PR #1634, opened 2019-06-30
Make configure work on Python 3 #295

pombredanne:
This is to ensure configure works without any side effects on all OSes in Python 2 and 3.


pombredanne:
At this stage we have enough green to merge. ... so I am going for it.

PR #1639, opened 2019-07-02
Adding Path.py as dependencies for porting path #295

Abhishek-Dev09:
As Path.py supports both Python 2 and 3. No need to tackle with bytes and unicode on linux on both version of Python.


pombredanne:
LGTM. There are some heisen failures on the CI, but that's OK

PR #1641, opened 2019-07-04
Use byte path only on Linux and Python 2 #295

Abhishek-Dev09:
We now use bytes in commoncode.fileutils only if  on Linux and Python 2 and unicode everywhere else.


pombredanne:
All green now... Merging! thanks.

PR #1646, opened 2019-07-11
Add new License rules and licenses

pombredanne:
This is an "omnibus" PR for miscellaneous new and improved licenses and license rules
The bulk of it are updates from the review of tests.

It also contains minor update to copyright detection.


pombredanne:
We are all green so I am merging this.

PR #1648, opened 2019-07-12
Port commoncode module to Python 3 #295

Abhishek-Dev09:
The objective of PR is to ensure that all the tests of commoncode pass on both Python 2 & 3 by use of conffest plugin. 


src/commoncode/fileset.py
```diff
@@ -44,8 +45,8 @@
     logger.setLevel(logging.DEBUG)
 
 
-POSIX_PATH_SEP = b'/' if on_linux else '/'
-EMPTY_STRING = b'' if on_linux else ''
+POSIX_PATH_SEP = b'/' if on_linux and py2 else '/'
```
pombredanne: Could you use instead `POSIX_PATH_SEP = b'/' if on_linux and py2 else u'/'` so it is very clear that we deal with bytes vs. unicode texts

src/commoncode/paths.py
```diff
@@ -38,15 +38,21 @@
...
 
-POSIX_PATH_SEP = b'/' if on_linux else '/'
-WIN_PATH_SEP = b'\\' if on_linux else '\\'
-EMPTY_STRING = b'' if on_linux else ''
+if on_linux and py2:
+   POSIX_PATH_SEP = b'/' 
+   WIN_PATH_SEP = b'\\' 
+   EMPTY_STRING = b''  
+else:
+   POSIX_PATH_SEP = '/'
```
pombredanne: Same as above please use an explicit `u` prefix in this case to make it very clear that we have bytes vs. unicode:    `POSIX_PATH_SEP = u'/'`

src/commoncode/text.py
```diff
@@ -181,10 +184,12 @@ def toascii(s, translit=False):
         else:
             converted = unicodedata.normalize('NFKD', s).encode('ascii', 'ignore')
     except:
-        converted = str(s.decode('ascii', 'ignore'))
+        if py2:
+            converted = compat.unicode(s.decode('ascii', 'ignore'))
+        else:
+            converted = compat.unicode(s)
```
pombredanne: You are no longer normalizing to ASCII on Python3.

src/commoncode/version.py
```diff
@@ -54,9 +55,9 @@ def VERSION_PATTERNS_REGEX():
 ]]
 
 
-POSIX_PATH_SEP = b'/' if on_linux else '/'
-EMPTY_STRING = b' ' if on_linux else ' '
-VERSION_PREFIX = b'v' if on_linux else 'v'
+POSIX_PATH_SEP = b'/' if on_linux and py2 else '/'
```
pombredanne: Same as above: use `u` prefix in these cases

tests/commoncode/test_command.py
```diff
@@ -165,4 +165,4 @@ class MockOs(object):
 
         unicode_path = u'foo\udcb1bar'
         command.update_path_environment(unicode_path, MockOs)
-        {'PATH': 'c:\\bin\\foo\xb1bar;foo\xb1bar;c:\\windows;C:Program Files'}
+        {'PATH': 'c:\\bin\\foo\xb1bar;foo\xb1bar;c:\\windows;C:Program Files'}
```
pombredanne: You are missing a trailing LF

tests/commoncode/test_fileset.py
```diff
@@ -30,14 +30,16 @@
...
+pytestmark = pytest.mark.scanpy3 #NOQA
 
 class FilesetTest(commoncode.testcase.FileBasedTesting):
     test_data_dir = os.path.join(os.path.dirname(__file__), 'data')
 
     def test_load(self):
         irf = self.get_test_loc('fileset/scancodeignore.lst')
         result = fileset.load(irf)
-        assert ['/foo/*', '!/foobar/*', 'bar/*', '#comment'] == result
+        assert [b'/foo/*', b'!/foobar/*', b'bar/*', b'#comment'] == result
```
pombredanne: How could this possibly work on Python 3?

tests/commoncode/test_paths.py
```diff
@@ -83,12 +86,18 @@ def test_safe_path_posix_style_absolute_trailing_slash(self):
 
     def test_safe_path_posix_style_french_char(self):
         test = paths.safe_path('/includes/webform.compon\xc3nts.inc/')
-        expected = 'includes/webform.componAnts.inc'
+        if py2:
+           expected = 'includes/webform.componAnts.inc'
+        else:
+           expected = 'includes/webform.compon_nts.inc'
```
pombredanne: Why do not we get the same transliteration on Python 2 and 3?

tests/commoncode/test_paths.py
```diff
@@ -83,12 +86,18 @@ def test_safe_path_posix_style_absolute_trailing_slash(self):
...
         assert expected == test
 
     def test_safe_path_posix_style_chinese_char(self):
         test = paths.safe_path('/includes/webform.compon\xd2\xaants.inc/')
-        expected = 'includes/webform.componS_nts.inc'
+        if py2:
+           expected = 'includes/webform.componS_nts.inc'
+        else:
+           expected = 'includes/webform.compon__nts.inc'
```
pombredanne: Same as above ... I would expected the same results on both Pythons.

azure-pipelines.yml
```diff
@@ -167,8 +167,18 @@ jobs:
...
+                    tests/commoncode/test_fileset.py \
                     tests/commoncode/test_fileutils.py \
-                    tests/commoncode/test_functional.py
+                    tests/commoncode/test_functional.py \
+                    tests/commoncode/test_ignore.py \
+                    tests/commoncode/test_paths.py \
+                    tests/commoncode/test_saneyaml.py \
+                    tests/commoncode/test_timeutils.py \
+                    tests/commoncode/test_urn.py \
+                    tests/commoncode/test_version.py
```
pombredanne: At this stage if all test modules are importable you may be able to use `bin/py.test -vvs tests/commoncode` more simply?

tests/commoncode/test_command.py
Abhishek-Dev09: no changes here !

tests/commoncode/test_fileset.py
Abhishek-Dev09: console says bytes like object required not str

tests/commoncode/test_filetype.py
```diff
@@ -73,7 +77,10 @@ def test_get_type(self):
         # symlinks and special files are not supported on win
         if on_posix:
             expected += [ ('2-SYMTYPE', 'l'), ('6-FIFOTYPE', 's'), ]
-
+        
+        if on_windows and py3:
+            expected += [ ('2-SYMTYPE', 'l') ]
```
Abhishek-Dev09: Yesterday, i am talking about this line.

tests/commoncode/test_paths.py
Abhishek-Dev09: According to console, it requires.

src/commoncode/text.py
Abhishek-Dev09: Normalising to ascii on Py3 throwing an error

tests/commoncode/test_fileset.py
pombredanne: So may be the file should be opened in binary mode only on Python 2?

src/commoncode/fileset.py
```diff
@@ -186,9 +187,12 @@ def load(location):
...
     assert (os.path.exists(fn) and os.path.isfile(fn)), msg
-    with open(fn, 'rb') as f:
-        return [l.strip() for l in f if l and l.strip()]
-
+    if on_linux and py2:
+        with open(fn, 'rb') as f:
+           return [l.strip() for l in f if l and l.strip()]
+    else:
+        with open(fn, 'r') as f:
+           return [l.strip() for l in f if l and l.strip()]
```
pombredanne: To avoid code duplication you could write instead:
```
    mode = 'rb' if on_linux and py2 else 'r'
    with open(fn, mode) as f:
       return [l.strip() for l in f if l and l.strip()]
```

src/commoncode/hash.py
```diff
@@ -26,9 +26,12 @@
 from __future__ import division
 from __future__ import print_function
 
+import binascii
+
```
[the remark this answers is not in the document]
pombredanne: I fixed this

PR #1651, opened 2019-07-16
Add fingerprint plugin

arnav-mandal1234:
Added new fingerprint plugin.


plugins/scancode-fingerprint/README.md
```diff
@@ -0,0 +1 @@
+A ScanCode scan plugin to generate fingerprints using Simhash algorithm
```
majurg: Can you add a bit more detail here? Maybe some short background on the algorithm or some links AND how to go about running it from the command line?

plugins/scancode-fingerprint/src/plugin_fingerprint/fingerprint.py
```diff
@@ -0,0 +1,148 @@
...
+        """
+        bytes = binascii.unhexlify(fingerprint_hex)
+        result = self.bitarray_from_bytes(bytes)
+
+        return result
+
+
+    def hamming_distance(self, fingerprint1, fingerprint2):
+        """
+        Return hamming distance between two given fingerprints
```
majurg: Can you add a bit more detail here in the docstring as to what hamming distance numbers mean what? For example, what would one expect the hamming distance to be for similar files and what would it be for completely different files. You can sort of get an idea of this looking at the test cases, but it would nice to have that information in the source code for future reference.

plugins/scancode-fingerprint/src/plugin_fingerprint/fingerprint.py
```diff
@@ -0,0 +1,148 @@
...
+HASH_LENGTH = 128
+SHINGLE_LENGTH = 3
+
+class Simhash:
+    """
+    Fingerprint class to generate fingerprints for files used for similarity matching
+    """
+    def __init__(self):
+        self.tokens = []
+
```
JonoYang: Minor formatting issue: there should be one new line between methods and two new lines between functions and classes.

PR #1657, opened 2019-07-19
Make plugincode importable on Python 3 #295

Abhishek-Dev09:
Objective of this PR is to ensure that plugincode is importable on Python 3 with help of conffest plugin.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,6 +26,7 @@ jobs:
               extractcode: bin/py.test -n 2 -vvs --test-suite=all tests/extractcode
               cluecode: bin/py.test -n 2 -vvs --test-suite=all tests/cluecode
               packagedcode: bin/py.test -n 2 -vvs --test-suite=all tests/packagedcode
+              plugincode: bin/py.test -n 2 -vvs tests/plugincode
 
     - template: etc/ci/azure-win.yml
       parameters:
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -191,6 +194,7 @@ jobs:
           python_architecture: x86
           test_suites:
               commoncode: Scripts\py.test -vvs tests\commoncode
+              plugincode: Scripts\py.test -vvs tests/plugincode
```
pombredanne: Please use a backslash in Windows paths

tests/plugincode/test_plugincode.py
```diff
@@ -0,0 +1,37 @@
...
+from plugincode import output
+from plugincode import post_scan
+from plugincode import pre_scan
+from plugincode import scan
+
+import pytest
+pytestmark = pytest.mark.scanpy3  # NOQA
+
+def test_dummy():
+    pass
```
pombredanne: To make it more explicit that we are only testing the import could instead move them to the test function?
Also you are missing basic imports for future and absolute imports

```
def test_plugincode_can_be_imported():
    import plugincode
    from plugincode import location_provider
    from plugincode import output_filter
    from plugincode import output
    from plugincode import post_scan
    from plugincode import pre_scan
    from plugincode import scan
```

Issue #1643, opened 2019-07-10
Revisit parsers in packagedcode


JonoYang:
In updating some package parsers for (#1588), I noticed that some of them are not being used at all despite being in the list of package parsers (e.g. https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/pypi.py)

pombredanne:
@JonoYang good catch!

pombredanne:
A good way to check would be to verify if we have at least one top level command line test for each package type.

linexb:
I have added the argument "--list-package" to show all supported package code through command.

The usage and result is like this:
```
./scancode  --list-packages
--------------------------------------------
Package: about
  class: packagedcode.about:AboutPackage
  metafiles: *.ABOUT

--------------------------------------------
Package: android
  class: packagedcode.models:AndroidApp
  extensions: .apk
  filetypes: zip archive

--------------------------------------------
Package: android-lib
  class: packagedcode.models:AndroidLibrary
  extensions: .aar
  filetypes: zip archive

--------------------------------------------
Package: axis2
  class: packagedcode.models:Axis2Mar
  metafiles: META-INF/module.xml
  extensions: .mar
  filetypes: java archive , zip archive

--------------------------------------------
Package: bower
  class: packagedcode.bower:BowerPackage
  metafiles: bower.json, .bower.json


```

pombredanne:
@linexb thanks... could you also add a tests for each know package type to make sure this is detected when using a command line scan?

linexb:
Here is all packages we supported now and the status means if we have the test cases for it or not.

Package: about
Status: Yes

--------------------------------------------
Package: android
Status: No

--------------------------------------------
Package: android-lib
Status: No

--------------------------------------------
Package: autotools
Status: No

--------------------------------------------
Package: axis2
Status: No

--------------------------------------------
Package: bazel
Status: No

--------------------------------------------
Package: bower
Status: Yes

--------------------------------------------
Package: buck
Status: No

--------------------------------------------
Package: cab
Status: No

--------------------------------------------
Package: cargo
Status: Yes

--------------------------------------------
Package: chef
Status: Yes

--------------------------------------------
Package: chrome
Status: No

--------------------------------------------
Package: composer
Status: Yes

--------------------------------------------
Package: conda
Status: Yes

--------------------------------------------
Package: cpan
Status: No

--------------------------------------------
Package: deb
Status: No

--------------------------------------------
Package: dmg
[...]

PR #1663, opened 2019-07-25
Add --list-package command line to show all supported packages #1643

linexb:
Add --list-package command line to show all supported packages


linexb:
@pombredanne  This pull request is ready. Please review. Thanks. :)

src/scancode/cli.py
```diff
@@ -62,6 +62,8 @@
 from commoncode.fileutils import POSIX_PATH_SEP
 from commoncode.timeutils import time2tstamp
 
+from packagedcode import PACKAGE_TYPES
```
pombredanne: IMHO we cannot import this globally... Instead make this an option in the packagedcode plugin and not in the main cli.py

src/scancode/cli.py
```diff
@@ -356,6 +379,12 @@ def print_options(ctx, param, value):
     help='Show the list of available ScanCode plugins and exit.',
     help_group=DOC_GROUP, cls=CommandLineOption)
 
+@click.option('--list-packages',
```
pombredanne: This should go in the packagedcode module and plugin instead

linexb: @pombredanne  Got it, thanks. I will update according to this idea.

src/scancode/cli.py
linexb: @pombredanne  Thanks, I have moved the the option in the packagedcode plugin instead.

linexb:
@pombredanne  The pull request is ready. The changes include:
1. Fix according to your comments to move the command in packagedcode plugin
2. Refine the recognize function of python package to make sure python scan works.
3. Add parse function of PKG-INFO file.

pombredanne:
@linexb Thank you!
Merging.

PR #1664, opened 2019-07-25
Port extractcode to Python 3 #295

Abhishek-Dev09:
The objective of PR is to ensure that all the tests of extractcode pass on both Python 2 & 3 by use of conffest plugin.

Signed-off-by: Abhishek Kumar <Abhishek.kasyap09@gmail.com>


src/commoncode/fileutils.py
```diff
@@ -185,6 +185,17 @@ def get_temp_dir(base_dir=scancode_temp_dir, prefix=''):
...
+    """
+    Return the `pth` path string either as encoded bytes if on Linux and using
+    Python 2 or as a unicode/text otherwise.
+    """
+    if on_linux and py2 and isinstance(pth, compat.unicode):
+        return fsencode(pth)
+    if not isinstance(pth, compat.unicode):
+        return fsdecode(pth)
+    return pth
+
```
pombredanne: @Abhishek-Dev09 This is wrong IMHO
We should instead have something like this:
```
    if on_linux and py2:
        if not isinstance(pth, compat.bytes):
            pth = fsencode(pth)
        return path
    if not isinstance(pth, compat.unicode):
        return fsdecode(pth)
    return pth

```

Abhishek-Dev09: This changes have been made by you only.

pombredanne: Sure, that does not mean that you should not fix it.

Abhishek-Dev09: i have fixed

pombredanne:
We are now green!
at last... There is a test not completed yet on the slower Appveyor But I am going to merge anyway.

src/commoncode/fileutils.py
pombredanne: You can never hurt me!

PR #1667, opened 2019-07-29
Update version range of dependency 'attrs'

qduanmu:
* attrs 'fields_dict' is imported since version 18.1.0


pombredanne:
@qduanmu thank you ++ and great catch!

PR #1668, opened 2019-07-29
Add an option to include the original matched license text

lemoshi:
Add an option to include the original matched license text so that there is no need to remove the highlighting marks from the scanning result if the user wants to get the original matched license text.

### DIFF
--- a/src/licensedcode/plugin_license.py
+++ b/src/licensedcode/plugin_license.py
@@ -87,6 +87,12 @@ class LicenseScanner(ScanPlugin):
             help='Include the detected licenses matched text.',
             help_group=SCAN_OPTIONS_GROUP),
 
+        CommandLineOption(('--origin-license-text',),
+            is_flag=True,
+            required_options=['license'],
+            help='Include the detected licenses matched text without highlighting.',
+            help_group=SCAN_OPTIONS_GROUP),
+
         CommandLineOption(('--license-url-template',),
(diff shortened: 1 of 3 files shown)


pombredanne:
@lemoshi thank you ++ for this.
There is one test failing but this i minor(that's the test checking the cli --help option which now has a new option alright)

Now, I would think we could make this the standard `--license-text` option and only return text highlighting with another new option that would be for debug?

We could also return the whole lines of text?

What do you think?

There is also a possible special case that could be interesting to handle where a given file is mostly made of license matches and is a "legal"-like file (see the --classify option for that and some code in `summarycode.classify`) ...such as a COPYING, LICENSE, etc... in this case I would think we would want to report the whole file as text.  

BTW, I would assume you are trying to automate the creation of some attribution notice text from a scan by collating all the detected clean texts, is this right?

pombredanne:
@lemoshi ping ... do my suggestion make sense? If this works for you I can take you code and update it accordingly

pombredanne:
@lemoshi I have merged your PR in #1682 and added a slightly different logic on top as described above. https://github.com/nexB/scancode-toolkit/pull/1668#issuecomment-516691676 
So I am closing this PR in favor of #1682 
Thank you ++

lemoshi:
@pombredanne sorry for the late response. It is a long story... Thanks for merging my PR and I do like your idea. This is a great project and I'm still learning it. Hope I can contribute more to it in future.

PR #1676, opened 2019-08-08
Port textcode to Python 3 #295

Abhishek-Dev09:
The objective of this PR is to ensure that all the tests of textcode pass on both Python 2 & 3 by use of conffest plugin.


tests/textcode/test_analysis.py
```diff
@@ -38,7 +38,12 @@
 from textcode.analysis import unicode_text_lines
 from textcode.analysis import numbered_text_lines
 from commoncode.fileutils import resource_iter
+from commoncode import compat
```
pombredanne: It would be great to sort imports (they were not sorted before, that's a mistake to fix)

tests/textcode/test_analysis.py
```diff
@@ -91,7 +96,8 @@ def test_numbered_text_lines_handles_sfdb(self):
         test_file = self.get_test_loc('analysis/splinefonts/Ambrosia.sfd')
         result = list(l for _, l in numbered_text_lines(test_file))
         expected_file = test_file + '.expected'
-        expected = open(expected_file, 'rb').read().splitlines(True)
+        mode = 'rb' if on_linux and py2 else 'r'
```
pombredanne: The more should be the same on all OSses. These are not paths here but file contents.... so there is no difference between python2 and 3 and linux or not.

Note: The approach to read as binary and split lines keeping ends what there to ensure that we would keep the CRL/LF if any if I recall OK. This is not essential to keep line endings but it may be difficult to avoid reading bytes in. I am not sure we can get text in at first until we have eventually detected encoding and so on.

tests/textcode/test_analysis.py
```diff
@@ -139,11 +145,11 @@ def test_numbered_text_lines_return_correct_number_of_lines(self):
         assert 2 == len(result)
 
     def test_as_unicode_converts_bytes_to_unicode(self):
-        test_line =  b'    // as defined in https://tools.ietf.org/html/rfc2821#section-4.1.2”.'
```
pombredanne: It is essentially to keep bytes in this test... the whole purpose is to tests conversion from bytes to unicode!

src/textcode/analysis.py
```diff
@@ -206,7 +208,7 @@ def break_numbered_unicode_text_lines(numbered_lines, split=u'([",\'])', max_len
         if len(line) > max_len:
             # spli then reassemble in more reasonable chunks
             splitted = splitter(line)
-            chunks = (splitted[i:i + chunk_len] for i in xrange(0, len(splitted), chunk_len))
+            chunks = (splitted[i:i + chunk_len] for i in sx.range(0, len(splitted), chunk_len))
```
pombredanne: you can use a plain `range` everywhere. That's simpler

src/textcode/analysis.py
```diff
@@ -146,8 +148,8 @@ def numbered_text_lines(location, demarkup=False, plain_text=False):
     if T.is_text:
         numbered_lines = enumerate(unicode_text_lines(location), 1)
         # text with very long lines such minified JS, JS map files or large JSON
-        locale = b'locale' if on_linux else u'locale'
-        package_json = b'package.json' if on_linux else u'package.json'
+        locale = b'locale' if on_linux and py2 else u'locale'
+        package_json = b'package.json' if on_linux and py2 else u'package.json'
```
pombredanne: This is a proper use of the py2 test here: we are dealing with paths :+1:

Abhishek-Dev09: thanks

src/textcode/analysis.py
pombredanne: you can use a plain `range` everywhere and do not import six. That's simpler

tests/textcode/test_pdf.py
```diff
@@ -32,6 +32,9 @@
 from textcode.analysis import numbered_text_lines
 
 
+import pytest
```
pombredanne: one line before and two after please

tests/textcode/test_pdf.py
```diff
@@ -32,6 +32,9 @@
...
+pytestmark = pytest.mark.scanpy3  # NOQA
+
```
pombredanne: add extra blank line

tests/textcode/test_sfdb.py
```diff
@@ -30,6 +30,8 @@
 from commoncode.testcase import FileBasedTesting
 from textcode import sfdb
 
+import pytest
+pytestmark = pytest.mark.scanpy3  # NOQA
```
pombredanne: Add extra blank line

Abhishek-Dev09: ok I pushing now

tests/textcode/test_markup.py
```diff
@@ -48,190 +48,189 @@ def test_jsp_demarkup(self):
         test_file = self.get_test_loc(u'markup/java.jsp')
         result = list(markup.demarkup(test_file))
         expected = [
-            u' version="1.0" encoding="ISO-8859-1"?>\n',
```
Abhishek-Dev09: I have also think like that but I thought I have not allowed to touch expectations

pombredanne:
We are all greae and the test did complete everywhere, but there is a glitch wrt. reporting the completion of Azure runes to GH. Merging now

Abhishek-Dev09:
Thanks for merging.

Issue #1673, opened 2019-08-02
Consolidation plugin should consolidate on known build files


JonoYang:
Related to #1553, it would be nice to be able to consolidate on package build files like we do with package metadata files.

pombredanne:
This is all working in develop ... closing!

PR #1677, opened 2019-08-08
Consolidate build files #1673 

JonoYang:
This PR adds initial build file detection to Package scanning and the ability to consolidate build files and their Resources to the consolidation plugin for #1673


JonoYang:
@pombredanne I've squashed the revert commit

pombredanne:
@JonoYang re
> I think adding a build field is a good idea now that we can detect them.

OK, go for it then, that will be cleaner than a package. But I am still merging still for now.
Thanks!

PR #1681, opened 2019-08-10
Port cluecode to Python 3 #295

Abhishek-Dev09:
The objective of this PR is to ensure that all the tests of cluecode pass on both Python 2 & 3 by use of conffest plugin.


tests/cluecode/test_copyrights_fosso.py
```diff
@@ -148,7 +148,7 @@ def build_copyright_test_methods_with_fossology_data():
 def make_test_func(test_file_loc, expected_file_loc, expected):
     def copyright_test_method(self):
         copyrights, _authors, _holders = cluecode_test_utils.copyright_detector(test_file_loc)
-        copyrights = [c.encode('utf-8') for c in copyrights]
+        copyrights = [c for c in copyrights]
```
pombredanne: `copyrights = [c for c in copyrights]` does nothing... so this is no longer needed

tests/cluecode/test_copyrights_fosso.py
```diff
@@ -168,7 +168,7 @@ def build_tests(clazz):
     Dynamically build test methods and attach these to the clazz test class.
     """
     for test_method, test_name in build_copyright_test_methods_with_fossology_data():
-        setattr(clazz, test_name, test_method)
+        setattr(clazz, test_name.decode(), test_method)
```
pombredanne: Function names are bytes on Python 2 and unicode on Python3

tests/cluecode/test_copyrights_fosso.py
```diff
@@ -78,53 +78,53 @@ def build_copyright_test_methods_with_fossology_data():
...
-            (b'Copyright  Copyright', b'Copyright'),
-            (b'Copyright Copyright', b'Copyright'),
-            (b'All rights reserved.', b''),
-            (b'Created 1991.', b''),
-            (b'Created 1990.', b''),
-            (b'copyright under', b''),
-            (b'@copyright{}', b''),
-            (b'. .', b'.'),
-            (b'', b''),
+            (r'. All rights reserved.', r'.'),
```
pombredanne: IMHO expect plain strings everywhere, not bytes and certainly not raw strings.

tests/cluecode/test_copyrights_fosso.py
Abhishek-Dev09: oh sorry,

src/textcode/markup.py
```diff
@@ -62,7 +63,7 @@ def is_markup(location):
     if location.endswith(extensions):
         return True
 
-    with open(location, 'rb') as f:
+    with io.open(location, 'r', encoding="ISO-8859-1") as f:
```
pombredanne: Why using this `ISO-8859-1` encoding?
We are reading arbitrary files: so we typically open source as binary and we can eventually convert them to text/unicode based on their encoding. We cannot use any arbitrary encoding by default at first

tests/cluecode/test_copyrights.py
```diff
@@ -33,7 +33,7 @@
 
 
 import pytest
-pytestmark = pytest.mark.scanslow
```
pombredanne: You need to keep this `pytestmark = pytest.mark.scanslow` please... these tests are slow and running only when the `--test-suite=all` option is used with pytest. So do not remove this

tests/cluecode/test_copyrights_fosso.py
```diff
@@ -33,11 +33,12 @@
 import pytest
 
 import cluecode_test_utils
+from commoncode import compat
+from commoncode.system import py3
 from commoncode.testcase import FileDrivenTesting
 from commoncode.text import python_safe_name
 
-
-pytestmark = pytest.mark.scanslow
```
pombredanne: Same as above do not remove this marker please.

tests/cluecode/test_copyrights_fosso.py
```diff
@@ -134,9 +136,10 @@ def build_copyright_test_methods_with_fossology_data():
         tfn = test_file.replace(test_data_dir, '').strip('\/\\')
         test_name = 'test_fossology_copyright_%(tfn)s' % locals()
         test_name = python_safe_name(test_name)
-        if isinstance(test_name, unicode):
+        if isinstance(test_name, compat.unicode):
             test_name = test_name.encode('utf-8')
-
+        
+        test_name = test_name.decode() if py3 else test_name
```
pombredanne: Rather than decoding... only encode above if on Python 2 e.g. something more or less like this would be better IMHO:
```
if py2 and isinstance(test_name, compat.unicode):
    test_name = test_name.encode('utf-8')
if py3 and not isinstance(test_name, compat.unicode):
    test_name = test_name.decode('utf-8')
```

tests/cluecode/test_copyrights_fosso.py
```diff
@@ -168,7 +170,7 @@ def build_tests(clazz):
     Dynamically build test methods and attach these to the clazz test class.
     """
     for test_method, test_name in build_copyright_test_methods_with_fossology_data():
-        setattr(clazz, test_name, test_method)
+    	setattr(clazz, test_name, test_method)
```
pombredanne: Please, use spaces and not tabs.

tests/cluecode/test_copyrights_fosso.py
```diff
@@ -77,54 +78,55 @@ def build_copyright_test_methods_with_fossology_data():
 
     copyregex = re.compile('<s>(.*?)</s>', re.DOTALL | re.UNICODE)
     for expected_file, test_file in zip(expected_files, files_to_test):
-
-        expected_text = open(expected_file, 'rb').read()
+        
+        import io
```
pombredanne: Make this import top level

tests/cluecode/test_copyrights_fosso.py
```diff
@@ -77,54 +78,55 @@ def build_copyright_test_methods_with_fossology_data():
...
+        expected_text = io.open(expected_file, 'r', encoding='utf-8').read()
```
pombredanne: Since you are chaing this code, it would be best to wrap this in a proper `with ...` block to ensure descriptors are properly closed.

azure-pipelines.yml
```diff
@@ -235,6 +238,7 @@ jobs:
                  bin/py.test  -vvs tests/typecode \
                     tests/extractcode \
                     tests/plugincode \
+                    tests/cluecode \
```
pombredanne: It would be best to be consistent and make cluecode always in the same position on all such blocks? In some case it comes last in some other case, next to last

src/textcode/markup.py
Abhishek-Dev09: To remove ascii decode error

tests/cluecode/test_copyrights_fosso.py
Abhishek-Dev09: Dont't understand

tests/cluecode/test_copyrights_fosso.py
Abhishek-Dev09: How do I marked that it has been ported?

tests/cluecode/test_copyrights.py
Abhishek-Dev09: How do I marked that it has been ported?

src/textcode/markup.py
pombredanne: that's not the right way. This has to be solved differently, possibly with chardet, unicode dammit, etc, strings, any routine we already have, but not this way.

tests/cluecode/test_copyrights.py
pombredanne: You need to add more than one marker there them with this:
`pytestmark = [pytest.mark.scanpy3, pytest.mark.scanslow]`

And you likely will need to have an extra test suite run for the slow Py3 tests in the pipeliens definition

tests/cluecode/test_copyrights_fosso.py
pombredanne: 1. move `import io` to the top of the module
2.  use this
```
with io.open(expected_file, 'r', encoding='utf-8') as i:
    expected_text = i.read()
```

Abhishek-Dev09: Why ``import io`` globally as it is not anywhere except here.

Abhishek-Dev09:
@pombredanne :Merge it but some files are left which dependent on scancode-module.

pombredanne:
All green. Merging!

PR #1682, opened 2019-08-11
Improve capture of license and notice texts 

pombredanne:
This PR builds on #1668 by @lemoshi to always return a matched license text or notice
without highlighting which part is matched exactly or not. 

And this also adds a new option to add highlighting to a matched license text


pombredanne:
@lemoshi  your review is much welcomed! This PR includes your commits in #1668 but changes the logic a bit. 
@sschuberth this would be the final bit for #260

Issue #1628, opened 2019-06-27
Add support for CRAN R packages


pombredanne:
### Description
R is a programming languages and CRAN its package repository.
https://cran.r-project.org/
Packages use a `DESCRIPTION` package manifest.
See https://github.com/mdlincoln/clipr 

https://github.com/philipmorrisintl/RPackUtils provides the features but is GPL. 

Reported by @bduranc

pombredanne:
> duranbc (Brian Duran)
Another two with similar circumstances: https://clearlydefined.io/definitions/git/github/cran/codetools/e10d13c12831c8cb869c7564a058eb183a884245 https://clearlydefined.io/definitions/git/github/cran/coin/1b75f892193967e5b4d783bb66225504da14e65d

linexb:
This was fixed and merged into develop branch. I would close ticket, please reopen if needed.

PR #1684, opened 2019-08-19
Create cran r package #1628 

linexb:
Create new cran r package type and its test cases


linexb:
@pombredanne A new request is ready, please review. :)

PR #1694, opened 2019-08-24
Port remaining parts of packagedcode to Python 3 #295

Abhishek-Dev09:
Port packagedcode to Python 3 with help of conffest plugin


src/packagedcode/gemfile_lock.py
```diff
@@ -26,8 +26,10 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import attr
```
pombredanne: This should come after standard imports

src/packagedcode/nuget.py
```diff
@@ -124,7 +124,7 @@ def _parse_nuspec(location):
     """
     if not location.endswith('.nuspec'):
         return
-    with open(location) as loc:
+    with open(location , "rb") as loc:
```
pombredanne: Can you elaborate why opening 'rb' is needed? Also please use single quotes.

src/scancode/cli_test_utils.py
```diff
@@ -147,7 +147,7 @@ def check_json_scan(expected_file, result_file, regen=False, remove_file_date=Fa
     """
     results = load_json_result(result_file, remove_file_date)
     if regen:
-        with open(expected_file, 'wb') as reg:
+        with open(expected_file, 'w') as reg:
```
pombredanne: You are likely to need something like this instead:
```
        if py2:
            mode = 'wb'
        if py3:
            mode = 'w'
        with open(expected_file, mode) as reg:
....
````

```

tests/packagedcode/data/chef/dependencies/metadata.rb.expected
```diff
@@ -35,16 +35,16 @@
   "manifest_path": null,
   "dependencies": [
     {
-      "purl": "pkg:chef/mingw",
-      "requirement": ">= 1.1",
+      "purl": "pkg:chef/seven_zip",
+      "requirement": null,
```
pombredanne: I am surprised as why we do not get stable ordering there: is this because we are not using an OrderedDict

tests/packagedcode/test_chef.py
```diff
@@ -28,11 +28,15 @@
 
 from collections import OrderedDict
 import os.path
+from unittest.case import skipIf
```
pombredanne: Use `pytest.xmark.skipif` instead as `skipIf` did not mesh well in the past with the other use of `pytestmark = pytest.mark.scanpy3`

tests/packagedcode/test_chef.py
```diff
@@ -46,7 +50,8 @@ def test_parse_from_rb(self):
         test_file = self.get_test_loc('chef/basic/metadata.rb')
         expected_file = self.get_test_loc('chef/basic/metadata.rb.expected')
         self.check_package(chef.parse(test_file), expected_file, regen=False)
-
+    
+    @skipIf(not py3, 'Python 3 only') # data are different for both Python And data generated by assigning regen = false
```
pombredanne: Add this whole comment `data are different for both Python And data generated by assigning regen = false ` to the skipif e.g.:
```
@pytext.xmarlk.skipif(not py3, reason='Python 3 only, data are different for both Python And data generated by assigning regen = false')
```

BUT why is this the case? that does not sound right

src/packagedcode/gemfile_lock.py
Abhishek-Dev09: This is no use now because i have problem that class gem is of unhashable type

src/packagedcode/nuget.py
Abhishek-Dev09: Read this https://github.com/martinblech/xmltodict/issues/77

tests/packagedcode/data/plugin/python-package-expected.json
```diff
@@ -89,46 +89,7 @@
     {
       "path": "setup.py",
       "type": "file",
-      "packages": [
```
Abhishek-Dev09: Why this has been deleted by assigning regen = true

src/packagedcode/gemfile_lock.py
Abhishek-Dev09: https://github.com/nexB/scancode-toolkit/pull/1694/files#diff-ebf0b433a58c067b70ad8235d3e08c5eR210 . This line gives `TypeError: unhashable type: 'Gem' `.So how to use attrs here hor hashing.

src/packagedcode/gemfile_lock.py
```diff
@@ -153,6 +155,12 @@ def __init__(self, name=None, version=None, platform=None):
         # a map of direct dependent Gems, keyed by name
         self.dependencies = OrderedDict()
 
+    def __lt__(self,other):
```
Abhishek-Dev09: We can't able to sort Gem object So this method needed to compare . Otherwise it gives `TypeError: '<' not supported between instances of 'Gem' and 'Gem' `

tests/packagedcode/data/chef/dependencies/metadata.rb.expected
Abhishek-Dev09: Leave it because you have to delete the test for py2

src/packagedcode/gemfile_lock.py
pombredanne: My point was only about the ordering of imports `This should come after standard imports`... this is not a standard library import and we sort imports in this order: standard, third-party then scancode-own modules.

src/packagedcode/gemfile_lock.py
pombredanne: :+1:

src/packagedcode/gemfile_lock.py
Abhishek-Dev09: Ok, first solve that error , then think about ordering . Because if you find other way , then we have to delete this import.

pombredanne:
All green. Merging!
Thank you ++

PR #1702, opened 2019-08-30
Rename license 'fsf-mit' to 'fsf-unlimited-no-warranty'

qduanmu:
*  This is identical to the SPDX identifier FSFULLR but
       with a simple GNU-style warranty disclaimer added.
       Also it is not really like the MIT license textually
       or in terms of relative permissiveness.

Reported-by: Richard Fontana <rfontana@redhat.com>


pombredanne:
@qduanmu @richardfontana Thank you!

So far the `FSFULLR` SPDX id has a `fsf-unlimited` id in Scancode. There are also several other FSF licenses that start with `fsf-` in ScanCode.  

You want to rename `fsf-mit` to use the `fsfullr` id as a prefix. Instead it would be much better to rename `fsf-mit` to use the  `fsf-unlimited-no-warranty` id: this way we keep things consistent and achieve the same results e.g. you have two license keys `fsf-unlimited-no-warranty` and `fsf-unlimited` that share the same prefix and this conveys that they share some text. 

Otherwise this would mean renaming more than one license key... and I would rather avoid this if possible: we have to thread lightly when we do renames of keys as the whole idea of using scancode's own keys is to make these keys stable and avoid changing them. (much the same way you have a stable primary key in a relational table to identify a record). 

Changing them means deprecating the old one and everywhere someone used the old id, the new one will need to be used (which already happened once to `fsf-unlimited` in the past.)

@DennisClark ping, you take is much welcomed too!

qduanmu:
@pombredanne Thank you for the detail, let's see if any different opinions from @richardfontana and @DennisClark , I will give an update later accordingly ;)

DennisClark:
Planning to do the following actions: 
* Deprecate `fsf-mit`
* Create new license `fsf-unlimited-no-warranty` in its place
That should take care of the issue.

pombredanne:
:+1: 
Merging!

PR #1705, opened 2019-08-30
System provided binaries plugin for Distro package

aj4ayushjain:
This is a python plugins which provides path to system package provided binaries which is to be used in distro packages.
Signed-off-by: aj4ayushjain <aj4ayushjain@gmail.com>


pombredanne:
Thank you ++ all clean!

Issue #1688, opened 2019-08-20
Check for Update - feature request


bobgob1:
Feature Request:

To facilitate running on the most recently released version of scancode-toolkit, add "Check for Software Update" under Help menu.  Or some similar implementation.  Without this it is too easy to continue running old versions because if you aren't actively involved in development, you probably don't go to github to check.

pombredanne:
@bobgob1 excellent suggestion... I think we can use something such as the pip nag that comes this way as a simple warning message that is issues no more than once per week.
```
You are using pip version 18.1, however version 19.2.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```

https://github.com/pypa/pip/blob/449d268b5a6a6e126e4f00f0c848f101e689c2a0/src/pip/_internal/utils/outdated.py

Actually we already have code in a branch in https://github.com/nexB/scancode-toolkit/commit/200d96196b050c3ae33aeb1cd309a72511059a0b by @jdaguil  so it should be easy to update and activate!

pombredanne:
This is merged in develop now: all yours to test there (though it may be hard to test until we have made releases at PyPI).
Do you think we should have a command line option to also force a check if a newer version is available?

Otherwise this is automatic and done once a week at most

bobgob1:
I like command line options as much as the next person, but perhaps a version check would be best done by overloading --version?   Would this mess anyone up who is parsing that info?  They can always just pick off the third field.
Something like:
scancode --version
ScanCode version 3.0.2 (A newer version, 3.0.3, is available at <link>>.
or 
ScanCode version 3.0.2 (This is the latest released version.)

pombredanne:
> Would this mess anyone up who is parsing that info? 

I do not think so. The actual current version is available in a structured form in the scans alright. I can add that to a new line too.

bobgob1:
What I would really like is if that version info  "A newer version ... is available" was part of the scancode output, because that's where I'll see it.

pombredanne:
@bobgob1 I will add this in the results too. Good point and there is a bug that needs to be fixed before we push a proper release candidate.

pombredanne:
One issue to fix is this message:
```
WARNING: You are using ScanCode Toolkit version 3.1.1.post147.a08176565, however the newer version 3.1.1 is available.
You should download and install the latest version of ScanCode with bug and security fixes and the latest license detection data for accurate scanning.
[...]

Issue #111, opened 2015-11-05
New Version notification within Scancode Tool


kaushikk:
When the user starts Scancode tool, display a message to user that a new version of scancode is available

Development teams using scancode need to be updated with the latest version of the tool. When starting scancode CLI, it will help if a notification message is printed or an extra parameter to command line like 'scancode --check-updates' may also help

pombredanne:
@kaushikk this makes a lot of sense. If you are familiar with the `pip` Python tool, they use a similar approach. Do you think that is something that you could help with?

jdaguil:
@kaushikk this is a nice enhancement. Currently if you did want to compare versions, the way to do this is to run `scancode --version` and compare it to the latest tagged release at https://github.com/nexB/scancode-toolkit/releases

pombredanne:
In terms of implementation, we could reuse or borrow and adapt the code from pip:
https://github.com/pypa/pip/blob/7cea748d0fb2e8980c2676304607426585550b41/pip/utils/outdated.py

This means also that wee should first make scancode available on Pypi: there is no issue per se there since it builds as a wheel all right. With that, the pip code can check on a remote package version and compare it to the locally installed version and alert that it is out of date easily.

Note that making scancode as a Pypi package would also require that all dependencies are also available on Pypi: this should be the case mostly, but some may not: in which case we will need push a Pypi package for these too.

mjherzog:
What are the other pros & cons of offering ScanCode on Pypi? In general it seems to make sense to do so since ScanCode and our other tools are open source and Python, but does this mean that we will redirect people to get it from Pypi as a primary source for stable releases instead of our GH repo?  Is installation from Pypi easier (or not) than our current installation process?

pombredanne:
@mjherzog there are no cons. Either you use the full distro from GH for regular usage, or you would use the Pypi version as a library when used in Python code.

mjherzog:
I assume that this will also work for AboutCode with the same caveats about dependencies.
Is this something that Li knows how to do easily from his other work on DJA ?

jdaguil:
@pombredanne @mjherzog 
I see two possible solutions to check if you have the latest release:
<ol>
<li> Github Releases API. This would allow you to compare versions by pulling the latest version from the API. It's not much work to implement, and it would address the issue reasonably soon.
<li>  Put project on Pypi and use pip. This sounds great; although it might take more research and time to implement.
</ol>

pombredanne:
[...]

PR #1708, opened 2019-09-02
New version notification #111 #1688

pombredanne:
Add support for new version notifications for #111 and #1688
The principle is to check once a week if a new release has been made on PyPI and issue a warning if there a new version. No warning or failure is reported if the PyPI servers cannot be reached.


pombredanne:
This is all green now. Merging this @jdaguil it took a while... but thank you.

PR #1711, opened 2019-09-03
Extractcode 7z system provided

aj4ayushjain:
This has the proper indentation along with the updated  MANIFEST.in file with license which was missing.

### DIFF
--- a/plugins-builtin/extractcode-7z-system_provided/src/extractcode_7z/__init__.py
+++ b/plugins-builtin/extractcode-7z-system_provided/src/extractcode_7z/__init__.py
@@ -54,7 +54,7 @@ class SevenzipPaths(LocationProviderPlugin):
 			
 			lib_dir = '/usr/libexec/p7zip'
 
-		else
+		else:
 			lib_dir = '/usr'
 
 	        locations = {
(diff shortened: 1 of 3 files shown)


plugins-builtin/extractcode-7z-system_provided/src/extractcode_7z/__init__.py
```diff
@@ -35,32 +35,31 @@
...
-	if distribution in debian_based_distro:
+		if distribution in debian_based_distro:
 		
-		lib_dir = '/usr/lib/p7zip'
+			lib_dir = '/usr/lib/p7zip'
 	
-	elif distribution in rpm_based_distro:
-		
-		lib_dir = '/usr/libexec/p7zip'
+		elif distribution in rpm_based_distro:
```
pombredanne: Culd you make sure that you use 4 spaces throughout and no tabs?

aj4ayushjain: Will take care of that!
What's reason behind this are my tabs long spaced(8 spaces) or something else?

pombredanne: The reason is simply that we never use tabs

PR #1725, opened 2019-09-12
Run Python 3 tests on travis #295

pombredanne:
This PR enables Python 3.6 tests on Travis.


pombredanne:
We are now all green which means it is merging time!

PR #1729, opened 2019-09-13
Extractcode libarchive system provided

aj4ayushjain:
This provides the support for the freebsd os in plugins which can be used for pip installation.

### DIFF
--- a/plugins-builtin/extractcode-libarchive-system_provided/src/extractcode_libarchive/__init__.py
+++ b/plugins-builtin/extractcode-libarchive-system_provided/src/extractcode_libarchive/__init__.py
@@ -54,18 +54,14 @@ class LibarchivePaths(LocationProviderPlugin):
             else:
                 raise Exception('Unsupported system: {}'.format(distribution))
 
-            lib_dll = path.join(lib_dir, 'libarchive.so.13')
-
         elif mainstream_system == 'freebsd':
             if path.isdir('/usr/local/'):
-                lib_dir = '/usr/local'
+                lib_dir = '/usr/local/lib'
             else:
-                lib_dir = '/usr'
-
-            lib_dll = path.join(lib_dir, 'lib/libarchive.so')
+                lib_dir = '/usr/lib'
 
         locations = {
             'extractcode.libarchive.libdir': lib_dir,
-            'extractcode.libarchive.dll': lib_dll,
+            'extractcode.libarchive.dll': path.join(lib_dir, 'libarchive.so'),
         }
         return locations
--- a/plugins-builtin/extractcode-libarchive-system_provided/src/extractcode_libarchive/__init__.py
+++ b/plugins-builtin/extractcode-libarchive-system_provided/src/extractcode_libarchive/__init__.py
@@ -25,39 +25,46 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
-from os.path import abspath
-from os.path import dirname
-from os.path import join
-
-from plugincode.location_provider import LocationProviderPlugin
-from plugincode.location_provider import location_provider_impl
-
+from os import path
 import platform
 
+from plugincode.location_provider import LocationProviderPlugin
+
+
 class LibarchivePaths(LocationProviderPlugin):
-	def get_locations(self):
-	
-		distribution=platform.linux_distribution()[0]	
-	
-		# List of various major distributions consisting of flavors
-		debian_based_distro=['Ubuntu','Mint','debian']
-	
-		rpm_based_distro=['Fedora','redhat']
+    def get_locations(self):
+        """
+        Return a mapping of {location key: location} providing the installation
+        locations of the libarchive shared library as installed on various Linux
+        distros or on FreeBSD.
+        """
+        mainstream_system = platform.system().lower()
+        if mainstream_system == 'linux':
+            distribution = platform.linux_distribution()[0].lower()
+            debian_based_distro = ['ubuntu', 'mint', 'debian']
+            rpm_based_distro = ['fedora', 'redhat']
 
-		if distribution in debian_based_distro:
-		
-			lib_dir = '/usr/lib/x86_64-linux-gnu'
-	
-		elif distribution in rpm_based_distro:
-		
-			lib_dir = '/usr/lib64'
+            if distribution in debian_based_distro:
+                lib_dir = '/usr/lib/x86_64-linux-gnu'
 
-		else:
-			#User defined if installation directory differs
-			lib_dir = '/usr'
+            elif distribution in rpm_based_distro:
+                lib_dir = '/usr/lib64'
 
-		locations = {
-			'extractcode.libarchive.libdir': lib_dir,
-			'extractcode.libarchive.dll': join(lib_dir, 'libarchive.so.13'),
-		}
-		return locations
+            else:
+                raise Exception('Unsupported system: {}'.format(distribution))
+
+            lib_dll = path.join(lib_dir, 'libarchive.so.13')
+
+        elif mainstream_system == 'freebsd':
+            if path.isdir('/usr/local/'):
+                lib_dir = '/usr/local'
+            else:
+                lib_dir = '/usr'
+
+            lib_dll = path.join(lib_dir, 'lib/libarchive.so')
+
+        locations = {
+            'extractcode.libarchive.libdir': lib_dir,
+            'extractcode.libarchive.dll': lib_dll,
+        }
+        return locations
--- a/plugins-builtin/extractcode-libarchive-system_provided/src/extractcode_libarchive/__init__.py
+++ b/plugins-builtin/extractcode-libarchive-system_provided/src/extractcode_libarchive/__init__.py
@@ -38,6 +38,7 @@ class LibarchivePaths(LocationProviderPlugin):
         locations of the libarchive shared library as installed on various Linux
         distros or on FreeBSD.
         """
+        system_arch = platform.machine()
         mainstream_system = platform.system().lower()
         if mainstream_system == 'linux':
             distribution = platform.linux_distribution()[0].lower()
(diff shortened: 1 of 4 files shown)


pombredanne:
@knobix you review would be great as @aj4ayushjain added support for FreeBSD there.

@aj4ayushjain are you sure that `lib_dir = 'lib'` is enough? I would have expected some absolute path of sorts on FreeBSD?

knobix:
> @aj4ayushjain are you sure that `lib_dir = 'lib'` is enough? I would have expected some absolute path of sorts on FreeBSD?

That's a good start but the `lib_dir` and the name of the `.so` need to be changed for FreeBSD. `libarchive` is shipped with FreeBSD and the location is `/usr/lib/libarchive.so` which is a symbolical link that points to the actual binary.

With FreeBSD 11.2- and 11.3-RELEASE `/usr/lib/libarchive.so` points to `/usr/lib/libarchive.so.6`. (= 3.3.3) 

With FreeBSD 12.0-RELEASE `/usr/lib/libarchive.so` points to `/usr/lib/libarchive.so.7`. (= 3.3.0)

With FreeBSD 13.0-CURRENT `/usr/lib/libarchive.so` points to `/usr/lib/libarchive.so.7` (= 3.4.0)

There's also an recent [package/port](https://www.freshports.org/archivers/libarchive/) of `libarchive` available. The default prefix for local packages is `/usr/local` so the constellation would be then:

`/usr/local/lib/libarchive.so` points to `/usr/local/lib/libarchive.so.13`. 

Thus it would be preferable to try to use the pkg/port version of `libarchive` in `/usr/local/lib/libarchive.so` first. If that version is not present then `libarchive` from the FreeBSD base system in `/usr/lib/libarchive.so` can be used as a fallback.

aj4ayushjain:
Thanks @knobix  for the review.
I just saw the paths from the https://www.freshports.org/archivers/libarchive/ page and was not sure of that.
What's the reason behind not to use the path to original binary `/usr/local/lib/libarchive.so.13`?

knobix:
> Thanks @knobix for the review.
> I just saw the paths from the https://www.freshports.org/archivers/libarchive/ page and was not sure of that.

Pleasure and thanks for the update! The given paths/files listed in the `pkg-plist` are relative to the installation prefix (usually `/usr/local`) which is the reason, why all those pathnames are somewhat shorter or missing the leading `/`.

> What's the reason behind not to use the path to original binary `/usr/local/lib/libarchive.so.13`?

I thought about it to reduce the need for maintenance. The version of the `.so` might be bumped in near future (e.g. `libarchive.so.14`) so one doesn't need to always change the source code to match the new name of the library.

aj4ayushjain:
> > What's the reason behind not to use the path to original binary `/usr/local/lib/libarchive.so.13`?
> 
> I thought about it to reduce the need for maintenance. The version of the `.so` might be bumped in near future (e.g. `libarchive.so.14`) so one doesn't need to always change the source code to match the new name of the library.

Thank you++ for the reply.
As i have used this path to original  elsewhere also so will update it soon.

plugins-builtin/extractcode-libarchive-system_provided/src/extractcode_libarchive/__init__.py
```diff
@@ -28,36 +28,61 @@
...
+                lib_dir = '/usr'
+
+            locations = {
+                'extractcode.libarchive.libdir': lib_dir,
+                'extractcode.libarchive.dll': join(lib_dir, 'libarchive.so.13'),
+            }
+       
+	elif mainstream_system == 'freebsd' :
+            
+            if path.isdir('/usr/local/') :
```
pombredanne: Why a blank line?

plugins-builtin/extractcode-libarchive-system_provided/src/extractcode_libarchive/__init__.py
```diff
@@ -28,36 +28,61 @@
...
-			#User defined if installation directory differs
-			lib_dir = '/usr'
+            else:
 
-		locations = {
-			'extractcode.libarchive.libdir': lib_dir,
-			'extractcode.libarchive.dll': join(lib_dir, 'libarchive.so.13'),
-		}
-		return locations
+                '''User defined flavor if installation directory differs'''
```
pombredanne: Why do you use triple quotes there?  a plain # would be better? also what do you mean exactly with this comment?

plugins-builtin/extractcode-libarchive-system_provided/src/extractcode_libarchive/__init__.py
```diff
@@ -28,36 +28,61 @@
...
+                'extractcode.libarchive.libdir': lib_dir,
+                'extractcode.libarchive.dll': join(lib_dir, 'libarchive.so.13'),
+            }
+       
+	elif mainstream_system == 'freebsd' :
+            
+            if path.isdir('/usr/local/') :
+
+                lib_dir = '/usr/local'
+
```
pombredanne: Same as above, this empty line is not needed there

plugins-builtin/extractcode-libarchive-system_provided/src/extractcode_libarchive/__init__.py
```diff
@@ -28,36 +28,61 @@
...
+            else :
+
```
pombredanne: Same as above, this empty line is not needed there

plugins-builtin/extractcode-libarchive-system_provided/src/extractcode_libarchive/__init__.py
```diff
@@ -28,36 +28,61 @@
...
+            if distribution in debian_based_distro:
 		
-			lib_dir = '/usr/lib/x86_64-linux-gnu'
+                lib_dir = '/usr/lib/x86_64-linux-gnu'
 	
-		elif distribution in rpm_based_distro:
-		
-			lib_dir = '/usr/lib64'
+            elif distribution in rpm_based_distro:
+
```
pombredanne: Why an empty line? IMHO not needed here

plugins-builtin/extractcode-libarchive-system_provided/src/extractcode_libarchive/__init__.py
```diff
@@ -28,36 +28,61 @@
...
-		debian_based_distro=['Ubuntu','Mint','debian']
+    def get_locations(self):
+
+        '''This defines the various mainstream system available 
+        freebsd(unix derivative),Linux(unix alternative)'''
+        
+        mainstream_system = platform.system()
+
+        if mainstream_system == 'Linux' :
+
```
pombredanne: Empty line not needed here

plugins-builtin/extractcode-libarchive-system_provided/src/extractcode_libarchive/__init__.py
```diff
@@ -28,36 +28,61 @@
...
 class LibarchivePaths(LocationProviderPlugin):
-	def get_locations(self):
-		curr_dir = dirname(abspath(__file__))
-		distribution=platform.linux_distribution()[0]	
-	
-		# List of various major distributions consisting of flavors
-		debian_based_distro=['Ubuntu','Mint','debian']
+    def get_locations(self):
+
+        '''This defines the various mainstream system available
```
pombredanne: We use triple double quotes """ for docstrings and no empty line before. We also use the imperative style and describe what is returned
So here this should be:
```
    def get_locations(self):
        """
        Return a mapping of {location key: location} providing the installation
        locations of the libarchive shared library as installed on various Linux
        distros or on FreeBSD.
        """
```

plugins-builtin/extractcode-libarchive-system_provided/src/extractcode_libarchive/__init__.py
aj4ayushjain: I had thought this makes code much more readable that was the reason of inline comments and blank line spacing but will remove it.

plugins-builtin/extractcode-libarchive-system_provided/src/extractcode_libarchive/__init__.py
aj4ayushjain: Will take care of these things from next time.
And will try to improve my imperative style comment.

aj4ayushjain: Taken Note off!!
Will take care of these things from next time.
And will try to improve my imperative style comment.

plugins-builtin/extractcode-libarchive-system_provided/src/extractcode_libarchive/__init__.py
aj4ayushjain: If the user has defined some other path of installation or some different flavor from unusual.

aj4ayushjain:
@pombredanne  I have added the support for the freebsd in all the plugins and addressed the issue #1736.

pombredanne:
@aj4ayushjain re

>  I have added the support for the freebsd in all the plugins and addressed the issue #1736.

excellent! do you know what would be the procedure to test this for @madscientist159 ?

aj4ayushjain:
> @aj4ayushjain re
> 
> > I have added the support for the freebsd in all the plugins and addressed the issue #1736.
> 
> excellent! do you know what would be the procedure to test this for @madscientist159 ?

Thank you.
By test you meant how he can use it to solve the issue with binaries  or something else?

pombredanne:
@aj4ayushjain By test I meant some simple instructions here on how to test scancode on power pc.

pombredanne:
@aj4ayushjain I am merging this now anyway.

aj4ayushjain:
@pombredanne I have provided some steps related to POWER pc system in issue there which could be used to run scancode and i need feedback from them if problems if any and too merge is good enough.

PR #1737, opened 2019-09-23
Consolidation fixes

JonoYang:
This PR fixes some bugs with the new consolidation feature.

- Consolidations without files are now properly not reported
- Directories that have a different majority origin than its parent will now be reported properly
- Combine a resource's license expression and holders once and save it to `extra_data`


pombredanne:
All good. Thanks! merging!

PR #1750, opened 2019-10-08
Don't visit files and subdir of ignored dirs

JonoYang:
This PR fixes the issue of ScanCode attempting to create resources for each file or subdirectory of an ignored directory. This fix is done by clearing the list of files and subdirectories returned for a top-level directory that we are ignoring when we walk a filesystem when creating a Codebase.

### DIFF
--- a/src/scancode/resource.py
+++ b/src/scancode/resource.py
@@ -437,6 +437,10 @@ class Codebase(object):
         # walk proper
         for top, dirs, files in os_walk(root.location, topdown=True, onerror=err):
             if skip_ignored(top):
+                # We clear out `dirs` and `files` to prevent `os_walk` from visiting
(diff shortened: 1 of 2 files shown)


src/scancode/resource.py
```diff
@@ -437,6 +437,10 @@ def create_resources(_seq, _top, _parent, _is_file):
         # walk proper
         for top, dirs, files in os_walk(root.location, topdown=True, onerror=err):
             if skip_ignored(top):
+                # We clear out `dirs` and `files` to prevent `os_walk` from visiting
+                # the files and subdirectories of directories we are ignoring
+                dirs[:] = []
+                files[:] = []
```
pombredanne: Interesting idiom! good find!

PR #1768, opened 2019-10-17
Minor typo fix in match.py

sbs2001:
Just noticed this minor typo.
Signed-off-by: Shivam Sandbhor <shivam.sandbhor@gmail.com>


pombredanne:
@sbs2001 Thanks! every little bit counts!
You added the signoff to the PR so that works, but in the future we prefer to have them in commit messages, FYI.

Issue #1773, opened 2019-10-18
Errors scanning log4j 


DennisClark:
Using the latest develop branch, as of 2019-10-18, on Mac OS 10.14.5, I ran a scan of `apache-log4j-2.12.1-src` and received the messages shown below.   Scan results attached. 
[apache-log4j-2.12.1-src-scan.json.zip](https://github.com/nexB/scancode-toolkit/files/3745657/apache-log4j-2.12.1-src-scan.json.zip)

```
./scancode -clip --consolidate -n 4 --max-in-memory 0 {{path}}/apache-log4j-2.12.1-src/ --json-pp {{path}}/apache-log4j-2.12.1-src-scan.json
Setup plugins...
Collect file inventory...
Scan files for: info, licenses, copyrights, packages with 4 process(es)...
[####################] 2960                                                                              
ERROR: failed to run post-scan plugin: consolidate:
Traceback (most recent call last):
  File "/Users/dclark1330/scancode-toolkit-develop/src/scancode/cli.py", line 1040, in run_codebase_plugins
    plugin.process_codebase(codebase, **kwargs)
  File "/Users/dclark1330/scancode-toolkit-develop/src/scancode/plugin_consolidate.py", line 207, in process_codebase
    consolidated_packages.append(c.to_dict())
  File "/Users/dclark1330/scancode-toolkit-develop/src/scancode/plugin_consolidate.py", line 125, in to_dict
    package = self.package.to_dict()
  File "/Users/dclark1330/scancode-toolkit-develop/src/packagedcode/models.py", line 302, in to_dict
    mapping['repository_download_url'] = self.repository_download_url()
  File "/Users/dclark1330/scancode-toolkit-develop/src/packagedcode/maven.py", line 121, in repository_download_url
    extension=qualifiers.get('type') or 'jar',
AttributeError: 'unicode' object has no attribute 'get'

Scanning done.
Some files failed to scan properly:
ERROR: failed to run post-scan plugin: consolidate:
Traceback (most recent call last):
  File "/Users/dclark1330/scancode-toolkit-develop/src/scancode/cli.py", line 1040, in run_codebase_plugins
    plugin.process_codebase(codebase, **kwargs)
  File "/Users/dclark1330/scancode-toolkit-develop/src/scancode/plugin_consolidate.py", line 207, in process_codebase
    consolidated_packages.append(c.to_dict())
[...]

PR #1774, opened 2019-10-19
Manipulate qualifiers after others #1773

JonoYang:
In a Package to_dict method, we normalize the qualifiers after the other fields are populated since some fields need to use qualifiers in its unnormalized form

### DIFF
--- a/src/packagedcode/models.py
+++ b/src/packagedcode/models.py
@@ -293,14 +293,13 @@ class BasePackage(BaseModel):
         Return an OrderedDict of primitive Python types.
         """
         mapping = attr.asdict(self, dict_factory=OrderedDict)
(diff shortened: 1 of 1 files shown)


JonoYang:
@pombredanne In #1773, we fail because a pom.xml has an attribute for `packaging`, whose value is `war`. `normalize_qualifiers()` modifies the qualifiers to serialize them as a string than a mapping, but this changes the Package's qualifier object from a mapping to a string, causing an exception to be thrown when we try to access qualifiers as a mapping when we generate `purl`, `repository_homepage_url`, or `api_data_url` (I forget which one uses qualifiers for its generation)

JonoYang:
This fix does not work. I will reopen when I get a solution.

Issue #1691, opened 2019-08-23
scancode with --summary throws warnings about missing logger handlers


armijnhemel:
### Description
Please leave a brief discription of the bug or feature request:

When running scancode with the --summary option I get warnings:

    No handlers could be found for logger "bs4.dammit"

Scancode was run as follows:

    $ ./scancode --full-root -l --license-text -c -e -u -n 8 --summary --json-pp=/tmp/output.json2 /tmp/busybox-1.28.0/

### System configuration
For bug reports, it really helps us to know:
* What OS are you running on? (Windows/MacOS/Linux) - Fedora 30
* What version of scancode-toolkit was used to generate the scan file? - develop branch, commit 483e47a69e24a0b9ac08f0861ce63dbb7be457f9
* What installation method was used to install/run scancode? (pip/source download/other) - source download

pombredanne:
Thank you for this. Dammit dammit!

Iamshankhadeep:
@pombredanne What do I need to do exactly?

pombredanne:
@Iamshankhadeep thanks for willing to help!
I guess you would need to try to reproduce this bug first with the help of @armijnhemel 
Then I am not sure how to fix this... it could be that this is a bug in bs4 https://pypi.org/project/beautifulsoup4/ ?

Iamshankhadeep:
thanks for replying @pombredanne . so, @armijnhemel how did you get to that error?

armijnhemel:
> thanks for replying @pombredanne . so, @armijnhemel how did you get to that error?

I ran scancode ;-)

My system is Fedora 30. The command I used can be found in the first post in this thread.

johnmhoran:
@pombredanne I am also now getting this alert running scans on my Windows 10 box -- pretty sure I only started seeing this recently.  Running on `develop` branch `version 3.1.1.post82.d7c826308,`.  Excerpt:

```
. . .
Setup plugins...
Collect file inventory...
Scan files for: info with 2 process(es)...
No handlers could be found for logger "bs4.dammit"
No handlers could be found for logger "bs4.dammit"
No handlers could be found for logger "bs4.dammit"
Scanning done.
. . .
```

pombredanne:
@Iamshankhadeep  use this.
[...]

PR #1775, opened 2019-10-19
Add test to check unwanted log messages is not printing to the screen…

Iamshankhadeep:
… #1691

Signed-off-by: Shankhadeep <shankhadeepdey99@gmail.com>


Iamshankhadeep:
@pombredanne but the test is not failing when the patch is not applied.
mainly because `results.output` does not contain `No handlers could be found for logger "bs4.dammit"`.

Issue #1780, opened 2019-10-22
Cannot call (easily) the main scancode function from Python 


pombredanne:
### Description
This has been discovered during a hacking session with  @valeriocos . See also  https://github.com/chaoss/grimoirelab-graal/issues/64

Note that we also faced #1778 reported by @AyanSinhaMahapatra

### System configuration
For bug reports, it really helps us to know:
* What OS are you running on? Linux
* What version of scancode-toolkit was used to generate the scan file? develop
* What installation method was used to install/run scancode? pip

Iamshankhadeep:
@pombredanne Can I work on it?

pombredanne:
@Iamshankhadeep this is a rather complex one: this about being able to call https://github.com/nexB/scancode-toolkit/blob/25d71f078f82d23a0b6ee0d0019e810430af4295/src/scancode/cli.py#L540 from another program for instance with `run_scan(<absolute location of a file>, license=True, copyright=True, return_results=True)` and get the results.

I know how to fix the problem ... but this requires introspection into how options are contributed by the all scancode plugins which is far from trivial. ... the `**kwargs` that are passed to `scancode()` by Click are not passed to `run_scan()` when you call it directly. This would likely mean digging in the innards of https://github.com/nexB/scancode-toolkit/blob/25d71f078f82d23a0b6ee0d0019e810430af4295/src/scancode/cli.py#L155 and/or may be more likely play with the values in https://github.com/nexB/scancode-toolkit/blob/25d71f078f82d23a0b6ee0d0019e810430af4295/src/scancode/cli.py#L222 to convert them to `kwargs`  or may be use something like that https://github.com/nexB/scancode-toolkit/blob/25d71f078f82d23a0b6ee0d0019e810430af4295/src/scancode/cli.py#L233 

You are welcome to have a stab at this... The first step would be to craft and commit a failing tests.

BUT I would suggest that you may look into something a tad easier like any of these two https://github.com/nexB/scancode-toolkit/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22+milestone%3A%22v3.1+%22

pombredanne:
This has been fixed and merged in develop. I am keeping this open to add the docs for this.

chinyeungli:
@AyanSinhaMahapatra Can you verify this is fixed and no special documentation is needed?

2400032450Deepak:
[...]

PR #1782, opened 2019-10-22
Allow to call "run_scan" as a function #1780

pombredanne:
This is the first setp and can be used this way:
```
>>> from scancode import cli
>>> rc, results = run_scan(<absolute loc a file or dir to scan>, license=Trye, copyright=True)
```
The options are overall the same as when calling the CLI


pombredanne:
@nishakm @valeriocos early feedback welcomed. This is not yet tested though

nishakm:
> @nishakm @valeriocos early feedback welcomed. This is not yet tested though

I am still unable to install the pip package at the tip of develop (using pip3 install . from the root of the git repo) Any suggestion on how to get around this?

pombredanne:
All green so I am merging now.

PR #1786, opened 2019-10-23
Add a separate boolean py2_with_linux

sbs2001:
We are using `py2 and on_linux`  repeatedly in  many places. I propose to have separate boolean  
`py2_with_linux = py2 and on_linux`  defined in our src/commoncode/system.py .

I also added some minor fixes, I came across.

Please share your opinions


pombredanne:
@sbs2001 thank you... but I am not sure this is worth the effort. Pretty soon we will stop supporting Python 2.7 in the main branches entirely.. so this code will be gone.

PR #1792, opened 2019-10-25
Update Consolidation Plugin Following Package Model Update

JonoYang:
This PR updates the Consolidation plugin to work with the new Package model. An extra step to get the package root has been added. Consolidation logic was also fixed, where we were not properly ignoring Package file count from the total file count when determining license-holder majorities.


pombredanne:
:+1:  Merging!

Issue #1794, opened 2019-10-25
Scancode reports on "originally licensed under X" as if it's a license notice


MartinPetkov:
### Description
Not sure if this is a "bug" or if there's any good way to solve it short of embedding an NLP parser, but it's at least an edge case. In https://github.com/gettalong/kramdown/blob/179b81dcf057f8079fd9df5296ba858114d30f7a/README.md, there's this text:
```
kramdown was originally licensed under the GPL until the 1.0.0 release. However, due to the many requests it is now released under the MIT license and therefore can easily be used in commercial projects, too.
```

Try scanning it:
```shell
$ git clone https://github.com/gettalong/kramdown
$ cd kramdown && git checkout 179b81dcf057f8079fd9df5296ba858114d30f7a
$ scancode --verbose --license -n 12 --json-pp ./results.json --only-findings --info --strip-root --license-text ./README.md
$ cat ./results.json
```

You see this:
```json
{
  "key": "gpl-1.0-plus",
  "score": 100,
  "name": "GNU General Public License 1.0 or later",
  "short_name": "GPL 1.0 or later",
  "category": "Copyleft",
  "is_exception": false,
  "owner": "Free Software Foundation (FSF)",
  "homepage_url": "http://www.gnu.org/licenses/old-licenses/gpl-1.0-standalone.html",
  "text_url": "http://www.gnu.org/licenses/old-licenses/gpl-1.0-standalone.html",
  "reference_url": "https://enterprise.dejacode.com/urn/urn:dje:license:gpl-1.0-plus",
  "spdx_license_key": "GPL-1.0-or-later",
  "spdx_url": "https://spdx.org/licenses/GPL-1.0-or-later",
  "start_line": 5,
  "end_line": 5,
  "matched_rule": {
    "identifier": "gpl_85.RULE",
    "license_expression": "gpl-1.0-plus",
    "licenses": [
      "gpl-1.0-plus"
    ],
    "is_license_text": false,
    "is_license_notice": true,
    "is_license_reference": false,
    "is_license_tag": false,
    "matcher": "2-aho",
    "rule_length": 4,
    "matched_length": 4,
    "match_coverage": 100,
    "rule_relevance": 100
  },
  "matched_text": "licensed under the GPL"
}
```
It (correctly imo) zeroes in on the "licensed under..." text, but it would be nice if it could take into account context and ignore this particular blurb. I understand if it's not feasible, but wanted to bring it up.

### System configuration
[...]

PR #1797, opened 2019-10-27
Add new rule for GPL historical note #1794

MartinPetkov:
This rule is from a scan of https://github.com/gettalong/kramdown/. Specifically, the README notes that it was originally GPL licensed. This blurb should not actually be matched.


pombredanne:
@MartinPetkov Thank you ++. :bowing_man:

PR #1799, opened 2019-10-29
Parse BUILD files

JonoYang:
This PR adds the Bazel BUILD file parser to packagedcode. The BUCK and Bazel parsers use the same code for parsing, so I made a class `StarlarkManifestPackage` and `BuckPackage` and `BazelPackage` are subclasses of that. I have also moved the test code and data for the BUCK and Bazel parser back into the test file/folder for `build.py` since  all the code is living in `build.py` now.


src/packagedcode/build.py
```diff
@@ -26,13 +26,19 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
+from collections import defaultdict
+from collections import OrderedDict
+import ast
```
pombredanne: nit: this could be sorted

PR #1807, opened 2019-11-02
Add notice about Python 3

filips123:
Add notice that Python 2.7 up to 3.6.8 is supported to README. This should be correct, but please review if any of installation methods don't work.

Also, this should also be added to [repository wiki](https://github.com/nexB/scancode-toolkit/wiki/Comprehensive-Installation). I will do this after I'm sure that all changes here are correct.

@pombredanne @Abhishek-Dev09

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -84,6 +84,7 @@ Pre-requisites:
 
   * On Ubuntu 14, 16 and 18 use:
     ``sudo apt-get install python3.6-dev xz-utils zlib1g libxml2-dev libxslt1-dev bzip2``
+    You may need to install Python PPA repository: https://askubuntu.com/a/865569
 
   * On Debian and Debian-based distros use:
     ``sudo apt-get install python3.6-dev xz-utils zlib1g libxml2-dev libxslt1-dev libbz2-1.0``
--- a/README.rst
+++ b/README.rst
@@ -55,8 +55,10 @@ Build and tests status
 Quick Start
 ===========
 
-Install Python 2.7 then download and extract the latest ScanCode release from
+Install Python 3.6.8 then download and extract the latest ScanCode release from
 https://github.com/nexB/scancode-toolkit/releases/ 
+
+Although ScanCode supports Python 2.7 up to 3.6.8, it is recommended to use the latest supported Python version.
 
 Then run ``./scancode -h`` for help.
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -68,29 +70,29 @@ Pre-requisites:
...
-    ``sudo apt-get install python-dev xz-utils zlib1g libxml2-dev libxslt1-dev bzip2``
+    ``sudo apt-get install python3.6-dev xz-utils zlib1g libxml2-dev libxslt1-dev bzip2``
 
   * On Debian and Debian-based distros use:
-    ``sudo apt-get install python-dev xz-utils zlib1g libxml2-dev libxslt1-dev libbz2-1.0``
+    ``sudo apt-get install python3.6-dev xz-utils zlib1g libxml2-dev libxslt1-dev libbz2-1.0``
 
   * On RPM distros use:
-    ``sudo yum install python-devel xz-libs zlib libxml2-devel libxslt-devel bzip2-libs``
+    ``sudo yum install python3.6-devel xz-libs zlib libxml2-devel libxslt-devel bzip2-libs``
```
pombredanne: Same as above, are you sure that `python3.6-devel` is the correct package name and is available?

README.rst
```diff
@@ -55,9 +55,11 @@ Build and tests status
 Quick Start
 ===========
 
-Install Python 2.7 then download and extract the latest ScanCode release from
+Install Python 3.6.8 then download and extract the latest ScanCode release from
 https://github.com/nexB/scancode-toolkit/releases/ 
 
+Although ScanCode supports Python 2.7 up to 3.6.8, it is recommended to use the latest supported Python version.
```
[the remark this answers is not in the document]
filips123: There is also Python 3.6.9. Is it supported?

I will fix those requested changes later today.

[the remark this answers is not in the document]
pombredanne: Any sub-version of 3.6 should be support (3.6.*)

filips123:
I will check if that package names are correct tomorrow.

README.rst
```diff
@@ -68,29 +70,29 @@ Pre-requisites:
...
   Next, download and extract the latest ScanCode release from
   https://github.com/nexB/scancode-toolkit/releases/
 
-* On Linux install the Python 2.7 "devel" and these packages using your
+* On Linux install the Python 3.6 "devel" and these packages using your
   distribution package manager:
 
   * On Ubuntu 14, 16 and 18 use:
-    ``sudo apt-get install python-dev xz-utils zlib1g libxml2-dev libxslt1-dev bzip2``
+    ``sudo apt-get install python3.6-dev xz-utils zlib1g libxml2-dev libxslt1-dev bzip2``
```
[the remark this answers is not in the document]
filips123: Ubuntu 14 and 16 do not have Python 3.6 in directly asvailible from `apt-get`. But it can be supported with [installing from PPA](https://askubuntu.com/a/865569).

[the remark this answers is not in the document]
pombredanne: :+1:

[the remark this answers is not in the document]
filips123: I added notice about PPA.

README.rst
filips123: Unfourtunately, Yum handles varsions a bit diffrent. Because of that, `python36-devel` does not exist and only Python 3.7 is availible.

(Tested on Fedora in Docker)

pombredanne: You rock :bowing_man:  ... but what about EPEL?

filips123: I think that EPEL also does not have Python 3.6 version of `python-devel`...

filips123: In case if Python 2.7 is still supported, I can update instructions to only use Python 2.7 on RPM.

But the real solution would be to update ScanCode to support newer versions of Python and then also update instructions to just install any (newer) Python 3 version.

pombredanne: >  to just install any (newer) Python 3 version.

Yes, but we will never support anything before Python 3.6

pombredanne:
@filips123 are we all set? Do you mind to rebase so I can review and  merge?

README.rst
filips123: But is now Python 3.7 supported?

pombredanne:
@filips123 ping

filips123:
The problem is Python 3.6 handling in distributions that use Yum. I think that they just have 3.7+ available.

The solution for this is to check if ScanCode supports 3.7 and newer versions (and obviously support them if they are not supported already), and then fix instructions to install any newer version.

filips123:
Recent README and documentation changes were already updated to recommend Python 3.6.8 so this PR is not needed anymore.

Issue #1821, opened 2019-11-08
consolidation plugin crashes with ZeroDivisionError exception


JonoYang:
The consolidation plugin crashed with a ZeroDivisionError when scanning some private code. This is related to the line of code: https://github.com/nexB/scancode-toolkit/blob/develop/src/scancode/plugin_consolidate.py#L368

I will have to check to see what exactly was scanned that triggered this, but I can put a check to see if the file count is zero before using it for division.

JonoYang:
This is no longer relevant since we are creating consolidations for anything with a holder detected in it.

PR #1830, opened 2019-11-13
Ensure current file count is not zero before determining majority license-holder during consolidation

JonoYang:
RE: #1821 is_majority throws a ZeroExceptionError when `current_file_only_count` is zero. I am including a guard to prevent us from using `current_file_only_count` if it is zero.


JonoYang:
@pombredanne I'll take a look. I encountered this error while scanning some private code and I am not sure what created the error.

JonoYang:
#1877 has been merged in, so this is no longer needed.

PR #1834, opened 2019-11-19
Add new license rule for unicode-mappings

qduanmu:


qduanmu:
Per feedback from our legal user, this license text is essentially the same as the "Unicode Mappings License", their texts cover the same thing, so I imported it as a new license rule of 'unicode-mappings'.

Issue #1833, opened 2019-11-18
Cannot install v3.1.1 release candidate from Pypi on Python2 


pombredanne:
### Description
That version was built on Python 3 and requires Python 3.6 or higher.

`Requires-Python: >=3.6` is in the generated metadata

pombredanne:
We no longer support Python 2 and the PyPI installation works now.

Issue #335, opened 2016-10-12
Support Python 64 bits on Windows 64


pombredanne:
Today only Python 32 bits is supported on Windows 64

Following a discussion in https://github.com/nexB/scancode-toolkit/pull/333
@sschuberth wrote:

> when will ScanCode run with 64-bit Python on Windows? 

@pombredanne replie:

> running on 64bits should not be a big issue, this is just about recompiling all the thirdparty and ensuring the tests pass: I have started working on getting 64 bits Python wheels here https://github.com/pombreda/thirdparty to have a fully automated rebuild of all thirddparties. There are some of them that can compile only under Mingw and I did not come down to these yet. If you are interested to chip in I could guide you alright. This is is mostly about native libraries and testing: the candidates are likely libarchive, libmagic and 7zip and it is very possible that no recompilation is needed at all since these are either spawned (7zip) or use with a ctypes binding (libarchive, libmagic)

pombredanne:
#616 is alos the ticket that was used to track this

pombredanne:
@chinyeungli Can you tell if this works fine for you now?

chinyeungli:
```
Python 3.6.0 (v3.6.0:41df79263a11, Dec 23 2016, 08:06:12) [MSC v.1900 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import struct
>>> print(struct.calcsize("P") * 8)
64
```
I am running 64bits, and I have no problem of running scancode-tk

pombredanne:
@chinyeungli Thanks
At this stage I am calling this done and closed!

PR #1835, opened 2019-11-19
Make PyPI installation work on Python 2 #1833 

pombredanne:
This PR is a fix for #1833 to properly install plugins on Python 2.
There is in fact some updates as plugins are moved to their own repos, we add new native wheels for windows64 and update the deps. And #335 to support Windows64.
Along the way there are a few other things fixed such as  #1971 and extractcode support for 7z on case insensitive macOS filesystems

It also tries to speed up the CI test runs by cleaning some of them.


pombredanne:
At last this is passing on all OSes. Merging!

Issue #1010, opened 2018-03-27
ScanCode should not consider scanning devices and special files


JonoYang:
This is a regression

pombredanne:
Thanks. That's a regression introduced with the new codebase model.

adityaviki:
@pombredanne 

What do you mean by "regression"? also, what type of file should not be considered for scanning?

pombredanne:
@adityaviki a "regression" is a general term for something that used to work and no longer works (it did "regress"). Here I am pretty sure that we did not attempt in the past to scan any character device file, FIFO and similar.
At least that must have been the case in v1.0
That said, we want to skip these: FIFO, character device files, symlinks, 

For instance you can see that the code in https://github.com/nexB/scancode-toolkit/blob/v1.2.0/src/commoncode/fileutils.py#L250 was skipping these special files (and walk() was used then by clip.py

JonoYang:
@pombredanne I've run scancode against `/dev` on my system using `sudo scancode -i /dev --json ~/Desktop/dev.json`, and scancode correctly did not scan the special files and only reported the directories under `/dev`.

pombredanne:
> but I am wondering if we should be more verbose about the fact we are skipping special files or if we should prevent scancode from attempting to create a Codebase from a path to a special file?

@JonoYang please create a separate ticket then and let's close this one

PR #1839, opened 2019-11-20
Add /dev and /sys to ignores #1010

JonoYang:
When we create a codebase, we only ignored the immaterial VCS files when we should have also ignored OS-related files at the same time. This PR adds `/dev/` and `/sys` to the list of patterns to ignore and also properly ignores directories when creating a codebase.


JonoYang:
Ignoring `/dev/*` and `/sys/*` is too heavy handed. It appears that we are ignoring the right things already, but the error handling isn't working properly during codebase creation.

PR #1840, opened 2019-11-20
Mark any time a resource has a majority during Consolidation

JonoYang:
This is useful for cases where we want increased granularity of where majorities occur. Since this is a new flag in  `extra_data`, this will not impact scan outputs or results.


JonoYang:
#1877 was merged into develop, so this PR is no longer relevant.

PR #1852, opened 2019-12-05
composer.lock Parser

JonoYang:
This enhances the current Composer package parser to also parse composer.lock files and return the detected required runtime and dev packages as Packages.

### DIFF
--- a/src/packagedcode/phpcomposer.py
+++ b/src/packagedcode/phpcomposer.py
@@ -34,6 +34,7 @@ import logging
 import sys
 
 import attr
+from packageurl import PackageURL
 from six import string_types
 
 from commoncode import filetype
@@ -69,8 +70,12 @@ if TRACE:
 
 @attr.s()
 class PHPComposerPackage(models.Package):
-    metafiles = ('composer.json',)
-    filetypes = ('.json',)
+    metafiles = (
+        'composer.json',
+        'composer.lock',
+        'installed.json',
+    )
+    filetypes = ('.json', '.lock')
     mimetypes = ('application/json',)
     default_type = 'composer'
     default_primary_language = 'PHP'
@@ -80,7 +85,8 @@ class PHPComposerPackage(models.Package):
 
     @classmethod
     def recognize(cls, location):
-        yield parse(location)
+        for package in parse(location):
+            yield package
 
     @classmethod
     def get_package_root(cls, manifest_resource, codebase):
@@ -135,6 +141,10 @@ def is_phpcomposer_json(location):
     return (filetype.is_file(location) and fileutils.file_name(location).lower() == 'composer.json')
 
 
+def is_phpcomposer_lock(location):
+    return (filetype.is_file(location) and fileutils.file_name(location).lower() == 'composer.lock')
+
+
 def parse(location):
     """
     Return a Package object from a composer.json file or None.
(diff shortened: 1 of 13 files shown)


src/packagedcode/phpcomposer.py
```diff
@@ -135,23 +146,30 @@ def is_phpcomposer_json(location):
...
-        return
-
-    with io.open(location, encoding='utf-8') as loc:
-        package_data = json.load(loc, object_pairs_hook=OrderedDict)
-
-    return build_package(package_data)
-
-
-def build_package(package_data):
+    is_json = is_phpcomposer_json(location)
```
pombredanne: This code below (e.g. the if json or is lock test) likely means that you would be better off with a different class or subclass... But since we are stuck with one class per Package type, I guess that the way to to go for now .
Now this is minor and feel free to ignore this but may be this would be a tad simpler code?
```
if is_phpcomposer_json(location):
    with io.open(location, encoding='utf-8') as loc:
        package_data = json.load(loc, object_pairs_hook=OrderedDict)
    yield build_package_from_json(package_data)

elif is_phpcomposer_lock(location):
    with io.open(location, encoding='utf-8') as loc:
        package_data = json.load(loc, object_pairs_hook=OrderedDict)

    for package in build_package_from_lock(package_data):
        yield package

```

src/packagedcode/phpcomposer.py
```diff
@@ -336,3 +354,56 @@ def parse_person(persons):
...
+    Yield a composer Package object from a package data mapping that originated
+    from a composer.lock file
+    """
+    packages = package_data.get('packages', []) + package_data.get('packages-dev', [])
+    for package in packages:
+        dependencies = []
+        for dep, req in package.get('require', {}).items():
+            dependencies.append(
+                models.DependentPackage(
+                    purl=PackageURL(type='composer', name=dep).to_string(),
```
pombredanne: The dependencies in a lock file would be special because "is_resolved" would be True. that is all the packages in the lock file are eventually the deps that are have been resolved?
(with care not to be mixing up the deps in the package reord of a lock file entry and a lock file entry which is itself a dependent Package of the root Package, e.g. the one in the composer.json)

tests/packagedcode/test_phpcomposer.py
```diff
@@ -58,35 +58,55 @@ def test_parse_person(self):
...
+        packages = list(phpcomposer.parse(test_file))
+        assert len(packages) > 0
+        package = packages[0]
         self.check_package(package, expected_loc, regen=False)
 
     def test_parse_slim(self):
         test_file = self.get_test_loc('phpcomposer/slim/composer.json')
         expected_loc = self.get_test_loc('phpcomposer/slim/composer.json.expected')
-        package = phpcomposer.parse(test_file)
+        packages = list(phpcomposer.parse(test_file))
```
pombredanne: Why testing a single package? and not all of them?

src/packagedcode/phpcomposer.py
```diff
@@ -135,23 +146,30 @@ def is_phpcomposer_json(location):
     return (filetype.is_file(location) and fileutils.file_name(location).lower() == 'composer.json')
 
 
+def is_phpcomposer_lock(location):
+    return (filetype.is_file(location) and fileutils.file_name(location).lower() == 'composer.lock')
```
pombredanne: nit picking: The parens wrapping is not needed

src/packagedcode/phpcomposer.py
```diff
@@ -336,3 +354,56 @@ def parse_person(persons):
                 url and url.strip('() '))
     else:
         raise ValueError('Incorrect PHP composer persons: %(persons)r' % locals())
+
+
+def build_package_from_lock(package_data):
+    """
+    Yield a composer Package object from a package data mapping that originated
+    from a composer.lock file
```
pombredanne: Is the logic to return one package for each entry of the lock file?

JonoYang: Yes. I report every package listed there.

src/packagedcode/phpcomposer.py
JonoYang: ack

tests/packagedcode/test_phpcomposer.py
JonoYang: I will update the PackageTester class to be able to check lists of packages instead of single packages

pombredanne:
Reviewed and LGTM. Merging. Thank you

Issue #1850, opened 2019-12-05
Parse composer.lock files


JonoYang:
It would be nice to be able to get info from these files during a scan

JonoYang:
This has been merged into develop.

PR #1857, opened 2019-12-11
1220 lockfile support

JonoYang:
Related to #1850 and #1852, this PR adds more lock file parsers to ScanCode.


pombredanne:
Thanks... Here are a couple sources of yarn lock test files to consider:
- https://github.com/imsnif/synp/tree/master/test/fixtures (also interesting because it's description is `Convert yarn.lock to package-lock.json and vice versa`
- https://github.com/yarnpkg/yarn/blob/master/__tests__/fixtures

JonoYang:
@pombredanne Thanks for the links. This test looks interesting: https://github.com/imsnif/synp/blob/master/test/fixtures/file-deps-yarn/yarn.lock

Should ScanCode consider handling dependencies where the requirement is pointing to a file?

pombredanne:
> Should ScanCode consider handling dependencies where the requirement is pointing to a file?

Not sure. What's your take?

JonoYang:
It would be useful to know that this is a dependency being used, but this is also the first time I seen something like this in a project, so I don't think this is that common.

pombredanne:
All good, but we need to use smaller test files... the yarn.lock sample  you have for tests is a tad big.
Merging though... you can fix it later

PR #1882, opened 2020-01-19
Add Keywords to ignore in spam Email detection

### LINKED ISSUE #997: Filter more junk emails
For instance:
testUser@gmail.com
some@other.com

Gaupeng:
Signed-off-by: Gaupeng <gaurang.r.rao@gmail.com>

Fixes #997 
Adds more keywords to ignore junk/spam emails during detection.
Also adds documentation for a previously defined function `classify_url`

### Tasks

  Run [tests](https://aboutcode.readthedocs.io/en/latest/scancode-toolkit/contribute/contrib_dev.html#running-tests) locally to check for errors. 


Gaupeng:
Oops, realised this is quite a blunder. Filters quite a few valid emails.

src/cluecode/finder_data.py
```diff
@@ -212,6 +227,11 @@ def classify(s, data_set, suffixes=None):
     if not s:
         return False
     s = s.lower().strip('/')
+    # Separate test for emails - need to ignore xyz@some.com, but not say, xyz@gruesome.com
+    if '@' in s and exact_match:
```
pombredanne: Could `ignored_hosts` or similar be a better name for this variable? e.g. using a name that describes what it is, not a name that describes what this used for.

src/cluecode/finder_data.py
```diff
@@ -212,6 +227,11 @@ def classify(s, data_set, suffixes=None):
...
+        email_domain = s.split('@')
```
pombredanne: Is it possible that there is more than on @ in that string? and if yes, would the segment on the right always be the domain?
What about using a `'.rpartition()` instead of `split()` and using proper names (e.g. host_name) for each part?

src/cluecode/finder_data.py
```diff
@@ -223,10 +243,13 @@ def classify(s, data_set, suffixes=None):
 
 classify_host = partial(classify, data_set=JUNK_HOSTS_AND_DOMAINS, suffixes=JUNK_DOMAIN_SUFFIXES)
 
-classify_email = partial(classify, data_set=JUNK_EMAILS, suffixes=JUNK_DOMAIN_SUFFIXES)
+classify_email = partial(classify, data_set=JUNK_EMAILS, suffixes=JUNK_DOMAIN_SUFFIXES, exact_match=JUNK_EXACT_DOMAINS)
 
 
 def classify_url(url):
+    '''
```
pombredanne: Please use double quotes for docstrings.

src/cluecode/finder_data.py
```diff
@@ -49,7 +51,6 @@ def set_from_text(text):
     example.com
     example.net
     example.org
-    test.com
```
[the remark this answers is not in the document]
Gaupeng: Firstly, thanks for the (very) thorough review, will look into everything you've suggested! 
I removed it, as it was being checked for using the `exact_match` test, but now I see cases (thanks to you!) where it could fail. Will work on it!

src/cluecode/finder_data.py
Gaupeng: Yes, sounds good to me, makes more sense to name it that way.

src/cluecode/finder_data.py
Gaupeng: Looking at the [Regular Expression for an email](https://en.wikipedia.org/wiki/Email_address#Syntax), the @ symbol is recommended to be used only once, however it can be used again within a quoted string. Short answer, yes, there could be a case where the last @ symbol is not followed by a domain, though highly unlikely. Do I have to assume we will be looking at emails such as those? 
> What about using a '.rpartition() instead of split() and using proper names (e.g. host_name) for each part?
Sure, that should do the job, provided the last @ is followed by a domain (a fairly strong assumption.)

src/cluecode/finder_data.py
Gaupeng: Noted, thanks!

Gaupeng:
@pombredanne, made changes as per your suggestions, and also added test cases. Thought I'd let you know, even though there was a function [`uninteresting_email_filters`](https://github.com/nexB/scancode-toolkit/blob/develop/src/cluecode/finder.py#L184), it was never actually added to the filter [here](https://github.com/nexB/scancode-toolkit/blob/develop/src/cluecode/finder.py#L163) while scanning emails (thus making it a function that's never called?)
Anyhow, I've made changes, and now I'm using that filter too. Was there a reason as to why that filter was not used?

Also, please check the changes, and let me know if I have to make further changes!
Thanks!

pombredanne:
@Gaupeng at this stage this would need to be rebased before a final review

src/cluecode/finder_data.py
```diff
@@ -200,18 +217,24 @@ def set_from_text(text):
...
     the data in string s is not junk. Return False if the data in string s is
-    classified as 'junk' or uninteresting.
+    classified as 'junk' or uninteresting. Return False is domain is in `ignored_hosts`.
     """
     if not s:
         return False
     s = s.lower().strip('/')
+    # Separate test for emails - need to ignore xyz@some.com, but not say, xyz@gruesome.com
+    if '@' in s and ignored_hosts:
+        host_name = s.rpartition('@')[-1]
```
pombredanne: Rather than using [-1] use something like this that is a tad more explicit:
`_name, _at, host_name = s.rpartition('@')`

tests/cluecode/test_finder.py
```diff
@@ -182,6 +182,23 @@ def test_emails_does_filter_junk_domains(self):
         result = find_emails_tester(test_file)
         assert expected == result
 
+    def test_emails_for_ignored_hosts(self):
+        """
+        Add tests for ignored_hosts filter, for email domain hosts.
+        """
```
pombredanne: Please avoid using docstrings in tests: either use longer, more ddescriptive test method names or use comments.
Test runner otherwise tend to report the test doc rather than the method name and this does snot help much in reports and failure traces.

src/cluecode/finder_data.py
```diff
@@ -223,10 +246,13 @@ def classify(s, data_set, suffixes=None):
 
 classify_host = partial(classify, data_set=JUNK_HOSTS_AND_DOMAINS, suffixes=JUNK_DOMAIN_SUFFIXES)
 
-classify_email = partial(classify, data_set=JUNK_EMAILS, suffixes=JUNK_DOMAIN_SUFFIXES)
+classify_email = partial(classify, data_set=JUNK_EMAILS, suffixes=JUNK_DOMAIN_SUFFIXES, ignored_hosts=JUNK_EXACT_DOMAIN_NAMES)
 
 
 def classify_url(url):
+    """
+    Return False if `url` found to be a candidate for a spam url. Return True otherwise.
```
pombredanne: What about this instead?
```
Return True if `url` is a good url or False if this is spam url.
```
BUT since this is completely unrelated to your main PR, may be this addition is not needed?

src/cluecode/finder_data.py
```diff
@@ -200,18 +217,24 @@ def set_from_text(text):
    .png
    .jpg
    .gif
+   .jpeg
 ''')))
 
 
-def classify(s, data_set, suffixes=None):
+def classify(s, data_set, suffixes=None, ignored_hosts=None):
```
pombredanne: Can you add tests for this method since you changed it?

src/cluecode/finder_data.py
```diff
@@ -200,18 +217,24 @@ def set_from_text(text):
    .png
    .jpg
    .gif
+   .jpeg
```
pombredanne: Can you elaborate why you made this addition and why this is related to spam Email detection?

src/cluecode/finder_data.py
```diff
@@ -35,6 +35,9 @@ def set_from_text(text):
 
 JUNK_EMAILS = set_from_text(u'''
     test@test.com
+    testuser
+    trialuser
+    sampleuser
```
pombredanne: How are these three actual emails?
Did you add tests for that?

AUTHORS.rst
```diff
@@ -16,3 +16,5 @@ The following organizations or individuals have contributed to ScanCode:
 - Tushar Mittal @techytushar
 - Martin Petkov @MartinPetkov
 - Mrinal Paliwal @mriiinal
+- Mankaran Singh @MankaranSingh
+- Gaurang Rao @Gaupeng
```
pombredanne: Is this a joint development? Why do you add @MankaranSingh name?

Gaupeng: Whoops, big oopsie on my side. WIll fix it.

src/cluecode/finder_data.py
Gaupeng: Since we're checking for junk suffixes, and we consider other image/gif formats to be candidates for junk suffixes, I assumed a .jpeg suffix would be considered here too, especially while classifying spam emails? Can revert it, if you disagree.

src/cluecode/finder_data.py
Gaupeng: Done, thank you!

src/cluecode/finder_data.py
Gaupeng: They're not, was my bad to include the here. Perhaps better to leave it out now, and work on them specifically later.

src/cluecode/finder_data.py
Gaupeng: I've added a few tests here: https://github.com/nexB/scancode-toolkit/pull/1882/files#diff-b64a4b3f81b0ea3294cba2a968debaa5R185
I'll add a few more with the next push!

PR #1893, opened 2020-02-10
add --replace-originals flag to extractcode

maxhbr:
This adds the flag `--remove-originals` to `extractcode`. This modifies the behavior to remove the archive that was extracted from the file tree.

Related to: https://github.com/nexB/scancode-toolkit/issues/14

### Tasks

  Run [tests](https://aboutcode.readthedocs.io/en/latest/scancode-toolkit/contribute/contrib_dev.html#running-tests) locally to check for errors. 

### DIFF
--- a/src/extractcode/extract.py
+++ b/src/extractcode/extract.py
@@ -97,7 +97,7 @@ An ExtractEvent contains data about an archive extraction progress:
 ExtractEvent = namedtuple('ExtractEvent', 'source target done warnings errors')
 
 
-def extract(location, kinds=extractcode.default_kinds, recurse=False):
+def extract(location, kinds=extractcode.default_kinds, recurse=False, replace_originals=False):
     """
     Walk and extract any archives found at `location` (either a file or
     directory). Extract only archives of a kind listed in the `kinds` kind tuple.
@@ -116,10 +116,26 @@ def extract(location, kinds=extractcode.default_kinds, recurse=False):
     archives If `recurse` is false, then do not extract further an already
     extracted archive identified by the corresponding extract suffix location.
 
+    If `replace_originals` is True, the archives will be extracted in place.
+
     Note that while the original file system is walked top-down, breadth-first,
     if recurse and a nested archive is found, it is extracted to full depth
     first before resuming the file system walk.
     """
+    events = list(extract_files(location, kinds, recurse))
+    if replace_originals:
+        for xevent in reversed(events):
+            if xevent.done:
+                source = xevent.source
+                target = xevent.target
+                if TRACE:
+                    logger.debug('extract:replace_originals: replace %(source)r by %(target)r' % locals())
+                fileutils.delete(source)
+                fileutils.copytree(target, source)
+                fileutils.delete(target)
+    return events
+
+def extract_files(location, kinds=extractcode.default_kinds, recurse=False):
     ignored = partial(ignore.is_ignored, ignores=ignore.default_ignores, unignores={})
     if TRACE:
         logger.debug('extract:start: %(location)r  recurse: %(recurse)r\n' % locals())
--- a/src/extractcode/extract.py
+++ b/src/extractcode/extract.py
@@ -116,7 +116,8 @@ def extract(location, kinds=extractcode.default_kinds, recurse=False, replace_or
     archives If `recurse` is false, then do not extract further an already
     extracted archive identified by the corresponding extract suffix location.
 
-    If `replace_originals` is True, the archives will be extracted in place.
+    If `replace_originals` is True, the extracted archives are replaced by the
+    extracted content.
 
     Note that while the original file system is walked top-down, breadth-first,
     if recurse and a nested archive is found, it is extracted to full depth
--- a/src/scancode/api.py
+++ b/src/scancode/api.py
@@ -312,7 +312,7 @@ def get_file_info(location, **kwargs):
     return result
 
 
-def extract_archives(location, recurse=True):
+def extract_archives(location, recurse=True, replace_originals=False):
     """
     Yield ExtractEvent while extracting archive(s) and compressed files at
     `location`. If `recurse` is True, extract nested archives-in-archives
@@ -323,5 +323,5 @@ def extract_archives(location, recurse=True):
     """
     from extractcode.extract import extract
     from extractcode import default_kinds
-    for xevent in extract(location, kinds=default_kinds, recurse=recurse):
+    for xevent in extract(location, kinds=default_kinds, recurse=recurse, replace_originals=replace_originals):
         yield xevent
--- a/src/scancode/extract_cli.py
+++ b/src/scancode/extract_cli.py
@@ -89,11 +89,12 @@ Try 'extractcode --help' for help on options and arguments.'''
 @click.option('--verbose', is_flag=True, default=False, help='Print verbose file-by-file progress messages.')
 @click.option('--quiet', is_flag=True, default=False, help='Do not print any summary or progress message.')
 @click.option('--shallow', is_flag=True, default=False, help='Do not extract recursively nested archives (e.g. not archives in archives).')
+@click.option('--replace-originals', is_flag=True, default=False, help='Replace extracted archives by the extracted content.')
 
 @click.help_option('-h', '--help')
 @click.option('--about', is_flag=True, is_eager=True, callback=print_about, help='Show information about ScanCode and licensing and exit.')
 @click.option('--version', is_flag=True, is_eager=True, callback=print_version, help='Show the version and exit.')
-def extractcode(ctx, input, verbose, quiet, shallow, *args, **kwargs):  # NOQA
+def extractcode(ctx, input, verbose, quiet, shallow, replace_originals, *args, **kwargs):  # NOQA
     """extract archives and compressed files found in the <input> file or directory tree.
 
     Use this command before scanning proper as an <input> preparation step.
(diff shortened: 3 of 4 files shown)


src/extractcode/extract.py
```diff
@@ -178,16 +178,22 @@ def extract_file(location, target, kinds=extractcode.default_kinds, verbose=Fals
                      + getattr(extractor, '__module__', '')
                      + '.' + getattr(extractor, '__name__', ''))
     if extractor:
-        yield ExtractEvent(location, target, done=False, warnings=[], errors=[])
+        yield ExtractEvent(location, target if not replace_originals else location, done=False, warnings=[], errors=[])
```
pombredanne: Why would the `target` (e.g. where we extract) become the `location` (what we extract) with `replace_originals`? that's mostly informative because that what events are about... but still?

src/extractcode/extract.py
```diff
@@ -178,16 +178,22 @@ def extract_file(location, target, kinds=extractcode.default_kinds, verbose=Fals
...
             warnings.extend(warns)
-            fileutils.copytree(tmp_tgt, target)
-            fileutils.delete(tmp_tgt)
+            if not replace_originals:
+                fileutils.copytree(tmp_tgt, target)
+                fileutils.delete(tmp_tgt)
+            else:
+                if TRACE:
+                    logger.debug('extract_file: replace original with extracted content: %(location)r' % locals())
+                fileutils.delete(location)
```
pombredanne: Are you sure you want to delete `location` and not the `target`?

maxhbr:
> BUT still there is something which may not be working OK. See https://ci.appveyor.com/project/nexB/scancode-toolkit/builds/30698997/job/ag1rs0yofobgsfd3#L4406

There was is an error in the recursion, I have to fix that.

maxhbr:
> Let me reformulate the intent and outcome to make sure we are in sync: this forces a deletion of the extraction target and a new re-extraction of any archive?

I think we are not in sync here. My goal is to change the behavior in the following way:
instead of extracting to
```
            a/a.tar.gz
            a/a.txt
            a/a.tar.gz-extract/a/b/a.txt
            a/a.tar.gz-extract/a/b/b.txt
            a/a.tar.gz-extract/a/c/c.txt
```
I want to extract "in place" and get:
```
            a/a.txt
            a/a.tar.gz/a/b/a.txt
            a/a.tar.gz/a/b/b.txt
            a/a.tar.gz/a/c/c.txt
```

src/extractcode/extract.py
maxhbr: As far as I understand the code, the variables could be the following:
- `location=a/a.tar.gz`
- `target=a/a.tar.gz-extract`

Since my changes extract the content of the file `location` to the folder `location` instead of the folder `target`. I thus think that this change here is correct.

maxhbr:
this code breaks at: https://github.com/nexB/scancode-toolkit/blob/16377111febfaef2fd0806200a43cd9f239f7d0a/src/extractcode/__init__.py#L102-L109

maxhbr:
@pombredanne I have updated the description and the intend of this PR. It now no longer tries to do "in place extraction" but instead just removes the original archives after they were extracted.

pombredanne:
> @pombredanne :
> 
> > Let me reformulate the intent and outcome to make sure we are in sync: this forces a deletion of the extraction target and a new re-extraction of any archive?
> 
> I think we are not in sync here. My goal is to change the behavior in the following way:
> instead of extracting to
> 
> ```
>             a/a.tar.gz
>             a/a.txt
>             a/a.tar.gz-extract/a/b/a.txt
>             a/a.tar.gz-extract/a/b/b.txt
>             a/a.tar.gz-extract/a/c/c.txt
> ```
> 
> I want to extract "in place" and get:
> 
> ```
>             a/a.txt
>             a/a.tar.gz/a/b/a.txt
>             a/a.tar.gz/a/b/b.txt
>             a/a.tar.gz/a/c/c.txt
> ```

@maxhbr I get it... and I am glad I asked. So this becomes a destructive extraction then, but I can see how this makes things simpler towards #14 alright!

So I wonder if an approach may be instead to do this:
1. extract things as usual
2. once done and if you have the arg to replace/overwrite
2.1. remove the `location`
2.2. move the `target` to the `location`

pombredanne:
> @pombredanne I have updated the description and the intend of this PR. It now no longer tries to do "in place extraction" but instead just removes the original archives after they were extracted.

 Even better! let me review this tomorrow

src/extractcode/extract.py
```diff
@@ -151,21 +151,21 @@ def extract(location, kinds=extractcode.default_kinds, recurse=False):
...
+            for xevent in extract_file(loc, target, kinds, remove_originals=remove_originals):
                 if TRACE:
                     logger.debug('extract:walk:extraction event: %(xevent)r' % locals())
                 yield xevent
 
             if recurse:
                 if TRACE:
                     logger.debug('extract:walk: recursing on target: %(target)r' % locals())
-                for xevent in extract(target, kinds, recurse):
+                for xevent in extract(target if remove_originals else localtion, kinds, recurse, remove_originals=remove_originals):
```
pombredanne: `localtion` likely would need to be `location`...

src/extractcode/extract.py
pombredanne: YOu do... now that I get the intent correctly/

src/scancode/api.py
```diff
@@ -323,5 +323,5 @@ def extract_archives(location, recurse=True):
     """
     from extractcode.extract import extract
     from extractcode import default_kinds
-    for xevent in extract(location, kinds=default_kinds, recurse=recurse):
+    for xevent in extract(location, kinds=default_kinds, recurse=recurse, remove_originals=remove_originals):
         yield xevent
```
pombredanne: Could it be that an alternative design would be to use the extract events location and target field (once the regular extraction took place) to actually to delete location and move/rename target in its place at once?

src/extractcode/extract.py
maxhbr: Good catch,  i though I had fixed that.

src/scancode/api.py
maxhbr: Sounds like a better idea. I will look into that! Thank you!

maxhbr:
@pombredanne I have now implemented your suggested way of using the events

src/extractcode/extract.py
```diff
@@ -97,7 +97,21 @@
...
+    if replace_originals:
+        for xevent in reversed(events):
+            if xevent.done:
+                source = xevent.source
+                target = xevent.target
+                if TRACE:
+                    logger.debug('extract:replace_originals: replace %(source)r by %(target)r' % locals())
+                fileutils.delete(source)
+                fileutils.copytree(target, source)
+                fileutils.delete(target)
```
pombredanne: Just FYI and nitpicking: You could have been using the standard library `os.rename/os.renames` instead rather than the `fileutils` copytree/delete. https://docs.python.org/3/library/os.html?highlight=os%20rename#os.rename 

On linux at least `os.rename` is using the Linux kernel `rename` "atomic" syscall behind the scenes which is best as it will pass or fail at once IMHO

If you want to go that route that would become:
```
                fileutils.delete(source)
                os.rename(target, source)
 ``` 

You would need to also add an `import os`  at the top https://github.com/nexB/scancode-toolkit/pull/1893/files#diff-16d915471be787f821dcd1d38891c2bdR32 

Again no change required, this is more of a FYI

File: src/extractcode/extract.py [hunk 6]
```diff
@@ -97,7 +97,21 @@
 ExtractEvent = namedtuple('ExtractEvent', 'source target done warnings errors')
 
 
-def extract(location, kinds=extractcode.default_kinds, recurse=False):
+def extract(location, kinds=extractcode.default_kinds, recurse=False, replace_originals=False):
```
pombredanne: Do you mind  moving up the docstring that is now attached to the new `extract_files` function below https://github.com/nexB/scancode-toolkit/pull/1893/files#diff-16d915471be787f821dcd1d38891c2bdR116?

You could also add a doc note about the new `replace_originals` argument 
there.

src/extractcode/extract.py
```diff
@@ -97,7 +97,21 @@
...
+    events = list(extract_files(location, kinds, recurse))
+    if replace_originals:
+        for xevent in reversed(events):
```
pombredanne: Is going reverse needed here? e.g. essentially bottom up instead of top down?
I sense this is needed but I am not 100% sure.

maxhbr: otherwise one would move `tmp.zip-extracted` before moving `tmp.zip-extracted/innerArchive.zip-extracted`

src/extractcode/extract.py
maxhbr: I also had that code, but I thought it looks more consistent if I use your utils

src/extractcode/extract.py
pombredanne: OK: that what I thought fuzzily somehow!

src/extractcode/extract.py
pombredanne: Both work. The fileutils ways have been heavily tested on all OSes too. So Let's keep it.

maxhbr:
@pombredanne the comments are resolved

JakobSchubert:
Hi @pombredanne,
i wanted to ask if you could trigger the merge, I would love to try out this feature.

Have a nice day,
Jakob

pombredanne:
@JakobSchubert sure thing!
@maxhbr thanks for the hard work :bowing_man:

PR #1941, opened 2020-02-24
Update exception message check in run_scan

JonoYang:
This PR changes the string to be checked in a TypeError exception message depending on the Python version used to run ScanCode during the plugin loading process since the TypeError exception message has changed between Python 2 and Python 3.


JonoYang:
The issue is user related and not related to scancode.

Issue #1723, opened 2019-09-11
Incorrect ScanCode version warning


majurg:
### Description
```(scancode-toolkit) [sesser@nexb ~/Code/scancode-toolkit] scancode -n 6 -clipeu --json-pp ~/Code/workbench-test-scans/sigar-1.6.4-clipeu.json ~/Downloads/sigar-sigar-1.6.4.tar.gz-extract/
Setup plugins...
Collect file inventory...
Scan files for: info, licenses, copyrights, packages, emails, urls with 6 process(es)...
[####################] 317                                            
Scanning done.
Summary:        info, licenses, copyrights, packages, emails, urls with 6 process(es)
Errors count:   0
Scan Speed:     11.51 files/sec. 106.40 KB/sec.
Initial counts: 369 resource(s): 317 file(s) and 52 directorie(s) 
Final counts:   369 resource(s): 317 file(s) and 52 directorie(s) for 2.86 MB
Timings:
  scan_start: 2019-09-11T125343.605919
  scan_end:   2019-09-11T125439.560438
  setup_scan:licenses: 27.94s
  setup: 27.94s
  scan: 27.54s
  output:json-pp: 0.29s
  output: 0.29s
  total: 56.33s
Removing temporary files...done.
WARNING: You are using ScanCode Toolkit version 3.1.1.post12.d32880591, however the newer version 3.1.1 is available.
You should download and install the latest version of ScanCode with bug and security fixes and the latest license detection data for accurate scanning.
Visit https://github.com/nexB/scancode-toolkit/releases for details.
(scancode-toolkit) [sesser@nexb ~/Code/scancode-toolkit]
```

From the above output, scancode gives a warning that I am using an out-of-date version even though I am on develop branch version: `3.1.1.post12.d32880591`

### System configuration
For bug reports, it really helps us to know:
* What OS are you running on? **Linux**
* What version of scancode-toolkit was used to generate the scan file? **3.1.1.post12.d32880591**
* What installation method was used to install/run scancode? **git**

pombredanne:
@lechasseur @armijnhemel FYI

pombredanne:
This is still not fixed.  I want this fixed before the upcoming release!

steven-esser:
@pombredanne I think this is fixed?

armijnhemel:
Looks to me like it is fixed indeed.

steven-esser:
@armijnhemel Thanks. Closing.

PR #1944, opened 2020-02-27
Fix outdated.py version checking logic

majurg:
* Update check_scancode_version() to properly handle git-based versions
* Add test case

Addresses: #1723

### DIFF
--- a/src/scancode/outdated.py
+++ b/src/scancode/outdated.py
@@ -155,6 +155,16 @@ def check_scancode_version(installed_version=scancode_version,
             % (installed_version, latest_version)
         )
 
+        # Our git version string is not PEP 440 compliant, and thus improperly parsed via
+        # most 3rd party version parsers. We handle this case by pulling out the "base"
+        # release version by split()-ting on "post".
+        # 
+        # For example, "3.1.2.post351.850399ba3" becomes "3.1.2"
+        if isinstance(installed_version, packaging_version.LegacyVersion):
+            installed_version = installed_version.split('post')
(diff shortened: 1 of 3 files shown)


majurg:
@pombredanne Not sure why this final travis check fails for the PR... Its some `awesome_bot` url checker on the about files that passes locally and on the "push" Travis build. 

May just implement about check instead of using this ruby application to check the About files.

pombredanne:
> @pombredanne Not sure why this final travis check fails for the PR... Its some `awesome_bot` url checker on the about files that passes locally and on the "push" Travis build.
> 
> May just implement about check instead of using this ruby application to check the About files.

@AyanSinhaMahapatra ^ does this ring a bell?

AyanSinhaMahapatra:
@majurg @pombredanne Sorry, I missed this mention. Is this creating problems i.e. failing on valid links?

majurg:
@AyanSinhaMahapatra I think this was just a Travis fluke. It seems like Travis has strange behavior w at times, when doing network checks/operations. Restarting the failed job once or twice usually fixes things.

src/scancode/outdated.py
```diff
@@ -155,6 +155,16 @@ def check_scancode_version(installed_version=scancode_version,
             % (installed_version, latest_version)
         )
 
+        # Our git version string is not PEP 440 compliant, and thus improperly parsed via
```
pombredanne: Good catch... the alternative was to be not semver compliant AFAICR.

Issue #15, opened 2015-07-05
Control what is extracted with the extractcode command


pombredanne:
`extractcode` supports selecting what is extracted but this is not exposed as a command line option. 
There should be a way to control what is extracted possibly with an expanded option `--extract=<kind>` or but making extraction a separate of a sub command

agneet42:
I'd to like to work on this. By my understanding, this aims to provide more user-options.
Could you guide me a little further?  @pombredanne

agschrei:
Considering that this issue has seemingly gone stale, I am not sure how useful my contribution here is.
However, I would also like to get involved on this before filing a feature request in a new issue.

One problem I continue to have with extractcode is that it tries to extract sparse archives that are sometimes contained within source distributions, for example in the docker-ce source:
`docker-ce-19.03.8/components/engine/vendor/archive/tar/testdata/pax-sparse-big.tar-extract/pax-sparse`

These seemingly innocuous archives bloat to several gigabytes in size when extracted and regularly fill our scratchpad.
At the same time, since these are test archives, they don't provide any additional value in terms of identifying licenses/copyrights, etc.

My simple solution for this problem would be to add a CLI flag to extractcode that allows excluding files matching certain regex file patterns from being extracted.

An example might be: 
```bash
extractcode  myArchiveWithSparseTestFiles.tar --exclude="*sparse*;pax*"
```

Where the exclusion patterns are separated by semicolons.
Basically, this is the same approach that already exists in `scancode` as the `--ignore` flag.

I also assume that this issue is not unique to our use-case so if you agree that such an option provides additional value to users, I am more than willing to contribute it upstream @pombredanne

majurg:
@agschrei There is an open PR that deals with a similar issue: https://github.com/nexB/scancode-toolkit/pull/1946

This is done by a prospective GSoC student and AFAIK is a working solution. There is some cleanup to do on the PR iteself, and I have pinged the original author to see if he has some extra code changes. Otherwise I will clean it up and hopefully have it merged soon. 

We use `extractcode` often internally and have the same need for this feature that you do.

majurg:
@agschrei If interested, you can test out that particular branch here: https://github.com/JRavi2/scancode-toolkit/tree/add-ignore-flag

You may want to test out this solution on your end to see if there are any bugs in a real use case.

agschrei:
[...]

PR #1946, opened 2020-03-02
Add --ignore flag to extractcode

### LINKED ISSUE #1870: Provide an option in `extractcode` to extract sepecific extension files
For instance, I would only want to extract `.rpm`, `.zip`, `.tar.gz`, but not `.jar` files.

The current `extractcode` does not have the ability for user to define which files do the user want to extract or ignore from the extraction.

JRavi2:
Signed-off-by: Ravi <jravi248@gmail.com>

Fixes #1870

### Tasks


JRavi2:
@majurg Thanks for the review. 
From what I've understood, the `--ignore` feature in scancode is executed by calling the `is_ignored()` function in commoncode/ignore.py. So in src/extractcode/archive.py/should_extract(), instead of using the `shouldIgnore` flag, should I call the `is_ignored()` function from commoncode, like scancode does?

JRavi2:
@majurg I've made some changes to try to match the behavior with that of scancode. Can you please review them?

Also, I'm not able to figure out:
- Why the Travis CI build is failing
- Why there is such a drastic decrease in coverage in the Codecov report

Can you please help me with these.
Once the changes are approved I'll add the tests in.

pombredanne:
@JRavi2 you wrote:

> I've made some changes to try to match the behavior with that of scancode. Can you please review them?
> 
> Also, I'm not able to figure out:
> 
>     * Why the Travis CI build is failing

That's not your fault...

>> Checking links in thirdparty/wheel-0.33.6-py2.py3-none-any.whl.ABOUT
>>  3. https://bitbucket.org/pypa/wheel 

This URL in the ABOUT file is dead.

>     * Why there is such a drastic decrease in coverage in the Codecov report

 That's a covecov issue

src/extractcode/extract.py
```diff
@@ -176,7 +176,7 @@ def extract_files(location, kinds=extractcode.default_kinds, recurse=False):
             if recurse:
                 if TRACE:
                     logger.debug('extract:walk: recursing on target: %(target)r' % locals())
-                for xevent in extract(target, kinds, recurse):
+                for xevent in extract(target, ignored_extensions, kinds, recurse):
```
pombredanne: Here and elsewhere I think using named kwargs is usually better and a more robust approach: so it is fine to apply that change as in:
`for xevent in extract(target=target, kinds=kinds, recurse=recurse, ignored_extensions=ignored_extensions):`
both here and elsewhere you made changes

JRavi2:
@pombredanne Thanks for the review. I'll soon add in commits for the suggested changes.

JRavi2:
@pombredanne Sorry for the delay. I've added the suggested changes.

pombredanne:
The test failure on Travis is unrelated and a false alarm, so I will ignore that

src/extractcode/archive.py
```diff
@@ -36,6 +36,7 @@
 from commoncode import functional
 from commoncode.system import on_linux
 from commoncode.system import py2
+from commoncode.ignore import is_ignored
```
pombredanne: Sorry to be nit picking but could you keep this import sorted? ignore comes before system

src/extractcode/archive.py
```diff
@@ -103,13 +104,15 @@ def can_extract(location):
...
-def should_extract(location, kinds):
+def should_extract(location, kinds, ignored_extensions=()):
     """
     Return True if this location should be extracted based on the provided
     kinds
     """
     location = os.path.abspath(os.path.expanduser(location))
-    if get_extractor(location, kinds):
+    ignored_extensions = {extension : 'User ignore: Supplied by --ignore' for extension in ignored_extensions}
+    shouldIgnore = is_ignored(location, ignored_extensions)
```
pombredanne: We tend to prefer snake_case vs camelCase ... here that would mean `shouldIgnore` should rather be `should_ignore`

src/scancode/api.py
```diff
@@ -312,7 +312,7 @@ def get_file_info(location, **kwargs):
     return result
 
 
-def extract_archives(location, recurse=True, replace_originals=False):
+def extract_archives(location, recurse=True, replace_originals=False, ignored_extensions = ()):
```
pombredanne: Nit picking again, I can spot a few extra spaces in `ignored_extensions = ()` ... may be rather `ignored_extensions=()`

JRavi2:
@pombredanne I need some guidance for the tests. 
I know how to add the tests. I want to know what tests to add.

pombredanne:
@JRavi2 There is two things to look for now:
1. add tests

> Now, if we were to run ./extractcode . --ignore *.tar, it would ignore b.tar and would extract a.zip, b.zip , but will ignore c.tar. Will this test be enough?

That's a start as an integration test but likely not enough. Think about adding also unit tests for each function that you modified to verify that they behave correctly and do what they are supposed to do.

2. are you sure that you are ignoring only extensions? Or rather ignoring glob patterns?

pombredanne:
in anycase you tests in https://github.com/nexB/scancode-toolkit/pull/1946#issuecomment-598773905 in a start alright :+1:

JRavi2:
@pombredanne I am using the `is_ignored` function from commoncode, which I believe uses the glob patters only, so this option would ignore glob patterns only.
Regarding the unit tests, where should the test for `should_extract` function in `archive.py` be added? \
I see that there are different classes to test the extraction of different types of archives, but there is none specifically to test `should_extract`.
I also see that currently there is no existing test for `extract_archives` function in `test_api.py`, so is the test for that required?
Also, thanks for your patience :)

JRavi2:
@pombredanne I've added two tests, please have a look at them. For the rest two functions (`should_extract`, `extract_archives`), I need help as asked in the previous comment.

JRavi2:
@rpotter12 I apologize for the late response. I've resolved the conflicts. Thanks :)

src/scancode/extract_cli.py
```diff
@@ -90,11 +90,12 @@ class ExtractCommand(utils.BaseCommand):
 @click.option('--quiet', is_flag=True, default=False, help='Do not print any summary or progress message.')
 @click.option('--shallow', is_flag=True, default=False, help='Do not extract recursively nested archives (e.g. not archives in archives).')
 @click.option('--replace-originals', is_flag=True, default=False, help='Replace extracted archives by the extracted content.')
+@click.option('--ignore', default=[], multiple=True, help='Ignore files with a particular extension')
```
rpotter12, line 93: add full stop at the end of help like others have :)

src/extractcode/extract.py
```diff
@@ -125,7 +125,7 @@ def extract(location, kinds=extractcode.default_kinds, recurse=False, replace_or
     """
     processed_events = []
     processed_events_append = processed_events.append
-    for event in extract_files(location, kinds, recurse):
+    for event in extract_files(location, kinds, recurse, ignored_extensions):
```
majurg, line 128: is `ignored_extensions` always a file extention? What if I do something like `--ignore=specific-file.tar.gz`

src/scancode/extract_cli.py
majurg, line 93: `--ignore` should work the same way as it does in scancode, meaning you can ignore whole filename or directories, not just file extensions.

src/scancode/extract_cli.py
```diff
@@ -159,8 +160,9 @@ def display_extract_summary():
     extract_result_with_errors = []
     unique_extract_events_with_errors = set()
     has_extract_errors = False
+    
     extractibles = extract_archives(
-        abs_location, recurse=not shallow, replace_originals=replace_originals)
+        abs_location, recurse=not shallow, replace_originals=replace_originals, ignored_extensions=ignore)
```
majurg, line 165: Same comment as above. We probably want to keep variable naming consistent. why have the variable `name `ignored_extensions` and not just `ignore`?

tests/extractcode/test_extract.py
```diff
@@ -1112,9 +1112,24 @@ def test_extract_zipslip_tar_posix(self):
 
         warns = [r.warnings for r in result if r.warnings]
         assert [] == warns
-
+
```
majurg, line 1115: Is this white space? if so, we should probably remove this.

tests/extractcode/test_extract.py
```diff
@@ -1112,9 +1112,24 @@ def test_extract_zipslip_tar_posix(self):
...
         assert [] == warns
-
+        
     def test_extract_always_returns_a_generator_and_not_a_list(self):
         # a test for #1996 to ensure that progress is displayed "progressively"
         test_dir = self.get_test_loc('extract/generator', copy=True)
         result = extract.extract(test_dir)
         assert isinstance(result, GeneratorType)
+    
+    def test_extract_ignore(self):
```
majurg, line 1122: We probably want to have more test cases here. We want to make sure `--ignore` can ignore:
1) full filenames
2) directory names
3) various patterns

We should at least have tests cases for all of the above. I would review the current set of extractcode tests and model them after that.

majurg:
@JRavi2 Ping!

JRavi2:
Sorry for the silence! 
This option would also work for ignoring particular files/directories as we are just providing it a glob pattern, so I should change the variable names to a more appropriate one. But, I can't change it to just `ignore` as it would conflict with another ignore being used [here](https://github.com/nexB/scancode-toolkit/blob/develop/src/extractcode/extract.py#L156). Does `ignore_pattern` sound okay? Or, is there another solution to this?

majurg:
@JRavi2 is that particular `ignore` used for anything?

@pombredanne Maybe you have some insight here.

JRavi2:
@majurg I believe this ignore is used to ignore certain default files while walking through the directory. So I just realized that, what we could also do, is just append to the list that it ignores, with the list that the user provides. 
But, the question for the name still remains.

agschrei:
So as announced in #15 I have found the time to run a real-world test on this PR using the [docker-ce git repository](https://github.com/docker/docker-ce).

The following example command is an attempt to cover all three types of arguments the `--ignore` flag should be able to handle

> We want to make sure `--ignore` can ignore:
> 
> 1. full filenames
> 2. directory names
> 3. various patterns

```bash
extractcode --verbose --ignore="*sparse*.tar" --ignore="*issue*.tar" --ignore="/docker-ce/components/engine/vendor/archive/tar/testdata/star.tar"  --ignore="**tarsum/testdata/**" <location of docker-ce clone on disk>
```

Running it behaves exactly as expected and excludes:  

- [x] the absolute path `/docker-ce/components/engine/vendor/archive/tar/testdata/star.tar`  
- [x] any tar archives with `sparse` in their name  
- [x] any tar archives with 'issue' in their name (this exclusion is actually necessary because these archives break libarchive on extraction)  
- [x] any file within the tarsum/testdata subdirectory  

@JRavi2 thank you for your efforts, this contribution is going to be extremely useful for anybody regularly scanning debian package sources.

majurg:
@agschrei Thanks for the feedback! This will be added soon then :)

JRavi2:
@agschrei Glad you found it useful :)
@majurg I'll add the tests in today as soon as possible. Though, I'm still not sure what to rename the variables to?

JRavi2:
@majurg I've added in most of the asked changes. Please review them. :)
The only thing left, if the changes are approved, would be the renaming of the parameters.

majurg:
@JRavi2 I think `ignore_pattern` works as a name, since this is what it actually is.

PR #1948, opened 2020-03-02
Add depth option to the scancode CLI

### LINKED ISSUE #1945: Add --depth option to the CLI
## Short Description
A --depth option for the CLI to limit the nesting level of the scan to a specified number of sub-directories .

## Possible Labels
- new feature
- nice to have

### Category: Scan Feature

The feature can be helpful when you want to scan only a particular number of sub-directories. For example, if you just want to scan the files in the current directory, setting depth to 1 would be more efficient than normal scan.

I have implemented this feature (partially) here: fe14d6ea5fda83e7d65ac8cd38c3d1ca3739a3a2
If you think it is a good idea, I will submit a PR after completing it and writing some tests.

Edit: Some of the CI checks are failing, mainly due to one error. I will get to it in a while.

Thank You.

hanif-ali:
Fixes #1945 

### Tasks


src/scancode/resource.py
```diff
@@ -434,19 +436,45 @@ def create_resources(_seq, _top, _parent, _is_file):
...
+            """Walk through the root_directory recursively upto max_depth nested subdirectories
+            and build the Resource tree. max_depth should be positive or -1 for walking through
+            maximum available nesting levels."""
+
+            # If depth is limited 
+            if max_depth >= 0: 
+                # Note down the directory separator and the depth of root directory
+                # This helps us calculate the depth of nested directories to eliminate
+                # the ones we do not want
+                dir_sep = os.path.sep
```
majurg: Why do wee need this `dir_sep` variable when we can refer to `os.path.sep`?

hanif-ali: Alright. Will get to it, hopefully by tomorrow. Thanks.

hanif-ali:
The tests are all passing locally but they are failing on the CI build. I replicated the environment locally but it still passes finely.

pombredanne:
@hanif-ali thank you for all this!
You wrote:
> The tests are all passing locally but they are failing on the CI build. I replicated the environment locally but it still passes finely.

Actually it could be that some the tests are passing on Python 2 and failing on Python 3?
Did you try on both versions locally?

hanif-ali:
@pombredanne yes. Actually the problem was I had an empty directory in the test resources, so it was not showing up in git. So I had different test results locally and on the Travis build.

src/scancode/resource.py
```diff
@@ -1392,7 +1420,8 @@ def __init__(self, location,
                  codebase_attributes=None,
                  full_root=False, strip_root=False,
                  temp_dir=temp_dir,
-                 max_in_memory=10000):
+                 max_in_memory=10000,
+                 depth=-1):
```
pombredanne: What about using `**kwargs` instead and not doing anything with it below?

hanif-ali: Yeah, that would be neater I think.

src/scancode/cli.py
```diff
@@ -406,10 +406,16 @@ def print_options(ctx, param, value):
     hidden=True,
     help_group=MISC_GROUP, sort_order=1000, cls=CommandLineOption)
 
+@click.option('--depth',
+    type=int, default=-1,
+    help='Scan upto depth level of nesting. Negative for maximum depth.'
```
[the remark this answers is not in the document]
hanif-ali: Thanks for this. I think the git-grep wording is good because all of the other switches are using the dash-separated format. I will make the changes.

hanif-ali:
@pombredanne I have made the changes you have mentioned. Thank you.

src/scancode/resource.py
```diff
@@ -434,19 +436,40 @@ def create_resources(_seq, _top, _parent, _is_file):
         # not keep parents already walked and we walk topdown.
         parent_by_loc = {root.location: root}
 
+
+        def _walk():
```
pombredanne: Can you articulate the benefit of having an inner closed function in this case?

hanif-ali: I created it with a depth parameter initially but kept it as an inner function after removing that parameter. I think it just makes it easier to think of this block of code that traverses through the directory. What do you think?

pombredanne: > I think it just makes it easier to think of this block of code that traverses through the directory.

That's a valid point. There is still something that tickles me a bit somehow... the fact that this is closed on available variables and does not receive arguments makes it look a bit like magic and does not promote readability IMHO.... It feels a bit like some magic happened because of the closures. I personally try to avoid the implicit nature of closures when possible and only use them when they are actually needed (typically when that inner function is a higher order function that returns function(s).  

Based on all that, I am not sure this help with readability and clarity in this case and I would be more comfy with either:
1. do not use a function
2. OR use a function with explicit arguments and not closures (such as here https://github.com/nexB/scancode-toolkit/pull/1948/files#diff-a8e64273dc1cb3ca2a1ff8127a989c6cR401 )

Now for the sake of testability in unit tests, it could make sense to have that as a top level function instead... It may be the sanest and best way. And then you can test this a lot!

src/scancode/resource.py
```diff
@@ -434,19 +436,40 @@ def create_resources(_seq, _top, _parent, _is_file):
...
+            root_dir_depth = root.location.count(os.path.sep)
+
+            # Default behaviour. current_depth is always less than max_depth so that 
+            # os_walk keeps running. When max_depth is limited, the current_depth is
+            # changed in each iteration and compared with max_depth
+            current_depth = -2
+
+            for top, dirs, files in os_walk(root.location, topdown=True, onerror=err):
+                # If depth is limited
+                if self.max_depth >= 0:
```
pombredanne: `self.max_depth` is accessed twice in the loop may be it would make sense to define `max_depth = self.max_depth` outside of the loop? and then use it?

src/scancode/resource.py
```diff
@@ -434,19 +436,40 @@ def create_resources(_seq, _top, _parent, _is_file):
...
+                    current_depth = top.count(os.path.sep) - root_dir_depth
```
pombredanne: Are you sure that `top.count(os.path.sep)` will always be correct?  What if somehow `top = /some//foo///bar`?

src/scancode/resource.py
```diff
@@ -434,19 +436,40 @@ def create_resources(_seq, _top, _parent, _is_file):
...
+            # Find root directory depth using path separator's count
+            root_dir_depth = root.location.count(os.path.sep)
+
+            # Default behaviour. current_depth is always less than max_depth so that 
+            # os_walk keeps running. When max_depth is limited, the current_depth is
+            # changed in each iteration and compared with max_depth
+            current_depth = -2
+
+            for top, dirs, files in os_walk(root.location, topdown=True, onerror=err):
+                # If depth is limited
```
pombredanne: This comment is not needed as the test below is quite obvious IMHO

src/scancode/resource.py
pombredanne: What happens if `self.max_depth == 0` ?

src/scancode/resource.py
```diff
@@ -1392,7 +1415,8 @@ def __init__(self, location,
                  codebase_attributes=None,
                  full_root=False, strip_root=False,
                  temp_dir=temp_dir,
-                 max_in_memory=10000):
+                 max_in_memory=10000,
+                 *args, **kwargs):
```
pombredanne: Just to be clear here: is your point that a virtualcodebase (e.g. --from-json) would never be subject to any max-depth filtering since it does not handle it at all? It may be OK, but it would need to be documented in the code and the CLI IMHO.

AyanSinhaMahapatra:
Yes! If the documentation is updated simultaneously with PRs making changes, that's great! In this case, as this scan option is in the  `CORE_GROUP`, you'd have to add a basic short description [here](https://github.com/nexB/scancode-toolkit/blob/develop/docs/source/rst_snippets/core_options.rst) and in a bit detail [here](https://github.com/nexB/scancode-toolkit/blob/develop/docs/source/cli-reference/core-options.rst). Here's a [RST Cheatsheet](https://github.com/ralsina/rst-cheatsheet/blob/master/rst-cheatsheet.rst) for you, keep in mind the [tests](https://scancode-toolkit.readthedocs.io/en/latest/contribute/contrib_doc.html#continuous-integration) and don't hesitate to ask for help!

src/scancode/cli.py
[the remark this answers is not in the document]
AyanSinhaMahapatra: And this is shown when the `--help` option is used, in the CLI. After this change, Perfect!

src/scancode/resource.py
hanif-ali: That would only select the specified directory/file.Not of much use but it is the minimum depth you can specify.

src/scancode/resource.py
hanif-ali: Yeah, that is what I thought. In case of VirtualCodebase, the directories are already scanned with a particular depth so using --from-json would go to the same original depth. Or would it make more sense to specify a different depth while scanning from JSON?

hanif-ali: Or maybe we can notify when --max-depth is used with --from-json. A warning message or something.

src/scancode/resource.py
hanif-ali: Yeah. Thanks.

src/scancode/resource.py
hanif-ali: I think that would not be a problem because `top` comes from `os.walk` which always uses the OS's path separator (`os.path.sep`). What would be the alternative implementation, in your opinion?

src/scancode/resource.py
hanif-ali: Yes. Thanks.

src/scancode/resource.py
hanif-ali: Factoring the walk function out seems to be the best choice here. We could have a top-level `depth_walk` function and that would also allow us to reuse it in other places.

hanif-ali:
Thank you @pombredanne and @AyanSinhaMahapatra for the help. I will get to these tonight.

hanif-ali:
@pombredanne Sorry for the delay on this one. I have acted upon the changes that you requested. 
As for the inner `_walk` function that you mentioned in the above comment, I have refactored it into a top level generator function  (`depth_walk`) with explicit arguments along with some unit tests. This makes it more testable and reusable, as you mentioned. The client code can interleave with the generator at each step and perform the desired actions (our `create_resource` calls, for instance). 
I also made the changes in the documentation as guided by @AyanSinhaMahapatra and edited the docstrings of `Codebase` and `VirtualCodebase` classes to clarify the added option.

docs/source/cli-reference/core-options.rst
```diff
@@ -119,3 +119,25 @@ Comparing Progress Message Options
...
+    Normally, the scan takes place upto the maximum level of nesting of directories possible. But
+    using the ``--max-depth`` option, the depth of the scan can be limited. This can reduce the time
+    taken for the scan when deeper directories are not relevant.
+
+    Note that the ``--max-depth`` option will be ignored if you are scanning from a JSON file using
+    the ``--from-json`` option. In that case, the original depth is used.
+
+    Some INTEGER values of the ``--max-depth INTEGER`` option:
+
+    - **4**     - Scan only upto 4 levels of subdirectories
```
pombredanne, line 137: Say I scan ~/foo/  would this mean ~/foo/one/two/three/four/file5 would be scanned? or that ~/foo/one/two/three/file4 would be scanned?

docs/source/cli-reference/core-options.rst
```diff
@@ -119,3 +119,25 @@ Comparing Progress Message Options
...
+    - **-1**    - Default, Scan upto maximum levels of subdirectories (no depth limit)
+    - **0**     - No files are scanned at all
```
pombredanne, line 139: What about making 0 the default instead?  I am not sure we would ever want "No files are scanned at all"? and it feels better than -1 may be?

src/scancode/resource.py
```diff
@@ -395,8 +432,8 @@ def _populate(self):
         def err(_error):
             """os.walk error handler"""
             self.errors.append(
-                'ERROR: cannot populate codebase: {}\n'.format(_error)
-                + traceback.format_exc())
+                ('ERROR: cannot populate codebase: %(_error)r\n' % _error)
```
pombredanne, line 435: Was there a problem with using format here? Did you had an issue from mot having a repr?

docs/source/cli-reference/core-options.rst
hanif-ali: Yeah, I think that makes much more sense. We can have positive integers for fixed depth while negative and zero for no limit and set 0 as the default.

src/scancode/resource.py
hanif-ali: My bad. That actually resulted from a mistake while solving a merge conflict. I have reverted it back to the `format` version.

docs/source/cli-reference/core-options.rst
hanif-ali: With a depth of 4, that would scan upto `file4`. The contents of `foo` will be treated as having a depth of 1. Thus the contents of `three` are having a depth of 4. We would not recurse any deeper than that.

hanif-ali:
@pombredanne Thanks for feedback :) I have made the changes as I have mentioned in the above comments.

docs/source/cli-reference/core-options.rst
pombredanne: Thank you fr the explanantion

hanif-ali: Yeah, I see. Would something like this be appropriate?
*Descend at most INTEGER levels of directories below the starting-point.  A zero or negative value indicates no limit.*
This is similar wording to what's on the manpage of `find` utility.

pombredanne: What about this: `Descend at most INTEGER levels of directories below and including the starting-point. Use zero for no limit.` ?

docs/source/cli-reference/core-options.rst
```diff
@@ -119,3 +119,28 @@ Comparing Progress Message Options
...
+    taken for the scan when deeper directories are not relevant.
+
+    Note that the ``--max-depth`` option will be ignored if you are scanning from a JSON file using
+    the ``--from-json`` option. In that case, the original depth is used.
+
+    Some INTEGER values of the ``--max-depth INTEGER`` option:
+
+    - **4**     - Scan only upto 4 levels of subdirectories
+    - **0**     - Default, Scan upto maximum levels of subdirectories (no depth limit)
+    - **-1**    - Same effect as when set to 0
```
pombredanne, line 139: Let's not keep option that have no effect around.

src/scancode/cli.py
```diff
@@ -406,10 +406,16 @@ def print_options(ctx, param, value):
     hidden=True,
     help_group=MISC_GROUP, sort_order=1000, cls=CommandLineOption)
 
+@click.option('--max-depth',
+    type=int, default=0, show_default=False, metavar='N',
+    help='For each <input> given on the command line, scan at most N levels '
+        'of directories. A negative value or 0 means no limit. 0 is the default',
+    help_group=CORE_GROUP, sort_order=10, cls=CommandLineOption)
+
```
pombredanne, line 414: What about this?

```Python
@click.option('--max-depth',
    type=int, default=0, show_default=False, 
    help='Maximum nesting depth of subdirectories to scan. '
        'Descend at most INTEGER levels of directories below and including '
        'the starting directory. Use 0 for no scan depth limit.',
    help_group=CORE_GROUP, sort_order=10, cls=CommandLineOption)


```
Also can you move this up right after the `--max-in-memory` option, since it is part of the core group?

src/scancode/cli.py
```diff
@@ -868,7 +875,8 @@ def echo_func(*_args, **_kwargs):
                 codebase_attributes=codebase_attributes,
                 full_root=full_root,
                 strip_root=strip_root,
-                max_in_memory=max_in_memory
+                max_in_memory=max_in_memory,
+                max_depth=max_depth
```
pombredanne, line 879: Can you add a trailing comma?

src/scancode/resource.py
```diff
@@ -120,6 +120,39 @@ class ResourceNotInCache(Exception):
 class UnknownResource(Exception):
     pass
 
+def depth_walk(root_location, max_depth, skip_ignored=lambda x: False, error_handler=lambda:None):
+    """ Generator that walks the `root_location` directory recursively upto `max_depth` of
```
pombredanne, line 124: To be consistent with other docstring, please wrap the first line as in:
```
    """
    Generator that walks the `root_location` directory recursively upto `max_depth` of ...
    ...
    """
```

src/scancode/resource.py
```diff
@@ -120,6 +120,39 @@ class ResourceNotInCache(Exception):
...
+
+    # Default behaviour. current_depth is always less than max_depth so that 
+    # os_walk keeps running. When max_depth is limited, the current_depth is
+    # changed in each iteration and compared with max_depth
+    current_depth = -2
+
+    for top, dirs, files in os_walk(root_location, topdown=True, onerror=error_handler):
+        # If depth is limited
+        if max_depth > 0:
+            current_depth = top.count(os.path.sep) - root_dir_depth
```
pombredanne, line 146: Are you sure `count(os.path.sep)` will work OK? what if the path contains a segment with multiple contiguous separators?
for instance what happens with this:
```
>>> for x in os.walk('samples///'):
...  print(x)
... 
('samples///', ['arch', 'JGroups', 'zlib'], ['screenshot.png', 'README'])
('samples///arch', [], ['zlib.tar.gz'])
```

src/scancode/resource.py
```diff
@@ -120,6 +120,39 @@ class ResourceNotInCache(Exception):
...
     pass
 
+def depth_walk(root_location, max_depth, skip_ignored=lambda x: False, error_handler=lambda:None):
+    """ Generator that walks the `root_location` directory recursively upto `max_depth` of 
+    nested subdirectories and yields a (top, dirs, files) tuple at each step. 
+
+    Arguments: 
+       - root_location: Absolute path for the directory to be walked
+       - max_depth: Positive integer for fixed depth limit. negative for no limit
+       - skip_ignored: Function that returns boolean indicating whether to ignore searching files
```
pombredanne, line 130: What about instead moving the err and is_ignored functions here from https://github.com/nexB/scancode-toolkit/pull/1948/files#diff-a8e64273dc1cb3ca2a1ff8127a989c6cR433 ... that way you do not need these args anymore?

src/scancode/resource.py
```diff
@@ -120,6 +120,39 @@ class ResourceNotInCache(Exception):
...
+    root_dir_depth = root_location.count(os.path.sep)
+
+    # Default behaviour. current_depth is always less than max_depth so that 
+    # os_walk keeps running. When max_depth is limited, the current_depth is
+    # changed in each iteration and compared with max_depth
+    current_depth = -2
+
+    for top, dirs, files in os_walk(root_location, topdown=True, onerror=error_handler):
+        # If depth is limited
+        if max_depth > 0:
```
pombredanne, line 145: `if max_depth` is enough here, since there is no negative anymore

src/scancode/resource.py
```diff
@@ -120,6 +120,39 @@ class ResourceNotInCache(Exception):
...
+    nested subdirectories and yields a (top, dirs, files) tuple at each step. 
+
+    Arguments: 
+       - root_location: Absolute path for the directory to be walked
+       - max_depth: Positive integer for fixed depth limit. negative for no limit
+       - skip_ignored: Function that returns boolean indicating whether to ignore searching files
+                      in the current location. No ignoring by default
+       - error_handler: Error handler callback. No action taken by default.
+    """
+
```
pombredanne, line 134: Can you add a bounds check on args that `max_depth >= 0`  here ?

src/scancode/resource.py
```diff
@@ -120,6 +120,39 @@ class ResourceNotInCache(Exception):
...
+    # Find root directory depth using path separator's count
+    root_dir_depth = root_location.count(os.path.sep)
+
+    # Default behaviour. current_depth is always less than max_depth so that 
+    # os_walk keeps running. When max_depth is limited, the current_depth is
+    # changed in each iteration and compared with max_depth
+    current_depth = -2
```
pombredanne, line 141: -2 does not feel right...see below another approach. where the default would be 0.

src/scancode/resource.py
```diff
@@ -120,6 +120,39 @@ class ResourceNotInCache(Exception):
...
+
+    for top, dirs, files in os_walk(root_location, topdown=True, onerror=error_handler):
+        # If depth is limited
+        if max_depth > 0:
+            current_depth = top.count(os.path.sep) - root_dir_depth
+
+        if skip_ignored(top) or current_depth >= max_depth:
```
pombredanne, line 148: instead of using -2, what about that?
`if skip_ignored(top) or (max_depth and current_depth >= max_depth):`

docs/source/cli-reference/core-options.rst
hanif-ali: So line `138` and `139` both need to be removed. Is that right?

src/scancode/cli.py
hanif-ali: Yeah sure. Will do that :)

src/scancode/cli.py
hanif-ali: Okay

src/scancode/resource.py
hanif-ali: Yeah, this is much neater

src/scancode/resource.py
hanif-ali: What should we do if the bound check fails? Should we raise an exception or default to having no limit at all? If we want to return an error, would it be a good idea to have the bound check earlier in `cli.py` and raise a `UsageError` if it fails?

hanif-ali: or do we need to have both?

src/scancode/resource.py
hanif-ali: We can do that for `skip_ignored` but `err` has to be bound to `Codebase` as it updates the `errors` attribute on the object. The idea of having them as option callback arguments was to allow some flexibility in case the functionality is reused somewhere else. What do you think?

src/scancode/resource.py
hanif-ali: The idea here is to normalize the `root_location` path first using `os.path.abspath` or `os.path.normpath`, so that any unnecessary slashes are removed. In our case, it being done in the `__init__` method of `Codebase`. That guarantees that any adjacent slashes are removed from the path.

hanif-ali:
@pombredanne Thanks a lot for the comments. I will get to these asap. I have highlighted a few questions in the inline comments.

src/scancode/resource.py
hanif-ali: `depth_walk` itself does not guarantee that the path is normalized or absolute. Do you think we should have another `abspath` inside here?

src/scancode/resource.py
```diff
@@ -120,6 +120,39 @@ class ResourceNotInCache(Exception):
 class UnknownResource(Exception):
     pass
 
+def depth_walk(root_location, max_depth, skip_ignored=lambda x: False, error_handler=lambda:None):
+    """ Generator that walks the `root_location` directory recursively upto `max_depth` of 
+    nested subdirectories and yields a (top, dirs, files) tuple at each step. 
+
+    Arguments: 
+       - root_location: Absolute path for the directory to be walked
+       - max_depth: Positive integer for fixed depth limit. negative for no limit
```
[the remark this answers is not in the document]
hanif-ali: Yes, sure

src/scancode/resource.py
pombredanne: not needed. Your answer above makes sense :+1:

docs/source/cli-reference/core-options.rst
pombredanne: Just the negative value... since this is not supported this is not needed. Also the ddefault may not need its own line of doc, so yes  138 and 139 can be removed

src/scancode/resource.py
pombredanne: OK, your call... but consider also YAGNI https://en.wikipedia.org/wiki/You_aren't_gonna_need_it ... it is easy to add if really needed, hard to remove otherwise.

src/scancode/resource.py
pombredanne: > What should we do if the bound check fails? Should we raise an exception

Fail loudly with an exception and you can also have a cli check if you like.

hanif-ali: Alright. I just added added these.

docs/source/cli-reference/core-options.rst
hanif-ali: I have removed these and added a usage example in this section.

src/scancode/resource.py
hanif-ali: Yeah, that is right. I have taken the `skip_ignored` function out, so we have one less argument. The same couldn't be done in a neat way for `err`.

hanif-ali: Yeah, you're right. I have taken `skip_ignored` out, so we have one less argument.

src/scancode/resource.py
pombredanne: :+1:

src/scancode/resource.py
pombredanne: What you have now works for me.

Issue #1894, opened 2020-02-10
Consolidating large codebases uses excessive amounts of memory


JonoYang:
I was using the consolidation plugin on a 18.3 mb scan with 14886 files and scancode managed to use 10+ gigs of ram. I have a feeling that it is this dict (https://github.com/nexB/scancode-toolkit/blob/develop/src/summarycode/plugin_consolidate.py#L391) bloating up horribly when there are a lot of copyright holders consolidated.

PR #1953, opened 2020-03-06
Fix memory leak in consolidation plugin

JonoYang:
This PR alleviates the memory bloat described in #1894 caused by an unnecessary dictionary exploding in size when processing a large codebase (https://github.com/nexB/scancode-toolkit/blob/develop/src/summarycode/plugin_consolidate.py#L391) This is done by creating consolidations at the same time we determine the highest common Resource for a given copyright holder.


pombredanne:
All set! Thanks!

PR #1970, opened 2020-03-12
Implement repository_download_url for more Packages

JonoYang:
This PR implements `repository_download_url()` for Cran and pypi packages.


pombredanne:
Thanks... yet there seem to be some tests that are still failing?

src/packagedcode/pypi.py
```diff
@@ -94,9 +94,9 @@ class PythonPackage(models.Package):
     extensions = ('.egg', '.whl', '.pyz', '.pex',)
     default_type = 'pypi'
     default_primary_language = 'Python'
-    default_web_baseurl = None
-    default_download_baseurl = None
-    default_api_baseurl = None
+    default_web_baseurl = 'https://pypi.org'
+    default_download_baseurl = 'https://pypi.io/packages/source'
```
pombredanne, line 98: @JonoYang can you check this https://github.com/pypa/pypi-legacy/issues/438#issuecomment-226940730 that was  suggested by @Kwpolska

src/packagedcode/pypi.py
```diff
@@ -105,6 +105,23 @@ def recognize(cls, location):
...
 
+    def repository_homepage_url(self, baseurl=default_web_baseurl):
+        if not self.name:
+            return
+        return '{}/project/{}'.format(baseurl, self.name)
+
+    def repository_download_url(self, baseurl=default_download_baseurl):
+        if not self.name or not self.version:
+            return
+        return '{baseurl}/{name[0]}/{name}/{name}-{version}.tar.gz'.format(baseurl=baseurl, name=self.name, version=self.version)
```
Kwpolska, line 116: Don’t assume there will always be a tarball available. It’s best to use the API and get the full download URL with the correct extension.

PR #2046, opened 2020-05-26
Add methods that returns where a Resource was extracted to or from

JonoYang:
This PR adds two new methods to the `Resource` class: `extracted_to` and `extracted_from`. These methods return the `Resource` where a given `Resource` was extracted to or extracted from, respectively. These methods only work on `Resource`s that were extracted using `extractcode`.


pombredanne:
See instead  #2215

PR #2047, opened 2020-05-26
Packagedcode to handle .gemspec file

### LINKED ISSUE #2035: Parse Gemspec Files

We need to add the ability to parse `.gemspec` files and have them apart of a `--package` scan.

rpotter12:
Fixes #2035 

Added gemspec parser to handle gemspec files.

### Tasks


rpotter12:
@majurg I will add tests for it in some time. Please review the code :)

src/packagedcode/rubygems.py
```diff
@@ -647,93 +648,133 @@ def normalize(gem_data, known_fields=known_fields):
...
+
+    dependencies = []
+
+    for individual in file_data:
+        # update the value of name
+        if 'name' in individual:
+            if '"' in individual:
+                name = individual.split('"')
+            elif "'" in individual:
+                name = individual.split("'")
```
majurg, line 681: What is this doing, exactly?

rpotter12: In some files string is like `spec.name = "name"` and in some files it is `spec.name = 'name'`.

This if-elif help in recognizing in `"` and `'`

src/packagedcode/rubygems.py
```diff
@@ -647,93 +648,135 @@ def normalize(gem_data, known_fields=known_fields):
...
 
 
-class GemSpec(object):
+def parse_spec(location):
+    """
+    Returns a dictionary which containing .gemspec file data.
+    """
+    file_data = []
+    with io.open(location, "r", encoding="utf-8") as f:
+        file_data = [line.rstrip('\n') for line in f]
```
JonoYang, line 671: Is it necessary to perform the `rstrip` for every line? Would `file_data = f.read()` work just as well?

src/packagedcode/rubygems.py
majurg: But is this really needed?

src/packagedcode/rubygems.py
rpotter12: @JonoYang What I am actually doing is that I am splitting it line by line and store in a list. `file_data = f.read()` will not give the desired output. 

The desired output can be obtained by 2 solutions:
```python
with io.open(file_name, "r", encoding="utf-8") as f:
    file_data = [line.rstrip() for line in f]
```
or
```python
with io.open(file_name, "r", encoding="utf-8") as f:
    file_data = f.read().splitlines()
```

Please tell me which will be the better approach for this??

src/packagedcode/rubygems.py
rpotter12: @majurg In our .json output file, the output is like `"name": "<name>"`. If we will not replace `"` or `'` the output will be like `"name: ""<name>""`
What I am actually doing is extracting the data between the `""` or `''`

I think this can be done in more efficient way:
```python
name = individual.split('= ')
name = name[1].strip('\'"')
data_dic.update({'name' : name})
```

What do you say??

src/packagedcode/rubygems.py
JonoYang: @rpotter12 Thanks for clearing that up. I was mistaken initially.

I think you could iterate through the file line by line doing this:

```
with io.open(file_name, "r", encoding="utf-8") as f:
    for line in f:
        <process line>
```

Though what you have works, I do not see a need for change.

majurg: @rpotter12 Does `.splitlines()` return a list of lines? if so, this is more readable to me IMO.

src/packagedcode/rubygems.py
```diff
@@ -647,93 +648,124 @@ def normalize(gem_data, known_fields=known_fields):
...
+            # >>> s = 'spec.name = "abc"'
+            # >>> name = s.split('= ')
+            # >>> name
+            # ['spec.name ', '"abc"']
+            # >>> name = name[1].strip('\'"')
+            # >>> name
+            # abc
+            name = individual.split('= ')
+            name = name[1].strip('\'"')
+            data_dic.update({'name' : name})
```
majurg, line 687: Do we really need an update call here? why not just `data_dic['name'] = name` 

Also, `data_dic` is not a great variable name,

src/packagedcode/rubygems.py
```diff
@@ -647,93 +648,124 @@ def normalize(gem_data, known_fields=known_fields):
     )
 
 
-def parse_spec(location):
-    pass
+# defaults fields of .gemspec file
+data_dic = {
```
majurg, line 652: Why is this a global? 

Also, this needs a better name like `gemspec_data` or similar

src/packagedcode/rubygems.py
rpotter12: @majurg You are right. We should use `data_dic['name] = name to update the value. This method is more faster than the previous one. 
References: https://stackoverflow.com/a/15456218/9380990

rpotter12:
@majurg @pombredanne Please review the PR again

rpotter12:
@majurg I have added tests. Please review :)

src/packagedcode/rubygems.py
```diff
@@ -648,92 +650,127 @@ def normalize(gem_data, known_fields=known_fields):
...
     """
+    name = gemspec_data.get('name')
+    description = gemspec_data.get('description')
+    homepage_url = gemspec_data.get('homepage')
+    package = RubyGem(
+        name=name,
+        description=description,
+        homepage_url=homepage_url
+    )
+    package.dependencies = gemspec_data.get('dependencies')
```
majurg, line 757: You will need to do a little more here to handle the dependencies here. See: https://github.com/nexB/scancode-toolkit/blob/72bef69a3c485380be164a58432dd8218ce4df84/src/packagedcode/models.py#L458

You have to create DependentPackage objects (similar to how you handle parties)

src/packagedcode/rubygems.py
```diff
@@ -648,92 +650,127 @@ def normalize(gem_data, known_fields=known_fields):
...
-    Represent a Gem specification.
+    Yield RubyGem Packages from gemspec.
     """
+    name = gemspec_data.get('name')
+    description = gemspec_data.get('description')
+    homepage_url = gemspec_data.get('homepage')
+    package = RubyGem(
+        name=name,
+        description=description,
+        homepage_url=homepage_url
```
majurg, line 755: We need a few more things here:
1) version information? do we have it in the gemspec file?
2) namespace information? do all gems listed in a gemspec file come from rubygems.org?

src/packagedcode/rubygems.py
```diff
@@ -648,92 +650,127 @@ def normalize(gem_data, known_fields=known_fields):
...
+
+    authors = gemspec_data.get('authors') or []
+    email = gemspec_data.get('email') or []
+
+    parties = []
+    if authors:
+        parties.append(
+            models.Party(
+                name=authors,
+                email=email,
```
majurg, line 767: Do the author names always correspond to the email information?

src/packagedcode/rubygems.py
majurg: Yes you are right here, I forget gemspec files are actual ruby code.

src/packagedcode/rubygems.py
rpotter12: I will look at this too :)

src/packagedcode/rubygems.py
rpotter12: In general author names are always correspond to the email information but we know that there are many ways to a particular thing so most of the time author names are correspond to the email :)

src/packagedcode/rubygems.py
rpotter12: > version information? do we have it in the gemspec file?

Yes, it is present in the files. It needs to be implemented :)

rpotter12:
@majurg Please review the PR again. :)

src/packagedcode/rubygems.py
```diff
@@ -32,6 +32,8 @@
 from os.path import abspath
 from os.path import expanduser
 
+import io
+import re
 import attr
 import saneyaml
 from packageurl import PackageURL
```
majurg, line 39: Please sort your imports alphabetically in the following way:
```
import built-in-python-libs

import 3rd-party-python-libs

import scancode-project-imports
```

src/packagedcode/rubygems.py
```diff
@@ -120,7 +122,8 @@ def recognize(cls, location):
 
         if location.endswith('.gemspec'):
             # TODO: implement me
```
majurg, line 124: Remove this TODO, since I believe this is what you are currently doing.

src/packagedcode/rubygems.py
```diff
@@ -255,7 +258,6 @@ def get_gem_metadata(location):
         if warnings:
             raise Exception('Failed to extract RubyGem .gem file.\n' + '\n'.join(warnings))
```
majurg, line 259: I would rather not have newlines in this Exception.

src/packagedcode/rubygems.py
```diff
@@ -648,92 +629,172 @@ def normalize(gem_data, known_fields=known_fields):
 
 
 def parse_spec(location):
-    pass
+    """
+    Returns a dictionary which containing .gemspec file data.
```
majurg, line 633: remove the word `which` here.

src/packagedcode/rubygems.py
```diff
@@ -648,92 +629,172 @@ def normalize(gem_data, known_fields=known_fields):
...
+            except IndexError:
+                dep_version = None
+
+            # >>> s = 's.add_runtime_dependency(%q<mechanize>, [">= 0"])'
+            # >>> d = re.split("'|<|>", s)
+            # >>> d
+            # ['s.add_runtime_dependency(%q', 'mechanize', ', ["', '= 0"])']
+            dependency = re.split("'|\"|<|>", individual)
+            dependencies[dependency[1]] = dep_version
+            continue
```
majurg, line 742: Why do we continue after each if statement? Doesnt this make it impossible to grab all the fields?

src/packagedcode/rubygems.py
```diff
@@ -648,92 +629,172 @@ def normalize(gem_data, known_fields=known_fields):
...
+                    purl=PackageURL(
+                        type='gem', name=dep_name).to_string(),
+                    scope='dependencies',
+                    is_runtime=True,
+                    is_optional=False,
+                    requirement=dep_version,
+                )
+            )
+    package.dependencies=package_dependencies
+    return package
```
majurg, line 797: You are returning the package here instead of Yielding (like your docstring states). Either convert this to a generator or modify your function's docstring.

src/packagedcode/rubygems.py
rpotter12: If we will not use continue the script will check all if statements for every element. This will increase the execution time. The script is grabbing data on a particular keyword which is starting from `.` like gemspec file contain string like `data = 's.name = "abc"'`, so we are searching by particular keyword `if '.name' in data:`. `continue` is used to reduce the execution time.

src/packagedcode/rubygems.py
rpotter12: Done :)

src/packagedcode/rubygems.py
rpotter12: Done :)

src/packagedcode/rubygems.py
rpotter12: Yes. It is corrected now :)

src/packagedcode/rubygems.py
rpotter12: Done :)

rpotter12:
@majurg All changes are done. Please review the PR again :)

src/packagedcode/rubygems.py
```diff
@@ -648,92 +628,172 @@ def normalize(gem_data, known_fields=known_fields):
...
+            # >>> name = s.split('= ')
+            # >>> name
+            # ['spec.name ', '"abc"']
+            # >>> name = name[1].strip('\'"%q{}')
+            # >>> name
+            # abc
+            name = individual.split('= ')
+            name = name[1].strip('\'"%q{}')
+            gemspec_data['name'] = name
+            continue
```
majurg, line 666: What are the continues for here? I forgot if I asked this in the past, but to me they seem a bit ugly. Is there a better way to do this?

src/packagedcode/rubygems.py
```diff
@@ -648,92 +628,172 @@ def normalize(gem_data, known_fields=known_fields):
...
+            # >>> d = re.split("'|<|>", s)
+            # >>> d
+            # ['s.add_runtime_dependency(%q', 'mechanize', ', ["', '= 0"])']
+            dependency = re.split("'|\"|<|>", individual)
+            dependencies[dependency[1]] = dep_version
+            continue
+
+    # stores dependencies list into dictionary
+    if not dependencies:
+        dependencies = None
```
majurg, line 745: This is redundant.

src/packagedcode/rubygems.py
```diff
@@ -648,92 +628,172 @@ def normalize(gem_data, known_fields=known_fields):
...
 def parse_spec(location):
-    pass
+    """
+    Returns a dictionary containing .gemspec file data.
+    """
+    file_data = []
+    with io.open(location, "r", encoding="utf-8") as f:
+        file_data = f.read().splitlines()
+
+    dependencies = OrderedDict()
```
majurg, line 638: Why do we define this way up here? If we want to define an empty dict, we should do it closer to where its actually used in code.

src/packagedcode/rubygems.py
```diff
@@ -648,92 +628,172 @@ def normalize(gem_data, known_fields=known_fields):
...
     """
+    name = gemspec_data.get('name')
+    version = gemspec_data.get('version')
+    description = gemspec_data.get('description')
+    homepage_url = gemspec_data.get('homepage')
+    package = RubyGem(
+        name=name,
+        version=version,
+        description=description,
+        homepage_url=homepage_url
```
majurg, line 763: Its slightly confusing to create a minimal package at first and keep adding to it. Why not create a mapping dict and just call `package = RubyGem(**mapped_data)` at the end?

src/packagedcode/rubygems.py
```diff
@@ -648,92 +628,172 @@ def normalize(gem_data, known_fields=known_fields):
...
+    )
+    # package.dependencies = gemspec_data.get('dependencies')
```
majurg, line 765: This should be removed (you reference this later in the code).

src/packagedcode/rubygems.py
rpotter12: I have used this dictionary in for loop. In python2 and python3  normal dictionary was stores data in different ways so I used this here to store in a particular order only.

src/packagedcode/rubygems.py
rpotter12: I don't the better way but I am removing continue in this :)

src/packagedcode/rubygems.py
rpotter12: @majurg I tried `package = RubyGem(**mapped_data)` but not able to implement it. 
I saw https://github.com/nexB/scancode-toolkit/pull/2062/files implementation, in this you have implemented by for loop. 
I think whatever is implemented in this is good and is similar to other packaged code also. :)

rpotter12:
@majurg Please review this again. :)

src/packagedcode/rubygems.py
```diff
@@ -648,92 +628,163 @@ def normalize(gem_data, known_fields=known_fields):
...
+        "description": None,
+        "homepage": None,
+        "license": None,
+        "dependencies": None
+        }
 
-class GemSpec(object):
+    dependencies = OrderedDict()
+
+    for individual in file_data:
```
pombredanne, line 653: Why use this "individual" name? What does the variable hold?

src/packagedcode/rubygems.py
```diff
@@ -648,92 +628,163 @@ def normalize(gem_data, known_fields=known_fields):
 
 
 def parse_spec(location):
-    pass
-
+    """
+    Returns a dictionary containing .gemspec file data.
+    """
+    file_data = []
```
pombredanne, line 634: Why declare this here? is there ever a case where if will not exist below?

src/packagedcode/rubygems.py
```diff
@@ -648,92 +628,163 @@ def normalize(gem_data, known_fields=known_fields):
...
+    with io.open(location, "r", encoding="utf-8") as f:
+        file_data = f.read().splitlines()
```
pombredanne, line 636: Is there a better name than `file_data` ?

src/packagedcode/rubygems.py
```diff
@@ -648,92 +628,163 @@ def normalize(gem_data, known_fields=known_fields):
...
+        "dependencies": None
+        }
 
-class GemSpec(object):
+    dependencies = OrderedDict()
+
+    for individual in file_data:
+        # update the value of name
+        if '.name ' in individual and 'basename' not in individual:
+            # >>> s = 'spec.name = "abc"'
```
pombredanne, line 656: What is the purpose of this comment? this just repast the code below.

src/packagedcode/rubygems.py
```diff
@@ -648,92 +628,163 @@ def normalize(gem_data, known_fields=known_fields):
...
+            # >>> name = s.split('= ')
+            # >>> name
+            # ['spec.name ', '"abc"']
+            # >>> name = name[1].strip('\'"%q{}')
+            # >>> name
+            # abc
+            name = individual.split('= ')
```
pombredanne, line 663: What if there is no = in the line? or if there is no space after the equal?
You should also consider using `.partition()` rather than split() as it has better guarantees on what is being returned

src/packagedcode/rubygems.py
```diff
@@ -648,92 +628,163 @@ def normalize(gem_data, known_fields=known_fields):
...
+            name = name[1].strip('\'"%q{}')
```
pombredanne, line 664: What if there is no `=`?
```
>>> 'foo'.split('= ')[1]
  File "<stdin>", line 1, in <module>
IndexError: list index out of range
```
What if the name starts with `q`?
```
>>> 'quiet'.strip('\'"%q{}')
'uiet'
```
Your should use a partition() or a .replace() instead for safety.

src/packagedcode/rubygems.py
```diff
@@ -648,92 +628,163 @@ def normalize(gem_data, known_fields=known_fields):
...
+            gemspec_data['name'] = name
+
+        # update the value of version
+        if '.version ' in individual:
```
pombredanne, line 668: What is there is no trailing space?

src/packagedcode/rubygems.py
```diff
@@ -648,92 +628,163 @@ def normalize(gem_data, known_fields=known_fields):
...
+            version = individual.split('= ')
+            version = version[1].strip('\'"%q{}')
```
pombredanne, line 670: Same comments as above. Furthermore, if this is the same code as above, make that a function.
Also that means you should rather to the parsing of a line once, and only then test for the attribute name

src/packagedcode/rubygems.py
```diff
@@ -648,92 +628,163 @@ def normalize(gem_data, known_fields=known_fields):
...
+            gemspec_data['version'] = version
+
+        # update the value of author
+        if '.authors ' in individual:
+            # >>> s = 's.authors = ["abc", "pqr", "xyz"]'
+            # >>> result = s.split('= ')
+            # ['s.authors ', '["abc", "pqr", "xyz"]']
+            # >>> result[1].strip('[]')
+            # '"abc", "pqr", "xyz"'
```
pombredanne, line 679: Same as above.

src/packagedcode/rubygems.py
```diff
@@ -648,92 +628,163 @@ def normalize(gem_data, known_fields=known_fields):
...
+    # stores dependencies list into dictionary
+    gemspec_data['dependencies'] = dependencies
+
+    return gemspec_data
+
+
+def build_packages_from_gemspec(gemspec_data):
     """
-    Represent a Gem specification.
+    Return a package built from gemspec.
```
pombredanne, line 743: What is `gemspec` ? instead say:
`Return a RubyGem package built from a `gemspec_data` mapping.`

src/packagedcode/rubygems.py
```diff
@@ -648,92 +628,163 @@ def normalize(gem_data, known_fields=known_fields):
...
-                    logger.warning('WARNING: {}: no license mapping for: "{}"'.format(self.filename, lic))
-        return mapped_licenses
+    dependencies = gemspec_data.get('dependencies')
+    package_dependencies = []
+    if dependencies:
+        for dep_name, dep_version in dependencies.items():
+            package_dependencies.append(
+                models.DependentPackage(
+                    purl=PackageURL(
+                        type='gem', name=dep_name).to_string(),
```
pombredanne, line 770: Is there a case where the version for this dep is "pinned" and set and is not an approximate range?

src/packagedcode/rubygems.py
```diff
@@ -648,92 +628,163 @@ def normalize(gem_data, known_fields=known_fields):
...
+                    scope='dependencies',
+                    is_runtime=True,
+                    is_optional=False,
+                    requirement=dep_version,
+                )
+            )
+    #package.dependencies=package_dependencies
```
pombredanne, line 777: Why this comment?

src/packagedcode/rubygems.py
```diff
@@ -648,92 +628,163 @@ def normalize(gem_data, known_fields=known_fields):
...
+
+    package = RubyGem(
+        name=name,
+        version=version,
+        description=description,
+        homepage_url=homepage_url,
+        parties = parties,
```
pombredanne, line 784: nit: no spaces around the = sign

PR #2060, opened 2020-06-04
Change expected data on macOS

Abhishek-Dev09:
Fix ar test with 7 zip on macOS
This PR is the extension of #1835 .


Abhishek-Dev09:
@pombredanne : All fixed on macOS except 1 test on windows.

pombredanne:
@Abhishek-Dev09 Thank you for that... you have made the tests pass but you have not fixed the bug... te bug is that some files are not correctly being extracted on macOS... If I were to merge this, this would hide the bug entirely.

PR #2075, opened 2020-06-18
Packagedcode to handle Ruby(.gemspec) and Cocoapods(.podspec)

### LINKED ISSUE #2035: Parse Gemspec Files

rpotter12:
Fixes #2035 
Packedcode to handle Ruby(.gemspec) and Cocoapods(.podspec).

1. Created `spec.py` file which  will extract all the necessary data.
2. Dependencies are extracted by using `gemfileparser`.
In this way we don't need to edit any library code.

### Tasks


src/packagedcode/gemspec.py
```diff
@@ -0,0 +1,265 @@
...
+from collections import OrderedDict
+import io
+import re
+import os
+import glob
+
+
+#!/usr/bin/env python
+#
+# Copyright 2015 Balasankar C <balasankarc@autistici.org>
```
pombredanne, line 36: Is this really GPL? if so we cannot use it in ScanCode as this not compatible with the Apache license. 
AFAIK @balasankarc kindly relicensed it to a choice of MIT or GPL on my request  https://github.com/balasankarc/gemfileparser/issues/4
So we would likely pick the MIT choice instead. @balasankarc can you make this a tad more explicit in the code comment header?
Also I wonder if we could instead contribute and further @balasankarc library and use it as a dep rather than forking it here.

@balasankarc what's your plan for https://github.com/balasankarc/gemfileparser long term? We could take over its maintenance if you like?

tests/packagedcode/test_rubygems.py
```diff
@@ -47,66 +47,39 @@
...
-            expected = json.load(ex)
-
-        assert sorted(expected.items()) == sorted(results.items())
-
     def test_rubygems_can_parse_gemspec_address_standardization_gemspec(self):
-        self.check_gemspec(
-            'rubygems/gemspec/address_standardization.gemspec',
-            'rubygems/gemspec/address_standardization.gemspec.expected.json')
-
-    def test_rubygems_can_parse_gemspec_arel_gemspec(self):
```
pombredanne, line 86: Why removing that test?

tests/packagedcode/test_rubygems.py
```diff
@@ -47,66 +47,39 @@
...
-        self.check_gemspec(
-            'rubygems/gemspec/arel.gemspec',
-            'rubygems/gemspec/arel.gemspec.expected.json')
+        test_file = self.get_test_loc('rubygems/gemspec/address_standardization.gemspec')
+        expected_loc = self.get_test_loc('rubygems/gemspec/address_standardization.gemspec.expected.json')
+        packages = rubygems.RubyGem.recognize(test_file)
+        self.check_packages(packages, expected_loc, regen=False)
 
-    def test_rubygems_modern_gemspec(self):
```
pombredanne, line 91: Why removing that test too a few other below?

rpotter12: These tests are added again :)

tests/packagedcode/test_rubygems.py
rpotter12: These tests are added again :)

src/packagedcode/gemspec.py
majurg: I agree with @pombredanne.

majurg: @rpotter12 We want to use `gemfileparser` as a library instead of forking its source code here.

rpotter12: @majurg we can't use https://github.com/balasankarc/gemfileparser library directly as it can only extract dependencies info only. So we have edited the existing library. I think @pombredanne wants to say that we can fork https://github.com/balasankarc/gemfileparser in @nexB and add our edited code and release a new library version so that we can extract all the data and use it in our library.

rpotter12: @pombredanne https://github.com/balasankarc/gemfileparser is in dual license - GPLv3 and MIT.

rpotter12:
@majurg https://github.com/nexB/scancode-toolkit/pull/2075#discussion_r447317279 Now nothing is edited from `gemfileparser`. I have made a different file which will handle gemspec and podspec all the data except dependency and dependencies will be handled by `gemfileparser` library. So there is no edited code in this PR :)

majurg:
@rpotter12 Yeah I see that you have done that change already. Not sure why GH has updated that comment... Anyway, it looks good.

src/packagedcode/rubygems.py
```diff
@@ -647,93 +626,67 @@ def normalize(gem_data, known_fields=known_fields):
...
+    package = RubyGem(
+        name=name,
+        version=version,
+        parties=parties,
+        homepage_url=homepage_url,
+        description=description,
+        declared_license=declared_license
+    )
+
+    dependencies = gemspec_data.get('dependencies')
```
pombredanne, line 670: As suggested for go.mod, consider this instead here and elsewhere:
`dependencies = gemspec_data.get('dependencies', {}) or {}` and remove `if dependencies:`

src/packagedcode/spec.py
```diff
@@ -0,0 +1,156 @@
...
+                for dependency in deps[key]:
+                    dependencies[dependency.name] = (dependency.requirement).split(',')
+        spec_data['dependencies'] = dependencies
+
+        return spec_data
+
+
+def get_stripped_data(line):
+    """
+    Return data after removing unnecessary special character and space.
```
pombredanne, line 116: is this a line? or data? or a string?

src/packagedcode/spec.py
```diff
@@ -0,0 +1,156 @@
...
+                spec_data['email'] = stripped_emails
+            elif '.author' in line:
+                authors = re.sub(r'/*.*author.*?=', '', line)
+                stripped_authors = get_stripped_data(authors)
+                stripped_authors = stripped_authors.replace(' => ', "=>")
+                stripped_authors = stripped_authors.strip()
+                stripped_authors = stripped_authors.split(',')
+                spec_data['author'] = stripped_authors
+            elif '.summary' in line:
+                summary = re.sub(r'/*.*summary.*?=', '', line)
```
pombredanne, line 90: Could a simpler `line.rpartition('=')` be all you need here and elsewhere?

src/packagedcode/spec.py
```diff
@@ -0,0 +1,156 @@
...
+                     ...
+                     42. Likes candies.
+                    DESC
+    Return description from podspec.
+    """
+    with io.open(location, encoding='utf-8', closefd=True) as data:
+        lines = data.readlines()
+    description = ''
+    for i, content in enumerate(lines):
+        if '.description' in content:
```
pombredanne, line 149: What is there is no `<<-DESC` starting on that line?

tests/packagedcode/data/rubygems/gemspec/arel.gemspec.expected.json
```diff
@@ -1,187 +1,62 @@
...
+    "qualifiers": {},
+    "subpath": null,
+    "primary_language": "Ruby",
+    "description": "Arel is a SQL AST manager for Ruby. It",
+    "release_date": null,
+    "parties": [
+      {
+        "type": "person",
+        "role": "author",
+        "name": "Aaron Patterson, Bryan Halmkamp, Emilio Tagua, Nick Kallen",
```
pombredanne, line 16: These were originally lists of values for authors and emails in https://github.com/nexB/scancode-toolkit/blob/157a64701f996676f7b252a514760bc9668c68aa/tests/packagedcode/data/rubygems/gemspec/arel.gemspec#L8
and in https://github.com/nexB/scancode-toolkit/pull/2075/files#diff-f2830f25610f80643082bfc08bb4ca40L9
why losing that?

tests/packagedcode/data/rubygems/gemspec/arel.gemspec.expected.json
```diff
@@ -1,187 +1,62 @@
...
+        "is_optional": false,
+        "is_resolved": false
+      }
+    ],
+    "contains_source_code": null,
+    "source_packages": [],
+    "purl": "pkg:gem/arel@2.0.7.beta.20110429111451",
+    "repository_homepage_url": "https://rubygems.org/gems/arel/versions/2.0.7.beta.20110429111451",
+    "repository_download_url": "https://rubygems.org/downloads/arel-2.0.7.beta.20110429111451.gem",
+    "api_data_url": "https://rubygems.org/api/v2/rubygems/arel/versions/2.0.7.beta.20110429111451.json"
```
pombredanne, line 60: :+1:

tests/packagedcode/data/rubygems/gemspec/cocoapods.gemspec.expected.json
```diff
@@ -0,0 +1,216 @@
...
+        "requirement": "~> 2.6",
+        "scope": "dependencies",
+        "is_runtime": true,
+        "is_optional": false,
+        "is_resolved": false
+      }
+    ],
+    "contains_source_code": null,
+    "source_packages": [],
+    "purl": "pkg:gem/cocoapods@Pod::VERSION",
```
pombredanne, line 211: Where does `Pod::VERSION` come from?

rpotter12: https://github.com/CocoaPods/CocoaPods/blob/master/cocoapods.gemspec 
They actually used this value in gemspec file.

tests/packagedcode/data/rubygems/gemspec/arel.gemspec.expected.json
rpotter12: It is not fixed that number of authors and emails are same. So I both emails and authors in string.

src/packagedcode/spec.py
rpotter12: This function is for `.podspec` files not for `.gemspec` files. According to the documentation in `.podspec` files, description will be written in this manner but in gemspec file it is not fixed how to write a description.

src/packagedcode/spec.py
rpotter12: This will be a line. I will update that in the next changes :)

src/packagedcode/spec.py
rpotter12: In parsing gemspec there is a line
`line = 'spec.summary = %q{The loading cat generator.}'`
To get desire data of summary we can use

1. summary = re.sub(r'/*.*summary.*?=', '', line) (which I used)
2. summary = line.rpartition('=') (you suggested)
3. summary = line.split('=')
and many more ways
So what should I use??

rpotter12:
> @rpotter12 What is the purpose of the `nose` dependency you have added?

@majurg Many tests were failing because of it. I think `gemfileparser` is using nose somewhere but has not mentioned it in the installation file. That's why I have to add that here to get installed. Maybe I have to add this in setup.py file too.

src/packagedcode/rubygems.py
```diff
@@ -647,93 +626,66 @@ def normalize(gem_data, known_fields=known_fields):
...
     """
+    gemspec_obj = ParseSpec()
+    gemspec_data = gemspec_obj.parse_spec(location)
+    
+    name = gemspec_data.get('name')
+    version = gemspec_data.get('version')
+    homepage_url = gemspec_data.get('homepage_url')
+    summary = gemspec_data.get('summary')
+    description = gemspec_data.get('description')
+    if len(summary)>len(description):
```
pombredanne, line 641: there should be a space there around the `>`

src/packagedcode/rubygems.py
```diff
@@ -647,93 +626,66 @@ def normalize(gem_data, known_fields=known_fields):
...
+        declared_license = declared_license.split(',')
+
+    author = gemspec_data.get('author') or []
+    email = gemspec_data.get('email') or []
+    parties = []
+    if author:
+        parties.append(
+            models.Party(
+                type=models.party_person,
+                name=','.join(author),
```
pombredanne, line 655: If author is a list, create as many Party as we have author.
Same for email: and since we cannot be sure that an author matches an email based on aligned lists, you should create one different Party for each author and each email.

src/packagedcode/rubygems.py
```diff
@@ -647,93 +626,66 @@ def normalize(gem_data, known_fields=known_fields):
...
-    def __init__(self, location):
-        """
-        Initialize from the gem spec or gem file at location.
-        """
-        spec = parse_spec(location)
-        self.location = location
-        self.description = spec.get('description')
-        self.summary = spec.get('summary')
-        self.author = spec.get('author')
-        self.authors = spec.get('authors')
```
pombredanne, line 669: Do you still handle these cases of author vs. authors?

src/packagedcode/spec.py
pombredanne: and what if this is using `<<-DESCRIPTION`? will this work too?

src/packagedcode/spec.py
```diff
@@ -0,0 +1,158 @@
...
+
+logger = logging.getLogger(__name__)
+
+if TRACE:
+    import sys
+    logging.basicConfig(stream=sys.stdout)
+    logger.setLevel(logging.DEBUG)
+
+
+class ParseSpec(object):
```
pombredanne, line 52: What do you gain from having a class there?

src/packagedcode/spec.py
```diff
@@ -0,0 +1,158 @@
...
+            elif '.homepage' in line:
+                homepage_url = line.rpartition('=')
+                spec_data['homepage_url'] = get_stripped_data(homepage_url[2])
+
+        parser = GemfileParser(location)
+        deps = parser.parse()
+        dependencies = OrderedDict()
+        for key in deps:
+            if deps[key]:
+                for dependency in deps[key]:
```
pombredanne, line 109: What about 
```
depends = deps.gett(key, []) or []
for dep in depends:
...
```

src/packagedcode/spec.py
```diff
@@ -0,0 +1,158 @@
...
+                    dependencies[dependency.name] = (dependency.requirement).split(',')
```
pombredanne, line 110: why the parens around `(dependency.requirement)` ?

src/packagedcode/spec.py
```diff
@@ -0,0 +1,158 @@
...
+        return spec_data
+
+
+def get_stripped_data(line):
+    """
+    Return line after removing unnecessary special character and space.
+    """
+    if '#' in line:
+        line = line[:line.index('#')]
+    stripped_data = line.strip()
```
pombredanne, line 122: what about chaining the calls instead?
may be more or less:
```Python
stripped_data = (
    line.strip()
    .replace("'", '')
    .replace('"', '')
    .replace('{', '')
    .replace('}', '')
    .replace('[', '')
    .replace(']', '')
    .replace('%q', '')
    .strip()
)
```
or a loop of sorts?
```Python
stripped_data = line.strip()
for strippable in ("'",'"', '{', '}', '[', ']', '%q',):
    stripped_data = stripped_data.replace(strippable, '')

return stripped_data.strip()
```

src/packagedcode/spec.py
```diff
@@ -0,0 +1,158 @@
...
+    https://guides.cocoapods.org/syntax/podspec.html#description
+    description is in the form:
+    spec.description = <<-DESC
+                     Computes the meaning of life.
+                     Features:
+                     1. Is self aware
+                     ...
+                     42. Likes candies.
+                    DESC
+    Return description from podspec.
```
pombredanne, line 145: This should be the first line of the docstring

tests/packagedcode/data/rubygems/gemspec/arel.gemspec.expected.json
```diff
@@ -1,187 +1,62 @@
...
-    "Emilio Tagua", 
-    "Nick Kallen"
-  ], 
-  "date": "2011-04-29 00:00:00 UTC", 
-  "requirements": null, 
-  "dependencies": [
-    "minitest (>= 2.0.2, development)", 
-    "hoe (>= 2.9.1, development)"
-  ], 
-  "files": [
```
pombredanne, line 27: Do we still capture these and test files below?

tests/packagedcode/data/rubygems/gemspec/arel.gemspec.expected.json
```diff
@@ -1,187 +1,62 @@
-{
-  "platform": "ruby",
```
pombredanne, line 2: Do we capture this too, as there are possible "java" values?

tests/packagedcode/data/rubygems/gemspec/arel.gemspec.expected.json
pombredanne: As suggested above, create instead as many Party as needed in this case.

tests/packagedcode/data/rubygems/gemspec/github.gemspec.expected.json
```diff
@@ -0,0 +1,70 @@
...
+      {
+        "purl": "pkg:gem/activerecord",
+        "requirement": "~>3.0.0",
+        "scope": "dependencies",
+        "is_runtime": true,
+        "is_optional": false,
+        "is_resolved": false
+      }
+    ],
+    "contains_source_code": null,
```
pombredanne, line 63: I think that in the general case a Gem contains source code, right? may be not for Pods though

tests/packagedcode/data/rubygems/gemspec/github.gemspec
```diff
@@ -0,0 +1,31 @@
...
+  s.executables   = `git ls-files -- bin/*`.split("\n").map{ |f| File.basename(f) }
+  s.require_paths = ["lib"]
+  
+  s.add_dependency "text-hyphen", "1.0.0"
+  s.add_dependency "text-format", "1.0.0"
+  s.add_dependency "highline", "~> 1.6"
+  s.add_dependency "json_pure", "~> 1.5.1"
+  s.add_dependency "launchy", "~> 2.0.2"
+
+  s.add_development_dependency "rake"
```
pombredanne, line 28: It looks like you are missing capturing these as "development" scope.

tests/packagedcode/data/rubygems/gemspec/rubocop.gemspec
```diff
@@ -11,12 +11,9 @@ Gem::Specification.new do |s|
...
   s.required_ruby_version = '>= 2.2.0'
   s.authors = ['Bozhidar Batsov', 'Jonas Arvidsson', 'Yuji Nakayama']
-  s.description = <<-DESCRIPTION
-    Automatic Ruby code style checking tool.
-    Aims to enforce the community-driven Ruby Style Guide.
-  DESCRIPTION
+  s.description = "Automatic Ruby code style checking tool. Aims to enforce the community-driven Ruby Style Guide."
 
-  s.email = 'rubocop@googlegroups.com'
+  s.email = ['rubocop@googlegroups.com']
```
pombredanne, line 16: Why did you change this test file? This is a real one at https://github.com/rubocop-hq/rubocop/blob/5cb29efb62f3ff36cbc9ad75d2c29392737734e2/rubocop.gemspec
We need to handle it, and not change it to fit the code ;)

tests/packagedcode/data/rubygems/gemspec/mecab-ruby.gemspec
```diff
@@ -0,0 +1,16 @@
...
+	s.author = 'Taku Kudo'
+	s.date = '2011-12-24'
+	s.description = "Ruby bindings for MeCab, a morphological analyzer."
+	s.email = 'taku@chasen.org'
+	s.extensions = [ 'extconf.rb' ]
+	s.files = [ 'AUTHORS', 'BSD', 'COPYING', 'GPL', 'LGPL',
+		'MeCab_wrap.cpp', 'README', 'bindings.html',
+		'extconf.rb', 'mecab-ruby.gemspec', 'test.rb' ]
+	s.has_rdoc = false
+	s.homepage = 'http://mecab.sourceforge.net/'
```
pombredanne, line 13: You are missing this in homepage_url below

tests/packagedcode/test_rubygems.py
```diff
@@ -47,66 +48,62 @@
...
-        self.check_gemspec(
-            'rubygems/gemspec/address_standardization.gemspec',
-            'rubygems/gemspec/address_standardization.gemspec.expected.json')
+        # dependency not in format as required in gemfilparser
+        test_file = self.get_test_loc('rubygems/gemspec/address_standardization.gemspec')
+        expected_loc = self.get_test_loc('rubygems/gemspec/address_standardization.gemspec.expected.json')
+        packages = rubygems.RubyGem.recognize(test_file)
+        self.check_packages(packages, expected_loc, regen=False)
 
+    @expectedFailure
```
pombredanne, line 62: Why expecting a failure there? what is the issue?

src/packagedcode/rubygems.py
rpotter12: I will change this :)

src/packagedcode/spec.py
rpotter12: I am following Cocoapods(https://guides.cocoapods.org/syntax/podspec.html#description) documentation. The general way is using `DESC` but using `DESC` will also work for `<<-DESCRIPTION` :)

tests/packagedcode/data/rubygems/gemspec/arel.gemspec.expected.json
rpotter12: No not for now

tests/packagedcode/test_rubygems.py
rpotter12: As I have mentioned in comment, `gemfileparser` is extracting for a particular type only. So the dependency which are in the gemspec file should be in our output file but it not extracting beacause `gemfileparser` does not contain the code for extracting.

tests/packagedcode/data/rubygems/gemspec/github.gemspec
rpotter12: I am not missing "development" scope. I am missing "add_dependency" because we fetching dependency from `gemfileparser` and it does contain code to handle these type of dependency. `gemfileparser` can only fetch .add_development_dependency and .add_runtime_dependency and with more limitation in it.

src/packagedcode/rubygems.py
rpotter12: No I am not handling anything like this

tests/packagedcode/data/rubygems/gemspec/github.gemspec
rpotter12: I am not missing "development" scope. I am missing "add_dependency" because we fetching dependency from gemfileparser and it does contain code to handle these type of dependency. gemfileparser can only fetch .add_development_dependency and .add_runtime_dependency and with more limitation in it.

rpotter12:
I will add ABOUT files shortly :)

rpotter12:
@pombredanne Please review the PR again. I will add ABOUT files shortly :)

setup.py
```diff
@@ -202,6 +202,8 @@ def read(*names, **kwargs):
         'toml >= 0.10.0',
         'pkginfo >= 1.5.0.1',
         'dparse >= 0.4.1',
+        'gemfileparser >= 0.6.2',
+        'nose >= 1.3.7',
```
pombredanne, line 206: Now that we have forked as gemfileparser2, we should leverage this an remove nose. I would rather not add nose at all

src/packagedcode/cocoapods.py
```diff
@@ -0,0 +1,193 @@
...
+from packagedcode import spec
+
+
+"""
+Handle cocoapods packages manifests for macOS and iOS
+including .podspec, Podfile and Podfile.lock files.
+See https://cocoapods.org
+"""
+
+# TODO: implementation to get dependency data using gemsfileparser
```
pombredanne, line 47: If you keep these for another PR later, it could make sense to enter a ticket?

src/packagedcode/rubygems.py
```diff
@@ -647,93 +626,73 @@ def normalize(gem_data, known_fields=known_fields):
...
+    version = gemspec_data.get('version')
+    homepage_url = gemspec_data.get('homepage_url')
+    summary = gemspec_data.get('summary')
+    description = gemspec_data.get('description')
+    if len(summary) > len(description):
+        description = summary
+
+    declared_license = gemspec_data.get('license')
+    if declared_license:
+        declared_license = declared_license.split(',')
```
pombredanne, line 645: Just curious: is this the podspec spec to have license always separated by comma?

src/packagedcode/rubygems.py
pombredanne: But they do exist in the wild, do they? And they were supported and tested before?

src/packagedcode/rubygems.py
```diff
@@ -647,93 +626,73 @@ def normalize(gem_data, known_fields=known_fields):
...
-        return email
-
-    def map_licenses(self):
-        licenses = self.spec.get('licenses', [])
-        if not isinstance(licenses, list):
-            licenses = [licenses]
-
-        mapped_licenses = []
-        for lic in licenses:
-            mapped_license = LICENSES_MAPPING.get(lic, None)
```
pombredanne, line 730: Do you get the license_expression from LICENSES_MAPPING?

src/packagedcode/spec.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    with io.open(location, encoding='utf-8', closefd=True) as data:
+        lines = data.readlines()
+
+    spec_data = {}
+
+    for line in lines:
+        if '.name' in line:
+            name = line.rpartition('=')
+            spec_data['name'] = get_stripped_data(name[2])
```
pombredanne, line 64: The benefit of partition is that you have a stable number of items returned so you could write that instead and avoid using a magic index for 2:
```
            _key, _sep, value = line.rpartition('=')
            spec_data['name'] = get_stripped_data(value) 
```

src/packagedcode/spec.py
```diff
@@ -0,0 +1,153 @@
...
+        elif '.version' in line and '.version.' not in line:
+            version = line.rpartition('=')
+            spec_data['version'] = get_stripped_data(version[2])
+        elif '.license' in line:
+            license_type = line.rpartition('=')
+            spec_data['license'] = get_stripped_data(license_type[2])
+        elif '.email' in line:
```
pombredanne, line 71: We have authors and emails (plural):
https://github.com/search?q="authors"+extension%3A.gemspec&type=Code
https://github.com/search?q="emails"+extension%3A.gemspec&type=Code

setup.py
rpotter12: I will modify gemfileparser2 and use that library here and remove these :)

src/packagedcode/cocoapods.py
rpotter12: No, I'm not keeping these for another PR. Actually at that time we were using gemfileparser library to get the data and in that library dependencies are fetched which are in a particular format only. 
I will use the gemfileparser2 library here and solve this :)

src/packagedcode/rubygems.py
rpotter12: No, not always separated by the comma. I have used this to make it a list because I was getting the correct license_expression value after doing this. I will look at this again :)

src/packagedcode/rubygems.py
rpotter12: I didn't use any LICENSES_MAPPING to get the license_expression. When I convert declared_license as a list, it gives the correct license_expression.

src/packagedcode/spec.py
rpotter12: Thanks, I will change my code according to this :)

src/packagedcode/spec.py
rpotter12: We are using a concept that a particular string is present in the line or not. and I think `.author` will work for both `authors` and `author` and similar to the email also. I think we will not have any error by using this. Can you tell me a case when it will detect wrong??

rpotter12:
@pombredanne Please ignore test files and checks for now. checks may be failing as gemfileparser2 has published version

tests/packagedcode/test_rubygems.py
```diff
@@ -46,67 +47,59 @@
...
-    def test_rubygems_rubocop_gemspec(self):
-        self.check_gemspec(
-            'rubygems/gemspec/rubocop.gemspec',
-            'rubygems/gemspec/rubocop.gemspec.expected.json')
-
-    def test_rubygems_gemspec_with_variables(self):
-        self.check_gemspec(
-            'rubygems/gemspec/with_variables.gemspec',
-            'rubygems/gemspec/with_variables.gemspec.expected.json')
+    if py3: # pytest fails in py2 because TypeError with StringIO
```
pombredanne, line 55: Could you use instead a `@skipIf(py2, 'Does not pass on Python2)` decorator OR better... may be the code can fixed upstream in gemfileparser?

rpotter12: I have used @pytest.mark.skipif to skip py2. Please review it again :)

pombredanne (approved):
All good! Thank you ++ for your preseverance!

PR #2077, opened 2020-06-20
Packagedcode to handle Python(requirements.txt)

### LINKED ISSUE #2037: Parse requirements.txt files

We need to add the ability to parse `requirements.txt` files and have them apart of a `--package` scan.

rpotter12:
Fixes #2037 

### Tasks

### DIFF
--- a/.github/ISSUE_TEMPLATE/--good-first-issue.md
+++ b/.github/ISSUE_TEMPLATE/--good-first-issue.md
@@ -66,7 +66,7 @@ Make sure all these boxes are checked before your pull request (PR) is ready to 
   Refer [Documentation](https://scancode-toolkit.readthedocs.io/en/latest/contribute/contrib_dev.html#running-tests)
 
 - [ ] 💾 **Commit and Push** your changes. Do check out this documentation on 
-  [writing good commit messages](https://aboutcode.readthedocs.io/en/latest/aboutcode-docs/writing_good_commit_messages.html) 
+  [writing good commit messages](https://aboutcode.readthedocs.io/en/latest/coding-principles/writing_good_commit_messages.html) 
 
 - [ ] 🔀 **Start a Pull Request**. There are two ways how you can start a pull request:
 


src/packagedcode/pypi.py
```diff
@@ -303,6 +305,33 @@ def parse_setup_py(location):
...
+def parse_requirements_txt(location):
+    """
+    Return a Package built from requirements.txt.
+    """
+    dependencies = OrderedDict()
+    with io.open(location, encoding='utf-8') as loc:
+        for req in requirements.parse(loc):
+            dependencies[req.name] = req.specs
+    package_dependencies = []
+    for dep_name, dep_specs in dependencies.items():
```
pombredanne, line 317: What benefits do you gain from reading your requirements in the `dependencies` mapping and then reding that mapping `dependencies.items()` here?

src/packagedcode/pypi.py
```diff
@@ -303,6 +305,33 @@ def parse_setup_py(location):
...
+        package_dependencies.append(
+            models.DependentPackage(
+                purl=PackageURL(
+                    type='pypi',
+                    name=dep_name
+                ).to_string(),
+                requirement=''.join(dep_specs[0]),
+                scope='dependencies',
```
pombredanne, line 325: Could there be a smarter way to infer a scope?
- Plain `requirements.txt` files would be `runtime` IMHO or whatever scope we use in for setup.py `install_requires`
-  `requirements_test.txt` or `requirements-tests.txt`  etc.. --> `test` scope?
-  `requirements_dev.txt` or `requirements-dev.txt`  etc.. --> `dev` scope?
...
etc?

src/packagedcode/pypi.py
```diff
@@ -303,6 +305,33 @@ def parse_setup_py(location):
...
+            dependencies[req.name] = req.specs
+    package_dependencies = []
+    for dep_name, dep_specs in dependencies.items():
+        package_dependencies.append(
+            models.DependentPackage(
+                purl=PackageURL(
+                    type='pypi',
+                    name=dep_name
+                ).to_string(),
+                requirement=''.join(dep_specs[0]),
```
pombredanne, line 324: Is there a case where the version is pinned with `==` ? if yes, then the version should be added to the `purl` and `is_resolved = True`

src/packagedcode/pypi.py
```diff
@@ -303,6 +305,33 @@ def parse_setup_py(location):
...
+                scope='dependencies',
+                is_runtime=True,
```
pombredanne, line 326: Based on the scope inference above, tests and dev would not be is_runtime

src/packagedcode/pypi.py
```diff
@@ -303,6 +305,33 @@ def parse_setup_py(location):
...
+                is_optional=False,
```
pombredanne, line 327: Dev and test reqs could be  optional.

src/packagedcode/pypi.py
```diff
@@ -303,6 +305,33 @@ def parse_setup_py(location):
...
+                is_resolved=False,
```
pombredanne, line 328: if we have an exact version pinned, its is resolved

tests/packagedcode/data/pypi/requirements.txt/gemfileparser-requirements-txt-expected.json
```diff
@@ -0,0 +1,52 @@
...
+      "purl": "pkg:pypi/setuptools",
+      "requirement": "==32.0.0",
+      "scope": "dependencies",
+      "is_runtime": true,
+      "is_optional": false,
+      "is_resolved": false
+    },
+    {
+      "purl": "pkg:pypi/nose",
+      "requirement": "==1.3.7",
```
pombredanne, line 39: See my comments about the case of `==` pinned requirements ... there we should also add that to the pulr version and mark this as resolved

PR #2078, opened 2020-06-20
Packagedcode to handle Python(requirements.txt)

### LINKED ISSUE #2037: Parse requirements.txt files

rpotter12:
Fixes #2037 

### Tasks

### DIFF
--- a/src/packagedcode/pypi.py
+++ b/src/packagedcode/pypi.py
@@ -83,7 +83,7 @@ if TRACE:
 
 @attr.s()
 class PythonPackage(models.Package):
-    metafiles = ('metadata.json', '*setup.py', 'PKG-INFO', '*.whl', '*.egg')
+    metafiles = ('metadata.json', '*setup.py', 'PKG-INFO', '*.whl', '*.egg', '*requirements*.txt', '*requirements*.in')
     extensions = ('.egg', '.whl', '.pyz', '.pex',)
     default_type = 'pypi'
     default_primary_language = 'Python'
@@ -112,6 +112,8 @@ def parse(location):
         file_name = fileutils.file_name(location)
         parsers = {
             'setup.py': parse_setup_py,
+            'requirements.txt': parse_requirements_txt,
+            'requirements.in': parse_requirements_txt,
             'metadata.json': parse_metadata,
             'PKG-INFO': parse_pkg_info,
             '.whl': parse_wheel,
@@ -187,11 +189,25 @@ def parse_dependencies(location, package):
             package.dependencies = dependencies
 
 
+def file_type(file_name):
+    """
+    Return a renamed filename.
+    """
+    req_file_types = ('.txt', '.in')
+    if file_name.endswith(req_file_types):
+        file_name = 'requirements.txt'
+
+    return file_name
+
+
 def parse_with_dparse(location):
     is_dir = filetype.is_dir(location)
     if is_dir:
         return
     file_name = fileutils.file_name(location)
+
+    file_name = file_type(file_name)
+
     if file_name not in (filetypes.requirements_txt,
                          filetypes.conda_yml,
                          filetypes.tox_ini,
@@ -211,9 +227,12 @@ def parse_with_dparse(location):
         package_dependencies = []
         for df_dependency in df_dependencies:
             specs = df_dependency.specs
+            is_resolved = False
             requirement = None
             if specs:
                 requirement = str(specs)
+                if '==' in requirement:
+                    is_resolved = True
             package_dependencies.append(
                 models.DependentPackage(
                     purl=PackageURL(
@@ -221,10 +240,19 @@ def parse_with_dparse(location):
                     scope='dependencies',
                     is_runtime=True,
                     is_optional=False,
-                    requirement=requirement,
+                    is_resolved=is_resolved,
+                    requirement=requirement
                 )
             )
         return package_dependencies
+
+
+def parse_requirements_txt(location):
+    """
+    Return a package built from requirements.txt.
+    """
+    package_dependencies = parse_with_dparse(location)
+    return PythonPackage(dependencies=package_dependencies)
 
 
 def parse_setup_py(location):
--- a/src/packagedcode/pypi.py
+++ b/src/packagedcode/pypi.py
@@ -209,9 +209,9 @@ def parse_with_dparse(location):
         return
     file_name = fileutils.file_name(location)
 
-    file_name = get_dependency_type(file_name)
+    dependency_type = get_dependency_type(file_name)
 
-    if file_name not in (filetypes.requirements_txt,
+    if dependency_type not in (filetypes.requirements_txt,
                          filetypes.conda_yml,
                          filetypes.tox_ini,
                          filetypes.pipfile,
@@ -223,31 +223,43 @@ def parse_with_dparse(location):
         mode = 'r'
     with open(location, mode) as f:
         content = f.read()
-        df = dparse.parse(content, file_type=file_name)
-        df_dependencies = df.dependencies
-        if not df_dependencies:
-            return
-        package_dependencies = []
-        for df_dependency in df_dependencies:
-            specs = df_dependency.specs
-            is_resolved = False
-            requirement = None
-            if specs:
-                requirement = str(specs)
-                if '==' in requirement:
-                    is_resolved = True
-            package_dependencies.append(
-                models.DependentPackage(
-                    purl=PackageURL(
-                        type='pypi', name=df_dependency.name).to_string(),
-                    scope='dependencies',
-                    is_runtime=True,
-                    is_optional=False,
-                    is_resolved=is_resolved,
-                    requirement=requirement
-                )
+        
+    df = dparse.parse(content, file_type=dependency_type)
+    df_dependencies = df.dependencies
+
+    if not df_dependencies:
+        return
+
+    package_dependencies = []
+    for df_dependency in df_dependencies:
+        specs = list(df_dependency.specs._specs)
+        is_resolved = False
+        requirement = None
+        purl = PackageURL(
+            type='pypi',
+            name=df_dependency.name
+        ).to_string()
+        if specs:
+            requirement = specs[0]
+            if specs[0].operator == '==':
+                is_resolved = True
+                purl = PackageURL(
+                    type='pypi',
+                    name=df_dependency.name,
+                    version=specs[0].version
+                ).to_string()
+        package_dependencies.append(
+            models.DependentPackage(
+                purl=purl,
+                scope='dependencies',
+                is_runtime=True,
+                is_optional=False,
+                is_resolved=is_resolved,
+                requirement=requirement
             )
-        return package_dependencies
+        )
+
+    return package_dependencies
 
 
 def parse_requirements_txt(location):
--- a/src/packagedcode/pypi.py
+++ b/src/packagedcode/pypi.py
@@ -191,7 +191,7 @@ def parse_dependencies(location, package):
 
 def file_type(file_name):
     """
-    Return a renamed filename.
+    Return a type of dependency file.
     """
     req_file_types = ('.txt', '.in')
     if file_name.endswith(req_file_types):
--- a/src/packagedcode/pypi.py
+++ b/src/packagedcode/pypi.py
@@ -223,7 +223,7 @@ def parse_with_dparse(location):
         mode = 'r'
     with open(location, mode) as f:
         content = f.read()
-        
+
     df = dparse.parse(content, file_type=dependency_type)
     df_dependencies = df.dependencies
 
@@ -240,14 +240,17 @@ def parse_with_dparse(location):
             name=df_dependency.name
         ).to_string()
         if specs:
-            requirement = specs[0]
-            if specs[0].operator == '==':
-                is_resolved = True
-                purl = PackageURL(
-                    type='pypi',
-                    name=df_dependency.name,
-                    version=specs[0].version
-                ).to_string()
+            requirement = str(df_dependency.specs)
+            for spec in specs:
+                operator = spec.operator
+                version = spec.version
+                if any(operator == element for element in ('==', '===')):
+                    is_resolved = True
+                    purl = PackageURL(
+                        type='pypi',
+                        name=df_dependency.name,
+                        version=version
+                    ).to_string()
         package_dependencies.append(
             models.DependentPackage(
                 purl=purl,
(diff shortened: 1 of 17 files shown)


src/packagedcode/pypi.py
```diff
@@ -303,6 +305,45 @@ def parse_setup_py(location):
...
+        runtime = False
+        optional = True
+    elif 'dev.txt' in location:
+        scope = 'dev'
+        runtime = False
+        optional = True
+    with io.open(location, encoding='utf-8') as loc:
+        for req in requirements.parse(loc):
+            resolved = False
+            if '==' in (req.specs)[0]:
```
JonoYang, line 327: I think `(req.specs)[0]` can be made into a variable since you reference it in multiple places after this, like line 334 and line 336.

Also, does this work if a requirement.txt requirement does not state a version? For example, consider the following as a requirements.txt:

```
setuptools==32.0.0
django
```

Would django be processed properly?

src/packagedcode/pypi.py
```diff
@@ -83,8 +84,8 @@ def logger_debug(*args):
 
 @attr.s()
 class PythonPackage(models.Package):
-    metafiles = ('metadata.json', '*setup.py', 'PKG-INFO', '*.whl', '*.egg')
-    extensions = ('.egg', '.whl', '.pyz', '.pex',)
+    metafiles = ('metadata.json', '*setup.py', 'PKG-INFO', '*.whl', '*.egg', '*.txt')
+    extensions = ('.egg', '.whl', '.pyz', '.pex', '.txt')
```
majurg, line 88: Not sure we want to have *.txt or .txt here. This will classify text files as requirements files, right?

src/packagedcode/pypi.py
```diff
@@ -112,6 +113,7 @@ def parse(location):
         file_name = fileutils.file_name(location)
         parsers = {
             'setup.py': parse_setup_py,
+            '.txt': parse_requirements_txt,
```
majurg, line 116: See comment above. I am skeptical with this approach (but I could be wrong)

@pombredanne thoughts?

src/packagedcode/pypi.py
```diff
@@ -303,6 +305,45 @@ def parse_setup_py(location):
...
+    runtime = True
+    optional = False
+    if 'test.txt' in location:
+        scope = 'test'
+        runtime = False
+        optional = True
+    elif 'dev.txt' in location:
+        scope = 'dev'
+        runtime = False
+        optional = True
```
majurg, line 323: I would add a newline here.

src/packagedcode/pypi.py
majurg: I agree with @JonoYang. I think your approach is slightly flawed here.

src/packagedcode/pypi.py
```diff
@@ -303,6 +305,45 @@ def parse_setup_py(location):
...
+        for req in requirements.parse(loc):
+            resolved = False
+            if '==' in (req.specs)[0]:
+                resolved = True
+            package_dependencies.append(
+                models.DependentPackage(
+                    purl=PackageURL(
+                        type='pypi',
+                        name=req.name,
+                        version=req.specs[0][1]
```
majurg, line 334: multiple list references like this are bad style.

tests/packagedcode/data/pypi/requirements.txt/gemfileparser-requirements-txt-expected.json
```diff
@@ -0,0 +1,52 @@
...
+  "vcs_url": null,
+  "copyright": null,
+  "license_expression": null,
+  "declared_license": null,
+  "notice_text": null,
+  "root_path": null,
+  "dependencies": [
+    {
+      "purl": "pkg:pypi/setuptools@32.0.0",
+      "requirement": "==32.0.0",
```
majurg, line 31: This is not correct. I would expect this value to be `setuptools`

rpotter12: @majurg Please see https://github.com/nexB/scancode-toolkit/pull/2077#discussion_r443118795 comment of @pombredanne

src/packagedcode/pypi.py
rpotter12: @JonoYang You were right. I have made some changes :)

tests/packagedcode/data/pypi/requirements.txt/gemfileparser-requirements-txt-expected.json
majurg: Ok. If `requirement` is not used for anything (now that all the package information is in the purl), why not delete this field? 

I guess mainly I am confused at to what `requirement` refers to. I interpret that field name to mean the name of the python package, but maybe this is wrong?

rpotter12:
```
purl = String(
    repr=True,
    label='Dependent package URL',
    help='A compact purl package URL. Typically when there is an unresolved requirement, there is no version. ' 'If the dependency is resolved, the version should be added to the purl')

requirement = String(
    repr=True,
    label='dependent package version requirement',
    help='A string defining version(s)requirements. Package-type specific.')
```
So both are important :)

majurg: Ah ok this is fine then :+1: thanks

src/packagedcode/pypi.py
ncoghlan: The main naming conventions I'm aware of are `requirements.txt`, `dev-requirements.txt` and `locked-requirements.txt` (I'm not sure how widespread the last one is, I just prefer using "requirements.txt" and "locked-requirements.txt" when using pip-compile)

rpotter12: @ncoghlan @majurg @pombredanne filename can be `requirements.txt`, `dev-requirements.txt`, `test-requirements.txt`, `requirements-dev.txt` or `requirements-test.txt` so what should I choose??

pombredanne: @rpotter12 IMHO all of them :) ... but what this really means is that the function that detects and parses requirements need to be flexible there and may not rely too much on the generic package type metadata.

rpotter12:
@majurg @pombredanne 
According to https://github.com/pyupio/dparse/blob/master/dparse/filetypes.py#L4 it parse files whose name is `requirements.txt`. I have done some minor changes and added test cases too :)

tests/packagedcode/data/plugin/help.txt
```diff
@@ -203,7 +203,7 @@ Package: nuget
 --------------------------------------------
 Package: pypi
   class: packagedcode.pypi:PythonPackage
-  metafiles: metadata.json, *setup.py, PKG-INFO, *.whl, *.egg
+  metafiles: metadata.json, *setup.py, PKG-INFO, *.whl, *.egg, *requirements.txt
```
majurg, line 206: What about `requirements.in` files and similar?

src/packagedcode/pypi.py
```diff
@@ -192,6 +194,10 @@ def parse_with_dparse(location):
     if is_dir:
         return
     file_name = fileutils.file_name(location)
+    
+    if file_name.endswith('requirements.txt') or file_name.endswith('requirements.in'):
+        file_name = "requirements.txt"
```
pombredanne, line 199: Please use single quotes consistently

src/packagedcode/pypi.py
```diff
@@ -211,22 +217,34 @@ def parse_with_dparse(location):
         package_dependencies = []
         for df_dependency in df_dependencies:
             specs = df_dependency.specs
+            is_resolved = False
             requirement = None
             if specs:
                 requirement = str(specs)
+                if "==" in requirement:
```
pombredanne, line 224: nit: quotes

src/packagedcode/pypi.py
```diff
@@ -192,6 +194,10 @@ def parse_with_dparse(location):
     if is_dir:
         return
     file_name = fileutils.file_name(location)
+    
+    if file_name.endswith('requirements.txt') or file_name.endswith('requirements.in'):
```
pombredanne, line 198: endswith() takes a tuple and does not need to be called twice. Also what about `requirements-dev.txt` and similar?
IMHO you should make this a separate function and you can then have a series of unit test for it.

src/packagedcode/pypi.py
```diff
@@ -83,7 +83,7 @@ def logger_debug(*args):
 
 @attr.s()
 class PythonPackage(models.Package):
-    metafiles = ('metadata.json', '*setup.py', 'PKG-INFO', '*.whl', '*.egg')
+    metafiles = ('metadata.json', '*setup.py', 'PKG-INFO', '*.whl', '*.egg', '*requirements*.txt', '*requirements*.in')
```
majurg, line 86: Do we have test cases for these filename recognitions yet? if not we should add them

rpotter12: https://github.com/nexB/scancode-toolkit/blob/83f29f9dbbca54a5d69db0d6d7617b8b4dd7021a/tests/packagedcode/test_pypi.py#L314-L333

I have added test here. I think this will be OK :)

majurg: nice

rpotter12:
@majurg @pombredanne If everything looks good please approve and merge this PR.:)

src/packagedcode/pypi.py
```diff
@@ -187,11 +189,25 @@ def parse_dependencies(location, package):
             package.dependencies = dependencies
 
 
+def file_type(file_name):
+    """
+    Return a renamed filename.
```
pombredanne, line 194: You are not returning a renamed filename, are you?
Rather you are returning a type of dependency file... so please update your doc accordingly.

src/packagedcode/pypi.py
```diff
@@ -112,6 +112,8 @@ def parse(location):
         file_name = fileutils.file_name(location)
         parsers = {
             'setup.py': parse_setup_py,
+            'requirements.txt': parse_requirements_txt,
+            'requirements.in': parse_requirements_txt,
```
pombredanne, line 116: Also note that for the future, I have submitted this PR upstream https://github.com/pyupio/dparse/pull/48

rpotter12: :+1:

tests/packagedcode/test_pypi.py
```diff
@@ -267,8 +268,66 @@ def test_pkginfo_parse_with_wheelfile_with_parse_function(self):
         expected_loc = self.get_test_loc('pypi/wheel/parse-wheel-expected.json')
         self.check_package(package, expected_loc, regen=False)
 
-    def test_parse_with_dparse(self):
```
pombredanne, line 270: IMHO it is a good idea to keep a test for this function too

tests/packagedcode/test_pypi.py
```diff
@@ -28,6 +28,7 @@
 
 from collections import OrderedDict
 import os
+import pytest
```
pombredanne, line 31: pytest imports goes after standard imports

src/packagedcode/pypi.py
```diff
@@ -187,11 +189,25 @@ def parse_dependencies(location, package):
             package.dependencies = dependencies
 
 
+def file_type(file_name):
```
pombredanne, line 192: What about this may be? which will also support adding other dparse parsers?
```
dependency_type_by_extensions = {
    ('.txt', '.in'): 'requirements.txt',
}

def get_dependency_type(file_name, dependency_type_by_extensions=dependency_type_by_extensions):
    """
    Return the type of a dependency as a string or None given a `file_name` string.
    """
    for extensions, dependency_type in dependency_type_by_extensions.items():
        if file_name.endswith(extensions):
            return dependency_type
```

rpotter12: I was also about to change this like this. This will be more easier.
I will change this shortly :)

rpotter12:
@pombredanne Please review the PR again. :)

src/packagedcode/pypi.py
```diff
@@ -187,11 +189,28 @@ def parse_dependencies(location, package):
...
+            return dependency_type
+
+
 def parse_with_dparse(location):
     is_dir = filetype.is_dir(location)
     if is_dir:
         return
     file_name = fileutils.file_name(location)
+
+    file_name = get_dependency_type(file_name)
```
pombredanne, line 212: may be rather:
`dependency_type = get_dependency_type(file_name)`

src/packagedcode/pypi.py
```diff
@@ -187,11 +189,28 @@ def parse_dependencies(location, package):
...
+
     if file_name not in (filetypes.requirements_txt,
```
pombredanne, line 214: `if dependency_type not in (filetypes.requirements_txt,`

src/packagedcode/pypi.py
```diff
@@ -211,22 +230,34 @@ def parse_with_dparse(location):
         package_dependencies = []
         for df_dependency in df_dependencies:
             specs = df_dependency.specs
+            is_resolved = False
             requirement = None
             if specs:
                 requirement = str(specs)
+                if '==' in requirement:
```
pombredanne, line 237: What about `===` and others quasi-equal in the latest specs?
Also since `df_dependencies.specs` is a list of Specifier packaging object https://github.com/pypa/packaging/blob/master/packaging/specifiers.py  why not use that instead of taking a `str()` on it?
You could check if the `.operator` matches a list of equality operators AND that there is only one specifier (which is the only valid case for `is_resolved` ) 
See https://github.com/pypa/packaging/blob/master/packaging/specifiers.py#L173

src/packagedcode/pypi.py
```diff
@@ -204,27 +223,51 @@ def parse_with_dparse(location):
...
+    for df_dependency in df_dependencies:
+        specs = list(df_dependency.specs._specs)
+        is_resolved = False
+        requirement = None
+        purl = PackageURL(
+            type='pypi',
+            name=df_dependency.name
+        ).to_string()
+        if specs:
+            requirement = specs[0]
```
pombredanne, line 243: what if there are more than one items in `specs?

pombredanne: I think you missed adding back something like:
```
requirement = str(df_dependency.specs)
```
?

src/packagedcode/pypi.py
```diff
@@ -204,27 +223,51 @@ def parse_with_dparse(location):
...
+            if specs[0].operator == '==':
```
pombredanne, line 244: what if there are more than one items in `specs`?

rpotter12: It is fixed now

Issue #2072, opened 2020-06-16
Dev utility: Create release scripts to build archives


pombredanne:
One archive for each OS/Python combo should be created and each should have its exact set of pinned requirements included in the built archive, no more and no less.

pombredanne:
See #2117

pombredanne:
So the way this would typically work out would be IMHO this:
1. there is a single MANIFEST.in file that excludes `tests/` and `docs` and includes the `thirparty/`
2. somehow there is a script that is launched and that accepts as option something like `--os window or linux or macox`, `--arch 32 or 64` and `--python 36, 37, etc` ... these should normally be the same value as the wheel tags. And also accept a requirements file as an input.
3. This script will first wipe clean the thirdparty/ directory 
4. Then it will  download there just the set of wheels, sdists and ABOUT files that are needed for the options that were chosen.  These wheels should be selected exactly from the provided requirement files and fetched from our own "github-repo-release-Pypi-repo" and saved in the thirdparty/ dir
5. finally the script generates an a `python setup.py sdist` that will include the current thirdparty and possibly renames that file so it contains the proper os/arch/python "tags" it its archive name. 

Another way can work too. That's one of the  ways :)

Abhishek-Dev09:
@altendky : Your suggestion are also welcome here.

Abhishek-Dev09:
@pombredanne : Only sdist/wheels are available in thirdparty not ABOUT,NOTICE because we are using `pip-download` here, that will be integrated with `romp` because `romp` will provide matrix for python/os/ach.

pombredanne:
The new https://github.com/nexB/scancode-toolkit/blob/1bddb92c70abcf26cf9435061e429bae62cbe040/etc/release/scancode-create-release.sh script creates these alright.

Issue #2071, opened 2020-06-16
 Dev utility: Create build loop in repo to build and upload wheels and sdist for all operating systems and Python versions in our "personal" PyPI


pombredanne:
Note that for the GitHub release upload part there are few nice tools and libraries to consider:
- @paulthomson https://github.com/google/github-release-retry
- @jwilges https://github.com/jwilges/monocat
- @YakDriver https://github.com/plus3it/satsuki

ncoghlan:
You may also want to look at https://github.com/joerick/cibuildwheel

pombredanne:
So the candidates for building the wheel on all OSes x Python versions combos are:
- https://github.com/joerick/cibuildwheel (thanks @ncoghlan ! )
- https://github.com/altendky/romp (with a more general purpose focus and Azure-only and a focus on running on-demand)

pombredanne:
See also #2087

pombredanne:
Eventually we built a tool that handles creation, documentation and publishing of wheels and source https://github.com/nexB/scancode-toolkit/tree/1bddb92c70abcf26cf9435061e429bae62cbe040/etc/release
And this is also part of the https://github.com/nexB/skeleton

For the release proper https://github.com/google/github-release-retry does the job until GH fixes their flakiness.

PR #2117, opened 2020-07-16
Expose all .whl,.ABOUT, .NOTICE,.LICENCE files and Create an archive

### LINKED ISSUE #2068: Dev utility: Create new repo to expose pre-built wheels, tarballs and ABOUT files as releases in a PyPI-like repo
Actually we already have that:

https://github.com/nexB/thirdparty-packages/

Abhishek-Dev09:
It will upload all `.whl`,`.ABOUT`, `.NOTICE`,`.LICENCE` files of thirdparty repo to new repo as an asset of specific OS and python and also generate archive for each Oses and all python

Release them as asset on Github for #2136

Fixes #2068 #2069 #2072


github_release.py
```diff
@@ -0,0 +1,157 @@
...
+import sys
+
+
+def release_asset(token, tag, repo, body, user, limit, file):
+    """
+    Release .whl,.ABOUT,.NOTICE,.LICENSE to github repository from 
+    given directory.
+    """
+
+    cwd = os.getcwd()
```
Abhishek-Dev09, line 43: This line will be removed

Abhishek-Dev09-zz:
> Can we put these in `etc/scripts/` like in nexB/scancode-toolkit#2118?

Yes, you can run from anywhere no problem

etc/scripts/deps_archive.py
```diff
@@ -0,0 +1,87 @@
...
+from __future__ import print_function
+
+import argparse
+from commoncode.system import on_windows
+import os
+from shutil import make_archive
+import sys
+
+
+def generate_OS_archive(input_dir, output):
```
pombredanne, line 36: Can you use all lowercase function names?

etc/scripts/github_release.py
```diff
@@ -0,0 +1,158 @@
...
+
+    parser.add_argument(
+        "--body_string",
+        help="Required (or use --body_file): Text describing the release. Ignored if the release already exists.",
+        type=str,
+    )
+
+    parser.add_argument(
+        "--retry_limit",
+        help="The number of times to retry creating/getting the release and/or uploading each file. ",
```
pombredanne, line 131: You have a trailing space in the help string here and most everywhere above.

etc/scripts/github_release.py
```diff
@@ -0,0 +1,158 @@
...
+import os
+import subprocess
+import sys
+
+python_version = str(sys.version_info[0]) + str(sys.version_info[1])
+py_abi = "{0}cp{1}{0}".format("*", python_version)
+
+
+def release_asset(token, tag, repo, body, user, limit, file):
+    """
```
pombredanne, line 41: Is `file` the best name? Is there a better name for that? This is a directory and one that contains assets?
 Suggest something more explicit

etc/scripts/github_release.py
```diff
@@ -0,0 +1,158 @@
...
+    Release .whl,.ABOUT,.NOTICE,.LICENSE to github repository from 
+    given directory.
+    """
+
+    os.environ["GITHUB_TOKEN"] = token
```
pombredanne, line 46: Why do you do that?

etc/scripts/github_release.py
```diff
@@ -0,0 +1,158 @@
...
+        or (
+            fnmatch.fnmatchcase(files, "*tar.gz*")
+            and not fnmatch.fnmatchcase(files, "*py2-ipaddress-3.4.1.tar.gz*")
+        )
+    ]
+    for deps in dependencies:
+        print(deps)
+        subprocess.run(
+            [
+                "python3",
```
pombredanne, line 62: What about calling this as a function instead of spawning a process?

etc/scripts/deps_archive.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    Generate an archive for dependencies for specific OS and
+    given version of python by taking directory as an input.
+    """
```
pombredanne, line 40: Can you remind me why and how this archive would be used?

etc/scripts/deps_archive.py
```diff
@@ -0,0 +1,87 @@
...
+
+    root_dir = os.path.expanduser(os.path.join("~", input_dir))
+    output_dir = os.path.expanduser(os.path.join("~", output))
```
pombredanne, line 43: why do you force these to be relative to my home directory? use abspath and expanduser, but do NOT join with `os.path.join("~"`

etc/scripts/github_release.py
Abhishek-Dev09: Github token is used as credential for accessing the repository. And github token must required by api i have used.

etc/scripts/deps_archive.py
Abhishek-Dev09: It will generate an archive for each os and each python. For example - macOS_py36.tar.gz, win3_py36.zip, win64_py.36zip, linux_py36.tar.gz . For more info see PR description

etc/scripts/deps_archive.py
Abhishek-Dev09: because it is recommended in python documentation of make_archive api.

etc/scripts/github_release.py
Abhishek-Dev09: FYI,` github token` is used as credential for repo where you release the asset. It required by github-release-retry api that you have recommended me to use in #2071

etc/scripts/github_release.py
Abhishek-Dev09: removing now

etc/scripts/github_release.py
Abhishek-Dev09: i think this is fine because `github_release_retry` is used in shell.

etc/scripts/deps_archive.py
Abhishek-Dev09: Doing now

etc/scripts/github_release.py
Abhishek-Dev09: done

etc/scripts/github_release.py
pombredanne: O, fair enough! :+1:

etc/scripts/deps_archive.py
pombredanne: I still do not know what these archives will be used for.... Are you saying that you plan to upload one such archive per OS/Python/scancode version to a GH release?
That archive will not be usable as a pip find-link then. And if there are dupes between OSes... that is going to be problematic. The archives we want to upload would be something built with `etc/release/` scripts instead... e.g. full sdist with a streamlined thirparty

etc/scripts/deps_archive.py
pombredanne: I am not convinced we need these `make_archive` and in anycase we do not want any `os.path.join("~"`... there if we were to create archives

etc/scripts/deps_archive.py
Abhishek-Dev09: But you said earlier `One archive for each OS/Python combo should be created and each should have its exact set of pinned requirements included in the built archive, no more and no less.`

pombredanne: Yes, but I would prefer to have the `sdist` being use for these archives that's using a proper manifest and does not include things we do not want there. For this you should expand on the release script instead

Abhishek-Dev09: FYI , this will give you 1 directory along with tar.gz ,if you are on linux /macOS.
if you are on window, you get 1 directory  along with zip.

BTW i got directory of wheels of 30.1 MB and tar.gz of 28 MB on my macOS.

etc/scripts/deps_archive.py
```diff
@@ -0,0 +1,110 @@
...
+#  ScanCode should be considered or used as legal advice. Consult an Attorney
+#  for any legal advice.
+#  ScanCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/scancode-toolkit/ for support and download.
+
+from __future__ import absolute_import
+from __future__ import print_function
+
+import argparse
+from commoncode.system import on_windows
```
pombredanne, line 30: Sort your imports: commoncode is not from the stdlib, but our own code and should go in a separate block.
See for instance https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/cargo.py#L36

etc/scripts/deps_archive.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def main_with_args(args: str) -> None:
+    parser = argparse.ArgumentParser(
+        description="""Creates a archive for specific OS and specific python.
+EXAMPLE:
+deps_archive.py \\
+  --input thirdparty \\
+  --req requirements.txt
+  --output_file macOS_py36 \\
```
pombredanne, line 71: Please remove that as it does help readability.

etc/scripts/deps_archive.py
```diff
@@ -0,0 +1,108 @@
...
+            '-r',
+            requirement,
+            '--dest',
+            archive_name,
+        ]
+    run(pip_agrs)
+    root_dir = os.path.abspath(archive_name)
+    output_dir = os.path.abspath(archive_name)
+    if on_windows:
+        make_archive(output_dir, 'zip', root_dir)
```
pombredanne, line 60: I am stil not clear about what these archives wrt. to what we need as `sdist`

Abhishek-Dev09: here you get all wheel/sdist deps packed like sdist file that required for ur python/os.

etc/scripts/github_release.py
```diff
@@ -27,127 +27,138 @@
 from __future__ import print_function
 
 import argparse
-from fnmatch import fnmatchcase
```
Abhishek-Dev09: why you remove this?

etc/scripts/github_release.py
```diff
@@ -27,127 +27,138 @@
...
+from pathlib import Path
 import os
-from subprocess import run
 import sys
 
+from github_release_retry import github_release_retry as grr
+
 from commoncode.fileutils import resource_iter
 
-python_version = str(sys.version_info[0]) + str(sys.version_info[1])
```
Abhishek-Dev09: Why remove this section ,it will help you filter the python 3 files if your directory contains multiple python deps.

etc/scripts/github_release.py
```diff
@@ -27,127 +27,138 @@
...
-    os.environ['GITHUB_TOKEN'] = token
-    thirdparty = list(resource_iter(asset_dir, with_dirs=False))
-    dependencies = [
-        files
-        for files in thirdparty
-        if fnmatchcase(files, '*py3*')
-        or fnmatchcase(files, py_abi)
-        or (
-            fnmatchcase(files, '*tar.gz*')
-            and not fnmatchcase(files, '*py2-ipaddress-3.4.1.tar.gz*')
```
Abhishek-Dev09: It will upload unwanted file also

etc/scripts/github_release.py
```diff
@@ -27,127 +27,138 @@
...
+            'Required unless you set the GITHUB_TOKEN environment variable as an alternative. '
+            'See for details: https://github.com/settings/tokens and '
+            'https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token'
+        )
+
     parser.add_argument(
-        '--body-string',
-        help='Required : Text describing the release. Ignored if the release already exists.',
+        '--token',
+        help=TOKEN_HELP,
```
Abhishek-Dev09, line 121: great detail here.

etc/scripts/github_release.py
```diff
@@ -27,127 +27,138 @@
...
-            body_string,
-            '--retry_limit',
-            retry_limit,
-            deps
-        ]
-        run(github_args)
-
-
-def main_with_args(args: str) -> None:
+    api = grr.GithubApi(
```
Abhishek-Dev09, line 64: Either you call API or cmd . result will be same.

etc/scripts/github_release.py
```diff
@@ -27,127 +27,138 @@
...
     )
 
     parser.add_argument(
-        '--retry-limit',
-        help='The number of times to retry creating/getting the release and/or uploading each file.',
+        '--description',
+        help='Text description for the release. Ignored if the release exists.',
         type=str,
-        default='10',
+        required=False,
```
Abhishek-Dev09, line 130: no need to mention that

etc/scripts/github_release.py
```diff
@@ -27,127 +27,138 @@
...
+        )
+
     parser.add_argument(
-        '--body-string',
-        help='Required : Text describing the release. Ignored if the release already exists.',
+        '--token',
+        help=TOKEN_HELP,
         type=str,
-        required=True,
+        required=False,
```
Abhishek-Dev09, line 123: no need to mention, by default it is false

etc/scripts/github_release.py
```diff
@@ -27,127 +27,138 @@
...
+        tag_name=args.tag_name,
+        description=args.description,
+        retry_limit=args.retry_limit,
+        token=token,
+        directory=args.directory,
+    )
 
 
-def main() -> None:
+def main():
```
Abhishek-Dev09, line 161: Why you remove anotations?

Issue #2087, opened 2020-06-23
Dev utility: build and download the sdist and wheels of a single package and all its dependencies


pombredanne:
As a developer of ScanCode, I want to download and or build as needed the wheels and sources of a given package or package/version. For this I expect to call a command line tool that would accept a package name or name==version as an argument and a target directory where the built wheels and tarballs would be downloaded.

As an output of this command, I expect to find in the target directory:

1. pre-built wheels
1.1. either one for each OS/Python version combo when the wheel contains native  
1.2. or one for each Python version if the wheel is Python-specific but not OS-specific
 1.3 or one for each OS if the wheel is OS-specific but not Python-specific
 
2. the corresponding source tarball or zip for each wheel (unless the wheel is a pure Python wheel that contains all its source code in which case this may not be needed)

3. one .ABOUT file for each wheel or source tarball/zip with all found details and URL
  NB: Since there is no place to download that from , it may need to be created from scratch as a placeholder empty file? TBD

4. corresponding NOTICE and LICENSE files
  NB: Since there is no place to download that from, it may need to be created from scratch as a placeholder empty file? TBD

pombredanne:
@Abhishek-Dev09 think about:
1. pip download and its flags to select a proper source and binary version such as   --platform,  --python-version,  --implementation  --abi --no-binary --only-binary
2. what you are doing with the pombreda/thirdparty* repos ... for the build that would need to be eventually automated and/or documented... or you could use romp? TBD

pombredanne:
See #2117

pombredanne:
This is going to be using the utilities listed in #2017 such a romp or https://github.com/pombreda/thirdparty-manylinux/ and https://github.com/pombreda/thirdparty
1. we need to provide a list of packages that need to be built as an input and IDEALLY that should be the same as our requirements file
2. the utilities either need to upload directly to our upcoming PyPI "GitHub release" repo at https://github.com/nexB/scancode-toolkit/issues/2068 OR would fetch these locally and will be uploaded after that to the PyPI "GitHub release" using the github_release script in https://github.com/nexB/scancode-toolkit/pull/2117
3. other files added such ABOUT files and license would be uploaded to the PyPI "GitHub release" using the github_release script in https://github.com/nexB/scancode-toolkit/pull/2117

pombredanne:
[...]

PR #2118, opened 2020-07-20
Automate the creation of requirement.txt and fetch deps

### LINKED ISSUE #2069: Dev utility: Create script to generate requirements with hashes
Start with support for Python 3.6 on the 4 OS variants.: win32, win64, linux 64, mac
Structure of freeze_and_update_reqs.py

- Input -a directory (such as thirdparty) and credentials to GH API and a github repo/release to use

- Function 

1.  It will create or update a requirement.txt with hashes from the dir 

2. It will upload all the files in the dir to the Github release

Abhishek-Dev09:
By running this script, we will able to create and update the requirements.txt by providing the input directory like thirdparty . 

This script will fetch dependencies with all corresponding files(.whl,.tar.gz,.LICENCE,.NOTICE,.ABOUT) from the given target . This feature will ask  ScanCode's developer to provide dependencies as package_name,version and target directory where the built wheels and tarballs would be fetched.

Fixes #2069 #2087


etc/scripts/freeze_and_update_reqs.py
```diff
@@ -0,0 +1,133 @@
...
+    dependencies = [
+        files
+        for files in thirdparty
+        if fnmatch.fnmatchcase(files, "*py3*")
+        or fnmatch.fnmatchcase(files, py_abi)
+        or (
+            fnmatch.fnmatchcase(files, "*tar.gz*")
+            and not fnmatch.fnmatchcase(files, "*py2-ipaddress-3.4.1.tar.gz*")
+        )
+    ]
```
pombredanne, line 56: This code snippet is the same as in your other script? if so may be you could have it only once in a shared module?

etc/scripts/freeze_and_update_reqs.py
```diff
@@ -0,0 +1,133 @@
...
+    if not (os.path.isdir("required_deps")):
+        os.mkdir("required_deps")
```
pombredanne, line 58: Why do you create that directory?

etc/scripts/scancode.py
```diff
@@ -0,0 +1,128 @@
+# -*- coding: utf-8 -*-
```
pombredanne, line 1: Please do not name this scancode ;) that's confusing!

etc/scripts/scancode.py
```diff
@@ -0,0 +1,128 @@
...
+    ]
+    if dependency:
+        whl = [
+            files for files in dependency if files.endswith(".whl")
+        ]  ## There are multiple version of same package So list of wheel will be considered.
+        sdist = [files for files in dependency if files.endswith(".tar.gz")]
+        about = [files for files in dependency if files.endswith(".ABOUT")]
+        notice = [files for files in dependency if files.endswith(".NOTICE")]
+        license = [files for files in dependency if files.endswith(".LICENSE")]
+        print(*whl, sep="\n")
```
pombredanne, line 56: why do you print all that ?

thirdparty/setuptools-41.2.0-py2.py3-none-any.whl.ABOUT
```diff
@@ -0,0 +1,15 @@
+about_resource: setuptools-41.2.0-py2.py3-none-any.whl
```
pombredanne, line 1: We should already have setuptools..... good catch. But please do not commit it there, craft another PR instead just for that.

etc/scripts/freeze_and_update_reqs.py
Abhishek-Dev09: To filter the dependencies of specific python version we want . Don't worry it will deleted later by scripts itself.

etc/scripts/scancode.py
Abhishek-Dev09: Suggest the better name.

etc/scripts/scancode.py
Abhishek-Dev09: Because scancode's developer want to fetch the deps in target directory. It will know developer by print the info of deps.

thirdparty/setuptools-41.2.0-py2.py3-none-any.whl.ABOUT
Abhishek-Dev09: Don't worry, i already mentioned in PR description.

etc/scripts/freeze_and_update_reqs.py
Abhishek-Dev09: Only some part same. if it used frequently later i will make the module.

etc/scripts/scancode.py
Abhishek-Dev09: `deps_info.py`

etc/scripts/freeze_and_update_reqs.py
pombredanne: OK, so please add a comment explaining that... and also this may not be created where you want. Make sure you use a well known location or better use some temp directory instead since this is temporary

etc/scripts/freeze_and_update_reqs.py
```diff
@@ -0,0 +1,133 @@
...
+        [
+            "pip-compile",
+            "--generate-hashes",
+            "--find-links",
+            "required_deps",
+            "--upgrade",
+            "--output-file",
+            output_file,
+            "--verbose",
+            "--upgrade-package",
```
pombredanne, line 71: why do you do this? we never want to do some unwanted `upgrade-package` IMHO and what is `"package_name",` below about?

etc/scripts/freeze_and_update_reqs.py
```diff
@@ -0,0 +1,133 @@
...
+    parser.add_argument(
+        "--output",
+        help="Output file name. Required if more than one input file is given. Will be derived from input file otherwise. ",
+        type=str,
+        default="requirements.txt",
+    )
+
+    parser.add_argument(
+        "--upgrade",
+        help="Upgrade all dependencies to new version. ",
```
pombredanne, line 108: nit: you have trailing whitespaces here and elsewhere

etc/scripts/freeze_and_update_reqs.py
Abhishek-Dev09: you mean to say `bin`

etc/scripts/freeze_and_update_reqs.py
```diff
@@ -0,0 +1,133 @@
...
+
+python_version = str(sys.version_info[0]) + str(sys.version_info[1])
+py_abi = "{0}cp{1}{0}".format("*", python_version)
+
+
+def generate_req_text(input_dir, output_file=False, package_name=False):
+    """
+    Generate a requirement.txt file of all dependencies present in thirdparty.
+    """
+    thirdparty = list(resource_iter(input_dir, with_dirs=False))
```
pombredanne, line 45: The list() wrapper is not needed here

etc/scripts/freeze_and_update_reqs.py
```diff
@@ -0,0 +1,133 @@
...
+import shutil
+import sys
+
+python_version = str(sys.version_info[0]) + str(sys.version_info[1])
+py_abi = "{0}cp{1}{0}".format("*", python_version)
+
+
+def generate_req_text(input_dir, output_file=False, package_name=False):
+    """
+    Generate a requirement.txt file of all dependencies present in thirdparty.
```
pombredanne, line 43: and what are the args?
this should IMHO be something such as:
```
Generate a requirement file at `output_file` of all dependencies wheels and sdists present in the `input_dir` directory.
If a `package_name` is provided it will be updated to its latest version.
```
though I am not sure this is true for package_name
Also `output_file=False, package_name=False` is weird. These should be None... and output_file should always be required

Also what is the base used there? the setup.py or some requirements.in file? I think we will need a requirements.in file

etc/scripts/scancode.py
pombredanne: that's too verbose by default. Hide the prints behind an option

thirdparty/setuptools-41.2.0-py2.py3-none-any.whl.ABOUT
pombredanne: I get that, but please provide a different PR for adding that back (and FYI https://github.com/zopefoundation/zc.lockfile/issues/22 )

etc/scripts/freeze_and_update_reqs.py
Abhishek-Dev09: Base used here : setup.py and no need of requirements.in file. And by default output_file is `requirements.txt` and it is default arguments.

etc/scripts/freeze_and_update_reqs.py
Abhishek-Dev09: will removed that

etc/scripts/freeze_and_update_reqs.py
Abhishek-Dev09: done

etc/scripts/scancode.py
Abhishek-Dev09: it shows ONLY .ABOUT/.NOTICE/.LICENCE/.whl/.tar.gz

etc/scripts/freeze_and_update_reqs.py
Abhishek-Dev09: this feature alow you to upgrade specfic package. See help by running this arg.

thirdparty/setuptools-41.2.0-py2.py3-none-any.whl.ABOUT
Abhishek-Dev09: removed and PR https://github.com/nexB/scancode-toolkit/pull/2123

etc/scripts/deps_info.py
```diff
@@ -0,0 +1,128 @@
...
+
+import argparse
+import fnmatch
+from commoncode.fileutils import resource_iter
+import sys
+
+
+def search_package(package_name, target, version):
+    """
+    Search specific package in given directory with all corresponding files.
```
pombredanne, line 37: This just print things, correct? It do not do nor return anything?

etc/scripts/deps_info.py
```diff
@@ -0,0 +1,128 @@
...
+        else:
+            print("Corresponding .NOTICE does not exits in target\n")
+        print("\n")
+    else:
+        print("Specified package does not exist\n")
+
+
+def main_with_args(args: str) -> None:
+    parser = argparse.ArgumentParser(
+        description="""Fetch a specific package with version in given target like thirdparty by default.
```
pombredanne, line 84: Thsi command seems if fetches nothing at all, does it?

etc/scripts/freeze_and_update_reqs.py
```diff
@@ -0,0 +1,134 @@
...
+import shutil
+import sys
+
+python_version = str(sys.version_info[0]) + str(sys.version_info[1])
+py_abi = "{0}cp{1}{0}".format("*", python_version)
+
+
+def generate_req_text(input_dir, output_file=None, package_name=None):
+    """
+    Generate a requirement file at `output_file`(by default requirements.txt) of all dependencies wheels and sdists present in the `input_dir` directory.
```
pombredanne, line 43: Please wrap this. Also if you want a default of `(by default requirements.txt)`... then set the default  in the function definition and not as None

etc/scripts/freeze_and_update_reqs.py
pombredanne: Please use a temp dir instead

etc/scripts/freeze_and_update_reqs.py
pombredanne: I known that but what is `"package_name"` I know of no package whose name is `"package_name"` 
If this is some kind of magic, explain it in comments, do not make me chase down the help of that command

pombredanne: In any case why would I want to upgrade a package there? Please elaborate and explain in comments

etc/scripts/freeze_and_update_reqs.py
```diff
@@ -0,0 +1,134 @@
...
+            "--output-file",
+            output_file,
+            "--verbose",
+            "--upgrade-package",
+            "package_name",
+            "--pip-args",
+            "--no-index",
+        ]
+    )
+    shutil.rmtree("required_deps")
```
pombredanne, line 78: I would rather you use a temp file with a "with" context manager to avoid leaving dangling files

etc/scripts/freeze_and_update_reqs.py
```diff
@@ -0,0 +1,134 @@
...
+    parser.add_argument(
+        "--deps_directory",
+        help="Required: Thirdparty Dependencies directory to be archived.",
+        type=str,
+        required=True,
+    )
+
+    parser.add_argument(
+        "--output",
+        help="Output file name. Required if more than one input file is given. Will be derived from input file otherwise.",
```
pombredanne, line 102: Avoid implicit, derived name. It is preferred to always require an input instead

etc/scripts/freeze_and_update_reqs.py
```diff
@@ -0,0 +1,134 @@
...
+        type=str,
+        default=None,
+    )
+
+    args = parser.parse_args()
+
+    tpdir = args.deps_directory
+    output_file = args.output
+    package_name = args.upgrade_package
+    upgrade = args.upgrade or None
```
pombredanne, line 125: This argument is not used at all... so what is it for?

etc/scripts/deps_info.py
Abhishek-Dev09: Yes, offcourse

etc/scripts/deps_info.py
Abhishek-Dev09: it description of function , you know this thing by running `--help` option.

etc/scripts/freeze_and_update_reqs.py
Abhishek-Dev09: upgrade is for upgrading all packages whereas upgrade package is for specific package,

etc/scripts/freeze_and_update_reqs.py
Abhishek-Dev09: It is not a file , it is directory

etc/scripts/freeze_and_update_reqs.py
Abhishek-Dev09: done

etc/scripts/freeze_and_update_reqs.py
Abhishek-Dev09: Package name may be intbitset, banal, urlpy . Anything deps that exist in thirdparty

etc/scripts/freeze_and_update_reqs.py
Abhishek-Dev09: Already set as defaults.

etc/scripts/freeze_and_update_reqs.py
Abhishek-Dev09: pip-tools uses `--output-file` for same purpose why should i not not use `--output`

pombredanne: My point is to avoid using a derived name. Instead make this always required and not implied.

etc/scripts/freeze_and_update_reqs.py
pombredanne: But where do you use it?

pombredanne: Is not an argument used anywhere in your functions

Abhishek-Dev09: It is an option just like ON/OFF . It is used here only , i have tested it it works, if you want i will paste output here.

etc/scripts/freeze_and_update_reqs.py
Abhishek-Dev09: Ok , your call i changing the option `--output` to `--req` . And making always required.

etc/scripts/freeze_and_update_reqs.py
pombredanne: You need to test this. The string `"package_name",` will NOT resolve to the variable `package_name`
`"package_name",` is of itself NOT the name a package...

etc/scripts/freeze_and_update_reqs.py
```diff
@@ -0,0 +1,135 @@
...
+            "--generate-hashes",
+            "--find-links",
+            "temp dir",
+            "--upgrade",
+            "--output-file",
+            req_file,
+            "--verbose",
+            "--allow-unsafe",
+            "--upgrade-package",
+            "package_name",
```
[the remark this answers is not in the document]
Abhishek-Dev09: Oh, yeah it is function parameter through arguments .

etc/scripts/deps_info.py
```diff
@@ -0,0 +1,128 @@
...
+EXAMPLE:
+scancode.py \\
```
Abhishek-Dev09, line 86: deps

etc/scripts/deps_info.py
```diff
@@ -0,0 +1,128 @@
...
+from __future__ import absolute_import
+from __future__ import print_function
+
+import argparse
+import fnmatch
+from commoncode.fileutils import resource_iter
+import sys
+
+
+def search_package(package_name, target, version):
```
pombredanne, line 35: Please `version=None` since this is optional

etc/scripts/deps_info.py
```diff
@@ -0,0 +1,128 @@
...
+#  ScanCode is a free software code scanning tool from nexB Inc. and others.
+#  Visit https://github.com/nexB/scancode-toolkit/ for support and download.
+
+from __future__ import absolute_import
+from __future__ import print_function
+
+import argparse
+import fnmatch
+from commoncode.fileutils import resource_iter
+import sys
```
pombredanne, line 32: Sort imports

etc/scripts/deps_info.py
pombredanne, line 37: Use  instead:
```
Search `package_name` (with an optional `version`) in the `target` directory. Print results on screen.
```

etc/scripts/deps_info.py
pombredanne: It should be instead:
```
Search PACKAGE_NAME (with an optional VERSION_OF_PACKAGE) in the TARGET_DIR directory. 
Print results on screen.
```

etc/scripts/deps_info.py
```diff
@@ -0,0 +1,128 @@
...
+    parser.add_argument(
+        "--fetch",
+        help="Required: Specific Dependencies  to be fetched.",
+        type=str,
+        required=True,
+    )
+
+    parser.add_argument(
+        "--target",
+        help=" a target directory where the built wheels and tarballs would be fetched.",
```
pombredanne, line 103: Use this:
`"Directory to search for package wheels and tarballs. [default: thirdparty]"`

etc/scripts/deps_info.py
```diff
@@ -0,0 +1,128 @@
...
+""",
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+    )
+
+    parser.add_argument(
+        "--fetch",
+        help="Required: Specific Dependencies  to be fetched.",
+        type=str,
+        required=True,
+    )
```
pombredanne, line 99: Consider this instead:
```

    parser.add_argument(
        "--name",
        help="Name of the package to search in the TARGET_DIR directory",
        type=str,
        required=True,
    )
```

etc/scripts/deps_info.py
```diff
@@ -0,0 +1,128 @@
...
+    parser.add_argument(
+        "--target",
+        help=" a target directory where the built wheels and tarballs would be fetched.",
+        type=str,
+        default="thirdparty",
+    )
+
+    parser.add_argument(
+        "--version",
+        help="Specify version of dependencies to be fetched.",
```
pombredanne, line 110: Use this `"Optional package version to search."`

etc/scripts/deps_info.py
```diff
@@ -0,0 +1,128 @@
...
+        type=str,
+        default=None,
+    )
+
+    args = parser.parse_args()
+
+    package_name = args.fetch
```
pombredanne, line 117: rename fetch everywhere to package_name

etc/scripts/deps_info.py
```diff
@@ -0,0 +1,128 @@
+# -*- coding: utf-8 -*-
```
pombredanne, line 1: Consider instead `search-package.py`

etc/scripts/deps_info.py
```diff
@@ -0,0 +1,128 @@
...
+
+def main_with_args(args: str) -> None:
+    parser = argparse.ArgumentParser(
+        description="""Fetch a specific package with version in given target like thirdparty by default.
+EXAMPLE:
+scancode.py \\
+  --fetch PACKAGE_NAME \\
+  --target TARGET_DIR \\
+  --version VERSION_OF_PACKAGE \\
+""",
```
pombredanne, line 90: Please remove this example as this is not helping with readability of the help and is redundant with argparse's printouts

etc/scripts/deps_info.py
```diff
@@ -0,0 +1,128 @@
...
+
+    parser.add_argument(
+        "--fetch",
+        help="Required: Specific Dependencies  to be fetched.",
+        type=str,
+        required=True,
+    )
+
+    parser.add_argument(
+        "--target",
```
pombredanne, line 102: Use `directory` instead

etc/scripts/freeze_and_update_reqs.py
pombredanne: By wrap, wrap the very long lines.

etc/scripts/freeze_and_update_reqs.py
```diff
@@ -0,0 +1,134 @@
...
+    ]
+    if not (os.path.isdir('temp dir')):
+        os.mkdir('temp dir')
+    for deps in dependencies:
+        copy(deps, 'temp dir')
+    pip_args = [
+        'pip-compile',
+        '--generate-hashes',
+        '--find-links',
+        'temp dir',
```
pombredanne, line 65: I mean to use this https://docs.python.org/3/library/tempfile.html?highlight=tempfile#module-tempfile 
No a dir name `temp dir`

etc/scripts/freeze_and_update_reqs.py
```diff
@@ -0,0 +1,134 @@
...
+        '--allow-unsafe',
+        '--pip-args',
+        '--no-index',
+    ]
+    if upgrade:
+        pip_args.append('--upgrade')
+    if package_name:
+        pip_args.extend(['--upgrade-package', package_name])
+    run(pip_args)
+    rmtree('temp dir')
```
pombredanne, line 77: With a proper https://docs.python.org/3/library/tempfile.html?highlight=tempfile#module-tempfile you will NOT need to rm anything, with will be handled for you if you use a context manager.
such as this example from the Python doc:
```
# create a temporary directory using the context manager
>>> with tempfile.TemporaryDirectory() as tmpdirname:
...     print('created temporary directory', tmpdirname)
 .... do your things here
```

etc/scripts/freeze_and_update_reqs.py
```diff
@@ -0,0 +1,134 @@
...
+def main_with_args(args: str) -> None:
+    parser = argparse.ArgumentParser(
+        description="""Generate a requirement file at `output_file`(by default requirements.txt) of all dependencies wheels and sdists present in the `input_dir` directory.
+    If a `package_name` is provided it will be updated to its latest version.
+EXAMPLE:
+freeze_and_update_reqs.py \\
+  --deps_directory DEPS_DIRECTORY \\
+  --output OUTPUT \\
+  --upgrade_package PACKAGE_NAME \\
+""",
```
pombredanne, line 89: Please remove this example.

etc/scripts/freeze_and_update_reqs.py
```diff
@@ -0,0 +1,134 @@
...
+    )
+
+    args = parser.parse_args()
+
+    tpdir = args.deps_directory
+    requirement_file = args.requirement
+    package_name = args.upgrade_package
+    upgrade = args.upgrade or False
+    generate_req_text(tpdir, requirement_file, package_name, upgrade)
+
```
pombredanne, line 127: prefer named keyword arguments for example:
`(tpdir=tpdir, requirement_file=requirement_file, package_name=package_name, upgrade=upgrade)
But do not use these names... use the same variable names on both sides:  e.g. if you use `args.deps_directory`  then use `deps_directory` and not `tpdir`

etc/scripts/freeze_and_update_reqs.py
```diff
@@ -0,0 +1,134 @@
...
+        '--deps_directory',
+        help='Required: Thirdparty Dependencies directory to be archived.',
+        type=str,
+        required=True,
+    )
+
+    parser.add_argument(
+        '--requirement',
+        help='Requirement file name. Required if more than one input file is given. Will be derived from input file otherwise.',
+        type=str,
```
pombredanne, line 103: Use instead this `help='Requirement file name.'`

etc/scripts/freeze_and_update_reqs.py
```diff
@@ -0,0 +1,131 @@
...
+#  Visit https://github.com/nexB/scancode-toolkit/ for support and download.
+
+from __future__ import absolute_import
+from __future__ import print_function
+
+import argparse
+from fnmatch import fnmatchcase
+import os
+from subprocess import run
+from shutil import copy, rmtree
```
pombredanne, line 33: nitpicking: your imports are not sorted. Try to use https://pypi.org/project/isort/

etc/scripts/freeze_and_update_reqs.py
```diff
@@ -0,0 +1,131 @@
...
+from commoncode.fileutils import resource_iter
+
+python_version = str(sys.version_info[0]) + str(sys.version_info[1])
+py_abi = '{0}cp{1}{0}'.format('*', python_version)
+
+
+def generate_req_text(find_links, req_file, package_name=None, upgrade=False):
+    """
+    Generate a requirement file at `req_file` of all dependencies wheels and 
+    sdists present in the `input_dir` directory.If a `package_name` is provided
```
pombredanne, line 46: There is no input_dir? 
`...sdists present in the `input_dir` directory...`
Do you mind doing an extra review pass on this PR and #2117 to make sure the docstrings are well synced with the actual function arguments and what the function does>

Abhishek-Dev09: I think, this PR has reviewed many times.

etc/scripts/freeze_and_update_reqs.py
Abhishek-Dev09: Oh , i missed that

etc/scripts/freeze_and_update_reqs.py
Abhishek-Dev09: Done

PR #2133, opened 2020-07-26
Add supoort for go.sum dependency lock files #2080

### LINKED ISSUE #2080: Parse Go(go.mod, go.sum) files
This is apart of GSoC 2020 phase 2.

We need to add the ability to parse Go(go.mod, go.sum) files and have them a part of a --package scan.

rpotter12:
Fixes #2080 

### Tasks


src/packagedcode/go_sum.py
```diff
@@ -0,0 +1,126 @@
...
+        """
+        with io.open(location, encoding='utf-8', closefd=True) as data:
+            lines = data.readlines()
+
+        gosum_data = []
+
+        for line in lines:
+            parsed_dep = self.parse_dep_type1(line)
+            if parsed_dep:
+                dep_data = [parsed_dep.group('namespace'), parsed_dep.group('name'), parsed_dep.group('version'), parsed_dep.group('checksum')]
```
pombredanne, line 118: May be you can wrap this long line? and below too?

src/packagedcode/go_sum.py
```diff
@@ -0,0 +1,126 @@
...
+                gosum_data.append(dep_data)
+            else:
+                parsed_dep = self.parse_dep_type2(line)
+                if parsed_dep:
+                    dep_data = [parsed_dep.group('namespace'), parsed_dep.group('name'), parsed_dep.group('version'), parsed_dep.group('checksum')]
+                    gosum_data.append(dep_data)
```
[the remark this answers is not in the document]
pombredanne: Also why not using returning an instance of self (which could have all the fields for ns, name, version and checksum and make the `def parse_gosum(self, location):` function a classmethod factory `def from_gosum(cls, location):`?
That would be more explicit IMHO than carrying around plain tuples of "gosum_data"... and ` build_gosum_package()` could then accept instead a Gosum object?

src/packagedcode/golang.py
```diff
@@ -141,3 +147,27 @@ def build_gomod_package(gomod_data):
         homepage_url=homepage_url,
         dependencies=package_dependencies
     )
+
+
+def build_gosum_package(gosum_data):
+    """
+    Return a Package object from a go.mod file or None.
```
pombredanne, line 154: > Return a Package object from a go.mod file or None.

That looks like copied from go.mod

src/packagedcode/golang.py
```diff
@@ -141,3 +147,27 @@ def build_gomod_package(gomod_data):
...
+    for namespace, name, version, _ in gosum_data:
+        package_dependencies.append(
+            models.DependentPackage(
+                purl=PackageURL(
+                    type='golang',
+                    namespace=namespace,
+                    name=name
+                ).to_string(),
+                requirement=version,
+                scope='exclude',
```
pombredanne, line 166: Are you sure?

src/packagedcode/golang.py
```diff
@@ -141,3 +147,27 @@ def build_gomod_package(gomod_data):
...
+                is_runtime=True,
+                is_optional=False,
+                is_resolved=False,
```
pombredanne, line 169: I though that the go.sum entire purpose be to store fully resolved deps?

src/packagedcode/golang.py
```diff
@@ -84,6 +85,11 @@ def recognize(cls, location):
             gomod_obj = GoMod()
             gomod_data = gomod_obj.parse_gomod(location)
             yield build_gomod_package(gomod_data)
+        
+        if fileutils.file_name(location).lower() == 'go.sum':
```
pombredanne, line 89: elif may be?

src/packagedcode/golang.py
pombredanne: Also when do you return None?

src/packagedcode/go_sum.py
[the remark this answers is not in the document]
rpotter12: > Why append a tuple vs. yielding it?

@pombredanne we didn't append any tuple here. We have created a list of group after parsing from regex and append that in `gosum_data`

src/packagedcode/golang.py
```diff
@@ -84,6 +85,11 @@ def recognize(cls, location):
...
+            gosum_obj = GoSum()
```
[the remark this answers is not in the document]
rpotter12: Nothing much. I have just given a name for this :)

src/packagedcode/go_sum.py
```diff
@@ -0,0 +1,126 @@
...
+if TRACE:
+    import sys
+    logging.basicConfig(stream=sys.stdout)
+    logger.setLevel(logging.DEBUG)
+
+
+@attr.s()
+class GoSum(object):
+    # Regex expressions to parse different types of dependency
+    parse_dep_type1 = re.compile(
```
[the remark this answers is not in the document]
rpotter12: Sure, I will add this in some time :)

src/packagedcode/go_sum.py
```diff
@@ -0,0 +1,126 @@
+
+# Copyright (c) 2019 nexB Inc. and others. All rights reserved.
```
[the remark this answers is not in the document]
rpotter12: :+1:

src/packagedcode/golang.py
rpotter12: Please ignore this for now. I have just created a simple initial file to get review why it is not working :)

src/packagedcode/go_sum.py
rpotter12: sure :+1:

src/packagedcode/golang.py
rpotter12: Please ignore this for now. I have just created a simple initial file to get review why it is not working :)

src/packagedcode/golang.py
rpotter12: Please ignore this for now. I have just created a simple initial file to get review why it is not working :)

src/packagedcode/golang.py
rpotter12: Please ignore this for now. I have just created a simple initial file to get review why it is not working :)

tests/packagedcode/data/golang/gosum/sample1/go.sum
```diff
@@ -0,0 +1,4 @@
+github.com/BurntSushi/toml v0.3.1 h1:WXkYYl6Yr3qBf1K79EBnL4mak0OimBfB0XUf9Vl28OQ=
+github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
```
pombredanne, line 2: How would you handle the dupes here, since they obviously about one and the same?
And also what to do about the checksums?

rpotter12: @pombredanne 
> How would you handle the dupes here, since they obviously about one and the same?

Do we really need to handle dupes here? I think checksums are different so can we can add them differently.

>And also what to do about the checksums?

I was thinking that we should add checksums option in `DependentPackage` in `models.py` :)

src/packagedcode/golang.py
```diff
@@ -70,10 +71,14 @@ class GolangPackage(models.Package):
 
     @classmethod
     def recognize(cls, location):
-        if fileutils.file_name(location).lower() == 'go.mod':
-            gomod_obj = GoMod()
-            gomod_data = gomod_obj.parse_gomod(location)
+        filename = fileutils.file_name(location).lower()
+        if filename == 'go.mod':
+            _obj = GoMod()
+            gomod_data = _obj.parse_gomod(location)
```
majurg, line 77: Why not make this line `gomod_data = GoMod().parse_gomod(location)`? This would keep the code consistent with the lines below.

rpotter12: I thought we change in another PR but we can do this in this PR too :)

rpotter12: @majurg I have updated my PR please review it again :)

src/packagedcode/go_sum.py
```diff
@@ -0,0 +1,133 @@
...
+                parsed_dep = cls.parse_dep_type2(line)
+
+            dep_data = [parsed_dep.group('namespace').strip(),
+                    parsed_dep.group('name').strip(),
+                    parsed_dep.group('version').strip()]
+
+            gosum_data.append(dep_data)
+
+        gosum_data.sort()
+        gosum_data = list(gosum_data for gosum_data, _ in itertools.groupby(gosum_data))
```
pombredanne, line 131: What is this doing?

src/packagedcode/go_sum.py
```diff
@@ -0,0 +1,133 @@
...
+        r'/'
+        r'(?P<name>[^\s]*)'
+        r'(\s)*'
+        r'(?P<version>(.*))'
+        r'(\s)*h1:'
+        r'(?P<checksum>(.*))'
+    ).match
+
+    @classmethod
+    def parse_gosum(cls, location):
```
pombredanne, line 110: What I meant earlier with a class method is for it to be a factory to return GoSum objects that would have actual real fields such as:
```
@attr.s()
class GoSum(object):
    namespace = attr.ib(default=None)
    name = attr.ib(default=None)
    version = attr.ib(default=None)
```
and not return a list of tuples

rpotter12: @pombredanne currently my go.sum parser return list of list which contain namespace, name, version in order like:
if this is the go.sum file:
```
github.com/BurntSushi/toml v0.3.1 h1:WXkYYl6Yr3qBf1K79EBnL4mak0OimBfB0XUf9Vl28OQ=
github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
github.com/cznic/golex v0.0.0-20181122101858-9c343928389c h1:G8zTsaqyVfIHpgMFcGgdbhHSFhlNc77rAKkhVbQ9kQg=
github.com/cznic/golex v0.0.0-20181122101858-9c343928389c/go.mod h1:+bmmJDNmKlhWNG+gwWCkaBoTy39Fs+bzRxVBzoTQbIc=
```
then the return value will be `[['github.com/BurntSushi', 'toml', 'v0.3.1'], ['github.com/cznic', 'golex', 'v0.0.0-20181122101858-9c343928389c']]`

You told me to use classmethod factory, so what output format you want from classmethod factory?? plz give a example

src/packagedcode/go_sum.py
rpotter12: @pombredanne It was handling dupes. This removes the duplicates form the list of lists

src/packagedcode/go_sum.py
pombredanne: I was suggesting to return a list of Gosum objects... I find it always better and easier to read and understand when the code uses objects with named attributes rather than to use nested lists.

NB: I pushed a small doc header to your branch

src/packagedcode/go_sum.py
pombredanne: OK, may you need to find a way to do this in a less clever, but easier to read and understand way? Or at the minimum add some explanatory comment?

tests/packagedcode/data/golang/gosum/sample1/go.sum
pombredanne: I am not that would work: 1. this is very specific to Go and 2. is the checksum really instead a resolved version for a Go module?

pombredanne: Another way to think of this go.sums checksums could be to add a qualifier to the Purl... with key to be determined.

src/packagedcode/go_sum.py
rpotter12: @pombredanne I have updated my PR. I think I have done changes according to you but I am not sure fully. Please review the PR again :)

tests/packagedcode/data/golang/gosum/sample1/go.sum
rpotter12: If everything is looking good in this PR, we can see for checksums in another issue.

src/packagedcode/go_sum.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def parse_gosum(location):
+    """
+    Return a list containing all the go.sum dependency data.
+    """
+    with io.open(location, encoding='utf-8', closefd=True) as data:
+        lines = data.readlines()
+
+    gosum_data = []
```
pombredanne, line 118: Just call this `gosums`: there is no need for a `_data`  suffix and the plural tells that this is some collection

src/packagedcode/go_sum.py
```diff
@@ -0,0 +1,133 @@
...
+
+    for line in lines:
+        parsed_dep = parse_dep_type1(line)
+        if not parsed_dep:
+            parsed_dep = parse_dep_type2(line)
+
+        dep_obj = GoSum(parsed_dep.group('namespace').strip(),
+                parsed_dep.group('name').strip(),
+                parsed_dep.group('version').strip()
```
pombredanne, line 127: that's great... but please use named keyword args and not positional args. ... more or less like this:
```Python
        gosum = GoSum(
            namespace=parsed_dep.group('namespace').strip(),
            name=parsed_dep.group('name').strip(),
            version=parsed_dep.group('version').strip(),
        )
```

rpotter12:
@pombredanne 
> you will want later to do that for the go_mod... and frankly you should not have two modules for that... a single go_mod.py for go.mod and go.sum support would be enough IMHO and would help avoid duplicated code.

Lets open a new ticket for this and will change these modules again and create a new PR for this..

src/packagedcode/go_sum.py
rpotter12: :+1:

src/packagedcode/go_sum.py
rpotter12: Okay...I have updated my PR

src/packagedcode/go_sum.py
```diff
@@ -0,0 +1,134 @@
...
+    logger.setLevel(logging.DEBUG)
+
+
+@attr.s()
+class GoSum(object):
+    namespace = attr.ib(default=None)
+    name = attr.ib(default=None)
+    version = attr.ib(default=None)
+
+    def __init__(self, namespace, name, version):
```
pombredanne, line 82: This method is not needed when you use attrs... it will be automatically created

src/packagedcode/golang.py
```diff
@@ -133,3 +138,28 @@ def build_gomod_package(gomod_data):
         homepage_url=homepage_url,
         dependencies=package_dependencies
     )
+
+
+def build_gosum_package(gosum_objs):
```
pombredanne, line 143: call this `gosums`

src/packagedcode/golang.py
```diff
@@ -133,3 +138,28 @@ def build_gomod_package(gomod_data):
...
+    """
+    Return a Package object from a go.sum file.
+    """
+    package_dependencies = []
+    for obj in gosum_objs:
```
pombredanne, line 148: Avoid using generic variable names like "obj" when possible... here we have a `Gosum` instance, so call it `gosum`

src/packagedcode/golang.py
```diff
@@ -133,3 +138,28 @@ def build_gomod_package(gomod_data):
...
+        package_dependencies.append(
+            models.DependentPackage(
+                purl=PackageURL(
+                    type='golang',
+                    namespace=obj.namespace,
+                    name=obj.name,
+                    version=obj.version
+                ).to_string(),
```
pombredanne, line 156: What about adding this method to Gosum?
```Python
class Gosum(...):
    ....
    @property
    def purl(self):
        return PackageURL(
                    type='golang',
                    namespace=self.namespace,
                    name=self.name,
                    version=self.version
                ).to_string()
```
then your code becomes:
```Python
models.DependentPackage(
    purl=gosum.purl, 
                ....
```

tests/packagedcode/data/golang/gosum/sample1/go.sum
pombredanne: OK, please enter a ticket for that part then.

tests/packagedcode/data/golang/gosum/sample1/output.expected.json
```diff
@@ -0,0 +1,54 @@
...
+        "scope": "dependency",
+        "is_runtime": true,
+        "is_optional": false,
+        "is_resolved": true
+      }
+    ],
+    "contains_source_code": null,
+    "source_packages": [],
+    "purl": null,
+    "repository_homepage_url": "https://pkg.go.dev/None/None",
```
pombredanne, line 50: I think you should be null in this case.

tests/packagedcode/data/golang/gosum/sample2/output.expected.json
```diff
@@ -0,0 +1,46 @@
...
+        "scope": "dependency",
+        "is_runtime": true,
+        "is_optional": false,
+        "is_resolved": true
+      }
+    ],
+    "contains_source_code": null,
+    "source_packages": [],
+    "purl": null,
+    "repository_homepage_url": "https://pkg.go.dev/None/None",
```
pombredanne, line 42: Same as above... this should be None

src/packagedcode/golang.py
```diff
@@ -70,10 +71,13 @@ class GolangPackage(models.Package):
 
     @classmethod
     def recognize(cls, location):
-        if fileutils.file_name(location).lower() == 'go.mod':
-            gomod_obj = GoMod()
-            gomod_data = gomod_obj.parse_gomod(location)
+        filename = fileutils.file_name(location).lower()
+        if filename == 'go.mod':
+            gomod_data = GoMod.parse_gomod(location)
```
pombredanne, line 76: Please avoid `_data` suffix here, it does not bring anything to the reader IMHO

src/packagedcode/golang.py
```diff
@@ -70,10 +71,13 @@ class GolangPackage(models.Package):
...
             yield build_gomod_package(gomod_data)
+        elif filename == 'go.sum':
+            gosum_objs = go_sum.parse_gosum(location)
```
pombredanne, line 79: Same as above avoid bland suffix that are not useful such as `_objs`

rpotter12: :+1:

src/packagedcode/golang.py
rpotter12: :+1:

rpotter12:
@pombredanne I have updated my PR. Please review it again.

src/packagedcode/golang.py
rpotter12: :+1:

src/packagedcode/go_sum.py
```diff
@@ -0,0 +1,138 @@
...
+    r'(\s)*'
+    r'(?P<version>(.*))'
+    r'(\s)*h1:'
+    r'(?P<checksum>(.*))'
+).match
+
+
+def parse_gosum(location):
+    """
+    Return a list containing all the go.sum dependency data.
```
pombredanne, line 117: What about:
```Return a list of GoSum from parsing the go.sum file at `location`.```

Comment (pombredanne) [inline src/packagedcode/go_sum.py:117]:
```suggestion
    Return a list of GoSum from parsing the go.sum file at `location`.
```

src/packagedcode/go_sum.py
```diff
@@ -0,0 +1,138 @@
...
+import io
+import itertools
+import logging
+import re
+
+import attr
+from packageurl import PackageURL
+
+"""
+This modules handles go.sum files from Go.
```
pombredanne, line 37:
```suggestion
This module handles go.sum files from Go.
```

src/packagedcode/golang.py
```diff
@@ -70,24 +71,30 @@ class GolangPackage(models.Package):
...
 
     @classmethod
     def get_package_root(cls, manifest_resource, codebase):
         return manifest_resource.parent(codebase)
 
     def repository_homepage_url(self, baseurl=default_web_baseurl):
-        return '{}/{}/{}'.format(baseurl, self.namespace, self.name)
+        if self.namespace and self.name:
+            return '{}/{}/{}'.format(baseurl, self.namespace, self.name)
+        return None
```
pombredanne, line 89:
```suggestion
```
returning None is not needed here

rpotter12:
@pombredanne I have commit your suggested changes. Please review again.

PR #2151, opened 2020-07-31
Add pip-tools and github_release_retry as req files

Abhishek-Dev09:
These tools can be used as tools for generating requirements.txt


pombredanne:
Can you try to install and run `restview`

PR #2152, opened 2020-08-01
Merge go.sum and go.mod modules in one module

### LINKED ISSUE #2150: Merge go_sum and go_mod modules
### Description

go_sum.py and go_mod.py contains some duplicate code and data is extracted in same format. We should merge these modules to avoid duplicates.

rpotter12:
Signed-off by: Rohit Potter rohitpotter12@gmail.com
'
Fixes #2150

### Tasks


src/packagedcode/go_mod.py
```diff
@@ -31,10 +30,48 @@
...
+                    name=self.name,
+                    version=self.version
+                ).to_string()
+
+    @property
+    def purl_gomod(self):
+        return PackageURL(
+                    type='golang',
+                    namespace=self.namespace,
+                    name=self.name
```
majurg, line 67: And here, gomod does not have version info?

rpotter12: @majurg go.mod file does not contains the version of module. We need to find the another way to get the version info.

majurg: No problem I just wanted to be sure.

rpotter12:
> I think we need tests for this, but the code looks good from here.

@majurg We don't need any test now. We already having tests for go.sum and go.mod. We just have merged the two files. Please review it again.

src/packagedcode/go_mod.py
```diff
@@ -31,10 +30,48 @@
...
+
+@attr.s()
+class GoModules(object):
+    namespace = attr.ib(default=None)
+    name = attr.ib(default=None)
+    version = attr.ib(default=None)
+    module = attr.ib(default=None)
+
+    @property
+    def purl_gosum(self):
```
pombredanne, line 54: How is this different from purl_gomod? IMHO you will get the same whether or not you provide a version

src/packagedcode/go_mod.py
```diff
@@ -31,10 +30,48 @@
...
+logger = logging.getLogger(__name__)
+
+if TRACE:
+    import sys
+    logging.basicConfig(stream=sys.stdout)
+    logger.setLevel(logging.DEBUG)
+
+
+@attr.s()
+class GoModules(object):
```
pombredanne, line 47: Why plural name?

src/packagedcode/go_mod.py
```diff
@@ -31,10 +30,48 @@
...
+    namespace = attr.ib(default=None)
+    name = attr.ib(default=None)
+    version = attr.ib(default=None)
+    module = attr.ib(default=None)
```
pombredanne, line 51: Can you provide an example of a value?

src/packagedcode/go_mod.py
```diff
@@ -73,140 +109,221 @@
...
+
+
+def parse_gomod(location):
+    """
+    Return a dictionary containing all the important go.mod file data.
+    """
+    with io.open(location, encoding='utf-8', closefd=True) as data:
+        lines = data.readlines()
+
+    gomods = {}
```
pombredanne, line 185: Why not use an object instead of a dict?

rpotter12: I was keeping some of the previous format. Now I have updated my PR again. :)

src/packagedcode/go_mod.py
rpotter12: I have provided all the examples in the comments. Please see below in the file. I have provided examples for both types of files.

src/packagedcode/go_mod.py
rpotter12: I was reading https://blog.golang.org/using-go-modules documentation. `Go Modules` was written in that documentation. So I used that name :)

src/packagedcode/go_mod.py
rpotter12: We don't need version number in `purl` because it was mentioned in the go modules that the version number which is written in the go.mod file is the minimum version required. But we need version number for requirements field. So I made it like this.

src/packagedcode/go_mod.py
pombredanne: OK, but your instance still represents a single module object, not many ones?

src/packagedcode/go_mod.py
pombredanne: ok, but if you where to use one function with an argument to `include_version=True` or `False`, you could have a single function and a bit less code?

rpotter12: I added new code. Please check, in that the way you were asking me to do it?

src/packagedcode/go_mod.py
rpotter12: I represents the Single module. I have fixed this.

src/packagedcode/go_mod.py
```diff
@@ -73,140 +105,221 @@
...
+        lines = data.readlines()
+
+    gosums = []
+
+    for line in lines:
+        parsed_dep = parse_dep_type1(line)
+        if not parsed_dep:
+            parsed_dep = parse_dep_type2(line)
+
+        dep_obj = GoModule(
```
pombredanne, line 316: Why not call that just dep? what does the _obj suffix brings you?

src/packagedcode/go_mod.py
```diff
@@ -73,140 +105,221 @@
...
+).match
+
+# dep_type2 example: github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCN....
+parse_dep_type2 = re.compile(
+    r'(?P<namespace>(.*))'
+    r'/'
+    r'(?P<name>[^\s]*)'
+    r'(\s)*'
+    r'(?P<version>(.*))'
+    r'(\s)*h1:'
```
pombredanne, line 297: Since the only difference between type 1 and 2 is this... what about that with a few other refinements:
```Python
get_dependency = re.compile(
    r'(?P<namespace>[^\s]*)'
    r'\/'
    r'(?P<name>[^\s]*)'
    r'\s*'
    r'(?P<version>[^\s]*)'
    r'\s*'
    r'(>P:<is_gomod>/go.mod)?
    r'\s*'
    r'h1:'(?P<checksum>[^\s]*)'
    r'\s*'
).match
```
This may need a few adjustments... but this will also tell you if this is a `go.mod` line (for the checksum, if an when we handle it in the future?)
And by ignoring the spaces everywhere in the capture groups the `.strip()` you used below are no longer needed.

src/packagedcode/go_mod.py
```diff
@@ -73,140 +105,221 @@
...
+        if not parsed_dep:
+            parsed_dep = parse_dep_type2(line)
+
+        dep_obj = GoModule(
+                namespace=parsed_dep.group('namespace').strip(),
+                name=parsed_dep.group('name').strip(),
+                version=parsed_dep.group('version').strip()
+            )
+
+        if dep_obj not in gosums:
```
pombredanne, line 322: Is this a possible case? or is this a side effect you treating lines with and without go.mod the same way?
If the later is true, then using the new proposed regex above and a test such as this may be more explicit. You can also then just yield too, rather than accumulate in a list
```
if parsed_dep.group('is_gomod'):
    continue
```

rpotter12: Since we are not using checksums in the dependency object, so maximum time duplicate values will be present.

src/packagedcode/go_mod.py
rpotter12: I tried this 
```
get_dependency = re.compile(
    r'(?P<namespace>[^\s]*)'
    r'\/'
    r'(?P<name>[^\s]*)'
    r'\s*'
    r'(?P<version>[^\s]*)'
    r'\s*'
    r'(>P:<is_gomod>/go.mod)?'
    r'\s*'
    r'h1:(?P<checksum>[^\s]*)'
    r'\s*'
).match
```
but it was not working. I am suggesting a new change. Maybe this looks good.

src/packagedcode/go_mod.py
rpotter12: I was specifying it more clearly. `dep` also looks good. I have updated my PR. Please review it again.

PR #2153, opened 2020-08-02
Packagedcode to handle Cargo.lock files

### LINKED ISSUE #2084: Parse Rust(cargo.lock) files
This is apart of GSoC 2020 phase 3.

We need to add the ability to parse Rust(cargo.lock) files and have them a part of a --package scan.

rpotter12:
Fixes #2084

### Tasks


src/packagedcode/cargo.py
```diff
@@ -70,33 +71,46 @@ def get_package_root(cls, manifest_resource, codebase):
...
-        return '{}/crates/{}/{}/download'.format(baseurl, self.name, self.version)
+        if self.name and self.version:
+            return '{}/crates/{}/{}/download'.format(baseurl, self.name, self.version)
+        return None
 
     def api_data_url(self, baseurl=default_api_baseurl):
-        return '{}/crates/{}'.format(baseurl, self.name)
+        if self.name:
+            return '{}/crates/{}'.format(baseurl, self.name)
+        return None
```
majurg, line 86: I would prefer if we followed this convention:
```
if not self.name:
    return None
return '{}/crates/{}'.format()
```
IMO this is easier to read and follow.

rpotter12: Okay. I will change it soon.

src/packagedcode/cargo.py
```diff
@@ -70,33 +71,46 @@ def get_package_root(cls, manifest_resource, codebase):
         return manifest_resource.parent(codebase)
 
     def repository_homepage_url(self, baseurl=default_web_baseurl):
+        if not self.name:
```
pombredanne, line 74: What about instead:
```
if self.name:
    return '{}/crates/{}'.format(baseurl, self.name)
```

src/packagedcode/cargo.py
```diff
@@ -70,33 +71,46 @@ def get_package_root(cls, manifest_resource, codebase):
...
+            return None
         return '{}/crates/{}'.format(baseurl, self.name)
 
     def repository_download_url(self, baseurl=default_download_baseurl):
+        if not self.name and not self.version:
+            return None
```
pombredanne, line 80: Same as above...

src/packagedcode/cargo.py
```diff
@@ -70,33 +71,46 @@ def get_package_root(cls, manifest_resource, codebase):
...
         return '{}/crates/{}/{}/download'.format(baseurl, self.name, self.version)
 
     def api_data_url(self, baseurl=default_api_baseurl):
+        if not self.name:
+            return None
```
pombredanne, line 85: Ditto

src/packagedcode/cargo.py
```diff
@@ -70,33 +71,46 @@ def get_package_root(cls, manifest_resource, codebase):
...
+    return (filetype.is_file(location) and fileutils.file_name(location).lower() == 'cargo.lock')
+
+
 def parse(location):
     """
-    Return a Package object from a Cargo.toml file or None.
+    Return a Package object from a Cargo.toml/Cargo.lock file or None.
     """
-    if not is_cargo_toml(location):
+    if not is_cargo_toml(location) and not is_cargo_lock(location):
```
pombredanne, line 101: What about this instead?
```
handlers = {'cargo.toml': build_cargo_toml_package, 'cargo.lock': build_cargo_lock_package}
filename = filetype.is_file(location) and fileutils.file_name(location).lower()
handler = handlers.get(filename)
if handler:
    return handler(toml.load(location, _dict=OrderedDict))
```

src/packagedcode/cargo.py
```diff
@@ -179,3 +193,29 @@ def parse_person(person):
...
 ).match
+
+
+def build_cargo_lock_package(package_data):
+    """
+    Return a Package object from a Cargo.lock package data mapping or None.
+    """
+
+    package_dependencies = []
+    core_package_data = package_data.get('package', {})
```
pombredanne, line 204: is this a mapping or a list?

rpotter12: Ah...It's a list. I will change it soon.

src/packagedcode/cargo.py
rpotter12: This also looks good. I add this code.

src/packagedcode/cargo.py
rpotter12: This is also good. Both types are meaningful. We can use anything.

src/packagedcode/cargo.py
rpotter12: ditto

src/packagedcode/cargo.py
rpotter12: ditto

src/packagedcode/cargo.py
pombredanne: MO is that positive tests (rather than a negative with `not`)  and avoiding returning None which is the default looks better, and is less code.

src/packagedcode/cargo.py
pombredanne: You could even use this pythonic construct:
```
return handler and handler(toml.load(location, _dict=OrderedDict))
```

rpotter12: Done

tests/packagedcode/data/cargo/cargo_lock/sample2/Cargo.lock
```diff
@@ -0,0 +1,13 @@
+# This file is automatically @generated by Cargo.
+# It is not intended for manual editing.
+[[package]]
+name = "autocfg"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f8aac770f1885fd7e387acedd76065302551364496e46b3dd00860b2f8359b9d"
```
pombredanne, line 7: I wonder if we are not missing out on something by skipping the `source` and `checksum`. Can you enter a follow up ticket to revisit this later? We might want to add some new fields to the `DependentPackage` for that as these look on the surface as being quite similar to what we have in a go.mod and go.sum? Or we may want to keep these as qualifiers? 
And thinking of it... there are also cases in Python and Ruby and npm where we get deps fetched from git repos

PR #2183, opened 2020-08-26
Revert to another version of ftfy that runs Py2 and 3

pombredanne:
This reverts ftfy to use a version that runs on Python2 and 3.
This helps make things work for the new repo approach 
crafted by @Abhishek-Dev09 


Abhishek-Dev09:
Looking fine!

Abhishek-Dev09 (approved):

setup.py
```diff
@@ -192,8 +192,8 @@ def read(*names, **kwargs):
         'dparse >= 0.4.1',
 
         # used to fix mojibake in Windows PE
-        'ftfy <  5.0.0; python_version == "2.7"',
-        'ftfy>=  5.0.0; python_version > "3"',
```
Abhishek-Dev09, line 196: But my github pypi contains this package.

pombredanne: That's OK, we can have newer and older versions alright there.

Issue #2070, opened 2020-06-16
Dev utility: Improve configure script to work of pinned requirements


pombredanne:
Using the "find link" at the new Github repo.
Something such as this:
```
$ pip install --upgrade banal --no-cache-dir --find-links=https://github.com/pombredanne/foobar/releases/tag/v0.1
```

PR #2184, opened 2020-08-26
Improve configure/release script to work with pinned requirements

Abhishek-Dev09:
Improve configure/release script to work of pinned requirements using github-pypi links.
Fixes : #2070 #2072


.travis.yml
```diff
@@ -93,8 +72,20 @@ before_install:
   - chmod +x ./docs/scripts/sphinx_build_link_check.sh
   - chmod +x ./docs/scripts/doc8_style_check.sh
 
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+    curl -LO https://raw.githubusercontent.com/GiovanniBussi/macports-ci/master/macports-ci;
```
pombredanne, line 76: What is this?

Abhishek-Dev09: I already told yoy about that error in travis ci . Thhere is bug that .congure run on py2 instead of py3 .

pombredanne: ok, got it: 
1. can you install it from here instead? https://github.com/nexB/macports-ci
2. can you explain with a # comment why you are doing there in the file?

etc/conf/requirements-py36_base.txt
```diff
@@ -0,0 +1,3 @@
+-r requirements-py36_all.txt
+-r dev/requirements_base.txt
```
pombredanne, line 2: Are you sure of this line?

etc/configure.py
```diff
@@ -12,9 +12,9 @@
 To use, create a configuration directory tree that contains any of these:
 
 * Requirements files named with this convention:
- - requirements_base.txt contains common requirements installed on all platforms.
- - requirements_win.txt, requirements_linux.txt, requirements_mac.txt and
-   requirements_posix.txt are os-specific requirements.
+ - requirements-py36_base.txt contains common requirements installed on all platforms.
```
pombredanne, line 15: Why do we need a -py36 there? we will not have a configure script for each Python version

etc/configure.py
```diff
@@ -104,18 +104,24 @@
     platform_names = tuple()
 
 
+if sys.version_info < (3, 6):
+    raise Exception('Python 2.7 is not supported!')
```
pombredanne, line 108: Say rather than only Python 3.6 and up is supported.

etc/configure.py
```diff
@@ -261,6 +272,9 @@ def create_virtualenv(std_python, root_dir, tpp_dirs=(), quiet=False):
         vcmd += ['-qq']
     # third parties may be in more than one directory
     vcmd.extend(build_pip_dirs_args(tpp_dirs, root_dir))
+    # Window doesn't support link as extra-search-dir
```
pombredanne, line 275: Is there a link  in the doc or a bug about this? Please add it there as a comment

etc/configure.py
pombredanne, line 15: What's the difference between a requirements files for 3.6, 3.7 and 3.8?

pombredanne (changes requested):
Thanks! 
I am not clear about why you need to have requirements files for only one Python version.

etc/conf/requirements-py36_base.txt
Abhishek-Dev09: Yes

etc/configure.py
Abhishek-Dev09: Requirements files difer in wheels that have native code  like intbitset is diff for py36,py37,py38 and many such more wheels like extractcode-7z

Abhishek-Dev09: Then what i use here ? is `requirements-(python-version)_base is fine?

.travis.yml
Abhishek-Dev09: Your repo is forked from that sources from where i have installed.

Abhishek-Dev09:
> Thanks!
> I am not clear about why you need to have requirements files for only one Python version.

Because all requirement files differ in native wheels like inbitset . I use py36 to avoid confusion.

etc/configure.py
Abhishek-Dev09: Dont worry it is comments . By the ./configure the detect the python version , it will find requirement files based on ur platform/os.

etc/configure.py
Abhishek-Dev09: i didn't get docs about this specifying same issue.

.travis.yml
Abhishek-Dev09: hey, now i m using nexb macport ci repo.

PR #2195, opened 2020-09-01
Bump dependencies #295

pombredanne:
Update dependencies to newer versions.


pombredanne:
This is stale. Closing as we are chaning the way we handle deps anytime now.

PR #2202, opened 2020-09-02
Fix Google Fuchsia license

tsteenbe:
Fix misspelling Fuschia -> Fuchsia and update URL

### DIFF
--- a/src/licensedcode/data/licenses/google-patent-license-fuchsia.LICENSE
+++ b/src/licensedcode/data/licenses/google-patent-license-fuchsia.LICENSE
@@ -1,0 +1,20 @@
+Additional IP Rights Grant (Patents)
+"This implementation" means the copyrightable works distributed by
+Google as part of the Fuchsia project.
+Google hereby grants to you a perpetual, worldwide, non-exclusive,
+no-charge, royalty-free, irrevocable (except as stated in this
+section) patent license to make, have made, use, offer to sell, sell,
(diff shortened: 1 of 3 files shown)


tsteenbe:
@pombredanne  I just notice there is also a `other_urls:` field for .yml files. Instead of removing the original link to https://github.com/flutter/flutter/blob/master/PATENTS shall I moved it to `other_urls:` field?

So google-patent-license-fuchsia.yml becomes:

```
key: google-patent-license-fuchsia
short_name: Google Patent License for Fuchsia
name: Google Patent License for Fuchsia
category: Patent License
owner: Google
homepage_url: https://fuchsia.googlesource.com/fuchsia/+/refs/heads/master/PATENTS
faq_url: https://en.wikipedia.org/wiki/Google_Fuchsia
other_urls:
    - https://github.com/flutter/flutter/blob/master/PATENTS
```

pombredanne:
Thank you ++ I important caveat though before I can merged: we cannot change a license key ever.
What we do instead is create a new one with the new key and mark the old one with `is_deprecated: yes` and add notes to it with  `notes: renamed to xxx` with the new license key ... this provides some guarantee of stability of license keys

pombredanne:
> Instead of removing the original link to https://github.com/flutter/flutter/blob/master/PATENTS shall I moved it to other_urls: field?

yes, that works

pombredanne:
@tsteenbe I applied the fixes to your branch directly as I want to merge this today.

PR #2230, opened 2020-09-15
Parse dependency if only name is present

### LINKED ISSUE #2229: Go parser cannot parse go.mod and go.sum files without a namespace
### Description

> Please leave a brief description of the bug or feature request:

Normally an entry to e.g. `go.sum` looks something like this:
```
github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
```
The Go parser parses `github.com/BurntSushi` into `namespace` and `toml` into `name`.

I have this entry:
```
go.opencensus.io v0.21.0/go.mod h1:mSImk1erAIZhrmZN+AvHh14ztQfjbGwt4TtuofqLduU=
```
Basically the `namespace` portion is missing.

For my concrete issue this is the corresponding source code: https://github.com/nexB/scancode-toolkit/blob/v3.2.1rc2/src/packagedcode/go_mod.py#L253-L254, but `namespace` is used through out the Go parser.

My proposal is to remove the `namespace` portion and parse everything into `name`, since I do not see in the source that `namespace` is used standalone somewhere.

### How To Reproduce

> Tell us how to reproduce the issue. 

### System configuration

* What OS are you running on? (Windows/MacOS/Linux) Ubuntu 20.04
* What version of scancode-toolkit was used to generate the scan file? v3.2.1rc2
* What installation method was used to install/run scancode? (pip/source download/other) `./configure`

rpotter12:
Fixes #2229

### Tasks

### DIFF
--- a/src/packagedcode/go_mod.py
+++ b/src/packagedcode/go_mod.py
@@ -250,13 +250,17 @@ Here are some example of usage of this module::
 # Regex expressions to parse go.sum file dependency
 # dep example: github.com/BurntSushi/toml v0.3.1 h1:WXkYY....
 get_dependency = re.compile(
-    r'(?P<namespace>[^\s]*)'
-    r'\/'
     r'(?P<name>[^\s]*)'
     r'\s*'
     r'(?P<version>[^\s]*)'
     r'\s*'
     r'h1:(?P<checksum>[^\s]*)\s*'
+).match
+
+name_parser = re.compile(
+    r'(?P<namespace>[^\s]*)'
+    r'\/'
+    r'(?P<name>[^\s]*)'
 ).match
 
 
@@ -273,9 +277,16 @@ def parse_gosum(location):
         line = line.replace('/go.mod', '')
         parsed_dep = get_dependency(line)
 
+        name = parsed_dep.group('name')
+        namespace = ''
+        if '/' in name:
+            dep_name = name_parser(name)
+            namespace = dep_name.group('namespace')
+            name = dep_name.group('name')
+
         dep = GoModule(
-                namespace=parsed_dep.group('namespace'),
-                name=parsed_dep.group('name'),
+                namespace=namespace,
+                name=name,
                 version=parsed_dep.group('version')
             )
 
--- a/src/packagedcode/go_mod.py
+++ b/src/packagedcode/go_mod.py
@@ -277,12 +277,8 @@ def parse_gosum(location):
         line = line.replace('/go.mod', '')
         parsed_dep = get_dependency(line)
 
-        name = parsed_dep.group('name')
-        namespace = ''
-        if '/' in name:
-            dep_name = name_parser(name)
-            namespace = dep_name.group('namespace')
-            name = dep_name.group('name')
+        ns_name = parsed_dep.group('name')
+        namespace, _, name = ns_name.rpartition('/')
 
         dep = GoModule(
                 namespace=namespace,
--- a/src/packagedcode/go_mod.py
+++ b/src/packagedcode/go_mod.py
@@ -123,17 +123,15 @@ For example:
 
 # Regex expressions to parse different types of go.mod file dependency
 parse_module_name = re.compile(
-    r'^module\s*'
-    r'(?P<namespace>[^\s]*)'
-    r'\/'
-    r'(?P<name>[^\s]*)\s*'
+    r'(?P<type>[^\s]*)'
+    r'(\s)*'
+    r'(?P<ns_name>[^\s]*)'
+    r'\s*(?P<version>[^\s]*)\s*'
 ).match
 
 parse_dep_link = re.compile(
     r'.*?(\s)*'
-    r'(?P<namespace>[^\s]*)'
-    r'\/'
-    r'(?P<name>[^\s]*)'
+    r'(?P<ns_name>[^\s]*)'
     r'\s*'
     r'(?P<version>[^\s]*)\s*'
 ).match
@@ -165,8 +163,12 @@ def parse_gomod(location):
 
         parsed_module_name = parse_module_name(line)
         if parsed_module_name:
-            gomods.name = parsed_module_name.group('name')
-            gomods.namespace = parsed_module_name.group('namespace')
+            ns_name = parsed_module_name.group('ns_name')
+            namespace, _, name = ns_name.rpartition('/')
+
+        if 'module' in line:
+            gomods.namespace = namespace
+            gomods.name = name
 
         if 'require' in line and '(' in line:
             for req in lines[i+1:]:
@@ -174,10 +176,12 @@ def parse_gomod(location):
                 if ')' in req:
                     break
                 parsed_dep_link = parse_dep_link(req)
+                ns_name = parsed_dep_link.group('ns_name')
+                namespace, _, name = ns_name.rpartition('/')
                 if parsed_dep_link:
                     require.append(GoModule(
-                            namespace=parsed_dep_link.group('namespace'),
-                            name=parsed_dep_link.group('name'),
+                            namespace=namespace,
+                            name=name,
                             version=parsed_dep_link.group('version')
                         )
                     )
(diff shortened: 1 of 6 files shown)


src/packagedcode/go_mod.py
```diff
@@ -250,15 +250,19 @@ def parse_gomod(location):
...
     r'\s*'
     r'h1:(?P<checksum>[^\s]*)\s*'
 ).match
 
+name_parser = re.compile(
+    r'(?P<namespace>[^\s]*)'
+    r'\/'
+    r'(?P<name>[^\s]*)'
+).match
+
```
pombredanne, line 265: I think I get your approach: get a `namespace/name` first and if it contains a `/` further split it afterwards below, right?
If so, let me suggest a slightly simpler approach for the second part below which would mean you do not need that second regex above

src/packagedcode/go_mod.py
```diff
@@ -273,9 +277,16 @@ def parse_gosum(location):
         line = line.replace('/go.mod', '')
         parsed_dep = get_dependency(line)
 
+        name = parsed_dep.group('name')
+        namespace = ''
+        if '/' in name:
+            dep_name = name_parser(name)
+            namespace = dep_name.group('namespace')
+            name = dep_name.group('name')
+
```
pombredanne, line 286: here is my suggestion:
```
ns_name = parsed_dep.group('name')
namespace, _, name = ns_name.rpartition('/')
```
the partition and rpartition functions are your friends most often over regex when splitting strings:
```
>>> 'foo'.rpartition('/')
('', '', 'foo')
>>> 'foo/bar'.rpartition('/')
('foo', '/', 'bar')
>>> 'foo/bar/baz'.rpartition('/')
('foo/bar', '/', 'baz')
```

pombredanne (changes requested):
Thank you! See my comments inline. Also do you tests the same thing for a go.mod such as there https://github.com/nexB/scancode-toolkit/issues/2229#issuecomment-692835210 ?

pombredanne:
@jmastr ping

src/packagedcode/go_mod.py
rpotter12: This will be more better than mine. Thanks :)

rpotter12:
@pombredanne I have updated the scripts. Please review it again.:)

src/packagedcode/go_mod.py
```diff
@@ -250,15 +250,19 @@ def parse_gomod(location):
...
-    r'(?P<namespace>[^\s]*)'
-    r'\/'
     r'(?P<name>[^\s]*)'
     r'\s*'
     r'(?P<version>[^\s]*)'
     r'\s*'
     r'h1:(?P<checksum>[^\s]*)\s*'
 ).match
 
+name_parser = re.compile(
```
pombredanne, line 260: I think this whole block is no longer needed, is it?

src/packagedcode/go_mod.py
```diff
@@ -250,15 +250,19 @@ def parse_gomod(location):
 # Regex expressions to parse go.sum file dependency
 # dep example: github.com/BurntSushi/toml v0.3.1 h1:WXkYY....
 get_dependency = re.compile(
-    r'(?P<namespace>[^\s]*)'
-    r'\/'
     r'(?P<name>[^\s]*)'
```
pombredanne, line 253: what if you named this `ns_name` or `namespace_name` rather than `name` since this is what it is now?

pombredanne (changes requested):
Thanks! I added some more comments. Could you also add tests with a go.mod?

rpotter12:
@pombredanne Please review it again :)

jmastr:
@rpotter12 Currently on my phone, but is your patch „only“ for go.sum? Can’t this happen in go.mod too? Sry, if I am missing something here.

rpotter12:
@pombredanne @jmastr 
1. https://github.com/census-instrumentation/opencensus-service/blob/57e037c599b75fe2dddf89cb9ada67bf4c8de9fc/go.mod
2. https://github.com/census-instrumentation/opencensus-service/blob/57e037c599b75fe2dddf89cb9ada67bf4c8de9fc/go.sum

I don't think we have to solve this because `go.opencensus.io` is not used like `go.opencensus.io v0.21.0/go.mod h1:mSImk1erAIZhrmZN+AvHh14ztQfjbGwt4TtuofqLduU=`

What do you say this??

jmastr:
> @pombredanne @jmastr
> 
>     1. https://github.com/census-instrumentation/opencensus-service/blob/57e037c599b75fe2dddf89cb9ada67bf4c8de9fc/go.mod
> 
>     2. https://github.com/census-instrumentation/opencensus-service/blob/57e037c599b75fe2dddf89cb9ada67bf4c8de9fc/go.sum
> 
> 
> I don't think we have to solve this because `go.opencensus.io` is not used like `go.opencensus.io v0.21.0/go.mod h1:mSImk1erAIZhrmZN+AvHh14ztQfjbGwt4TtuofqLduU=` as mentioned in above examples.
> 
> What do you say this??

@rpotter12 I have a mod file from google like this:
```
go.opencensus.io v0.22.4
```
https://github.com/googleapis/google-cloud-go/blob/v0.66.0/go.mod#L14

rpotter12:
@jmastr Okay then. I have solved this problem. You can review it :)

jmastr:
LGTM

rpotter12:
@pombredanne Please review this too :)

pombredanne (changes requested):
@rpotter12 Thank you! I have a few comments on the regex that I had not noticed before.

src/packagedcode/go_mod.py
```diff
@@ -123,17 +123,15 @@ def purl(self, include_version=True):
...
 # Regex expressions to parse different types of go.mod file dependency
 parse_module_name = re.compile(
-    r'^module\s*'
-    r'(?P<namespace>[^\s]*)'
-    r'\/'
-    r'(?P<name>[^\s]*)\s*'
+    r'(?P<type>[^\s]*)'
+    r'(\s)*'
+    r'(?P<ns_name>[^\s]*)'
+    r'\s*(?P<version>[^\s]*)\s*'
```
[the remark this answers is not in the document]
rpotter12: This suggestion doesn't work. I don't know why.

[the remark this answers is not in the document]
rpotter12: Ah....we need to add `r'\s?' in line 129`

src/packagedcode/go_mod.py
```diff
@@ -165,19 +163,25 @@ def parse_gomod(location):
 
         parsed_module_name = parse_module_name(line)
         if parsed_module_name:
-            gomods.name = parsed_module_name.group('name')
-            gomods.namespace = parsed_module_name.group('namespace')
+            ns_name = parsed_module_name.group('ns_name')
+            namespace, _, name = ns_name.rpartition('/')
+
+        if 'module' in line:
```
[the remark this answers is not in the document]
rpotter12: @pombredanne There are three types of lines:
1. module github.com/alecthomas/participle
2. require github.com/davecgh/go-spew v1.1.1
3. exclude github.com/davecgh/go-spew v1.1.1

so `parsed_module_name` parse all the three types of lines.

src/packagedcode/go_mod.py
```diff
@@ -123,17 +123,15 @@ def purl(self, include_version=True):
 
 # Regex expressions to parse different types of go.mod file dependency
 parse_module_name = re.compile(
-    r'^module\s*'
-    r'(?P<namespace>[^\s]*)'
-    r'\/'
-    r'(?P<name>[^\s]*)\s*'
+    r'(?P<type>[^\s]*)'
+    r'(\s)*'
```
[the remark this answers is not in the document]
rpotter12: I thought there can be multiple spaces also. So I did it like this.

rpotter12:
@pombredanne Please review the PR again. My latest changes are more clear then previous ones :)

rpotter12:
@pombredanne Please review the PR again. My latest changes are more clear then previous ones :)

rpotter12:
In https://github.com/nexB/scancode-toolkit/pull/2230/files#diff-081f69678b5dd083e0c798c6e4d24645R126 we should use a+ but it doesn't work for me. 

I think it will work for all the cases because r'(?P<type>[^\s]+)(\s)+(?P<ns_name>[^\s]+)\s?(?P<version>(.*))' in this \s? is used because module line is not having version number so version will be optional in that case and require and exclude line (?P<version>(.*)) will detect all the characters till end.

pombredanne (approved):

pombredanne:
Thanks!

Issue #2200, opened 2020-09-01
New and improved licenses


pombredanne:
From @tsteenbe :

> Licenses I would like to have added:
> https://github.com/jsr107/jsr107spec/blob/v1.0.0/LICENSE.txt (similar to jsr-107-jcache-spec.LICENSE)

> Eclipse Foundation Specification License https://www.eclipse.org/legal/efsl.php

> Liberation Font Software License Agreement (GPL font exception) https://pagure.io/liberation-fonts/blob/cf3040c146097d6b7361e2823d5c6b5c309a456a/f/master/License.txt

> Jetty Project Corporate Contributor License Agreement V1.1 https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/LICENSE-CONTRIBUTOR/ccla-template.txt

> BaKoMa Fonts License website: https://www.ctan.org/tex-archive/fonts/cm/ps-type1/bakoma

> OGC Document Notice https://www.ogc.org/ogc/Document

> Oracle Technology Network License Agreement for Oracle Java SE This is the new one for 2019 https://www.oracle.com/downloads/licenses/javase-license1.html

> Updates to license definitions:
>
> https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/ogc.LICENSE has 2 terms whilst https://www.ogc.org/ogc/software/1.0 has 3 terms.

> Test cases (e.g. for the files below ScanCode had issues detecting various licenses)
>
> https://github.com/aseprite/freetype2/blob/77f0814a3162a945d6d6704f545827f883e451e5/src/psaux/psread.h#L9-L34 (detection of FTL in license header)
> https://github.com/bitwiseshiftleft/sjcl/blob/1.0.7/README/COPYRIGHT  (detection of all OSS licenses)
> https://github.com/typetools/checker-framework/blob/checker-framework-2.0.0/LICENSE.txt (detection of all OSS licenses)
>

pombredanne:
All fixed in develop. Thank you!

PR #2225, opened 2020-09-10
Report SPDX keys everywhere

pombredanne:
See #532

### Tasks

* [x] Reviewed [contribution guidelines](https://github.com/nexB/scancode-toolkit/blob/develop/CONTRIBUTING.rst)
* [x] PR is descriptively titled 📑 and links the original issue above 🔗
* [ ] Tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR
  Run [tests](https://scancode-toolkit.readthedocs.io/en/latest/contribute/contrib_dev.html#running-tests) locally to check for errors. 
* [x] Commits are in uniquely-named feature branch and has no merge conflicts 📁

Issue #1336, opened 2019-01-24
No spdx short identifier for license x11-fsf?


dankegel:
Is it expected that there is no mapping from x11-fsf to a spdx code?

https://enterprise.dejacode.com/licenses/public/x11-fsf/ lists no spdx short identifier :-(

Where does one report such a thing?  I'll start here, since I first noticed it via scancode.

On an Ubuntu machine:

```
$ scancode --version
ScanCode version 2.9.9
$ scancode -l --json-pp - --copyright /usr/share/doc/libxcb1/copyright
...
      "licenses": [
        {
          "key": "x11-fsf",
          "score": 96.5,
          "name": "X11-Style (FSF)",
          "short_name": "X11-Style (FSF)",
          "category": "Permissive",
          "is_exception": false,
          "owner": "Free Software Foundation (FSF)",
          "homepage_url": "",
          "text_url": "",
          "reference_url": "https://enterprise.dejacode.com/urn/urn:dje:license:x11-fsf",
          "spdx_license_key": "",
          "spdx_url": "",
          "start_line": 16,
          "end_line": 41,
          "matched_rule": {
            "identifier": "x11-fsf.RULE",
            "license_expression": "x11-fsf",
            "licenses": [
              "x11-fsf"
            ],
...
```
Expected nonempty spdx_license_key.

DennisClark:
I recommend that you go to https://github.com/spdx/license-list-XML/blob/master/CONTRIBUTING.md  and request the new SPDX license.

pombredanne:
@dankegel Thanks for the report. There are two issues:
1. there is a bug. The license is not an `x11-fsf` license but a very close one called `xfree86-1.0` and this is the bug.
2. neither `x11-fsf` nor `xfree86-1.0` are yet known from SPDX. ScanCode/DejaCode casts a much wider net (there are only about ~300ish licenses in SPDX and 1200+ in ScanCode) .
I wished they would all be added in SPDX and that's something we discussed with SPDX but short term the thing missing is time to review all these on the SPDX side. That said, the slightly more recent license `xfree86-1.1` is listed at SPDX.... but the `xfree86-1.0` is not and still quite common and still used in many X11-related projects. So we should make the requests.... help is much welcomed!

pombredanne:
@dankegel the commit above fix the detection issue and returns now a proper `xfree86-1.0` license.  You can try it out in the  branch of the PR #1337

pombredanne:
[...]

Issue #1217, opened 2018-10-19
Should license detection also return SPDX license expressions?


pombredanne:
Today we have this:
- when the output is SPDX, we return proper SPDX license expressions using SPDX license ids
- otherwise (JSON, etc) we return
  - license expressions using ScanCode own license keys
  - a list licenses and for these that are known at SPDX the corresponding SPDX license key

In this later case I wonder if it could be useful to also return license expressions using SPDX license ids.
This would be eventually a new option and would be returned under its own file-level attribute.
This could also possibly apply to a companion to the `license_expression` returned with `packages` for  a package manifest. Now, of the 1301 licenses that scancode knows, only 375 are referenced at SPDX, therefore there will be quite a few "LicenseRef" in this case. See #532 for this issue and https://github.com/nexB/scancode-toolkit/issues/532#issuecomment-431420151 in particular.

@sschuberth @mjherzog @DennisClark @armijnhemel Any feedback?

sschuberth:
Maybe @mnonnenmacher also has an opinion about this as we were just today discussing this in some other context.

armijnhemel:
it makes sense to me. What I would love to see is the version of the SPDX standard that is used, so I don't have to guess :-)

pombredanne:
@armijnhemel good point! We can set the version of the SPDX license list in the scan headers.

pombredanne:
This is going to be useful for ClearlyDefined and @dabutvin ... so this is going into 3.1 alright at the minimum.

pombredanne:
This is being merged in develop and will be in v32.0

PR #2247, opened 2020-09-24
Revert "Make SPDX license key as default #2200"

fviernau:
All other .yml files always use IDs which are on the SPDX license list
as value for the field `spdx_license_key`, thus this entry is an
exception. Furthermore it is the only ID starting with 'LicenseRef-'.

Since this change seems not to bring much value, undo it for simplicity.

This reverts commit aec1535d5c3ba8e40d622accbaeea6c430dcafb3.

Signed-off-by: Frank Viernau <frank.viernau@here.com>

### DIFF
--- a/src/licensedcode/data/licenses/here-proprietary.yml
+++ b/src/licensedcode/data/licenses/here-proprietary.yml
@@ -3,4 +3,5 @@ short_name: HERE Proprietary
 name: HERE Proprietary License
 category: Commercial
 owner: HERE Global B.V.
-spdx_license_key: LicenseRef-Proprietary-HERE
+other_spdx_license_keys:
+    - LicenseRef-Proprietary-HERE


fviernau:
@pombredanne @tsteenbe - I was not aware of your rationale for this change, so this is a speculative fix.
It can at least be used to get the discussion started.

tsteenbe:
@fviernau  This is correct entry per SPDX spec - with "LicenseRef-" prefix we indicate that this is non-official SPDX identifier conform https://github.com/spdx/spdx-spec/blob/v2.2/chapters/appendix-IV-SPDX-license-expressions.md

fviernau:
> @fviernau This is correct entry per SPDX spec - with "LicenseRef-" prefix we indicate that this is non-official SPDX identifier conform https://github.com/spdx/spdx-spec/blob/v2.2/chapters/appendix-IV-SPDX-license-expressions.md

Sure, that's out of question that it's a valid ID. It just breaks several assumption the ORT code made at various places.

fviernau:
@pombredanne  was it planned to define more 'LicenseRef-' IDs in ScanCode?

pombredanne:
@fviernau FWIW I had made [this ](https://github.com/nexB/scancode-toolkit/commit/aec1535d5c3ba8e40d622accbaeea6c430dcafb3) on @tsteenbe 's request I am fine to revert it and in all case ......

And re:
> was it planned to define more 'LicenseRef-' IDs in ScanCode?

.... we are finally adopting having an explicit scancode SPDX namespace everywhere per #532 with #2225 such that `spdx_license_key` becomes a mandatory field of the [License object](https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/models.py#L94) and is always stored explicitly in the .yml of a license.

This makes things cleaner and simpler to always return SPDX license expressions #1217 and fix some weird issues and head scratcher such as #1336

fviernau:
Thanks @pombredanne! One aspect of the doubts I had was that the original change was actually renaming the ID from
`LicenseRef-scancode-here-proprietary` to `LicenseRef-Proprietary-HERE'. Well, the prefix `LicenseRef-scancode` comes from `ORT` but the scancode specific part `here-proprietary` gets changed effectively in ORT.

I have investigated further on ORT side, and it seems likely that there is an easy work-around for now. So we could also drop this PR. However, maybe it would be cleaner to do the `LicenseRef-` story all at once, rather than now just for the HERE license. 
I suppose that the `LicenseRef-` identifiers in `ScanCode` would anyway share a common `ScanCode` specific prefix, thus `LicenseRef-Proprietary-HERE' would need to change anyway again as part of #1217 or #1336. 

As a renaming of license IDs comes with some maintainance effort to me it seems reasonable to undo it now, and then define the lines refs as part of the tickets you linked.

Still, it's just my preference. I'd merge this PR only if you agree that it all makes sense and most importantly that there will be a new 3.2 release containing this PR. Would that be the case @pombredanne ?

pombredanne:
> Still, it's just my preference. I'd merge this PR only if you agree that it all makes sense and most importantly that there will be a new 3.2 release containing this PR. Would that be the case @pombredanne ?

Yes this is still in time for 3.2 :)

As for using `LicenseRef-scancode-here-proprietary` vs. `LicenseRef-Proprietary-HERE` this is really your call.
Technically we should have all `LicenseRef` in a namespace o avoid dangling references, and therefore it should be `LicenseRef-scancode-here-proprietary` unless you guys create your own namespace and I use that here instead ... e.g. something such as `LicenseRef-HERE-Proprietary` where `HERE` would your own namespace.
Anything flies for me as long as we are in sync!

pombredanne:
@fviernau note that for now your commits fail the tests with a duplicated rule: `here-proprietary_4.RULE` already exists with https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/rules/spdx_license_id_licenseref-proprietary-here_for_here-proprietary.RULE
```
>           assert result.exit_code == expected_rc, error
E           AssertionError: 
E             Failure to run: scancode --package /home/travis/build/nexB/scancode-toolkit/tests/scancode/data/rpm/fping-2.4-0.b2.rhfc1.dag.i386.rpm --json /tmp/scancode-tk-tests -9HaAIa/Yft69k/td/tf.results.json --test-mode
E             output:
E             Setup plugins...
E             Removing temporary files...done.
E             Usage: scancode [OPTIONS] <OUTPUT FORMAT OPTION(s)> <input>...
E             Try the 'scancode --help' option for help on options and arguments.
E             
E             Error: Invalid value: ERROR: failed to setup plugin: scan:licenses:
E             Traceback (most recent call last):
E               File "/home/travis/build/nexB/scancode-toolkit/src/scancode/cli.py", line 702, in run_scan
E                 plugin.setup(**requested_options)
E               File "/home/travis/build/nexB/scancode-toolkit/src/licensedcode/plugin_license.py", line 130, in setup
E                 get_index(return_value=False)
E               File "/home/travis/build/nexB/scancode-toolkit/src/licensedcode/cache.py", line 68, in get_index
E                 _LICENSES_BY_KEY_INDEX = get_cached_index(cache_dir, check_consistency)
E               File "/home/travis/build/nexB/scancode-toolkit/src/licensedcode/cache.py", line 249, in get_cached_index
E                 idx = LicenseIndex(rules, _spdx_tokens=spdx_tokens)
E               File "/home/travis/build/nexB/scancode-toolkit/src/licensedcode/index.py", line 245, in __init__
E                 rules, _legalese=_legalese, _spdx_tokens=_spdx_tokens)
E               File "/home/travis/build/nexB/scancode-toolkit/src/licensedcode/index.py", line 528, in _add_rules
E                 raise AssertionError(msg)
E             AssertionError: Duplicate rules: 
E             file:///home/travis/build/nexB/scancode-toolkit/src/licensedcode/data/rules/here-proprietary_4.RULE
E             file:///home/travis/build/nexB/scancode-toolkit/src/licensedcode/data/rules/spdx_license_id_licenseref-proprietary-here_for_here-proprietary.RULE

```

pombredanne:
@fviernau re:
> One aspect of the doubts I had was that the original change was actually renaming the ID from LicenseRef-scancode-here-proprietary to LicenseRef-Proprietary-HERE. Well, the prefix LicenseRef-scancode comes from ORT but the scancode specific part here-proprietary gets changed effectively in ORT.

Please ignore the case suggestions here. As far I am concerned the case of scancode license are always lower case and the case should never matter for SPDX either


> I have investigated further on ORT side, and it seems likely that there is an easy work-around for now (oss-review-toolkit/ort#3124). So we could also drop this PR. However, maybe it would be cleaner to do the LicenseRef- story all at once, rather than partially now just for the HERE license and later the rest. I suppose that the LicenseRef- identifiers in ScanCode would anyway share a common ScanCode specific prefix, thus LicenseRef-Proprietary-HERE would need to change anyway again as part of #1217 or #1336.

> As a renaming of license IDs comes with some maintainance effort, it seems reasonable to me to undo that rename now, and then define the LicenseRef- identifiers all at once as part of the tickets you linked.


We can handle it either way, so your call on what LicenseRef you want to have here.

pombredanne:
BTW the use of `LicenseRef-scancode-xxx` has been in use for a long while in the SPDX output https://github.com/nexB/scancode-toolkit/blob/0c08442e46606938547a2e4fec70780c6ca6f345/src/formattedcode/output_spdx.py#L249 and was contributed by @furuholm with https://github.com/nexB/scancode-toolkit/commit/ed9f861a72fd3d6134bbc8a21b85ea2b2cca8024

pombredanne:
@fviernau please tell me what you want me to do here

fviernau:
I've found an implemented handling for this in ORT, so if I'm not missing on something it is not required anymore. 
I'd have the slight tendency to merge it for simplicity.
Anyhow, it's only a slight tendency - let's do what you prefer / think what fits into ScanCode best. You know it best...

pombredanne:
@fviernau this is your license key :) so I am fine either way ... if I merge this, your SPDX key in ScanCode will become "LicenseRef-scancode-here-proprietary" as soon as I merge https://github.com/nexB/scancode-toolkit/pull/2225

fviernau:
Let's merge it then.

Issue #2256, opened 2020-09-29
NPM dist_mapper fails to decode integrity value properly


majurg:
While the `dist` value is not usually present in NPM packages in the wild, it does show up in some situations. Unfortunately (probably since we moved to python3), this code does not work properly when the `integrity` value is a string. 

To reproduce (using the docstring example from https://github.com/nexB/scancode-toolkit/blob/c3c92ff121632ea5db835f1c460c7d483a91a5d6/src/packagedcode/npm.py#L567):
```
(scancode-toolkit) [sesser@nexb-workstation ~/Code/nexB/scancode-toolkit] python
Python 3.6.11 (default, Jul 17 2020, 15:21:46)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> dist = {"integrity": "sha512-VmqXvL6aSOb+rmswek7prvdFKsFbfMshcRRi07SdFyDqgG6uXsP276NkPTcrD0DiwVQ8rfnCUP8S90x0OD+2gQ==", "shasum": "a124386bce4a90506f28ad4b1d1a804a17baaf32", "dnl_url": "http://registry.npmjs.org/npm/-/npm-2.13.5.tgz"}
>>> from packagedcode.npm import dist_mapper                                                                                               
>>> dist_mapper(dist, None)                                                                                                                
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/sesser/Code/nexB/scancode-toolkit/src/packagedcode/npm.py", line 580, in dist_mapper
    sha512 = b64value.decode('base64').encode('hex')
AttributeError: 'str' object has no attribute 'decode'
```

majurg:
This is fixed by #2261, closing.

PR #2261, opened 2020-09-30
Fix `dist_mapper()` in packagedcode/npm.py

majurg:
* Properly handle base64 encoded strings in the npm dist_mapper()
  function.
* Add test case

Addresses: #2256

### DIFF
--- a/src/packagedcode/npm.py
+++ b/src/packagedcode/npm.py
@@ -577,9 +577,9 @@ def dist_mapper(dist, package):
     integrity = dist.get('integrity') or None
     if integrity:
         algo, _, b64value = integrity.partition('-')
-        assert 'sha512' == algo
         algo = algo.lower()
-        sha512 = hashlib.sha512(base64.b64decode(b64value)).hexdigest()
+        assert 'sha512' == algo
+        sha512 = base64.b64decode(b64value).hex()
         package.sha512 = sha512
 
     sha1 = dist.get('shasum')
--- a/src/packagedcode/npm.py
+++ b/src/packagedcode/npm.py
@@ -26,9 +26,11 @@ from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import base64
 from collections import defaultdict
 from collections import OrderedDict
 from functools import partial
+import hashlib
 import io
 import json
 import logging
@@ -577,7 +579,7 @@ def dist_mapper(dist, package):
         algo, _, b64value = integrity.partition('-')
         assert 'sha512' == algo
         algo = algo.lower()
-        sha512 = b64value.decode('base64').encode('hex')
+        sha512 = hashlib.sha512(base64.b64decode(b64value)).hexdigest()
         package.sha512 = sha512
 
     sha1 = dist.get('shasum')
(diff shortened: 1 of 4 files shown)


pombredanne (changes requested):
Thanks! See my notes inline wrt the sha512 encoding, as I am not sure this is correct.

src/packagedcode/npm.py
```diff
@@ -577,7 +579,7 @@ def dist_mapper(dist, package):
         algo, _, b64value = integrity.partition('-')
         assert 'sha512' == algo
         algo = algo.lower()
-        sha512 = b64value.decode('base64').encode('hex')
+        sha512 = hashlib.sha512(base64.b64decode(b64value)).hexdigest()
```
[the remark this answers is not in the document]
majurg: Yes I was thinking about this last night and realized all I have to do is b64decode the value to get the proper `sha512` (and not rehashing)

src/packagedcode/npm.py
```diff
@@ -577,7 +579,7 @@ def dist_mapper(dist, package):
         algo, _, b64value = integrity.partition('-')
         assert 'sha512' == algo
         algo = algo.lower()
```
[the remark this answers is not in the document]
majurg: good point, I will make this change

src/packagedcode/npm.py
```diff
@@ -26,9 +26,11 @@
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import base64
 from collections import defaultdict
 from collections import OrderedDict
 from functools import partial
+import hashlib
```
pombredanne, line 33: Do you still need hashlib now?

majurg: Nope, the latest commit removes this.

src/packagedcode/npm.py
```diff
@@ -575,10 +576,16 @@ def dist_mapper(dist, package):
...
         algo, _, b64value = integrity.partition('-')
-        assert 'sha512' == algo
         algo = algo.lower()
-        sha512 = b64value.decode('base64').encode('hex')
+        assert 'sha512' == algo
+
+        decoded_b64value = base64.b64decode(b64value)
+        if type(decoded_b64value) == str:
+            sha512 = decoded_b64value.encode('hex')
+        elif type(decoded_b64value) == bytes:
```
pombredanne, line 585: That may not work between Py2 and Py3.
You may instead try this:
```
        if isinstance(decoded_b64value, string_types):
            sha512 = decoded_b64value.encode('hex')
        elif isinstance(decoded_b64value, binary_type):
....
```

pombredanne: Note that I think that b64decode results may always be bytes on Python 2 or Python 3 so you may not need the type test

majurg: I made this change in response to test failures on the python2 side. The latest code now works on both py3 and py2. `b64decode` returns bytes on python3 and string in python2, hence my code above. 

Unless I am missing something, why would `isinstance` be preferable here?

pombredanne: > Unless I am missing something, why would isinstance be preferable here?

that's catching subclasses too and I use it as an automatism but that's not the main point. Rather using `six.string_types` and `six.binary_types` is what matters more IMHO for portability.

pombredanne (approved):
Thanks for all the fixes. LGTM! :+1:

pombredanne:
Merging now that we are all green. :green_apple:  Thank you ++

PR #2307, opened 2020-10-30
Use jsonstreams library in JSON output plugin

JonoYang:
This PR updates the JSON output plugin to use the jsonstreams library to write scancode results to disk. This is done for memory efficiency, where we can stream JSON output to a file without having to create it in memory and then dumping it to disk.


pombredanne (changes requested):
@JonoYang I see that @dcbaker kindly released a 0.5.0 of jsonstreams with your patches and more.
This PR looks good otherwise and I will merge ASAP, but I would rather use the official upstream than a temporary advanced patched version

JonoYang:
@pombredanne I've got the new release of jsonstreams from pypi and updated the wheel in the thirdparty directory.

pombredanne (approved):
Thanks!
This is missing a changelog entry FWIW.

PR #2310, opened 2020-11-01
Fix for issue 1876 - Updated the batch file to replace "%" in arguments with an arbitrary …

### LINKED ISSUE #1876: Scanning a path with "%" fails on Windows
### Description

At least with ScanCode 3.0.2, scanning a path that contains "%" (the percent character) fails on Windows:

```
$ scancode --json-pp - --license NPM\%40xtuc\long\4.2.2\package.json
Usage: scancode [OPTIONS] <input> <OUTPUT FORMAT OPTION(s)>
Try 'scancode --help' for help on options and arguments.

Error: Invalid value for "input": Path "NPM\0xtuc\long\4.2.2\package.json" does not exist.
```

Note the `%40xtuc` in the input path vs. `0xtuc` in the error message.

Background: To guard against characters that are invalid / special on some file systems, [ORT](https://github.com/heremaps/oss-review-toolkit) percent-encodes charaters like `@` when downloading source code in order to scan it. At the example of the `long` NPM package version 4.2.2 from the `@xtuc` scope, it gets downloaded into a directory named "NPM\%40xtuc\long\4.2.2" where `@` is percent-encoded as `%40`.

### How To Reproduce

This can *only* be reproduced on *Windows*, it works on Linux. Also, I was only able to reproduce it with ScanCode version 3.0.2, as versions 3.1.x have [another issue](https://github.com/nexB/scancode-toolkit/issues/1875) on Windows.

1. Create a directory with "%40" as part of its name.
2. Create a file to scan within that directory.
3. Point ScanCode at that file to scan.

### System configuration

* What OS are you running on? Windows 10 x64
* What version of scancode-toolkit was used to generate the scan file? 3.0.2
* What installation method was used to install/run scancode? Git

Duncan-Howe:
…placeholder "##PC##" during the "collectargs" routine and then restoring the "%" signs prior to passing to the scancode executable

Signed-off-by: Duncan Howe <duncan.howe@live.co.uk>

Fixes #1876

### Tasks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,5 @@ pyvenv.cfg
 lib64
 tcl
 /.env
+
+*.orig
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,8 @@
 Changelog
 =========
-
+vNext
+-----
+ - Updated scancode.bat to handle % signs in the arguments #1876
 
 v3.2.2rc3 (2020-09-21)
 ----------------------
--- a/scancode.bat
+++ b/scancode.bat
@@ -24,6 +24,4 @@ if %errorlevel% neq 0 (
 @rem without this things may not always work on Windows 10, but this makes things slower
 set PYTHONDONTWRITEBYTECODE=1
 
-@rem Revert the placeholders back to single percent signs before passing through to ScanCode executable
-echo "%SCANCODE_ROOT_DIR%Scripts\scancode %*" >> d:\ScanCode.txt
 "%SCANCODE_ROOT_DIR%Scripts\scancode" %*
--- a/scancode.bat
+++ b/scancode.bat
@@ -17,7 +17,7 @@ set "SCANCODE_CMD_LINE_ARGS= "
 @rem Capture the current argument value
 	set CurrentParam=%1
     if ""%1""=="""" goto continue
-	@rem Replace a % sign with an arbitrary placeholder sting
+	@rem Replace a % sign with an arbitrary placeholder string
 	@rem Note: the double percent sign (%%) is required to escape the percent sign.
 	@rem The statement "set CurrentParam=!CurrentParam:%%=##PC##!" takes the value
 	@rem of CurrentParam, replaces any percent signs with ##PC## and assigns it back to CurrentParam
--- a/scancode.bat
+++ b/scancode.bat
@@ -3,6 +3,9 @@
 
 @rem  A wrapper to ScanCode command line entry point
 
+@rem Delayed expansion is required to allow the content of the variables to be manipulated prior to use
+setlocal EnableDelayedExpansion
+
 set SCANCODE_ROOT_DIR=%~dp0
 set SCANCODE_CONFIGURED_PYTHON=%SCANCODE_ROOT_DIR%Scripts\python.exe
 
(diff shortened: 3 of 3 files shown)


sschuberth:
Hmm, maybe before trying to fix the logic to "collect all command line arguments in a variable" we should ask @pombredanne *why* this collecting is even necessary. More concretely, why don't we just use `%*` instead of creating `SCANCODE_CMD_LINE_ARGS`?

pombredanne:
@Duncan-Howe Thank you ++
I wonder how to add a test for this?
@sschuberth I am no expert in batch files and cmd: this was to "slurp" arguments is as old as scancode itself.
If I recall correctly this is an idiom that I first saw in Tomcat https://github.com/apache/tomcat80/blob/trunk/bin/startup.bat#L47
I am fine with replacing this with anything that works nicely everywhere :)

pombredanne:
For reference https://github.com/nexB/scancode-toolkit/issues/1876

pombredanne:
@Duncan-Howe Do you mind to also update the changelog?

Duncan-Howe:
@pombredanne What version number should I use as the heading for CHANGELOG or would vNext be OK?

sschuberth:
> I am no expert in batch files and cmd: this was to "slurp" arguments is as old as scancode itself.

If the aim is to forward *all* args passed to the batch file, then IMO `%*` could be used and no complicated replacement of placeholders should be necessary. The story becomes more complicated if you want to forward e.g. all but the first argument, but this is not the case here AFAIU the code. BTW, the Tomcat code says to "Get remaining unshifted command line arguments", but there actually are no unshifted arguments before that line, so also the Tomcat code could be much simpler IMO. Tomcat probably also copied this from somewhere without exactly knowning what they are doing :wink:

sschuberth:
So, @Duncan-Howe, would you mind giving the `%*` approach a try?

Duncan-Howe:
@sschuberth Sure thing - I will refactor the batch file to use "%*" to pass the arguments through

pombredanne:
> @pombredanne What version number should I use as the heading for CHANGELOG or would vNext be OK?

@Duncan-Howe Yes, that would work fine :+1:

pombredanne:
> Tomcat probably also copied this from somewhere without exactly knowning what they are doing wink

I would call this me doing [cargo culting](https://en.wikipedia.org/wiki/Cargo_cult) at its best :D
This is 15 years old  :P https://github.com/apache/tomcat/blob/94400a237d15768ca227dd8c0f4dd5716b48e388/bin/startup.bat#L33

Duncan-Howe:
Updated scancode.bat to remove a number of lines of code that weren't really doing much :-)
Now the arguments are passed straight through using the "%*" syntax so any number of arguments should work (validation of the arguments is delegated to the ScanCode executable)

Duncan-Howe:
The updated batch file seems to perform in exactly the same way as previously when executed from ORT and produces the same result files for my own application scans

sschuberth:
> The updated batch file seems to perform in exactly the same way as previously

Great! Thanks a lot for getting rid of this cruft in the batch file 😄

pombredanne (approved):
LGTM and merging now!
Thank you ++

Issue #1532, opened 2019-04-16
Add `extra_data` field to Package


JonoYang:
The `extra_data` field would be a dictionary whose keys and values are strings.

In the case of parsing package manifest files, it would be convenient to be able to carry over parsed package data to the Package class so we can use it to fill in some fields on the Package class.

JonoYang:
This would be especially convenient in the case of parsing ABOUT files. It would be nice to roll up the information from an ABOUT file to the `about_resource` that the ABOUT file is for, but we need to parse the ABOUT file to know about the `about_resource`.

pombredanne:
@JonoYang Thanks! I have been resisting adding this... but this is a good use case, so let's do it!

pombredanne:
Actually this has been reverted. So I am leaving this opened for now.

JonoYang:
`extra_data` has not been added to `Package` since we do not a problem that would be solved by having the `extra_data` field. `extra_data` was added in some branches, but removed since more pertinent fields are available for use.

pombredanne:
This is all merged now. Closing

PR #2331, opened 2020-11-21
Add extra_data field to Package model #1532

JonoYang:
This PR reintroduces the `extra_data` field to the `Package` class.
It is being readded because we need a place to store the value for
`about_resource` so we can return it when `get_package_root()` is
called on a `Resource` that is an ABOUT file.

There are several other cases where keeping such extra data is
useful and needed for a specific Package type.


src/packagedcode/about.py
```diff
@@ -57,12 +57,8 @@ class AboutPackage(models.Package):
     def recognize(cls, location):
         yield parse(location)
 
-    @classmethod
-    def get_package_root(cls, manifest_resource, codebase):
-        # FIXME: this should have been already stored with the Package itself as extra_data
-        with io.open(manifest_resource.location, encoding='utf-8') as loc:
-            package_data = saneyaml.load(loc.read())
-        about_resource = package_data.get('about_resource')
+    def get_package_root(self, manifest_resource, codebase):
```
JonoYang, line 40: @pombredanne I've changed `get_package_root()` in `AboutPackage` from a classmethod to a regular instance method because I would like to access the `extra_data` field to get the `about_resource` value. I also modified https://github.com/nexB/scancode-toolkit/blob/31e3fee011dfb7fb671e9cdeb0ab7fc23f6c7a04/src/packagedcode/plugin_package.py#L137 to instantiate the package rather than just get the package class so the `extra_data` field is accessable. However, this introduces inconsistency on how the `get_package_root` method is called across packages, where this package requires the package instance to use the method whereas other packages do not.

pombredanne:
All green... lgtm and merging at last!

src/packagedcode/about.py
pombredanne: Hum... I had not noticed that part. Let me review first before merging then

pombredanne: This has been fixed now ... I also improved get_package_instance significantly.

pombredanne (approved):
All green. Merging now!

PR #2332, opened 2020-11-21
Changed the start and end time format in output after scanning

nitesh585:
The time format in output after scanning is 
time1
where the value after "T" does not easily signify the time in any format.

After editing the  "/scancode-toolkit/lib/python2.7/commoncode/timeutils.py" file the output is.
time2

This is my first contribution. So, please guide me, if something went wrong.


pombredanne:
@nitesh585 Sorry for the late review and thank you for this PR. I guess you are trying to switch to use ISO time stamps? https://en.wikipedia.org/wiki/ISO_8601 that would make sense indeed. The original rationale for using these time stamps is that they could be used as path on disk and colons are illegal in Windows filenames.

pombredanne:
I am closing this for now.
Thanks again.

PR #2385, opened 2021-02-08
Dockerfile: add -y flag to apt-get, so that it can run noninteractivly

maxhbr:
The Dockerfile had a call to `apt-get install` without the `-y`, resulting in a problem since it was waiting for the approval.

Also it had CRLF line endings, fixed that for hygiene.


pombredanne (approved):
Thank you ++
LGTM!

PR #2407, opened 2021-02-18
Reordered test assertions

### LINKED ISSUE #2394: RFC: Fix direction ALL test assertions  org-wide with pytest approach changed in latest version
See 
https://github.com/pytest-dev/pytest/issues/3333
https://github.com/pytest-dev/pytest/pull/6673/files

In recent pytest versions 5.x+ the ways asserts failures are reported changed (rather dramatically if I may say) where the order is now:
- `assert result == expected`
before and that was our org-wide convention we use the reverse:
- `assert expected == result`

This impacts significantly the way we look at test failures traces: the diffs are now unreadable and you need to make the mental switch that plus means minus and minus means plus to make sense of a failure trace.

I always thought that it is more natural to write assertion in this new way e.g. `assert result == expected` but that was never the convention in unittest, and pytest and related. 
To recover sanity and assert again safely that `plus == plus` and `minus=minus` I suggest we adopt the way:

1. define a new org-wide standard to use ths way
2. progressively update ALL our many tests to use  `assert result == expected` 

As an example of the current problem here, this snippet is a failure trace where we expected an `mit` license_expression to be detected but instead an `unknown` license_expression was returned:
```
E                     "homepage_url": null,
E                     "keywords": [],
E         -           "license_expression": "unknown",
E         ?                                  ^^^^^^^
E         +           "license_expression": "mit",
E         ?                                  ^^^
E                     "md5": null,
E                     "name": "laravel",
E                     "namespace": "laravel",
``` 

Yet, the way this looks is that an `unknown` license was not detected and an `mit` was returned instead.
Before we bump to use pytest 5.x the same failure was looking this way:

```
E                     "homepage_url": null,
E                     "keywords": [],
E         -           "license_expression": "mit",
E         ?                                  ^^^
E         +           "license_expression": "unknown",
E         ?                                  ^^^^^^^
E                     "md5": null,
E                     "name": "laravel",
E                     "namespace": "laravel",
``` 

and this i how we want things to look again hence the push to reorder all our test assertions

adityaviki:
Signed-off-by: Aditya Viki <adityaviki01@gmail.com>

Fixes #2394


pombredanne:
@adityaviki Thank you++ 
Did you check the pending test failures here?

tests/licensedcode/test_match_spdx_lid.py
```diff
@@ -400,7 +400,7 @@ def test_get_expression_without_and_should_not_return_unknown(self):
         unknown_symbol = get_unknown_spdx_symbol()
         line_text = '* SPDX-License-Identifier:     GPL-2.0+ BSD-2-Clause'
         expression = get_expression(line_text, licensing, spdx_symbols, unknown_symbol)
-        assert unknown_symbol != expression
+        assert expression == unknown_symbol
```
rpotter12: @adityaviki Maybe you have used `==` instead of `!=`

pombredanne (approved):
Thank you ++ that's ambitious and now all tests pass!

tests/licensedcode/test_match_spdx_lid.py
pombredanne: @rpotter12 great catch! :bow:

pombredanne:
@adityaviki That's lot of attention to details to get this right on all the tests :bow:  great job!

Issue #2235, opened 2020-09-21
Support exporting license and copyright information in debian copyright format to integrate with REUSE


msohn:
## Short Description

In order to use results from scancode in projects adopting REUSE [1] it would be nice if copyright and license information could be exported in debian copyright format [2] which is used in REUSE for non-intrusive bulk licensing. This is useful
e.g. in projects implemented in golang where source code of dependencies is pulled into the `vendor/` folder of the
project's repository.

[1] https://reuse.software/faq/#bulk-license
[2] https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/#disclaimer-field

## Possible Labels

- new feature
- enhancement
- integration
- plugin

## Select Category

-  [ x ] Enhancement
-  [ ] Add License/Copyright
-  [ ] Scan Feature
-  [ ] Packaging
-  [ ] Documentation
-  [ ] Expand Support
-  [ x ] Other

## **Describe the Update**

Export of copyright and license information in debian copyright format which is used by REUSE.

## **How This Feature will help you/your organization**

Automate compliance with REUSE especially for golang projects which tend to have
a lot of (source code of) dependencies in the `vendor/` folder. Both creation of initial
`.reuse/dep5` file and updating it whenever dependencies change in order to declare copyright
and license information for such dependencies is a lot of work when done manually.

We currently use a poor-man's partial solution using askalono which can only generate
the `Files:`  and `License:` fields of debian copyright format. 

## **Possible Solution/Implementation Details**

Without knowing any details about scancode's implementation I guess this could
be implemented as a post scan plugin.

## **Can you help with this Feature**

I am python 3 literate so I could try to help if you can provide hints how to start that.
I have no clue about the current scancode-toolkit implementation.

pombredanne:
@msohn that would be awesome indeed. The way it would typically be done might be a output plugin.
And in terms of building blocks the key elements would be:
1. collecting matched license expressions and license texts with `--license  --license-text`
2. collecting copyrights with --copyright and ideally except for these part of licenses with `--filter-clues`
3. using the `debut` library that knows everything about Debian files and Debian copyright files (See https://github.com/nexB/debut 
 and https://github.com/nexB/debut/blob/master/src/debut/copyright.py#L311 ) and that is already available in ScanCode and if it needs patching we can patch it alright in a snap.

[...]

PR #2417, opened 2021-02-28
Add minimal Debian copyright file output #2235

pombredanne:
This is a minimal debian copyright file formatted output
It is dumb for now and creates one paragraph per file

### Tasks

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -34,6 +34,7 @@ including:
 - https://github.com/nexB/thirdparty-packages/pypi/
 - https://github.com/nexB/scancode-plugins/
 - https://github.com/nexB/scancode-thirdparty-src/
+- https://github.com/nexB/thirdparty-packages/
 
 You may also contact us to request the source code by email at info@nexb.com or
 by postal mail at:
--- a/etc/release/utils_thirdparty.py
+++ b/etc/release/utils_thirdparty.py
@@ -1745,7 +1745,13 @@ def get_remote_repo(remote_links_url=REMOTE_LINKS_URL):
 
 
 def get_remote_package(name, version, remote_links_url=REMOTE_LINKS_URL):
-    return get_remote_repo(remote_links_url).get_package(name, version)
+    """
+    Return a PypiPackage or None.
+    """
+    try:
+        return get_remote_repo(remote_links_url).get_package(name, version)
+    except RemoteNotFetchedException as e:
+        print(f'Failed to fetch remote package info: {e}')
 
 
 _PYPI_REPO = None
@@ -1759,7 +1765,13 @@ def get_pypi_repo(pypi_simple_url=PYPI_SIMPLE_URL):
 
 
 def get_pypi_package(name, version, pypi_simple_url=PYPI_SIMPLE_URL):
-    return get_pypi_repo(pypi_simple_url).get_package(name, version)
+    """
+    Return a PypiPackage or None.
+    """
+    try:
+        return get_pypi_repo(pypi_simple_url).get_package(name, version)
+    except RemoteNotFetchedException as e:
+        print(f'Failed to fetch remote package info: {e}')
 
 ################################################################################
 #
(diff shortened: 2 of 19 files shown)


jelmer:
Thanks for adding this! Some initial observations (maybe for a follow-up DEP-5):

* This doesn't use the standardized DEP-5 terms for licenses - it should possibly have a dictionary mapping scancode license names to DEP-5 ? https://dep-team.pages.debian.net/deps/dep5/
* It doesn't recognise GPL-2.0+, instead saying "gpl-2.0 or gpl-3.0"
* it should exclude any full license files (currently includes e.g. COPYING)
* the actual license headers duplicate some lines from the original file

See e.g. https://pastebin.com/KLhCt1WD (generated by scancode) with https://salsa.debian.org/python-team/packages/dulwich/-/blob/5d9a44c4a56f01e2e9eed4a8f0e488bc7025ab93/debian/copyright (actual file)

pombredanne:
@jelmer thanks. 
so 1. https://salsa.debian.org/python-team/packages/dulwich/-/tree/5d9a44c4a56f01e2e9eed4a8f0e488bc7025ab93 is the exact tree you scanned
2. Let me attach the pastebin here
[dulwich.copyright.txt](https://github.com/nexB/scancode-toolkit/files/6057988/dulwich.copyright.txt)




>    This doesn't use the standardized DEP-5 terms for licenses - it should possibly have a dictionary mapping scancode license names to DEP-5 ? https://dep-team.pages.debian.net/deps/dep5/

Yes we need to draft a mapping. The question is also about all other licenses.
Say I have `GPL2+ WITH <some exception that ScanCode knows about>`
What should we dump then?
- use SPDX? : `GPL2+ WITH LicenseRef-scancode-exception-ScanCode-knows-about`
- use scancode : `GPL2+ WITH exception-ScanCode-knows-about`
What do we do about the exception-ScanCode-knows-about? a License paragraph?


>   It doesn't recognise GPL-2.0+, instead saying "gpl-2.0 or gpl-3.0"

GPL-2.0+ should be recognized OK, and I am not sure I saw exactly that in a dulwich scan rather mixing up things with some LGPL, and that's a bug. I looks like some dulwich notices are one of a kind, FWIW. That's still a bug on our side.

> it should exclude any full license files (currently includes e.g. COPYING)

We track the percentage of license text in a file and whether we matched against a full ext or not. Excluding full license files will need to be done in context of the license key (e.g. the mapping above). Here BTW, there is an important header in COPYING that states this is a choice of license that still may need to be reported?
So in any case we would need to have a way to exclude the standard full texts and only keep notices? 


>    the actual license headers duplicate some lines from the original file

Yes, that's another bug, likely in the `.dumps()` method of debian-inspector.

jelmer:
On Mon, Mar 01, 2021 at 11:08:35AM -0800, Philippe Ombredanne wrote:
> @jelmer thanks. 
> so 1. https://salsa.debian.org/python-team/packages/dulwich/-/tree/5d9a44c4a56f01e2e9eed4a8f0e488bc7025ab93 is the exact tree you scanned
> 2. Let me attach the pastebin here
> [dulwich.copyright.txt](https://github.com/nexB/scancode-toolkit/files/6057988/dulwich.copyright.txt)
> 
> >    This doesn't use the standardized DEP-5 terms for licenses - it should possibly have a dictionary mapping scancode license names to DEP-5 ? https://dep-team.pages.debian.net/deps/dep5/
> 
> Yes we need to draft a mapping. The question is also about all other licenses.
> Say I have `GPL2+ WITH <some exception that ScanCode knows about>`
> What should we dump then?
> - use SPDX? : `GPL2+ WITH LicenseRef-scancode-exception-ScanCode-knows-about`
> - use scancode : `GPL2+ WITH exception-ScanCode-knows-about`
> What do we do about the exception-ScanCode-knows-about? a License paragraph?

DEP-5 is explicit about this:

An exception or clarification to a license is signaled in plain text,
by appending with keywords exception to the short name. This document
provides a list of keywords that must be used when referring to the
most frequent exceptions. When exceptions other than these are in
effect that modify a common license by granting additional
permissions, you may use an arbitrary keyword not taken from the below
list of keywords. When a license differs from a common license because
of added restrictions rather than because of added permissions, a
distinct short name should be used instead of with keywords exception.

Only one exception may be specified for each license within a given
license specification. If more than one exception applies to a single
license, an arbitrary short name must be used instead.

There are only two reserved exceptions at the moment - OpenSSL and
Font. Other exceptions aren't defined - what you call them is up to
you, though it may be good to follow SPDX where possible?

> >   It doesn't recognise GPL-2.0+, instead saying "gpl-2.0 or gpl-3.0"
> GPL-2.0+ should be recognized OK, and I am not sure I saw exactly that in a dulwich scan rather mixing up things with some LGPL, and that's a bug. I looks like some dulwich notices are one of a kind, FWIW. That's still a bug on our side.
They are a bit different since it's dual-licensed under GPL2+ and
Apache-2. However, I would expect at most two licenses to show up
rather than four. It currently seems to match some lines multiple
times.

> > it should exclude any full license files (currently includes e.g. COPYING)

> We track the percentage of license text in a file and whether we matched against a full ext or not. Excluding full license files will need to be done in context of the license key (e.g. the mapping above). Here BTW, there is an important header in COPYING that states this is a choice of license that still may need to be reported?
> So in any case we would need to have a way to exclude the standard full texts and only keep notices? 
Yeah, that's what's done for debian/copyright files - lintian actually
warns when you reference full license files in debian/copyright.

Cheers,

Jelmer

-- 
Jelmer Vernooĳ <jelmer@jelmer.uk>
PGP Key: https://www.jelmer.uk/D729A457.asc

PR #2419, opened 2021-03-02
Allow calling copyright detection from text lines

jelmer:
Factor out detect_copyrights_from_lines(), allowing the copyright code to be
used without pulling in a lot of other libraries (like typecode).

### DIFF
--- a/src/cluecode/copyrights.py
+++ b/src/cluecode/copyrights.py
@@ -17,7 +17,6 @@ from time import time
 from cluecode import copyrights_hint
 from commoncode.text import toascii
 from commoncode.text import unixlinesep
-from textcode import analysis
 
 # Tracing flags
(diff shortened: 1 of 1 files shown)


pombredanne:
Thank you ++
Can you add you DCO signoff?
(Alternatively it is OK to just paste here and I will push that too in the merge commit.)

jelmer:
On Wed, Mar 03, 2021 at 09:35:26AM -0800, Philippe Ombredanne wrote:
> Thank you ++
> Can you add you DCO signoff?
> (Alternatively it is OK to just paste here and I will push that too in the merge commit.)
Done.

-- 
Jelmer Vernooĳ <jelmer@jelmer.uk>
PGP Key: https://www.jelmer.uk/D729A457.asc

pombredanne:
All good. Merging!Thank you. Note that the test failures are not your making.

Issue #942, opened 2018-02-17
Add support for chromium and fuschia package manifest "README"


pombredanne:
See  https://github.com/fuchsia-mirror/third_party-libvpx/blob/master/README.fuchsia and this: https://src.chromium.org/viewvc/chrome/trunk/src/third_party/README.chromium.template

techytushar:
@pombredanne do we have to edit these templates and place them inside packagescan folder?

pombredanne:
@techytushar this would be creating a new package object. Since the code is in flux until the #275 work is merged, I would not work on this just now.

pombredanne:
Another example: https://chromium.googlesource.com/android_tools/+/9a70d48fcdd68cd0e7e968f342bd767ee6323bd1/sdk/platform-tools/systrace/catapult/telemetry/third_party/tsproxy/README.chromium

pombredanne:
@JonoYang I think this is implemented now... Can you confirm?

PR #2450, opened 2021-03-22
Add ability to parse README.* manifest files

steven-esser:
* Add packagedcode module to handle README.* manifest files, which
  contain package information
* Add test cases for many varieties found in the wild

Addresses: #942


src/packagedcode/readme.py
```diff
@@ -0,0 +1,103 @@
...
+    @classmethod
+    def recognize(cls, location):
+        yield parse(location)
+
+    @classmethod
+    def get_package_root(cls, manifest_resource, codebase):
+        return manifest_resource.parent(codebase)
+
+    def compute_normalized_license(self):
+        return models.compute_normalized_license(self.declared_license)
```
steven-esser, line 52: `compute_normalized_license()` does not seem to work for strings like `BSD` and `Apache`. I am guessing this is by design to limit false-positives. However, these are primarily how licenses are listed in `README.*` files.

@pombredanne What would be the best way to do detection on the `declared_license` field here?

pombredanne (approved):
LGTM! merging!

src/packagedcode/readme.py
pombredanne: @majurg we would need a mapping and override compute_normalized_license() here

PR #2475, opened 2021-04-01
Improve README.* package parsing

steven-esser:
Add a README_MAPPING dictionary, used to map various README.* metadata
key-value pairs to scancode Package model values. A number of test cases
have been added as well, to reflect different mappings.

Addresses: #942


JonoYang:
@majurg Looks good. I ran the new version of the README.* parser on some private readme files and noticed that the returned package data does not include a name. A lot of the README.* files I have do not have a name field, or use some other field to specify name. I am wondering if there is a good way to extrapolate Package name for a README.* file from path or another value.

steven-esser:
In a lot of cases the readme does not contain any structured data so no “key” can be mapped as a name. 

Grabbing the parent directory name as a fallback is a good idea and would probably handle many of these cases

JonoYang:
@majurg I've added a bit of code to your PR that sets the Package name to the parent directory basename of the README.* file we are parsing if a Package name was not detected in the README.* file.

PR #2476, opened 2021-04-02
Factor out a detect_copyrights_from_lines() function that doesn't rely on textcode

jelmer:
Factor out a detect_copyrights_from_lines() function that doesn't rely on textcode.

### DIFF
--- a/src/cluecode/copyrights.py
+++ b/src/cluecode/copyrights.py
@@ -17,7 +17,6 @@ from time import time
 from cluecode import copyrights_hint
 from commoncode.text import toascii
 from commoncode.text import unixlinesep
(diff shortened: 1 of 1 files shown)


pombredanne (changes requested):
LGTM!
Thanks. Do you mind to add a DCO signoff?
Either in your commit or in the may text of the PR.
We could also add a simple test?

jelmer:
I've added Signed-Off-By - I can have a look at adding a test later this week.

pombredanne:
Thanks! I can merge as is alright too.

pombredanne:
@jelmer Thanks!

PR #2479, opened 2021-04-07
restore git diff functionality (and fix CI breakage)

tardyp:
all aboutcode repositories have "* binary" in .gitattribute

This has the side effet of disabling the git diff functionality from the commandline.

git diff will always refuse to report diff and say the binary files differ

I can see with the initial CI results that the issue is that windows mess up with the test data when checking it out, and the test fail.

The proposed strategy for that is to explicitly declare the test data as binary with `**/data/**` heuristic.
I applied `git add --renormalize .` to make sure no other data would be affected, and eventually added some more rules for obvious binary data, and forced the .bat to be crlf

ftFinally I let renormalize touch the few files which were still crlf, so that the repo is consitently 'lf'.


tardyp:
I can see the reason for that is test failures. I'll try to fix them while maintaining git diff functionality..

src/licensedcode/data/rules/biosl-4.0_1.yml
```diff
@@ -3,9 +3,9 @@ is_license_notice: yes
 relevance: 100
 notes: seen in https://raw.githubusercontent.com/davidtiefenthaler/BOPT-DemandForecast/main/README.md
 ignorable_copyrights:
-    - Copyright (c) 2009, 2018 Robert Bosch GmbH
+    - Copyright (c) 2009, 2018 Robert Bosch GmbH and its subsidiaries
```
tardyp, line 6: I have no idea why the crlf change made this test fail:
 tests/licensedcode/test_detection_validate.py::TestValidateLicenseExtended::test_validate_detect_biosl_4_0_1_rule

I can reproduce locally, and this change makes the test pass.

tardyp: @pombredanne I see from other PR failures that this is actually a breakage of the develop mainline. not sure why..

pombredanne: This is unrelated to your code!
Sorry fr the confusion, this is fixed mnow.

.gitattributes
```diff
@@ -1,4 +1,13 @@
-# Ignore all Git auto CR/LF line endings conversions
-* binary
+# make sure end of lines are always lf
+*		text eol=lf
+*.bat		text eol=crlf
+
+# data in the tests should be kept as is
+**/data/** binary
```
pombredanne, line 6: Many of the data are text and not binary, but I guess that's minor!

pombredanne (approved):
Thank you++
This looks like a much better solution. Merging!

pombredanne:
@tardyp I guess we will need to propagate that to all other repos then?
Which may be made easier with the https://github.com/nexB/skeleton

PR #2486, opened 2021-04-10
Skip scanning PEM certificates #620

### LINKED ISSUE #620: Do not scan PEM certificate for license or copyright
See for instance https://github.com/nexB/scancode-toolkit/blob/8559b54c9e6b78345f46e0c5ba9b165668be7486/tests/cluecode/data/copyrights/copyright_3a3b02ce_0-a_b_ce.0

ashwanthdurairaj:
skip those files when spotted at the first place

Signed-off-by: ashwanthdurairaj <ashwanthdurairaj2003@gmail.com>

Fixes #620

### Tasks

### DIFF
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -1056,7 +1056,9 @@ def scan_codebase(codebase, scanners, processes=1, timeout=DEFAULT_TIMEOUT,
     # location, resource id)
 
     # NOTE: we never scan directories
-    resources = ((r.location, r.rid) for r in codebase.walk() if r.is_file)
+    skip_file_types = ['.pem', '.cer', '.crt', '.key']
+
+    resources = ((r.location, r.rid) for r in codebase.walk() if r.is_file and r.location[-4 : ] not in skip_file_types)
 
     use_threading = (processes >= 0)
     runner = partial(scan_resource, scanners=scanners,


tardyp:
Hi @ashwanthdurairaj.

This implementation seems a bit specific to me. you are implementing a very specific exclusion inside a very generic piece of code.

Maybe you can implement that in a plugin?

  there is a --exclude command line argument. Maybe we can rather use this method.

Should we update this plugin to exclude some file extensions?

Also you will need to create some unit tests to verify your feature.

ashwanthdurairaj:
@tardyp Thanks for the feedback. I will try to implement that in a plugin.

ashwanthdurairaj:
@tardyp Thanks for the feedback!. I totally agree with your point on my implementation. But is there any plugin for the command --exclude? Because the terminal says there is no such option called --exclude. In which plugin do you exactly want me to implement this?

tardyp:
hi @ashwanthdurairaj  Its actually --ignore and the plugin is src/scancode/plugin_ignore.py

ashwanthdurairaj:
Thanks @tardyp I will look into it

pombredanne:
@ashwanthdurairaj Thanks!

I think you should start this way:
1. in https://github.com/nexB/typecode add tests data with a few of sample certificate files
2. add a new `is_certificate` function there that will return True when the file is a certificate
3. This may need to be exposed in the scan results somehow. The exact place is TBD.
3. Once we have that we could then consider having a new way to possibly ignore some files based on some flags or related conditions

Does this make sense?

ashwanthdurairaj:
@pombredanne Thanks for the feedback. It makes sense. I will look into it.

ashwanthdurairaj:
@tardyp @pombredanne implementing the changes in the plugin would definitely skip the certificates, but this happens only when the user calls the --ignore command right? Also, I feel that implementing these changes at the package level would do good. Why not implement them in resource.py file in commoncode though?? While we implement the identification part in typecode (contenttype.py), the exclusion part can be implemented in resource.py (in Codebase.walk() function to be specific) by importing typecode there, (when the is_certificate function in contenttype.py returns True). We can also create test cases by creating another is_certificate function in tests. Your views on this??

pombredanne:
Optional exclusion for a given type should not be implemented in a type-specific way IMHO.

pombredanne:
I am closing this for now as this needs some more design: it would be best to have a design discussion in #620 first. I hope this makes sense for you.

PR #2531, opened 2021-05-23
scancode_config: improve DistributionNotFound

priv-kweihmann:
In before get_distribution was run in the same scope as the import of
pkg_resources, but in the parent scope of the function
DistributionNotFound was used in the exception handler, leading to the
effect that on a failed import an undefined symbol was part of the ast.

Improve the handling by handing DistributionNotFound in the correct
inner scope, while handling ImportError of the outer one

Signed-off-by: Konrad Weihmann <kweihmann@outlook.com>

### Tasks

### DIFF
--- a/src/scancode_config.py
+++ b/src/scancode_config.py
@@ -72,14 +72,17 @@ def _create_dir(location):
 # current installation location. This is where the source code and static data
 # lives.
 
-
+# in case package is not installed or we do not have setutools/pkg_resources
+# on hand fall back to this version
+__version__ = '21.3.31'
 try:
     from pkg_resources import get_distribution, DistributionNotFound
-    __version__ = get_distribution('scancode-toolkit').version
-except (DistributionNotFound, ImportError):
-    # package is not installed or we do not have setutools/pkg_resources
-    # on hand
-    __version__ = '21.3.31'
+    try:
+        __version__ = get_distribution('scancode-toolkit').version
+    except DistributionNotFound:
+        pass
+except ImportError:
+    pass
 
 system_temp_dir = tempfile.gettempdir()
 scancode_src_dir = dirname(__file__)


src/scancode_config.py
```diff
@@ -75,8 +75,11 @@ def _create_dir(location):
 
 try:
     from pkg_resources import get_distribution, DistributionNotFound
-    __version__ = get_distribution('scancode-toolkit').version
-except (DistributionNotFound, ImportError):
+    try:
+        __version__ = get_distribution('scancode-toolkit').version
+    except DistributionNotFound:
+        __version__ = '21.3.31'
```
pombredanne: I wonder if we could may be define this default before the try block and only once instead?

pombredanne (changes requested):
Thank you! This makes sense. Yet still see my small nit inline.
But how did you manage to find this?

src/scancode_config.py
priv-kweihmann: Sure we could - will spin up a v2 later

priv-kweihmann: Is this better now?

priv-kweihmann:
> Thank you! This makes sense. Yet still see my small nit inline.
> But how did you manage to find this?

I was playing with an update of Jinja2, but when running with the original code with Jinja2!=2.10 I only got `SyntaxError: DistributionNotFound` or something like that... so it walked down that rabbit hole :-).

That's my complete patchset to make scancode work with the latest greatest of dependencies -> https://github.com/priv-kweihmann/meta-sca/commit/8dd7c1ef28a0de94ec4e5580a1029265a799f08b

pombredanne (approved):

pombredanne:
@priv-kweihmann https://github.com/nexB/scancode-toolkit/commit/0aa65604d769c0db60761d17e948e7e271cdd355 :)

PR #2548, opened 2021-06-11
Add new flag in License Data Model definition

### LINKED ISSUE #1675: RFC: Revamp "unknown" license detection
### [Context and Problem](#context-and-problem)

The whole idea of detecting unknown licenses and reporting these with the real detection is a bit weird and confusing. 

We should make this area clearer and cleaner by cleaning up license keys in use and changing the way "unknown" are reported, named and detected. Since we eventually detect many short mentions and do diffs, we may at time catch too many of such "unknown" or not enough and returning these in the same data structure as the detected licenses may be confusing.

I think we can do better.

First let's refine what we mean by unknown: these is a detected texts that is highly likely to be a license text or notice, but that cannot be properly matched to a known, `named` license (one with a proper, not `unknown` license key.

When doing a full scrub on a codebase they are important as they are the things that needs a detailed review.

This is a proposal to revamp and clarify the way detect, process and report these.

There are several things to consider:
- improve the data model for the License "YAML" data
- use fewer license keys for unknown things
- improve the way we deal with multiple "low score" detected licenses in a single file
- report unknowns in a separate section of the scan results, not mixed with the main license detection
- improve the detection of unknown licenses
- follow license referenced in references such as "see COPYING for license" to report the detected license in the referenced file if any. 
- properly handle "intro texts" that are used generically to introduce license terms

### [Improve the License data model definition](#license-data-model-definition) (e.g. the License "YAML" data)

Unknown means for now that this is something matched to a license rule tagged with the "unknown" license key. It would help to clarify which licenses are "special" licenses by having proper attributes. Today we can only identify some "unknown" license based on its key.

Beyond the regular "named" licenses, we have a few "special" license keys that could be made more explicit.

They could be separated in these categories:

1.  "generic" licenses: `commercial-license` `proprietary-license` `generic-cla` `other-copyleft` `other-permissive` `public-domain` `public-domain-disclaimer` `warranty-disclaimer`

All of these are used as an alternative to a named license (and sometimes for rarer licenses not worthy of a named entry). These are all generic "catch-all" license keys. Each match can reference a different text or notice: they do not have a self-standing reference license text stored with the .LICENSE file. We could make these more explicit by adding a `is_generic` license attribute, to mark them as being different from the default "named" licenses.

2.  "unknown" licenses: `unknown` `free-unknown` `unknown-spdx` `unknown-license-reference`

These are used to depict a detection that is likely about licenses but is NOT matched to a proper rule for a named or generic license.
Instead this is matched to a rule tagged with one of these `unknown` license keys.

3.  :heavy_check_mark: Also we have `unknown-license-reference` that are more like generic intro texts about licensing. The .RULE file names have been prefixed with "lead-in".
For instance "Software License agreement" would be such as an intro text. 

##### So to handle these all these, I suggest we could add these [new model attributes](#new-model-attributes) :

~~In the License (and also to the License Rule):~~

- ~~`is_generic` tells that a license is generic (case 1.)~~
- ~~`is_unknown` tells that a license is about some unnamed license. (case 2.)~~

~~In the License Rule:~~

- ~~`is_license_intro` to tell that a license detection rule is some license intro text. (case 3.). This would be used in conjunction with the `unknown` license key.~~

All done:
**:heavy_check_mark:  implemented with `is_generic` flag
:heavy_check_mark:  implemented with `is_unknown` flag for licenses and `has_unknown` property for rules
:heavy_check_mark:  This has been implemented with `is_license_intro` flag** 

### [License keys cleanup and retirement](#license-keys-cleanup-and-retirement) : use fewer license keys for unknown things.

- The `free-unknown` is a weird one that should be retired: we should return either `unknown` or `other-copyleft` or `other-permissive` instead.

- `unknown-spdx` is used to report some unknown license in an SPDX license expression. Since this is only produced by the SPDX expression detection it would unlikely be part of other "unknown" detection. It should be tagged with `is_unknown`

- `unknown-license-reference` is special if and only if it has a `referenced_filenames`. So it could be folded in `unknown` too.
The ones that have a `referenced_filenames` will be dealt with accordingly. Otherwise any rule can be tagged with `is_license_reference` to the same effect.

- We have also a category of unknown license rules that have been for now renamed with the prefix `lead-in_` that are not really depicting unknown licenses but rather are intro or title texts used before a license text or notice. For instance `Software license agreement` or `As a special exception` or `is under the following license` etc.  These are really special and are of interest if and only if they show up with otherwise unknown license texts. When they are detected just before an actual known, named license text or notice, reporting them as unknown is noisy and instead their match should be folded in the main match they introduce. We should tag these as `unknown` and `is_license_intro`.

#### improve the way we [deal with multiple "low score"](#deal-with-multiple-low-score) detected licenses in a single file

We could merge matches to different detected "rules" in the same text region when they point to the related license in one match.

We could also handle the cases where we have some mentions of bare GPL (detected as gpl-1.0-plus) together with some GPL version: these could be merged in some case as one match to the versioned GPL. This is mostly for FSF licenses 

### [Report unknowns separately](#report-unknowns-separately)

We should use a separate section of the scan results to report "unknown" license detections, not mixed with the main license detection for clarity.
This could be new section as "unknown_license_references" where we report the matched text and positions but we cannot report a specific matched license rule or key.

### [Improve the detection of unknown licenses](#improve-the-detection-of-unknown-licenses)

Beside or as a replacement to the actual detection of the "unknown" license rules, we should have a new way and more efficient way to detect unknown licenses using ngrams. The process would be roughly:

- run the regular license detection.
- remove from results and keep aside any match with a low coverage below a threshold (eventually merge them too)
- collect all the spans of scanned text that are not matched to any license
- run these spans through an automaton index that will contain ngrams from all regular license texts and rules
- merge all these matched spans (and the possible also the weak matches) in a single matched span.
  Consider also including any "unknown" rule matches.
- splits in multiple spans based on having some large enough gaps in the match.
- report these as "unknown_license_references" 

### [Follow license references to another file](#follow-license-references-to-another-file)

This is for references such as "see COPYING for license" to report the detected license in the referenced file if any e.g. mentions to look for license details in another file.

We should have a way to follow license references using the `referenced_filenames` attribute and find the detected license in these filenames. And when such a conclusive reference is found positively, we should update the match to use the referenced license(s) that were detected.

This is tracked in https://github.com/nexB/scancode-toolkit/issues/1364

Note that for now, this does NOT include following URLs which would imply having network access.

### [Properly handle "intro texts"](#properly-handle-intro-texts) that are used generically to introduce license terms

The are "lead-in"-like intro texts and are detected alone and qualified as  "unknown" today. For instance "License agreement" is a rule that would be detected as such.

We should have a way to detect such a text fragment and to discard or merge such a detection if this is immediately tied to an actual "named" license detection. This would avoid many unknown detections.

### LINKED ISSUE #2549: Add new is_unknown flag in data model definition
Add a new flag for licenses that are tagged as "unknown" i.e. not matched with a known named license. 
This will help in segregating the special license keys with the known ones.
## Possible Labels

- new feature

## Select Category

-  [x] Enhancement
-  [ ] Add License/Copyright
-  [x] Scan Feature
-  [ ] Packaging
-  [ ] Documentation
-  [ ] Expand Support
-  [ ] Other

akugarg:
Signed-off-by: akugarg <akanksha.garg2k@gmail.com>

Fixes #2549 

### Tasks


AyanSinhaMahapatra:
@akugarg btw when you [link the issue using the keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) it'll be closed. This is a general big-picture issue, so feel free to create smaller issues and link to them.

Also, the commit messages are the same here and they should be more specific and different IMO. Atleast:

- Add `is_unknown` flag in data model
- Add `is_unknown` flag to data files of unknown rules

If you could rebase to change the messages that'd be great.

akugarg:
@AyanSinhaMahapatra I have made the requested changes.

AyanSinhaMahapatra:
1. Did you do this with the script checking if the `license-expression` in .yml file has `unknown` in it?

From discuss@gitter
Akanksha Garg:
are there instances where license can be considered unknown and may not be indicated in file name?

Ayan Sinha Mahapatra:
File name is just a naming convention. The key value pairs in the .yml files is what are used in scancode, and not the filename. And these are structured .yml files so faily easy to read and search in them specificly.

I see two issues:

- I see cases of adding `is_unknown` flag in false-positive rules (i.e. `is_false_positive` flag set to true True), which should not be the case. 
- There are files which don't have `unknown` in their names, but have `unknown` in their license expressions also. These are files with `is_license_intro` set to True and some other cases. 

2. Paste the script here for future references.

akugarg:
import os, glob

os.chdir("git/scancode-toolkit/src/licensedcode/data/rules")
print("Changed")
to_find="unknown"
fp="is_false_positive: yes"
for file in glob.glob("*.yml"):
    f=0
    file1 = open(file,"r+")
    for line in file1:
        if fp in line:
            break
        else:
            if to_find in line:
                file1.write("is_unknown: yes")
                break

Used the above script for finding the "unknown" tag in license expression and for ignoring the false-positive case.
@AyanSinhaMahapatra Hope this works.
I have not yet added for category "unknown-spdx"

AyanSinhaMahapatra:
@akugarg btw, in addition to processing the `unknown-spdx` case, we also need to 

1. add the `is_unknown` flag to the License class also, like in the Rule, and add them to the respective .yml files in the licenses folder. Like [here](https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/free-unknown.yml)
2. Add a test to the validate_licenses funtion, (which is run when loading the licenses/rules) as a sanity check to make sure every rule with `is_unknown` set true has `unknown` in it's license expression. This means adding the test to the Rule.validate() function here: https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/models.py#L866

akugarg:
For adding to license class I have to add similar tag  like ones present here https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/models.py#L102 set to default = false 
right?

AyanSinhaMahapatra:
Yes!

akugarg:
When I am trying to add flag for license class and trying to run a scan for samples directory it is showing this error..
error
For this https://github.com/nexB/scancode-toolkit/blob/develop/samples/JGroups/LICENSE

AyanSinhaMahapatra:
@akugarg you could push your changes for addition of the flag to License Model if you want us to have a look at it (there's no point guessing where the failure is without the code), and always run the specific tests for licensedcode first and fix failures there before actually running a whole scan.

Also, you could also try to fix your tests first for the addition of `is_exception` to the Rule model only, instead of trying to debug both tasks at once.

akugarg:
OKay! I will first fix the test cases for flag addition to rule class only.

akugarg:
@AyanSinhaMahapatra Please have a look at this once!

AyanSinhaMahapatra:
@akugarg The test expectations need to be changed in many cases like [this](https://dev.azure.com/nexB/scancode-toolkit/_build/results?buildId=3978&view=logs&j=a4427093-5c58-53ff-5e9f-a15ae68585b0&t=5e3c828d-e114-5dc5-9182-4c462c05ab86&l=8666). Btw, note that you can set `regen=True` in the tests to regenerate and then check the diff carefully, to see if the changes make sense, it's faster. Be careful to not commit regen as True, and do this one function at a time.

There's also another test to be added, which fetches the corresponding license object for `is_unknown` rules and checks if the `is_unknown` falg for the license is set True (and category: Unstated License), 

We're nearly there! :+1:

akugarg:
@AyanSinhaMahapatra do you mean adding a validating test like you mentioned earlier ? I have already made a commit for that https://github.com/nexB/scancode-toolkit/pull/2548/commits/1d7872d57d794c79bf63b6b5aae7e77c21eaaad8
Or something else?

AyanSinhaMahapatra:
This test at 1d7872d checks if a rule with `is_unknown` set to True has `unknown` in it's license-expression. 

We should add two more tests also in the same validate function:

1. If a license rule has `unknown` in it's license-expression, it has  `is_unknown` set to True (checking the opposite).
2. If a license rule has `is_unknown` set to true, then get the license object that it is matched to (they are the license data files you added `is_unknown` to), and check if the license object has `is_unknown` set to True. This checks the `is_unknown` flag addition in the License model.

src/licensedcode/models.py
```diff
@@ -343,6 +343,10 @@ def validate(licenses, verbose=False, no_dupe_urls=False):
                 )
             if not lic.owner:
                 error('No owner')
+                
+            if lic.is_unknown:
+                if not "unknown" in lic.name:
```
AyanSinhaMahapatra: You should check in lic.key and not in lic.name, this is failing tests [here](https://dev.azure.com/nexB/scancode-toolkit/_build/results?buildId=4015&view=logs&j=a4427093-5c58-53ff-5e9f-a15ae68585b0&t=5e3c828d-e114-5dc5-9182-4c462c05ab86&l=1173)

output.json
```diff
@@ -0,0 +1,2346 @@
+{
```
AyanSinhaMahapatra: I think this file was commited by mistake?

AyanSinhaMahapatra (changes requested):

output.json
akugarg: Yup , Deleted

pombredanne (changes requested):
Thanks!
All your YAML files seem to be missing a trailing LF. The content does not look in the same order as the object too.

How did you create these? e.g. did you load all the Rule objects them dump them back?
Which script did you use?

akugarg:
> import os, glob
> 
> os.chdir("git/scancode-toolkit/src/licensedcode/data/rules")
> print("Changed")
> ans=0
> to_find="unknown"
> fp="is_false_positive: yes"
> for file in glob.glob("*.yml"):
> f=0
> file1 = open(file,"r+")
> for line in file1:
> if fp in line:
> break
> else:
> if to_find in line:
> ans=ans+1
> file1.write("is_unknown: yes")
> break
> print(ans)
> 
> Used the above script for finding the "unknown" tag in license expression and for ignoring the false-positive case.
> @AyanSinhaMahapatra Hope this works.
> I have not yet added for category "unknown-spdx"

I used this.

akugarg:
> import os, glob
> 
> os.chdir("git/scancode-toolkit/src/licensedcode/data/rules")
> print("Changed")
> to_find="unknown"
> fp="is_false_positive: yes"
> for file in glob.glob("*.yml"):
> f=0
> file1 = open(file,"r+")
> for line in file1:
> if fp in line:
> break
> else:
> if to_find in line:
> file1.write("is_unknown: yes")
> break
> 
> Used the above script for finding the "unknown" tag in license expression and for ignoring the false-positive case.
> @AyanSinhaMahapatra Hope this works.
> I have not yet added for category "unknown-spdx"

Used this one .

akugarg:
In order means that instead of having the flag added at end in YAML files as done here we should have inserted in between (may be after license expression or some flags) according to as defined in object definition right?
Also can I proceed as follows :
add this flag above other flags instead in definition itself.
then add this flag after license expressions for the required files.

AyanSinhaMahapatra:
@akugarg I think you should instead use [load_rules()](https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/models.py#L626) and then check for unknown in license expression explicitly, set the `is_unknown` flag to be `True` for them, and then use [Rule.dump()](https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/models.py#L1180) to overwrite the rules that should have `is_unknown` as `True`. This should take care of both the order and missing line returns.

There are similar functions for licenses also.

You should also merge to latest before doing this (delete the commits adding the is_unknown flag to .yml files by rebasing, merge to latest develop, and then add the changes as suggested), some more unknown rules were added.

src/licensedcode/models.py
```diff
@@ -754,6 +760,10 @@ class BasicRule(object):
     # detected immediately after.
     is_license_intro = attr.ib(default=False, repr=False)
 
+    #unknown licenses are the detected texts are highly likely to be a license
```
pombredanne: Please add one space after the # comment sign

src/scancode/api.py
```diff
@@ -239,6 +239,7 @@ def _licenses_data_from_match(
         result['short_name'] = lic.short_name
         result['category'] = lic.category
         result['is_exception'] = lic.is_exception
+        result['is_unknown']=lic.is_unknown
```
pombredanne: The spacing needs some fixing:
```suggestion
        result['is_unknown'] = lic.is_unknown
```

pombredanne:
@akugarg this is ready to merge imho but you have one failing test

akugarg:
@pombredanne I am not able to figure out what's causing this failing check or if it is something related to what I added.

akugarg:
@AyanSinhaMahapatra 
All green! :)

src/scancode/api.py
AyanSinhaMahapatra: I think this should be `result['has_unknown']=match.rule.has_unknown` right @pombredanne ?

src/scancode/api.py
```diff
@@ -270,6 +271,7 @@ def _licenses_data_from_match(
         matched_rule['is_license_reference'] = match.rule.is_license_reference
         matched_rule['is_license_tag'] = match.rule.is_license_tag
         matched_rule['is_license_intro'] = match.rule.is_license_intro
+        matched_rule['has_unknown'] = match.rule.has_unknown
```
AyanSinhaMahapatra, line 274: I think it should be present only once, and probably outside `matched_rule` since we are reporting this as a property of the match, and not rule.

AyanSinhaMahapatra (commented):
Looking good now.

src/scancode/api.py
akugarg: For license class "is_unknown" attribute is still there! So we need to remove that also?

akugarg: https://github.com/nexB/scancode-toolkit/pull/2548/files?file-filters%5B%5D=.csv&file-filters%5B%5D=.py#diff-af1613c351a0c29790c9720346fc4e323a7e0470fd3a6cc6ae8c965ca121fd3cR112
this one.

AyanSinhaMahapatra: No that stays. The point was to get [`has_unknown`](https://github.com/nexB/scancode-toolkit/pull/2548/files?file-filters%5B%5D=.csv&file-filters%5B%5D=.py#diff-af1613c351a0c29790c9720346fc4e323a7e0470fd3a6cc6ae8c965ca121fd3cR873) from getting it's license object and from there as opposed to just checking in `license_expression` (which is a hackish way to do it).

src/scancode/api.py
pombredanne: @AyanSinhaMahapatra It is fine to keep for now at the rule level. This is the way for other attributes.

pombredanne (approved):
There are a few nits for your consideration, but this is good to go otherwise! Thanks

src/licensedcode/models.py
akugarg: But this doesn't exist anymore!

pombredanne: My bad... this was outdated indeed!
I pushed a few updates to your branch BTW.

pombredanne (approved):
All good!
Merging

PR #2582, opened 2021-07-06
Detect installed windows programs

JonoYang:
This PR enhances Windows package detection by introducing  new functions that detects installed programs from Windows registry files.


src/packagedcode/__init__.py
```diff
@@ -82,6 +82,7 @@
     win_pe.WindowsExecutable,
     readme.ReadmePackage,
     build.MetadataBzl,
+    windows_helpers.MsiInstallerPackage,
```
pombredanne: what about windows.py rather than windows_helpers.py?

src/packagedcode/windows_helpers.py
```diff
@@ -0,0 +1,311 @@
+import os
```
pombredanne: You are missing some comment/license header

src/packagedcode/windows_helpers.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+logger = logbook.Logger(__name__)
+TRACE = False
+
+
+MSIINFO_BIN_LOCATION = 'packagedcode_msitools.msiinfo'
+
+
+def get_msiinfo_bin_location():
```
pombredanne: Consider using a cascade of possible locations for this like in https://github.com/nexB/extractcode/blob/632f6a8b91bc7d81a6d5ab5743b34779d42c349d/src/extractcode/sevenzip.py#L66

src/packagedcode/windows_helpers.py
```diff
@@ -0,0 +1,311 @@
...
+    if stderr:
+        error_message = f'Error encountered when reading MSI information from {location}: '
+        error_message = error_message + stderr
+        raise MsiinfoException(error_message)
+    return parse_msiinfo_suminfo_output(stdout)
+
+
+def msi_parse(location):
+    """
+    TODO: get proper package name and version from MSI
```
pombredanne: Do you think you could add this before we merge?

pombredanne: BTW if you ever need to extract some version from a name, this may help https://github.com/nexB/commoncode/blob/452c2b513231460d5fa97bccb0acfafa16356682/src/commoncode/version.py
so may libversion from repology?

src/packagedcode/windows_helpers.py
pombredanne: Feel free to split msi vs. registry in tow different  python modules such as msi.py and winreg.py

src/packagedcode/windows_helpers.py
```diff
@@ -0,0 +1,311 @@
+import os
+from pathlib import PureWindowsPath
+
+import attr
+import logbook
```
pombredanne: Let's not use a new logging lib for now.

tests/packagedcode/data/windows_helpers/msi/7z1900-x64.msi.expected
```diff
@@ -0,0 +1,58 @@
+{
```
pombredanne: This is too big, LGPL-licensed. To make things simpler I would consider tow things:
1. have tests that work off the msiinfo output only (eventually refactoring your code along these lines to separate running msitools from parsing the output)
2. have a few minimal tests with tiny .msi files to test the binary parsing proper

src/packagedcode/windows_helpers.py
```diff
@@ -0,0 +1,311 @@
...
+from regipy.exceptions import NoRegistrySubkeysException
```
pombredanne: If you split the registry code in a separate module (or not) ... it would make sense to have regipy as an optional extra dep and guard at least the imports inside the functions using them or with a try/except to ensure that the module always imports at the minimum.

src/packagedcode/windows_helpers.py
```diff
@@ -0,0 +1,311 @@
...
+        if not os.path.exists(software_registry_loc):
+            continue
+        for installed_program in reg_parse(software_registry_loc):
+            installed_program.populate_installed_files(root_dir)
+            yield installed_program
+
+
+@attr.s()
+class InstalledWindowsProgram(models.Package):
+    filetypes = ('SOFTWARE',)
```
pombredanne: Let's not put something here if we do not know

src/packagedcode/windows_helpers.py
```diff
@@ -0,0 +1,311 @@
...
+    mimetypes = ('application/octet-stream',)
+    default_type = 'windows-program'
```
pombredanne: What about just "windows" ? We need to find a good set of types here and for msi and the related win_pe and possibly MS Cabinets.

src/packagedcode/windows_helpers.py
```diff
@@ -0,0 +1,311 @@
...
+    def recognize(cls, location):
+        for installed in reg_parse(location):
+            yield installed
+
+    def populate_installed_files(self, root_dir):
+        install_location = self.extra_data.get('install_location')
+        if not install_location:
+            return
+
+        # Manipulate install_location to fit extracted Docker image structure
```
pombredanne: .. or a VM image

src/packagedcode/windows_helpers.py
```diff
@@ -0,0 +1,311 @@
...
+        install_location = PureWindowsPath(install_location)
+        # Remove leading drive letter ("C:\\") and replace all spaces with underscore
+        install_location_no_root = install_location.relative_to(*install_location.parts[:1])
+        install_location_no_root = str(install_location_no_root)
+        install_location_no_root = install_location_no_root.replace(' ', '_')
```
pombredanne: Why this replace?

src/packagedcode/windows_helpers.py
```diff
@@ -0,0 +1,311 @@
...
+        install_location_in_image = PureWindowsPath(root_dir).joinpath('Files', PureWindowsPath(install_location_no_root))
```
pombredanne: I would use posix path all the way and if you get Windows path from the registry, then convert them first to POSIX the only thing we could remove is the drive letter. 
If there is some issues with inconsistencies wrt. alignment with scancode.io extracted paths, that's a bug to solve there.

tests/packagedcode/data/windows_helpers/registry/Software_Delta-installed.expected
```diff
@@ -0,0 +1,310 @@
...
+    "notice_text": null,
+    "root_path": null,
+    "dependencies": [],
+    "contains_source_code": null,
+    "source_packages": [],
+    "extra_data": {},
+    "purl": "pkg:windows-program/Python%203.9.5%20Tcl/Tk%20Support%20%2864-bit%29@3.9.5150.0",
+    "repository_homepage_url": null,
+    "repository_download_url": null,
+    "api_data_url": null
```
pombredanne: What about the installed files?

pombredanne (changes requested):
Some nit for your consideration!

pombredanne:
Thanks! let me review now... Note that I created https://github.com/mkorman90/regipy/ as an example of a way to get fewer deps with regipy (and explicit support for Python 3.6 and 3.9 who both work fine with it)

src/packagedcode/__init__.py
JonoYang: The msi package code has been placed in msi.py

src/packagedcode/windows_helpers.py
JonoYang: I've made the split and added the header to both files.

src/packagedcode/windows_helpers.py
JonoYang: Done :+1:

src/packagedcode/windows_helpers.py
JonoYang: I left this as `windows-program`

src/packagedcode/windows_helpers.py
JonoYang: I've removed the import to logbook, but it will still be installed as a dependency for regipy

src/packagedcode/windows_helpers.py
JonoYang: Thanks. I modified the code there to get version number from the `subject` line of the `msiinfo` results here: https://github.com/nexB/scancode-toolkit/blob/cab062d14f06273bc173b651e57ebb95ffa6aa32/src/packagedcode/msi.py#L148

src/packagedcode/windows_helpers.py
JonoYang: ack

src/packagedcode/windows_helpers.py
JonoYang: I've made the changes here: https://github.com/nexB/scancode-toolkit/blob/cab062d14f06273bc173b651e57ebb95ffa6aa32/src/packagedcode/win_reg.py#L257

src/packagedcode/windows_helpers.py
JonoYang: This was a workaround that I have now removed. Extractcode would replace spaces in file/directory names and the code would not find the file on my system.

JonoYang: https://github.com/nexB/extractcode/issues/27

tests/packagedcode/data/win_reg/get_installed_packages_docker/expected-results
```diff
@@ -0,0 +1,50 @@
...
+  "copyright": null,
+  "license_expression": null,
+  "declared_license": null,
+  "notice_text": null,
+  "root_path": null,
+  "dependencies": [],
+  "contains_source_code": null,
+  "source_packages": [],
+  "extra_data": {
+    "install_location": "C:\\Program Files\\Test\\",
```
JonoYang, line 66: Consider posix-izing the path

Issue #2446, opened 2021-03-19
Copyright detection range is sometime approximative


tdruez:
input.txt:

```
Files: src/macos/src/macos_main.cpp
Copyright: 2001, Eric Lavigne
License: curl

Files: docs/examples/rtsp.c
Copyright: 2011, Jim Hollinger
```

`scancode -ce --json-pp - input.txt`

Current results:
```json
"copyrights": [
    {
      "value": "Copyright 2001, Eric Lavigne",
      "start_line": 2,
      "end_line": 3
    },
    {
      "value": "Copyright 2011, Jim Hollinger",
      "start_line": 6,
      "end_line": 6
    }
],
```

Expected results:
```json
"copyrights": [
    {
      "value": "Copyright 2001, Eric Lavigne",
      "start_line": 2,
      "end_line": 2
    },
    {
      "value": "Copyright 2011, Jim Hollinger",
      "start_line": 6,
      "end_line": 6
    }
],
```

The first copyright is detected on line 2 to 3, while only present on line 2, where the second copyright is properly detected on line 6 only.

sritasngh:
@tdruez can I work on this issue?

itssingh:
Input text: 
```
Files: src/macos/src/macos_main.cpp
Copyright: 2001, Eric Lavigne 
License: curl
All rights reserved

Copyright (c) 1991, 1992 Paul Kranenburg <pk@cs.few.eur.nl>
Copyright (c) 1993 Branko Lankester <branko@hacktic.nl>
Copyright (c) 1993 Ulrich Pegelow <pegelow@moorea.uni-muenster.de>
Copyright (c) 1995, 1996 Michael Elizabeth Chastain <mec@duracef.shout.net>
Copyright (c) 1993, 1994, 1995, 1996 Rick Sladkey <jrs@world.std.com>
Copyright (C) 1998-2001 Wichert Akkerman <wakkerma@deephackmode.org>
All rights reserved.
```

Result:
>  "copyrights": [
        {
          "value": "Copyright 2001, Eric Lavigne <pk@cs.few.eur.nl>",
          "start_line": 2,
          "end_line": 4
        },
        {
          "value": "Copyright (c) 1991, 1992 Paul Kranenburg <pk@cs.few.eur.nl>",
          "start_line": 6,
          "end_line": 12
        },
        {
          "value": "Copyright (c) 1993 Branko Lankester <branko@hacktic.nl>",
          "start_line": 6,
[...]

Issue #2172, opened 2020-08-25
Incorrect copyright detection


pombredanne:
### Description

With `+// Copyright 2018-2019 @paritytech/substrate-light-ui authors & contributors` the copyright is detected as :

```
      "copyrights": [
        {
          "value": "Copyright 2018-2019",
          "start_line": 1,
          "end_line": 1
        }
      ],
      "holders": [],
      "authors": [
        {
          "value": "& contributors",
          "start_line": 1,
          "end_line": 1
        }
      ],
```

pombredanne:
For instance from https://github.com/paritytech/substrate-light-ui/blob/8bd2362a0c68b1d6835e2a19d73e0dc896cff5e3/packages/electron-app/src/main/app/menu.ts

FlorianSW:
Just to clarify: I opened a similar issue with ClearlyDefined, and it seems the underlying issue is this one as well :) The issue in CD would be: https://github.com/clearlydefined/service/issues/767

Just for reference and one more example. If it is not related, sorry for the noise :)

pombredanne:
@FlorianSW Thank you for the report!
Your case of https://github.com/yargs/y18n/blob/45d2568800d6c57be045e76dc4984b2ef3432233/LICENSE#L1
`Copyright (c) 2015, Contributors` is actually slightly different from this case here, but that's OK.

When using this  `SCANCODE_DEBUG_COPYRIGHT=yes scancode -c --json-pp -  menu.ts`:

- The case `Copyright 2018-2019 @paritytech/substrate-light-ui authors & contributors`: is parsed as: ` [('Copyright', 'COPY'), ('2018-2019', 'YR'), ('paritytech/substrate-light-ui', 'NN'), ('authors', 'AUTHS'), ('&', 'CC'), ('contributors', 'CONTRIBUTORS')]` 

- Your new case `Copyright (c) 2015, Contributors` is parsed as `[('Copyright', 'COPY'), ('(c)', 'COPY'), ('2015,', 'YR'), ('Contributors', 'CONTRIBUTORS')]`

Therefore the grammar rule that will collect them alright is different. The key is that the word `Contributors` is is parsed as it's own tag as it is special and can have different meanings in different contexts

FlorianSW:
Oh, ok, thanks for your answer. Should I open a new issue for that or is it ok to have it within this one for now as well? :)

pombredanne:
@FlorianSW No need for another ticket, they are fine together

[...]

Issue #1565, opened 2019-05-14
Copyright is incomplete when there is a newline between


MMarwedel:
Hi,
if there is a newline between the copyright and the name of the company, the company does not show up in the "copyrights": as value.
Sample which goes wrong (actually not exactly the file I scanned, but same formatting and from the same company):
https://github.com/espressif/arduino-esp32/blob/master/tools/sdk/include/newlib/sys/features.h
"copyrights": [
        {
          "value": "COPYRIGHT (c) 1989-2014.",
          "start_line": 4,
          "end_line": 4
        }
      ],

Sample which works properly:
https://github.com/esp8266/Arduino/blob/master/tools/sdk/libc/xtensa-lx106-elf/include/sched.h
"copyrights": [
        {
          "value": "COPYRIGHT (c) 1989-2010. On-Line Applications Research Corporation",
          "start_line": 4,
          "end_line": 5
        }
      ],

pombredanne:
@MMarwedel Thank you for this. New, empty lines are used to decide which lines should be considered for a copyright detection. And empty line is this context is considered as a break and not part of the detection. The difficulty is to determine when this new lines is relevant and a break and when this is not not. In the vast majority of the cases, this is a break. 
Is this something you have seen often or just a few odd times?

MMarwedel:
I agree, its difficult to decide which part of the text belongs to which other part. In fact I only found one file with some "(c) DATE" but no autor or company following in the results.
I however cant say if I missed some
"(c) DATE author1 and

author2"

cases.

pombredanne:
as weird as it sounds the likely best way to fix this is to fix this in newlib with a patch that would remove the empty line. I ran a quick test with a naive way to consider an empty as possibly part of a copyright statement and it created massive test failures. This is really uncommon.  
I sent a patch: https://github.com/espressif/arduino-esp32/pull/2787

As for cases "(c) DATE author1 and

author2"
they would be really weird and would be even harder to parse. It really looks like the established convention is not to put empty lines in the middle of a copyright statements. So if there is one, we have to assume that this is either a mistake and we can help fix this or this is not part of such statement.

MMarwedel:
Okay, patching this in newlib, I guess is the best solution. The right place might be https://www.sourceware.org/newlib/ but I failed to find the file there and provide a link, so espressif was just a random other place to find the file online.

pombredanne:
[...]

PR #2590, opened 2021-07-11
Improve copyrights detection

pombredanne:
This PR makes several improvement to copyright detection

- lines are now tracked correctly #2446
- there is a massive performance boost: as an example a run of the copyright test suite now takes about 55s and used to take 178s in v21.6.7 which means that copyright detection takes roughly 1/3 of what it used to take to run. This comparison was made using the same machine, OS and Python version. This is mostly due to the adoption of the pygmars
- consider multiple lines with one empty line in some cases #1565
- various minor copyright detection improvements such as #2172

And other minor improvements


pombredanne:
This is all green. Merging now

PR #2595, opened 2021-07-13
2594 report proper bazel rules

JonoYang:
This PR introduces a fix that prevents the bazel/buck parser from reporting all build rules, only limiting reporting to build rules that end with `binary` or `library`.

I've also set `regen=False` to the package cli tests as some of them were set to True.


pombredanne (approved):
LGTM. Merging. Thanks!

PR #2606, opened 2021-07-27
Parse mui files

JonoYang:
This PR adds more valid extensions to the WinPE Package parser so we can collect metadata from more Windows files.


pombredanne (approved):
LGTM! Thank you ++

PR #2607, opened 2021-07-27
Update debian package getter

JonoYang:
The Debian package getter has been updated to report detected packages from individual Debian status files in the `/var/lib/dpkg/status.d/` directory if `/var/lib/dpkg/status` does not exist. This helps in the case of some distroless images, where `/var/lib/dpkg/status` does not exist, but there are individual status files for packages in the `/var/lib/dpkg/status.d/` directory.


pombredanne (approved):
All green! Thanks!

PR #2616, opened 2021-07-30
Follow license reference to another file

### LINKED ISSUE #1364: Follow references to license in another file:  ScanCode reports the "SEE LICENSE IN <filename>" text in an NPM package.json as "Unkown"
For non-SPDX / propretary licenses, NPM [suggests](https://docs.npmjs.com/files/package.json#license) to use a value of "SEE LICENSE IN <filename>" for the "license" key in `package.json`. If the license file is called `LICENSE`, that text ends up to be "SEE LICENSE IN LICENSE". At least ScanCode 2.9.7 reports this string as an "Unknown" license when scanning `package.json`. IMO, that text should not trigger a license finding in this case.

akugarg:
Signed-off-by: akugarg <akanksha.garg2k@gmail.com>

Fixes #1364 

### Tasks


akugarg:
Hey! @AyanSinhaMahapatra  How we will go about adding same logic for packagecode like we discussed yesterday?

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +118,57 @@ def get_scanner(
             license_text_diagnostics=license_text_diagnostics,
             license_url_template=license_url_template
         )
+    def process_codebase(self, codebase, **kwargs):
```
AyanSinhaMahapatra, line 121: One line should be present between functions in the same class declaration.

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +118,57 @@ def get_scanner(
...
+    modified = False
+
+    for match in matches:
+        ref_files=match.rule.referenced_filenames
+        if len(ref_files) != 0:
+            for i in range(len(ref_files)):
+                if not ref_files[i].startswith('usr/share/common-licenses'):
+                 new_loc=find_reference_file(location,ref_files[i])
+                 if new_loc != None:
+                   new_lic=get_licenses(new_loc, min_score=0)
```
AyanSinhaMahapatra: This is not the correct way to do this.

Currently you are:

1. Finding the directory locally with the location.
2. Doing a license detection on the file that you have found.

The approach should have been, as mentioned [here](https://github.com/nexB/scancode-toolkit/issues/1364#issuecomment-888988049) to 

1. get the parent resource from the current resource
2. check if the referenced filename is present in the child resources of this parent resource.
3. if resource object with referenced filename present, get the licenses from the from the resource object. 
4. then add the matches and license expression.

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +118,57 @@ def get_scanner(
...
+                   licenses.extend(new_lic['licenses'])
+                   license_expressions.extend(new_lic['license_expressions'])
+                   modified = True
```
AyanSinhaMahapatra: The flag seems redundant, we can save resource directly without using flags.

tests/licensedcode/test_plugin_license.py
```diff
@@ -67,6 +67,13 @@ def test_license_option_reports_license_texts_diag_long_lines():
     test_loc = test_env.get_test_loc('plugin_license/text_long_lines/scan-diag.expected.json')
     check_json_scan(test_loc, result_file, regen=False)
 
+def test_license_match_reference():
```
AyanSinhaMahapatra, line 75: Also add test to check this does not perform any change if the file is not found.

tests/licensedcode/data/plugin_license/license_reference/scan/license-notice.txt
```diff
@@ -0,0 +1 @@
+For licensing information, see the file 'LICENCE' in this directory.
```
AyanSinhaMahapatra: would be nice to also have the package.json mentioned [here](https://github.com/nexB/scancode-toolkit/issues/1364#issuecomment-869995820) and the corresponding LICENSE file as test.

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +118,57 @@ def get_scanner(
...
+    license_expressions = resource.license_expressions
+    if not licenses:
+        return 
+
+    location = resource.location
+    
+    from licensedcode import cache
+    from scancode.api import get_licenses
+    idx = cache.get_index()
+    matches = idx.match(
```
AyanSinhaMahapatra: The licenses we have above from `licenses = resource.licenses` are the same list of LicenseMatch objects that you'd get here in `matches`. This is a post processing step, we already have scanned for the licenses in the get_scanner function.

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +118,57 @@ def get_scanner(
...
+                   new_lic=get_licenses(new_loc, min_score=0)
+                   licenses.extend(new_lic['licenses'])
+                   license_expressions.extend(new_lic['license_expressions'])
+                   modified = True
+
+    if modified:
+        codebase.save_resource(resource)
+    return resource
+
+def find_reference_file(location,referenced_filename):
```
AyanSinhaMahapatra: two lines between functions which are not present inside a class declaration.

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +118,57 @@ def get_scanner(
...
+    
+    modified = False
+
+    for match in matches:
+        ref_files=match.rule.referenced_filenames
+        if len(ref_files) != 0:
+            for i in range(len(ref_files)):
+                if not ref_files[i].startswith('usr/share/common-licenses'):
+                 new_loc=find_reference_file(location,ref_files[i])
+                 if new_loc != None:
```
AyanSinhaMahapatra: `if new_loc:` should be enough?

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +118,57 @@ def get_scanner(
...
+    returns the found location
+    """
+    file_name=referenced_filename
+    par_dir=fileutils.parent_directory(location)
+    for root, dirs, files in os.walk(par_dir):
+      if file_name in files:
+         path_file = os.path.join(root,file_name)
+         return path_file
+        
+    return None
```
AyanSinhaMahapatra: Need not return `None` explicitly. (?)

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +118,57 @@ def get_scanner(
...
+    def process_codebase(self, codebase, **kwargs):
+
+        for resource in codebase.walk(topdown=False):
+            match_reference_license(resource,codebase)
+
+
+def match_reference_license(resource, codebase):
+    """
+    Find instances for any licenses in referenced filenames
+    """
```
AyanSinhaMahapatra, line 135: If not file, return. Like [here](https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/plugin_package.py#L102)

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +118,57 @@ def get_scanner(
             license_text_diagnostics=license_text_diagnostics,
             license_url_template=license_url_template
         )
+    def process_codebase(self, codebase, **kwargs):
+
```
AyanSinhaMahapatra, line 125: Return if single resource like [here](https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/plugin_package.py#L88)

src/licensedcode/plugin_license.py
AyanSinhaMahapatra: Question for @pombredanne, does `process_codebase` functions run after all scanners have completed running, i.e. when `process_codebase` is called for a resource for a scan plugin, have all resources already been populated with LicenseMatch objects detected in them? (The above is assuming that)

AyanSinhaMahapatra (changes requested):

AyanSinhaMahapatra:
Also, for packagedcode, say for npm, see [here](https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/npm.py#L99). We have a function converting a list of license declarations to a license-expression. It would have been ideal if we return some form of License Detections here instead of just the expressions (which is WIP), but in a short term we could check first for these kind of license references to file (and report only the license conclusion without the unknown).

src/licensedcode/plugin_license.py
akugarg: But the licenses thing here is array of dic which is returned by api.py [here](https://github.com/nexB/scancode-toolkit/blob/develop/src/scancode/api.py#L281)  with keys and values. From this how can we fetch referenced_filenames ?

src/licensedcode/plugin_license.py
pombredanne: > does process_codebase functions run after all scanners have completed running, i.e. when process_codebase is called for a resource for a scan plugin, have all resources already been populated with LicenseMatch objects detected in them? (The above is assuming that)

- For a scanner plugin, a `process_codebase` function is guaranteed to be called after this scanner has completed for all files. (it may be that all scanners have been running for that file, but that's not guaranteed)
- For a post-scan plugin, a `process_codebase` function is guaranteed to be called after all the scanners have completed for all files.

pombredanne: So for the license scanner, yes, it has finished running on all files when its `process_codebase` is called.

src/licensedcode/plugin_license.py
pombredanne: @akugarg good point!
the `referenced_filenames` should be returned by the API. Let's update this possibly in a separate PR.
Alternatively (or for now), given a rule identifier, you could look up in https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/index.py#L173 ... `rules_by_rid`....

But really what we need is have this be returned by the API function ....  `rules_by_rid` is not queryable.... so we would need to add a new data structure to the `LicenseIndex` such as more or less  `rule_identifiers = {rue.identifier: rule for rule in self.rules_by_rid}`

src/licensedcode/plugin_license.py
AyanSinhaMahapatra: This makes sense. Thanks @pombredanne

src/licensedcode/plugin_license.py
AyanSinhaMahapatra: Thanks @pombredanne I was searching for this but couldn't find. 

Btw, if using `rules_by_rid` for now, we may not do this for all the matches, but only check if `match.is_unknown` is `True`? But yes, having this in the API and in the LicenseIndex would be great.

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +118,57 @@ def get_scanner(
...
+        location=location, min_score=0)
+    
+    modified = False
+
+    for match in matches:
+        ref_files=match.rule.referenced_filenames
+        if len(ref_files) != 0:
+            for i in range(len(ref_files)):
+                if not ref_files[i].startswith('usr/share/common-licenses'):
+                 new_loc=find_reference_file(location,ref_files[i])
```
AyanSinhaMahapatra: Indentation in this `if` block is not proper.

src/licensedcode/plugin_license.py
akugarg: @AyanSinhaMahapatra @pombredanne I have created a separate PR for returning **referenced_filenames** by api.py

src/licensedcode/plugin_license.py
akugarg: > This is not the correct way to do this.
> 
> Currently you are:
> 
> 1. Finding the directory locally with the location.
> 2. Doing a license detection on the file that you have found.
> 
> The approach should have been, as mentioned [here](https://github.com/nexB/scancode-toolkit/issues/1364#issuecomment-888988049) to
> 
> 1. get the parent resource from the current resource
> 2. check if the referenced filename is present in the child resources of this parent resource.
> 3. if resource object with referenced filename present, get the licenses from the from the resource object.
> 4. then add the matches and license expression.

The parent that we get from Resource object is also a Resource object right? When I try to get children from parent object(using resource.parent to get parent and then using https://github.com/nexB/commoncode/blob/main/src/commoncode/resource.py#L1188) it throws attribute error altho if if parent is also a Resource objecr we can use [this](https://github.com/nexB/commoncode/blob/main/src/commoncode/resource.py#L1188) to get children

AyanSinhaMahapatra: It could be an iterable (? I'm not sure, doesn't look like it). You should be able to for loop through the children like [this](https://github.com/nexB/commoncode/blob/main/src/commoncode/resource.py#L1171) anyway.

akugarg: Yup tried this only ,but it is showing attribute error.
_par = resource.parent
for  child in par.children(codebase):_

AyanSinhaMahapatra: Could you paste the error here please. Also try these checking functions 1. https://github.com/nexB/commoncode/blob/main/src/commoncode/resource.py#L1211 2. https://github.com/nexB/commoncode/blob/main/src/commoncode/resource.py#L1199 3. https://github.com/nexB/commoncode/blob/main/src/commoncode/resource.py#L1182

akugarg:
```
 plugin.process_codebase(codebase, **kwargs)
  File "/home/akanksha/Desktop/git/scancode-toolkit/src/licensedcode/plugin_license.py", line 129, in process_codebase
    match_reference_license(resource,codebase)
  File "/home/akanksha/Desktop/git/scancode-toolkit/src/licensedcode/plugin_license.py", line 153, in match_reference_license
    child = par.children
AttributeError: 'function' object has no attribute 'children'
```

akugarg: Same attribute error for all all these functions . One point is if we try to access these func from resource directly it works! Ex: _for child in resource.children(codebase)_ this will work perfectly.
Seems like parent is not returning a Resource object ???

AyanSinhaMahapatra: >Yup tried this only ,but it is showing attribute error.
>par = resource.parent

Ah. `resource.parent` is a `function`, not the `Resource` object. `resource.parent(codebase)` actually returns the parent `Resource` object you need. :P These are functions, not property, and you need to pass `codebase` also.

akugarg: oops ! right silly mistake ;(

akugarg:
@AyanSinhaMahapatra @pombredanne Please have a look!
Also Can we keep the reference part for npm pacakges in a different PR from this one ?

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +119,57 @@ def get_scanner(
...
+    license_expressions = resource.license_expressions
+    if not licenses:
+        return 
+
+    reference_licenses=[]
+    reference_license_expressions=[]
+    unique_files = []
+
+    for license in licenses:
+        ref_files=license['matched_rule']['referenced_filenames']
```
pombredanne: The spaces around `=` are not correct

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +119,57 @@ def get_scanner(
...
+        if len(ref_files) != 0:
```
pombredanne: Why would this len check be needed? what will happen if the len is 0? just nothing IMHO you would not iterate on the list since it is empty.

src/licensedcode/plugin_license.py
pombredanne: Also call this
```suggestion
        referenced_filenames = license['matched_rule']['referenced_filenames'] 
```

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +119,57 @@ def get_scanner(
...
+            for i in range(len(ref_files)):
```
pombredanne: Why no just iterate on referenced_filenames? This is Python ;)
```suggestion
            for referenced_filename in referenced_filenames:
```

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +119,57 @@ def get_scanner(
...
+                if ref_files[i].startswith('usr/share/common-licenses') or ref_files[i] in unique_files:
```
pombredanne: What's unique files?
Could you just instead ensure that all rules referenced_filenames are unique at construction and validation time?

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +119,57 @@ def get_scanner(
...
+                    continue
+                unique_files.append(ref_files[i])
```
pombredanne: Please do not iterate using [i] indices.

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +119,57 @@ def get_scanner(
...
+                new_resource = find_reference_licenses(ref_files[i], resource, codebase)
```
pombredanne: Please use named kwargs rather than position args

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +119,57 @@ def get_scanner(
...
+    """
+    if not resource.is_file:
+        return
+
+    licenses = resource.licenses
+    license_expressions = resource.license_expressions
+    if not licenses:
+        return 
+
+    reference_licenses=[]
```
pombredanne: what about rather thsi name? also please use proper spacing around = here and in many other places
```suggestion
    referenced_licenses = []
```

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +119,57 @@ def get_scanner(
...
+    reference_license_expressions=[]
```
pombredanne:
```suggestion
    referenced_license_expressions = []
```

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +119,57 @@ def get_scanner(
...
+    license_expressions.extend(reference_license_expressions)
+    codebase.save_resource(resource)
+    return resource
+
+
+def find_reference_licenses(referenced_filename, resource, codebase):
+    """
+    Returns the resource which references to referenced_filename if found
+    """
+    par = resource.parent(codebase)
```
pombredanne: why using a shorthand par rather than `parent`?

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +119,57 @@ def get_scanner(
...
+
+    licenses.extend(reference_licenses)
+    license_expressions.extend(reference_license_expressions)
+    codebase.save_resource(resource)
+    return resource
+
+
+def find_reference_licenses(referenced_filename, resource, codebase):
+    """
+    Returns the resource which references to referenced_filename if found
```
pombredanne: Use imperative style `Return` not `Returns`
And `Returns the resource which references to referenced_filename if found` is hard to understand. I am not sure what you return... is this a `Resource` object? What are the three arguments you pass `referenced_filename`, `resource`, `codebase`?
One suggestion could be
```suggestion
    Return a Resource matching the `referenced_filename` path or None, given a ``resource`` Resource (what is this resource... you need to explain in a ``codebase`` Codebase.
```

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +119,57 @@ def get_scanner(
...
+    """
+    par = resource.parent(codebase)
+    
+    for child in par.children(codebase):
+        file_name = fileutils.file_name(child.path)
```
pombredanne: we have a name attribute alright  in a Resource per https://github.com/nexB/commoncode/blob/b21f9f13b8d596be8246a48abf2c7f7d607cb5db/src/commoncode/resource.py#L966
But the name is not enough... `referenced_filename` can be a path too and you should just check that the path.endswith it

That said there are also many other things that would need to be handled (possibly in a second stage)... what if you have multiple `COPYING` files in a codebase and that `referenced_filename` is `"COPYING"`?
Which one would be selected in this case?
Which one is the correct one?

src/licensedcode/plugin_license.py
pombredanne: Why do you skip path that `startswith('usr/share/common-licenses')`? it does not make sense to me

pombredanne (changes requested):
Thanks! beside the review comments, I would also like to see some unit tests for `find_reference_licenses`

src/licensedcode/plugin_license.py
```diff
@@ -10,8 +10,11 @@
 from functools import partial
 
 import attr
+import os
```
pombredanne: This is a standardlib import that should be moved up with standard imports

src/licensedcode/plugin_license.py
```diff
@@ -10,8 +10,11 @@
...
 
+from commoncode import fileutils
 from commoncode.cliutils import PluggableCommandLineOption
+from licensedcode.models import BasicRule, License
```
pombredanne: Please only one import per line

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +119,57 @@ def get_scanner(
...
+    def process_codebase(self, codebase, **kwargs):
+        
+        if codebase.has_single_resource:
+            return
+
+        for resource in codebase.walk(topdown=False):
+            match_reference_license(resource,codebase)
+
+
+def match_reference_license(resource, codebase):
```
pombredanne, line 130: This would need unit tests

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +119,57 @@ def get_scanner(
...
+    """
+    Find instances for any licenses in referenced filenames
```
pombredanne: Please state what you return and what arg you require.

src/licensedcode/plugin_license.py
akugarg: umm didn't got your point ;( . Suppose there are two licenses in licenses attribute and both of them reference to same file we need to report it only once right?

akugarg: @AyanSinhaMahapatra suggested me this in https://github.com/nexB/scancode-toolkit/issues/1364#issuecomment-869995820

src/licensedcode/plugin_license.py
akugarg: Yes, we are only checking in the same directory as for now

src/licensedcode/plugin_license.py
akugarg: Ig I have already added some test cases!

AyanSinhaMahapatra: The tests currently added are whole scans and not unittests for just `match_reference_license` and `find_reference_licenses` function

src/licensedcode/plugin_license.py
AyanSinhaMahapatra: I think @pombredanne asked not to have this afterwards.

src/licensedcode/plugin_license.py
AyanSinhaMahapatra: Also not sure why the resource object is being returned, it is already saved in the codebase.

src/licensedcode/plugin_license.py
AyanSinhaMahapatra: You can create [mock codebase](https://github.com/nexB/scancode-analyzer/blob/main/tests/test_analyzer_plugin.py#L200) from JSONs and also create mock resources directly like [this](https://github.com/nexB/scancode-analyzer/blob/main/tests/test_analyzer_plugin.py#L175). This is the only way to get resource/codebases outside of scans. Now use these to test various  cases of the function `match_reference_license`, where they are able to follow and update.

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +117,56 @@ def get_scanner(
...
+
+def find_reference_licenses(referenced_filename, resource, codebase, **kwargs):
+    """
+    Return the resource object for matching referenced-filename given a resource in codebase
+    """
+    parent = resource.parent(codebase)
+    
+    for child in parent.children(codebase):
+        path = child.path
+        if path.endswith(referenced_filename):
```
AyanSinhaMahapatra: This `endswith` makes this rigid wrt to the file extension, which maybe we should not do. 

For example if a Rule has only referenced a file called `COPYING` and the actual file is `COPYING.rst` or `COPYING.md` we still should be able to find this file. This is also the case if we have a rule with `see license in LICENSE` which will match even if there are extensions, in absence of more specific rules with those extensions.

Often `LICENSE` is in the extension and not the name, that should also be followed through correctly.

All these cases should be a unit test, for this function. Similar to [this comment](https://github.com/nexB/scancode-toolkit/pull/2616/files#r690223997).

There should also be a test then of two files with the same name being present, say there are two files with `LICENSE` in it's name `apache-2.0.LICENSE` is present in the same directory, in addition to a `bsd-new.LICENSE` file being present, then this function would just return on the first License it encounters, where we should ideally add to the license-expression both. This should also be a unit test.

@pombredanne does these make sense?

src/licensedcode/plugin_license.py
akugarg: I have made it now separate process instead of having it nested

src/licensedcode/plugin_license.py
akugarg: OK! I have now added a check for checking base name too

src/licensedcode/plugin_license.py
akugarg: Removed this for now!

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +117,57 @@ def get_scanner(
             license_text_diagnostics=license_text_diagnostics,
             license_url_template=license_url_template
         )
+
+    def process_codebase(self, codebase, **kwargs):
+        
+        if codebase.has_single_resource:
+            return
```
pombredanne, line 124: There is  still a possibility of a file making a reference to itself?
I created #2665 as a follow up

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +117,57 @@ def get_scanner(
...
+
+        for resource in codebase.walk(topdown=False):
+            match_reference_license(resource,codebase)
+
+
+def match_reference_license(resource, codebase):
+    """
+    Return the updated license matches having reference to another files
```
pombredanne: We need to improve this doc... this could be made clearer?
```suggestion
    Return the ``resource`` Resource updating and saving it in place, after adding new
    license matches (licenses and license_expressions) following their Rule
    ``referenced_filenames`` if any. Return None if this is not a file Resource.
```

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +117,57 @@ def get_scanner(
...
+
+    licenses.extend(referenced_licenses)
+    license_expressions.extend(referenced_license_expressions)
+    codebase.save_resource(resource)
+    return resource
+
+
+def find_reference_licenses(referenced_filename, resource, codebase, **kwargs):
+    """
+    Return the resource object for matching referenced-filename given a resource in codebase
```
pombredanne: What about this instead?
```suggestion
    Return a Resource matching the ``referenced_filename`` path or filename given a ``resource`` in ``codebase``.
    Return None if the ``referenced_filename`` cannot be found in the same directory as the base ``resource``.
    ``referenced_filename`` is the path or filename referenced in a LicenseMatch of ``resource``,
```

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +117,57 @@ def get_scanner(
...
+                
+    for referenced_filename in referenced_filenames:
+        new_resource = find_reference_licenses(referenced_filename=referenced_filename, resource=resource, codebase=codebase)
+        if new_resource:
+            referenced_licenses.extend(new_resource.licenses)
+            referenced_license_expressions.extend(new_resource.license_expressions)
+
+    licenses.extend(referenced_licenses)
+    license_expressions.extend(referenced_license_expressions)
+    codebase.save_resource(resource)
```
pombredanne: IMHO if there are no changes, no save should be done.

pombredanne (changes requested):
I have a few final review items for you. Thanks!

src/licensedcode/plugin_license.py
akugarg: If it references to itself, licenses would have been already detected right? If I am not wrong it will do double counting?

pombredanne: Good point! silly me!

src/licensedcode/plugin_license.py
```diff
@@ -116,3 +117,73 @@ def get_scanner(
...
+            return
+
+        for resource in codebase.walk(topdown=False):
+            match_reference_license(resource,codebase)
+
+
+def match_reference_license(resource, codebase):
+    """
+    Return the ``resource`` Resource updating and saving it in place, after adding new
+    icense matches (licenses and license_expressions) following their Rule
```
pombredanne:
```suggestion
    license matches (licenses and license_expressions) following their Rule
```

pombredanne (approved):
LGTM... Thank you ++

Issue #2615, opened 2021-07-30
win_pe Package parser is not identifying binaries 


JonoYang:
I am scanning a Windows codebase and there are files with the extensions that are prefixed with extensions recognized by the win_pe parser, like `.dll`, `.sys`, etc. which has a hex string as a suffix. For example, the extensions in question look like the following: `.dll_ebdc1de9`, `.mui_71b54a25`, etc. The extensions set in the win_pe parser should have a wildcard character after the extension to capture these.

JonoYang:
I've added the wildcard character after the extensions handled by the win_pe parser. This should help with the issue.

PR #2623, opened 2021-08-02
Collect InstalledWindowsProgram installed files #2615

JonoYang:
This PR introduces more file extensions that can be detected using the WinPE package parser. We also check the registry location `\\Wow6432Node\\Microsoft\\Windows\\CurrentVersion\\Uninstall` for installed programs. `populate_installed_files` has also been updated to include a `root_prefix` argument. This argument allows us to set a root directory to the relative package file paths generated by `populate_installed_files`. This helps us in the case when analyzing Windows Docker images because the rootfs for Windows is stored in the `Files` subdirectory of the layer directory rather than the layer directory itself.


pombredanne (approved):
LGTM... Thanks!

Issue #1888, opened 2020-01-31
Add support for CycloneDX Output format


eastmadc:
Similar to the support of exporting to SPDX output format.  It would be helpful to add support to export into the increasingly popular CycloneDX format.

[Specification](https://github.com/CycloneDX/specification)
[Example Project](https://github.com/CycloneDX/cyclonedx-node-module)

agschrei:
Having support for this output format would be a great help to anybody that wants to ingest ScanCode results into downstream tools like DependencyTrack or enrich them with data from other scanners/sources.  

I'm assuming many people have a business need for this, so I'm leaving this comment here in the hopes that it gains traction.  

Also, looking at the plugin architecture, it seems to me like implementing this as a post-scan plugin is possible.   
Could anybody with more hands-on experience on ScanCode plugins comment on that perhaps?

pombredanne:
@eastmadc @agschrei yes that would be fairly straightforward to implement especially since CycloneDX embases Package URL that were born here. The approach is to implement an output plugin like in https://github.com/nexB/scancode-toolkit/blob/develop/src/formattedcode/output_yaml.py
You can see the simple commit stream here https://github.com/nexB/scancode-toolkit/issues/2208 covering the whole thing (or this simple single commit https://github.com/nexB/scancode-toolkit/commit/310f4501a720fdb201e86e4d63eafb3e22e347b3 )

And this library is likely the way to go too https://github.com/CycloneDX/cyclonedx-python

pombredanne:
@stevespringett FYI ^ :)

BTW in  https://github.com/CycloneDX/cyclonedx-python/blob/master/setup.py there are several deps which are pinned which makes it somewhat impractical to reuse this as a library. I will likely provide a patch for your consideration to makes this easier to reuse.

pombredanne:
@agschrei Are you willing to chip in to help implement this? I can provide hand holding!

agschrei:
@pombredanne I'll hopefully find some time this weekend to look into it. I can't guarantee anything just yet because of university exams, but will report back on Monday.   

Given you know the codebase like the back of your hand, how do you estimate this feature in terms of effort/time required?

pombredanne:
> I'll hopefully find some time this weekend to look into it. I can't guarantee anything just yet because of university exams, but will report back on Monday.

That's very gentle of you!

> Given you know the codebase like the back of your hand, how do you estimate this feature in terms of effort/time required?

[...]

PR #2698, opened 2021-09-08
Add CycloneDx output option

agschrei:
Implements  #1888 

### Tasks


agschrei:
Looking at the jobs that are currently failing it seems that the failures are related to the introduction of the `output_format_version` key in the JSON output which is independent of the changes this PR introduced.  
See for example [this job](https://dev.azure.com/nexB/scancode-toolkit/_build/results?buildId=4623&view=logs&j=9a634e43-2dfe-534b-56d2-0b7ffad252eb&t=28d573b5-74f3-578a-960b-e035b88f9a42&l=4408)  

I'm not sure how to proceed here.

@pombredanne could you kindly provide some guidance?

JonoYang:
@agschrei Thank you for contributing a CycloneDX output plugin! We will review the plugin in depth, but it looks good so far. We are looking into the test error you mentioned, as it is affecting other branches as well.

pombredanne:
@agschrei Thank youn ++ for this... let me review.

tests/formattedcode/data/cyclonedx/expected.xml
```diff
@@ -0,0 +1 @@
...gz</url></reference><reference type="bom"><url>https://registry.npmjs.org/npm/2.13.5</url></reference></externalReferences></component><component type="library" bom-ref="pkg:npm/abbrev@1.0.7"><name>abbrev</name><version>1.0.7</version><description>Like ruby's abbrev module, but in js</description><author>Isaac Z. Schlueter i@izs.me</author><scope>required</scope><purl>pkg:npm/abbrev@1.0.7</purl><hashes /><licenses><license><id>ISC</id><url>http://fedoraproject.org/wiki/Licensing:MIT#Old_Style_with_legal_disclaimer_2</url></license></licenses><externalReferences><reference type="distribution"><url>https://registry.npmjs.org/abbrev/-/abbrev-1.0.7.tgz</url></reference><reference type="vcs"><url>http://github.com/isaacs/abbrev-js</url></reference><reference type="website"><url>https://www.npmjs.com/package/abbrev</url></reference><reference type="distribution"><url>https://registry.npmjs.org/abbrev/-/abbrev-1.0.7.tgz</url></reference><reference type="bom"><url>https://registry.npmjs.org/abbrev/1.0.7</url></reference></externalReferences></component></components><dependencies><dependency ref="pkg:npm/npm@2.13.5"><dependency ref="pkg:npm/abbrev@1.0.7" /></dependency></dependencies></bom>
```
pombredanne: Would it be possible to get a pretty-printed XML output by default?

tests/formattedcode/data/cyclonedx/expected.json
```diff
@@ -0,0 +1,122 @@
...
+    },
+    {
+      "name": "abbrev",
+      "version": "1.0.7",
+      "author": "Isaac Z. Schlueter i@izs.me",
+      "description": "Like ruby's abbrev module, but in js",
+      "purl": "pkg:npm/abbrev@1.0.7",
+      "licenses": [
+        {
+          "license": {
```
pombredanne: It looks like you are taking the declared license for licenses. This would be the license that's asserted in a package manifest, not detected and not normalized. You should instead use the license_expression. And likely convert this to an SPDX expression?

tests/formattedcode/data/cyclonedx/expected.json
```diff
@@ -0,0 +1,122 @@
...
+        {
+          "url": "https://registry.npmjs.org/abbrev/-/abbrev-1.0.7.tgz",
+          "type": "distribution"
+        },
+        {
+          "url": "https://registry.npmjs.org/abbrev/1.0.7",
+          "type": "bom"
+        }
+      ],
+      "type": "library",
```
pombredanne: Would not these field come further up, e.g. before the references?

docs/source/rst_snippets/output_format_options.rst
```diff
@@ -31,4 +31,8 @@ All Scan Output Options
 --html-app FILE         Write scan output as a mini HTML
                         application to FILE.
 
+--cyclonedx FILE        Write scan output as a CycloneDx 1.3 BOM in XML format to FILE
```
pombredanne: I wonder if it would make more sense to have `--cyclonedx` produce JSON and `--cyclonedx-xml` produce XML, e.g. promoting using JSON rather than XML?

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+import attr
+from commoncode.cliutils import OUTPUT_GROUP
+from commoncode.cliutils import PluggableCommandLineOption
+from formattedcode import FileOptionType
+from plugincode.output import OutputPlugin
+from plugincode.output import output_impl
+from typing import FrozenSet, List, Tuple
+
+
+def _get_set_of_known_licenses_and_spdx_license_ids() -> Tuple[List, FrozenSet[str]]:
```
pombredanne: I am not against using type annotations per se, but for now we are not using them anywhere in ScanCode and until we are doing type checking in the CI and at runtime, what is the value of these beside documentation?

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+    """load all licenses and all SPDX Ids known to ScanCode
+       this will also load scancode licenserefs, so we filter those
+    """
+    from licensedcode.models import get_all_spdx_keys, load_licenses
+    licenses = load_licenses(with_deprecated=True)
+    spdx_keys = filter(lambda x: "LicenseRef" not in x,
```
pombredanne: Why do you filter LicenseRef here? that's pretty essential to license detection accuracy as you are putting aside ~1290 licenses out of 1879, e.g. you are discarding roughly over 65% of the licenses. Also you likely do not want to keep deprecated licenses `with_deprecated=True` above anyway.

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+from plugincode.output import OutputPlugin
+from plugincode.output import output_impl
+from typing import FrozenSet, List, Tuple
+
+
+def _get_set_of_known_licenses_and_spdx_license_ids() -> Tuple[List, FrozenSet[str]]:
+    """load all licenses and all SPDX Ids known to ScanCode
+       this will also load scancode licenserefs, so we filter those
+    """
+    from licensedcode.models import get_all_spdx_keys, load_licenses
```
pombredanne: nitpicking: We prefer to have one import per line.

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+from formattedcode import FileOptionType
+from plugincode.output import OutputPlugin
+from plugincode.output import output_impl
+from typing import FrozenSet, List, Tuple
+
+
+def _get_set_of_known_licenses_and_spdx_license_ids() -> Tuple[List, FrozenSet[str]]:
+    """load all licenses and all SPDX Ids known to ScanCode
+       this will also load scancode licenserefs, so we filter those
+    """
```
pombredanne: nitpicking: We use rather this style for docstrings:
```
    """
    Return a tuple of (list of License, set of SPDX license ids).
    The SPDX license ids set does not include LicenseRef from ScanCode or other origins 
    """ 
```
key points:
- Start on the next line after """, Capitalize the first word.
- Use imperative style, and explain first what is returned

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+    from licensedcode.models import get_all_spdx_keys, load_licenses
+    licenses = load_licenses(with_deprecated=True)
+    spdx_keys = filter(lambda x: "LicenseRef" not in x,
+                       get_all_spdx_keys(licenses))
+    return (licenses, frozenset(spdx_keys))
+
+
+known_licenses, spdx_ids = _get_set_of_known_licenses_and_spdx_license_ids()
+
+hash_type_mapping = {
```
pombredanne: nitpicking: What about this instead:
```
cdx_hash_types_by_scancode_hash

File: src/formattedcode/output_cyclonedx.py [hunk 10]
```diff
@@ -0,0 +1,659 @@
...
+@attr.s
+class CycloneDxHashObject():
+    alg: str = attr.ib()
+    content: str = attr.ib()
+
+
+@attr.s
+class CycloneDxExternalRef:
+    url: str = attr.ib()
+    type: str = attr.ib(validator=attr.validators.in_(["vcs", "issue-tracker", "website",
```
pombredanne: Nitpicking: for such list we have adopted the black style (but have not yet switched to it... and we still use single quotes)... so that would likely come out this way:
```Python
    type: str = attr.ib(validator=attr.validators.in_([
        'vcs',
        'issue-tracker',
        'website',
        'advisories',
        'bom',
        'mailing-list',
        'social',
        'chat',
        'documentation',
        'support',
        'distribution',
        'license',
        'build-meta',
        'build-system',
        'other'
    ]))
```

but then, it could make sense to declare this as a constant instead, may be like this:

```Python
    reference_types = frozenset([
        'vcs',
        'issue-tracker',
        'website',
        'advisories',
        'bom',
        'mailing-list',
        'social',
        'chat',
        'documentation',
        'support',
        'distribution',
        'license',
        'build-meta',
        'build-system',
        'other'
    ])

    type: str = attr.ib(validator=attr.validators.in_(reference_types))

```

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+    "bug_tracking_url": "issue-tracker",
+    "code_view_url": "other",
+    "vcs_url": "vcs",
+    "repository_homepage_url": "website",
+    "repository_download_url": "distribution",
+    "api_data_url": "bom"
+}
+
+
+def get_external_refs(package: dict) -> List[CycloneDxExternalRef]:
```
pombredanne: Could this be a class method of CycloneDxExternalRef, since this is essentially a `from_package()` factory?

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+    ext_refs = []
+
+    for key in url_type_mapping:
+        ref = get_external_ref_from_key(package, key)
+        if ref is not None:
+            ext_refs.append(ref)
+    return ext_refs
+
+
+def get_hashes_list(package: dict) -> List[CycloneDxHashObject]:
```
pombredanne: Same as above, could this be a class method of CycloneDxHashObject, since this is essentially a from_package() factory?

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+            if alg is not None and digest is not None:
+                hashes.append(CycloneDxHashObject(
+                    hash_type_mapping[alg], digest))
+    return hashes
+
+
+def get_author_from_parties(parties: List[dict]) -> str:
+    if parties is not None:
+        try:
+            # assume there can only ever be one author and get that with next()
```
pombredanne: This assumption does not hold. There is no bound on the number of authors in a parties list. I wish it were, but packages models are all over the map data-structure wise?

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+            author = next(filter(lambda p: p.get("role") == "author", parties))
+            return ' '.join(filter(None, (author.get("name"),
+                                          author.get("email"),
+                                          author.get("url"))))
+        except StopIteration:
+            return None
+
+
+"""
```
pombredanne: May be this should be the docstring of the CycloneDxOutput class instead?

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+                    hash_type_mapping[alg], digest))
+    return hashes
+
+
+def get_author_from_parties(parties: List[dict]) -> str:
+    if parties is not None:
+        try:
+            # assume there can only ever be one author and get that with next()
+            author = next(filter(lambda p: p.get("role") == "author", parties))
+            return ' '.join(filter(None, (author.get("name"),
```
pombredanne: nit: let's use black wrapping style

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+Output plugin to write scan results in CycloneDX format.
+For additional information on the format,
+please see https://cyclonedx.org/specification/overview/
+"""
+
+
+@output_impl
+class CycloneDxOutput(OutputPlugin):
+    options = [
+        PluggableCommandLineOption(('--cyclonedx', 'output_cyclonedx',),
```
pombredanne: nit: let's use black wrapping style

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+
+    def process_codebase(self, codebase, output_cyclonedx, **kwargs):
+        bom = build_bom(codebase)
+        write_results(bom, output_cyclonedx, output_json=False)
+
+
+@output_impl
+class CycloneDxJsonOutput(OutputPlugin):
+    options = [
+        PluggableCommandLineOption(('--cyclonedx-json', 'output_cyclonedx_json',),
```
pombredanne: nit: let's use black wrapping style

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+        map(
+            lambda entry:
+            CycloneDxDependency(ref=entry[0], dependsOn=entry[1]),
+            dependencies.items()
+        )
+    )
+
+
+def merge_components(existing: CycloneDxComponent, new: CycloneDxComponent):
+    """merges two components and returns a new component"""
```
pombredanne: nit: use our docstring style

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+    return list(
+        map(
+            lambda entry:
+            CycloneDxDependency(ref=entry[0], dependsOn=entry[1]),
+            dependencies.items()
+        )
+    )
+
+
+def merge_components(existing: CycloneDxComponent, new: CycloneDxComponent):
```
pombredanne: What is the purpose of merging? can this happen between two packages that have different purl? should this be checked?

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+                components.append(component)
+                ref_component_map[purl] = component
+            else:
+                merge_components(ref_component_map[purl], component)
+
+    return components
+
+
+def build_bom(codebase) -> CycloneDxBom:
+    # TODO: find out if we can always expect that header to be present
```
pombredanne: It is guaranteed to be always there, but there could be multiple ones( it is a list), possibly each list item from a different tool.

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+
+    scancode_version = scancode_header.get("tool_version")
+
+    tool_header = get_tool_header(scancode_version)
+    bom_metadata = CycloneDxMetadata(tools=[tool_header])
+
+    files = OutputPlugin.get_files(codebase)
+
+    # get all packages that are not None
+    packages = [package for file in files for package in file.get("packages")]
```
pombredanne: nit: for consistency throughout, use single quotes for now everywhere except for docstrings

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+    # associate dependency relationship by purl of dependent package
+    dep_map = dict([(package.get("purl"), package.get("dependencies"))
+                    for package in packages])
+
+    components = generate_component_list(packages)
+    # associate components by purl
+    comp_map = dict(map(lambda c: (c.purl, c), components))
```
pombredanne: what about naming this `components_by_purl`, this means no comment is needed.
Also what if there are several times the same purl for different packages instances?

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+    elif new.hashes is not None:
+        merge_lists(existing.hashes, new.hashes)
+
+    if existing.properties is None:
+        existing.properties = new.properties
+    elif new.properties is not None:
+        merge_lists(existing.properties, new.properties)
+
+
+def generate_component_list(packages) -> List[CycloneDxComponent]:
```
pombredanne: could this be a class factory method of CycloneDxComponent instead? may be named `from_packages()`?

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+    ref_component_map = {}
```
pombredanne: nit: rather than suffix a name with its type (e.g. `_map`) we prefer to use a more descriptive name such as `components_by_purl`

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+            dict_repr["bom-ref"] = dict_repr["bom_ref"]
+            del dict_repr["bom_ref"]
+        return dict_repr
+
+
+def write_results_json(bom, output_file):
+    json.dump(bom, output_file, indent=2, sort_keys=False, cls=CycloneDxEncoder)
+
+
+class XmlSerializer():
```
pombredanne: why the trailing `()`?

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+    def _get_root_element(self) -> ET.Element:
+        bom_element = ET.Element('bom', {
+            'xmlns': "http://cyclonedx.org/schema/bom/1.3",
+            'version': '1',
+            'serialNumber': self.bom.serialNumber
+        })
+        return bom_element
+
+    def _get_tool_element(self) -> ET.Element:
+        tool = ET.Element("tool")
```
pombredanne: nit: let's use single quotes consistently

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+
+def get_external_ref_from_key(package: dict, key: str) -> CycloneDxExternalRef:
+    url = package.get(key)
+    type = url_type_mapping[key]
+    if url is not None and type is not None:
+        return CycloneDxExternalRef(url=url, type=type)
+    return None
+
+
+def get_license_entry_from_license_expression(license_expression: str) \
```
pombredanne: nit: please avoid using backslash for line continuations. Wrapping black-style is best. 90'sih line length is OK too.
Also, could this be instead a class  method on the CycloneDxLicenseEntry class ?

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+       this will also load scancode licenserefs, so we filter those
+    """
+    from licensedcode.models import get_all_spdx_keys, load_licenses
+    licenses = load_licenses(with_deprecated=True)
+    spdx_keys = filter(lambda x: "LicenseRef" not in x,
+                       get_all_spdx_keys(licenses))
+    return (licenses, frozenset(spdx_keys))
+
+
+known_licenses, spdx_ids = _get_set_of_known_licenses_and_spdx_license_ids()
```
pombredanne: IMHO, avoid to make such long running call at an import time. 

Use the ` licensedcode.cache` module instead to get what you need with `get_licenses_db()` to get a mapping of `{license key: License object}`

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+    if isinstance(declared_license, str):
+        # if license key is in list of known spdx ids, set as id
+        if declared_license in spdx_ids:
+            lic.id = declared_license
+        # if we match this regex assume we are dealing with a URL
+        elif declared_license is not None and url_pattern.match(declared_license):
+            lic.url = declared_license
+        else:
+            lic.name = declared_license
+    # some declared_license entries are also expressed as dicts
```
pombredanne: .... and some as list or anything. By construction the data structure is whatever existed upstream. I think you should not use the declared license in the output but instead only the `license_expression` which is the detection that took place

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+    lic_expr = package.get("license_expression")
+    entry = get_license_entry_from_license_expression(lic_expr)
+    if entry is not None:
+        licenses = [entry]
+        if entry.license is not None:
+            seen_ids.add(entry.license.id)
+    else:
+        licenses = []
+
+    declared_license = package["declared_license"]
```
pombredanne: Do not use both declared_license and license expression, as this is redundant.

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+        lic_type = declared_license.get("type")
+        if lic_type in spdx_ids:
+            lic.id = lic_type
+        else:
+            lic.name = lic_type
+        lic.url = declared_license.get("url")
+    return CycloneDxLicenseEntry(license=lic, expression=None)
+
+
+def get_licenses(package: dict) -> List[CycloneDxLicenseEntry]:
```
pombredanne: Could this be a factory class method on CycloneDxLicenseEntry instead?

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+    """query our list of known licenses to see if we can resolve
+    the license_expression to a single entry"""
+
+    # exit early if we don't have a valid license_expression
+    if license_expression is None:
+        return None
+
+    license = known_licenses.get(license_expression)
+    if license is not None:
+        # attempt to set either official OSI URL or any license text URL
```
pombredanne: IMHO a better URL is https://github.com/nexB/scancode-toolkit/blob/develop/src/scancode/api.py#L141

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+    predicate = lambda el: not (el is None or
+                                (isinstance(el, list) or isinstance(el, dict))
+                                and len(el) == 0)
+    if hasattr(obj, '__dict__'):
+        obj = vars(obj)
+    obj_dict = {k: v for k, v in obj.items() if predicate(v)}
+    return obj_dict
+
+
+class CycloneDxEncoder(json.JSONEncoder):
```
pombredanne: Could we instead have an `to_dict()` methods on objects to serialize and avoid having JSON-specific serialization code?

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+    """Custom encoder that removes fields which are either empty or None.
+       Additionally renames the `bom_ref` field to `bom-ref` for output"""
+
```
pombredanne, line 737: Is this the standard for CycloneDX to skip empty fields?

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+class CycloneDxMetadata():
+    timestamp = attr.ib(
+        default=datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%SZ'))
+    tools: List[dict] = attr.ib(factory=list)
+    properties: List[CycloneDxAttribute] = attr.ib(factory=list)
+
+
+@attr.s
+class CycloneDxHashObject():
+    alg: str = attr.ib()
```
pombredanne, line 76: may be using a full name doe not hurt, like algorithm? unless this is the cdx name of course.

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+    if license_expression is None:
+        return None
+
+    license = known_licenses.get(license_expression)
+    if license is not None:
+        # attempt to set either official OSI URL or any license text URL
+        url = license.osi_url
+        if license.text_urls:
+            url = license.text_urls[0]
+        lic = CycloneDxLicense(id=license.spdx_license_key,
```
pombredanne: Do you need an SPDX license expression? if so we can get this directly instead.

What is CycloneDxLicenseEntry? a single license or an expression?

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,659 @@
...
+
+
+def get_hashes_list(package: dict) -> List[CycloneDxHashObject]:
+    hashes = []
+    for alg in hash_type_mapping:
+        if alg in package:
+            digest = package[alg]
+            if alg is not None and digest is not None:
+                hashes.append(CycloneDxHashObject(
+                    hash_type_mapping[alg], digest))
```
pombredanne: What about this instead? Also note how I use named keyword arguments which helps if we ever go through some name refactoring and avoid the order dependency implied by positional arguments.

```
    for scancode_algo, cyclonedx_algo in hash_type_mapping.items():
        digest = package.get(scancode_algo)
        if digest:
            hashes.append(CycloneDxHashObject(algo=cyclonedx_algo, content=digest))
```

pombredanne (changes requested):
This is a beautiful PR!
I came with a few nitpickings for your consideration. One thing I am not sure I get is how licenses or license expressions are stored in CycloneDX.
We can have also some chat at your convenience; we meet at https://gitter.im/aboutcode-org/discuss

agschrei:
> This is a beautiful PR!  

Thanks, I have to say that your contribution guide is really helpful for first-time contributors.  

> I came with a few nitpickings for your consideration.

I'll have to set some time aside to implement these changes.

Looking at your comments most of the open points should be quick fixes,  
but there's some other stuff like the declared licenses topic that might require additional back and forth.

So there's a good chance I'll try to find you on gitter like you suggested to clarify on that.

> One thing I am not sure I get is how licenses or license expressions are stored in CycloneDX.

The way I understood the CycloneDx spec, a component may have a list of `License` entries where each entry may either have    
- a license expression as a string
- a `License Object` as the spec calls it   

but obviously not both.  
So I tried to implement that as an explicit union type with `CycloneDxLicenseEntry`.


I used Atlassian's great schema viewer to explore the CycloneDx 1.3 schema [here](https://json-schema.app/view/%23/%23%2Fdefinitions%2Fcomponent/%23%2Fdefinitions%2FlicenseChoice?url=https%3A%2F%2Fraw.githubusercontent.com%2FCycloneDX%2Fspecification%2Fmaster%2Fschema%2Fbom-1.3.schema.json) if you're interested.

src/formattedcode/output_cyclonedx.py
agschrei: this is the cdx name, so I'd petition to keep it like this so that the default __dict__ repr returns what we want the JSON key to be without modification

pombredanne:
> The way I understood the CycloneDx spec, a component may have a list of `License` entries where each entry may either have
> 
>     * a license expression as a string
> 
>     * a `License Object` as the spec calls it
> 
> 
> but obviously not both.
> So I tried to implement that as an explicit union type with `CycloneDxLicenseEntry`.
> 
> I used Atlassian's great schema viewer to explore the CycloneDx 1.3 schema [here](https://json-schema.app/view/%23/%23%2Fdefinitions%2Fcomponent/%23%2Fdefinitions%2FlicenseChoice?url=https%3A%2F%2Fraw.githubusercontent.com%2FCycloneDX%2Fspecification%2Fmaster%2Fschema%2Fbom-1.3.schema.json) if you're interested.

It makes sense. So I think that we should instead likely:
1. use only the license_expression field, as a bona fide SPDX expression
2. in order to also provide information about all the LicenseRef-scancode which are numerous, we could provide some 'license' externalReference for each non-SPDX license (or each license) as explained in https://json-schema.app/view/%23/%23%2Fdefinitions%2Fcomponent/%23%2Fdefinitions%2FexternalReference?url=https%3A%2F%2Fraw.githubusercontent.com%2FCycloneDX%2Fspecification%2Fmaster%2Fschema%2Fbom-1.3.schema.json
3. we could consider also keep the matched license notice text somewhere in the reference comment, but I do not think this can work as each license match may have a different notice even when matched to the same expression

tests/formattedcode/data/cyclonedx/expected.xml
agschrei: I noticed that `lxml` was already a dependency in the requirements.txt anyway, so I refactored to use that instead. Has a couple of advantages over `xml`, the biggest one being a pretty-print option.

tests/formattedcode/data/cyclonedx/expected.json
agschrei: I've moved these up so they no longer get lost behind the ext references which can get quite large

docs/source/rst_snippets/output_format_options.rst
agschrei: I've switched these two flags like you suggested

src/formattedcode/output_cyclonedx.py
agschrei: Depending on the IDE that you are developing in these can significantly improve autocompletion suggestions, so I have made a habit of using them.  
I've also noticed that the codebase doesn't make use of them, but `typing` was already part of the requirements.txt so I figured you are not directly opposed to adopting them.

src/formattedcode/output_cyclonedx.py
agschrei: I assumed so. For npm, as an example, many projects list multiple authors.  
The tricky bit here is that cdx expects a string as the value for the 'author' field.  
My guess is that cdx expects there to only ever be one author, hence referring to the singular.  

If we want to map multiple authors to this field, how would you like to proceed?

My suggestion is mapping authors in a CSV-like fashion:
`"author": " 'author a', 'author b' "`
Do you see a more elegant soluton here? I'm not happy with this, but it beats leaving out authors in the mapping.

pombredanne:
@agschrei thanks for the updates. Let me review.

src/formattedcode/output_cyclonedx.py
pombredanne: fair enough.  :+1:

src/formattedcode/output_cyclonedx.py
pombredanne: I would guess that a better approach would be to list all authors indeed, and just just their plain names, one per line

pombredanne:
@agschrei gentle ping. I would like to merge this soon enough.

src/formattedcode/output_cyclonedx.py
agschrei: I'm not sure if CycloneDx defines a preference on that, but intuitively it made sense for me to prune anything that is either empty or None as such entries add no value to the BOM

src/formattedcode/output_cyclonedx.py
agschrei: that does look a lot cleaner, incorporated this into the changes

src/formattedcode/output_cyclonedx.py
agschrei: that's how we are doing it now to ensure all authors are considered

src/formattedcode/output_cyclonedx.py
agschrei: true, accidentally separated this from the class

src/formattedcode/output_cyclonedx.py
agschrei: Since we use the `purl` as a `bom-ref` it has to be unique, merging boms that have the same purl ensures this.  
This method now checks if the `purl` values are indeed identical before merging and raises a `ValueError` if they are not.

Please note that the only location where we call this helper doesn't catch that because the context in which it is invoked already ensures that both components have the same purl

src/formattedcode/output_cyclonedx.py
agschrei: the approach that I am taking here:
I'm filtering the headers for the `scancode-toolkit` header and only forwarding that to the CycloneDx output.  Is this an acceptable solution for you?

src/formattedcode/output_cyclonedx.py
agschrei: this is now enforced across all strings

src/formattedcode/output_cyclonedx.py
agschrei: I've renamed this to be more expressive of its intent.  
At this point thanks to merging of components with the same purl we are already guaranteed to only have one component per purl.

src/formattedcode/output_cyclonedx.py
agschrei: I have a question about this:

In one of the larger tests I ran the `license_expression` produced by scancode and the declared package license looked like this:

```json
 "purl": "pkg:npm/has@1.0.3",
      "licenses": [
        {
          "expression": "mit AND (mit AND (mit AND unknown))"
        },
        {
          "license": {
            "id": "MIT"
          }
        }
      ],
```

I know too little about your license detection engine to explain where this detection comes from, but in that case having the declared license alongside the expression can be helpful.  

The approach I am currently taking here is:
- add the license expression as a license entry
- only add declared license(s) if they differ from the license expression

Given the example above, is this a viable option?

src/formattedcode/output_cyclonedx.py
agschrei: I didn't know about this one, but we are now using it instead of the OSI URL

agschrei:
Most of the requested changes are now implemented in the way I understood them.  
Please let me know if something still irks you about the modifications.

The only major thing that is still missing is the correct handling of `LicenseRef-*` SPDX IDs as suggested above, I will hopefully finish that tomorrow.  

I'll reach out to you on the gitter channel to double-check my solution then.

pombredanne:
I triggered a re-run of the test failures which are due to Azure flakiness, not a real error.

pombredanne:
and there is a new lint/check that was merged recently for the documentation and you are failing it... that's an easy fix though:

> Run ./scripts/doc8_style_check.sh
source/rst_snippets/output_format_options.rst:34: D001 Line too long
source/cli-reference/output-format.rst:585: D001 Line too long
Error: Process completed with exit code 1.

agschrei:
I just fixed the docs sections that were failing the lint check.  

Hopefully everything passes now.

pombredanne:
@agschrei gentle ping... there are few nitpickings left for your review

src/formattedcode/output_cyclonedx.py
agschrei: taken into consideration not just here but for all members

src/formattedcode/output_cyclonedx.py
agschrei: not sure if I understood you correctly, but I am now getting the licenses from the cache and extracting the spdx IDs from there

src/formattedcode/output_cyclonedx.py
agschrei: I am currently using the scancode license expression here as is, because the BOM format doesn't specify anything further about the expression field. But if we can get an SPDX license expression from that directly I would be in favor of that. Where do I need to look? Does the codebase offer some kind of utility to convert to that?

A `CycloneDxLicenseEntry` is a union type that can either hold a single expression or a  `CycloneDxLicense` that has an (spdx) id, name and optional URL. CycloneDx also allows to include the license text here, but we currently don't map this as I believe including the license text for every finding would only needlessly bloat the report.  
My argument here is that if a user / downstream tool wants the license text they can just grab it from the provided URL

src/formattedcode/output_cyclonedx.py
agschrei: @pombredanne I am open to this but would like to understand your reasoning about it first. Is there some downside to the serialization code that I am not seeing?  

I solved it this way because this encoder throws away any fields with empty or None values and renames the `bom_ref` field to `bom-ref` for output no matter what we throw in, so it works generically for all objects we work with here

src/formattedcode/output_cyclonedx.py
agschrei: fixed this

src/formattedcode/output_cyclonedx.py
agschrei: this is now obsolete as I am getting the licenses from the cache (no longer filtering on LicenseRef prefix) as you suggested bellow.

tests/formattedcode/data/cyclonedx/expected.json
agschrei: @pombredanne I already mentioned this a bit further down, but my reasoning was that the license expression inferred by scancode can sometimes be needlessly complex, see this example from the tests:
```json
 "purl": "pkg:npm/has@1.0.3",
      "licenses": [
        {
          "expression": "mit AND (mit AND (mit AND unknown))"
        },
        {
          "license": {
            "id": "MIT"
          }
        }
      ],
```

Granted this can be normalized to `mit AND unknown` but at that point it might still be helpful to have the declared license alongside it.

If license_expression and declared_license are one and the same, we only keep one of the entries, so this approach should never lead to duplicate information.

agschrei:
@pombredanne I was still waiting on feedback for some of the points you raised. I have now resolved everything that is fixed/considered, but for the points that still remain open I either need additional input or just a nod that this is how you want to proceed. Could you please have a look at the comments there?

Also, should I rebase this onto develop in the meantime to see if it plays nice with the latest changes there?

src/formattedcode/output_cyclonedx.py
agschrei: I was going for the low-hanging fruit here as we probably can't cover everything.  
But as raised [here](https://github.com/nexB/scancode-toolkit/pull/2698#discussion_r737205459) I'm currently keeping declared licenses around in cases where the detection differs from the declared license

src/formattedcode/output_cyclonedx.py
pombredanne: My point was also to move this line into a function and not call the function at import  time

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,774 @@
...
+                        map(
+                           lambda license: license.spdx_license_key,
+                           licenses.values()
+                       )
+                    )
+    return (licenses, frozenset(spdx_keys))
+
+
+known_licenses, spdx_ids = _get_set_of_known_licenses_and_spdx_license_ids()
+spdx_licenseref_prefix = "LicenseRef-scancode-"
```
pombredanne: You should get the SPDX license key instead from the license records. This "LicenseRef-scancode-" cannot be used safely as the set of licenses evolves continuously

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,774 @@
...
+class CycloneDxLicense:
+    id: str = attr.ib(default=None)
+    name: str = attr.ib(default=None)
+    url: str = attr.ib(default=None)
+
+
+@attr.s
+class CycloneDxLicenseEntry:
+    @classmethod
+    def from_declared_license(cls, declared_license) -> 'CycloneDxLicenseEntry':
```
pombredanne: You cannot use the "declared license" safely: this is a data structure which can be anything and is whatever was upstream in the package manifest. It could be a list, string, mapping... anything declared upstream.
You want instead to use the "license_expression" (and likely the SPDX license expression).

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,774 @@
...
+        if `license_expression`is None, return None
+        """
+
+        # exit early if we don't have a valid license_expression
+        if license_expression is None:
+            return None
+
+        from scancode.api import SCANCODE_LICENSEDB_URL
+
+        license = known_licenses.get(license_expression)
```
pombredanne: This will rarely work, as a license expression is not a license, but eventually multiple licenses. The license_expression would need to be broken down in its keys (using the `license_expression` library`) and  each key looked up in the license db.
Is there a way to store an SPDX expression directly in the CycloneDX doc instead

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,774 @@
...
+            lic = CycloneDxLicense(id=license.spdx_license_key, name=license.name, url=url)
+            return CycloneDxLicenseEntry(license=lic, expression=None)
+        else:
+            return CycloneDxLicenseEntry(license=None, expression=license_expression)
+
+    @classmethod
+    def from_package(cls, package: dict) -> List['CycloneDxLicenseEntry']:
+        """
+        Return a list of CycloneDxLicenseEntry instances built from package data
+        Attempt to map `license_expression` and `declared_license` data to CycloneDx format
```
pombredanne: As mentioned elsewhere, do not use `declared_license`, use only `license_expression`.

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,774 @@
...
+        """
+        # store previously encountered IDs to identify duplicates
+        seen_ids = set()
```
pombredanne: You cannot deduplicate based on license ids. Licenses are in an expression that can be logically simplified, and logically deduplicated, but doing it on the keys of an expression could lead to this issues: say the expression is: `(MIT OR GPL-2.0-only) AND GPL-2.0-only` ... your approach would likely damage the meaning of the original expression. You cannot process license ids when expressions are used throughout here or you risk returning incorrect or misleading results.

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,774 @@
...
+            lic_entry = CycloneDxLicenseEntry.from_declared_license(declared_license)
+            if lic_entry is not None:
+                id = lic_entry.license.id if lic_entry.license is not None else None
+                if id not in seen_ids and id is not None:
+                    seen_ids.add(id)
+                    licenses.append(lic_entry)
+
+        return licenses
+
+    license: CycloneDxLicense = attr.ib(default=None)
```
pombredanne: Please move attributes declaration at the top of the class

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,774 @@
...
+
+@attr.s
+class CycloneDxAttribute:
+    name: str = attr.ib()
+    value: str = attr.ib()
+
+
+@attr.s
+class CycloneDxMetadata:
+    timestamp = attr.ib(default=datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%SZ'))
```
pombredanne: This is executed at import time and is a global: all the object will shared the same value. This is not what you want ... you need a factory https://www.attrs.org/en/stable/api.html?highlight=factory#attr.Factory

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,774 @@
...
+    tools: List[dict] = attr.ib(factory=list)
```
pombredanne: list is likely not what you want: See https://www.attrs.org/en/stable/api.html?highlight=factory#attr.Factory

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,774 @@
...
+    properties: List[CycloneDxAttribute] = attr.ib(factory=list)
+
+
+@attr.s
+class CycloneDxHashObject:
+    __cdx_hash_types_by_scancode_hash = {
```
pombredanne: Why the __ prefix?

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,774 @@
...
+    }
+    @classmethod
+    def from_package(cls, package: dict) -> List['CycloneDxHashObject']:
+        hashes = []
+        for scancode_alg, cyclonedx_alg in cls.__cdx_hash_types_by_scancode_hash.items():
+            digest = package.get(scancode_alg)
+            if digest:
+                hashes.append(CycloneDxHashObject(alg=cyclonedx_alg, content=digest))
+        return hashes
+    alg: str = attr.ib()
```
pombredanne, line 76: Please put attributes at the top

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,774 @@
...
+
+    def _add_license_refs(self):
+        """
+        Build CycloneDxExternalRef entries for scancode SPDX LicenseRefs
+        Modify the component in-place
+        """
+        licenses = filter(None, map(lambda lic: lic.license, self.licenses))
+        license_ids = map(lambda lic: lic.id, licenses)
+        licenseref_ids = filter(
+            lambda lic_id: lic_id is not None and lic_id.startswith("LicenseRef-scancode-"),
```
pombredanne: As example elsewhere, you cannot craft SPDX ids this way. You need to grab the stored spdx_license_key of a license

src/formattedcode/output_cyclonedx.py
pombredanne: This may be a problem. ScanCode reports all instances of each package so there can be duplicated purls (and this is evolving further with https://github.com/nexB/scancode-toolkit/pull/2748 )
The thing is that each package instance may have the same purl but can different attributes (for instanced when derived from a certain manifest vs. another format of style such as a Gemfile, gemspec and installed metadata.

src/formattedcode/output_cyclonedx.py
```diff
@@ -0,0 +1,774 @@
...
+    elif new.hashes is not None:
+        merge_lists(existing.hashes, new.hashes)
+
+    if existing.properties is None:
+        existing.properties = new.properties
+    elif new.properties is not None:
+        merge_lists(existing.properties, new.properties)
+
+
+def truncate_none_or_empty_values(obj) -> dict:
```
pombredanne: Why do you need this? Could you also have some tests for this?

src/formattedcode/output_cyclonedx.py
pombredanne: > I am open to this but would like to understand your reasoning about it first. Is there some downside to the serialization code that I am not seeing?

We also use a `to_dict()` function throughout the codebase as a convention. 
A JSONEncoder makes the serialization JSON-specific and impossible to use with other presentations such as YAML and I would rather have an explicit field-by-field mapping rather than something implicit. 
Also why would you rename fields?  Does the format demands to have "bom-ref" vs. "bom_ref" ?

tests/formattedcode/data/cyclonedx/expected.json
pombredanne: > my reasoning was that the license expression inferred by scancode can sometimes be needlessly complex, see this example from the tests:

There is no such thing as a "needlessly complex expression". There are two cases:
1. an expression is correct and this should be the default case
2. an expression is not correct and this a bug

In the case of the `has` npm package this is a bug and we should treat this correctly 
Can you enter a bug?

https://github.com/tarruda/has/blob/8146666148f1812f90e737aeefcded17a8cf81f4/package.json#L24 

> If license_expression and declared_license are one and the same, we only keep one of the entries, so this approach should never lead to duplicate information.

As explained above the `declared_license` is not something that is a reliable data point as this has not gone through detection and normalization using the license detection engine, it should never be use as the license in a CycloneDX document.

pombredanne (changes requested):
Sorry for the late review!
See my comments inline for your consideration.

pombredanne:
> my reasoning was that the license expression inferred by scancode can sometimes be needlessly complex, see this example from the tests:

Note that we are working on introducing a "primary license" for packages... which would address your concern. In the meantime, I will merge your branch patched to use the `license_expression` and not the `declared_license`.

pombredanne:
@agschrei I merged the latest develop and updated your code according to some of the comments above. Please review before I merge!
Thank you for your patience.

The key changes are outlined in the commit message of https://github.com/nexB/scancode-toolkit/pull/2698/commits/953a40edd60551477986150e71083e09fce407e0

## Refine CycloneDX support #1888

- Only deal with an SPDX license_expression and not plain licenses or
declared licenses, because this is the only correct license information.

- Removing most type hints since we are not yet checking these.

- Remove dependency lookup which could be misleading. Instead add warning
if dependencies are skipped.

- Add docstring to most methods and functions.

- Streamline nested list comprehensions and filter/map constructions and
replace these with simpler constructs.

- Replace JSON encoder by a plain to_dict() serialization throughout.

- Move attrs attributes declarations towards the top of classes
declarations rather than at the bottom.

- Move move method-like functions to their respective objects, including
moving XML serialization functions to to_xml_element() methods for each
model object.

- Use factories for fields defaults such as URN and timestamp.

- Split method that work on ScanCode packages lists in two: one operating
on a single package mapping and one working on the packages list.

- Update tests to work on larger set of expected data.

- Always return all attributes, including empty values for now.

pombredanne:
All green now. Merging!

PR #2736, opened 2021-10-14
Add first draft ROADMAP

pombredanne:

### DIFF
--- a/ROADMAP.rst
+++ b/ROADMAP.rst
@@ -1,0 +1,108 @@
+ScanCode IO/TK Roadmap
+========================
+
+Top Issues
+---------------
+
+1. Primary license detection, top issue.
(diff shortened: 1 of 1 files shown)


AyanSinhaMahapatra:
Hey, there's this roadmap https://github.com/nexB/scancode-toolkit/blob/develop/docs/source/contribute/roadmap.rst here which we could update and merge, and the docs here https://scancode-toolkit.readthedocs.io/en/latest/contribute/roadmap.html

pombredanne:
@AyanSinhaMahapatra good point. Do you think we should keep the roadmap in the doc or at the root or both?
I would think both with the main doc being at the root?

AyanSinhaMahapatra:
@pombredanne I think we should merge if possible and keep it at the root (and include it from there in the docs). We could also refer to milestones from the document, and keep the contents of this new roadmap under the next milestone (https://github.com/nexB/scancode-toolkit/milestone/14).

PR #2763, opened 2021-11-19
api.get_licenses: clarify and improve docstring for "min_score" argument

zacchiro:
- add a missing "not" to correctly capture min_score semantics
- call the argument "min_score", as per current name
- apply minor style improvements


pombredanne:
@zacchiro Hey! Thanks. Great catch!

PR #2765, opened 2021-11-22
Add new licenses and new detection rules

pombredanne:
This PR provides general purpose license detection improvements mostly in the form of new licenses and new license detection rules.

PR #2773, opened 2021-11-30
Allow license rules to require the presence of certain defining keywords

### LINKED ISSUE #2637: allow license rules to require the presence of certain defining keywords
## Short Description
In many license expression rules, certain words or phrases carry a much higher significance than others. For example, `This software is distributed under a` seems a lot less important than `MIT License` for a rule that describes an MIT license notice. I think it would be valuable if these "defining words" could optionally be marked and, if declared, would prevent matches from happening unless those defining phrases are present in the matched text.

## Possible Labels

- new feature
- license scan

## Select Category

-  [x] Enhancement
-  [ ] Add License/Copyright
-  [x] Scan Feature
-  [ ] Packaging
-  [ ] Documentation
-  [ ] Expand Support
-  [ ] Other

## **Describe the Update**
This is a feature proposal which I haven't been able to find any discussions of when searching through the issues (it's quite possible that I have just missed it!). Anyhow, I think it would be really valueable if it was possible in a license expression rule to mark certain "defining words" (or phrases) as required to be present in the scanned text in order for a match to be reported.

As an example, I've seen quite a lot of false positives that I believe could be eliminated if these crucial phrases were enforced. For example, the following text

    ## License

    This SDK is distributed under the Apache License, Version 2.0, see LICENSE for more information.
    
was matched by `mit_923.RULE`:

    License
    Distributed under the MIT License. See LICENSE for more information.
    
even though it's pretty clear to a human reader that there are certain crucial aspects of this rule that doesn't match. That is to say, some words in the rule are more significant in others, and are essentially *defining* for the license expression. In this case, of course it is `MIT` that would be that *defining phrase*.

As can be clearly seen from a scan match:

          "matched_rule": {
            "identifier": "mit_923.RULE",
            ..
          },
          "matched_text": "License\n\n[This] [SDK] [is] distributed under the [Apache] License, [Version] [2].[0], see LICENSE for more information."
          
the defining aspect of the license expression (`MIT`) is missing from the match.

Now, I think it would be really cool and, I imagine, would reduce false positive by quite a significant number, if it was possible to mark these *defining words/phrases* (mandatory words, keywords, or whatever one migth want to call them) in a rule.

I believe it would reduce the number of false positives quite substantially although that remains to be seen.

## **Possible Solution/Implementation Details**
Although I'm unsure of the feasibility of implementing such a solution, it would be nice if it was possible to highlight these words directly in the license `RULE` file, for example surrounded by some markup:

    License
    Distributed under the {{MIT}} License. See LICENSE for more information.

The semantics of this would be that the rule would _never_ match (irrespective of score) if `MIT` wasn't present in the matched text.

I suppose an alternative realization, although not as appealing, would be to include these mandatory keywords/phrases as an attribute in the `.yml` file:

``` yaml
license_expression: mit
is_license_notice: yes
relevance: 100
referenced_filenames:
    - LICENSE
key_phrases:
    - MIT
```

or something to that effect.

I have spent zero time in the codebase, but if given a proper introduction I suppose I might be able to help out.

mrombout:
Fixes #2637

This PR depends on https://github.com/nexB/scancode-toolkit/pull/2765/commits/97a1a57eec13566328c26e67df6af1bb409144a9, contained in https://github.com/nexB/scancode-toolkit/pull/2765. I have cherry-picked that commit into this PR as 2428723959a2cdc488148e5ecc88807a5d358ebe.

### Tasks


AUTHORS.rst
```diff
@@ -43,6 +43,7 @@ The following organizations or individuals have contributed to ScanCode:
 - Maximilian Huber @maxhbr 
 - Michael Herzog @mjherzog
 - Michael Rupprecht @michaelrup
+- Mike Rombout @mrombout
```
pombredanne, line 47: Can you add `- Jiyeong Seok @dd-jy`   too there?

CHANGELOG.rst
```diff
@@ -49,6 +49,9 @@ License detection:
   - XXXX new license detection rules have been added, and
   - XXXX existing license rules have been updated.
 
+- Key phrases can now be defined in RULEs by surrounding one or more words with
```
pombredanne, line 52: BTW, does this also applies to .LICENSE texts? (not I would want to encourage this in license texts...)

src/licensedcode/data/rules/false-positive_ansible_2.RULE
```diff
@@ -5,4 +5,4 @@
   # - GPLv3
   # - Apache
   # - CC-BY
-license: {{ license }}
+license: \{\{ license \}\}
```
mrombout: In this case the `{{ license }}` is not meant to be a key phrase. As a workaround I've "escaped" the `{{` with a `\`, this as not explicit meaning in the parsing code, it simply trips up the regular expression used to find the tags.

pombredanne: IMHO, do not escape things as it looks weird... just remove the braces entirely.

tests/licensedcode/test_match.py
```diff
@@ -873,10 +1026,10 @@ class TestCollectLicenseMatchTexts(FileBasedTesting):
 
     def test_get_full_matched_text_base(self):
         rule_text = u'''
-            Copyright {{some copyright}}
-            THIS IS FROM {{THE CODEHAUS}} AND CONTRIBUTORS
-            IN NO EVENT SHALL {{THE CODEHAUS}} OR ITS CONTRIBUTORS BE LIABLE
-            EVEN IF ADVISED OF THE {{POSSIBILITY OF SUCH}} DAMAGE
+            Copyright [[some copyright]]
```
mrombout, line 1051: Simply replaced them as a workaround to avoid conflict with the key phrase syntax, do the `{{` and `}}` characters have another special meaning here?

tests/licensedcode/data/datadriven/lic1/boto-boto3-cw-example-creating-alarm.txt.yml
```diff
@@ -0,0 +1,5 @@
+license_expressions:
```
mrombout, line 1: Should be only `cc-by-nc-sa-4.0`, missing rule.

pombredanne: correct. Adding a new rules for cc-by-nc-sa-4.0 such as these would cure this

- `is_license_notice: yes`

```
   This work is licensed under a {{Creative Commons Attribution-NonCommercial-ShareAlike 4.0}}
   International License (the "License"). You may not use this file except in compliance with the
   License. A copy of the License is located at http://creativecommons.org/licenses/by-nc-sa/4.0/.

   This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
   either express or implied. See the License for the specific language governing permissions and
   limitations under the License.
```

and `is_license_reference: yes` and `relevance: 100`
```
{{Creative Commons Attribution-NonCommercial-ShareAlike 4.0}} International License
```

pombredanne (changes requested):
Thanks!
See a few comments in line.

AUTHORS.rst
pombredanne: @dd-jy I assume you are OK?

tests/licensedcode/test_match.py
pombredanne: > do the {{ and }} characters have another special meaning here?

In a primitive iteration of license detection there was a notion of "template" where `{{10 some text}}` would mean that up to 10 words could be missed and the match still be valid.
- https://github.com/nexB/scancode-toolkit/blob/v1.0.0/src/licensedcode/models.py#L256
- https://github.com/nexB/scancode-toolkit/blob/v1.0.0/src/textcode/analysis.py#L146
- https://github.com/nexB/scancode-toolkit/blob/v1.0.0/src/textcode/analysis.py#L193

This was byzantine and did not work as expected at all!

pombredanne (approved):
This is overall awesome!
I'd like to check if the impact on indexing is not too bad... but this is not a critical process anymore.

src/licensedcode/match.py
```diff
@@ -1389,6 +1395,55 @@ def filter_already_matched_matches(matches, query):
     return kept, discarded
 
 
+def filter_key_phrase_spans(matches):
+    """
+    Return a filtered list of kept LicenseMatch matches and a list of
+    discardable matches by removing all matches that do not contain all key
+    phrases required by the rule.
```
pombredanne, line 1402: What if there is only one match and this does not contain the keyphrases?
Would you discard it, therefore leaving no matches?

src/licensedcode/match.py
```diff
@@ -1485,6 +1540,10 @@ def _log(_matches, _discarded, msg):
         all_discarded.extend(discarded)
         _log(matches, discarded, 'HIGH ENOUGH SCORE')
 
+    matches, discarded = filter_key_phrase_spans(matches)
```
pombredanne: I think this filter should be positioned much earlier in the process otherwise here is the risk:
1. there was a smalller match that was contained
2. it is filtered because of containment in a larger match
3. the larger is discarded because of not having keyphrases
4. there is no match left for this region and we now have a false negative

pombredanne (changes requested):
I added some more comments for your review.

src/licensedcode/data/rules/false-positive_ansible_2.RULE
mrombout: Okay, removed them!

tests/licensedcode/test_match.py
mrombout: But this feature is not in anymore, is this test still relevant/correct? Since I changed the characters to `[[` and `]]` (which means nothing), and it didn't break :thinking: Same for the other similar ones.

tests/licensedcode/data/datadriven/lic1/boto-boto3-cw-example-creating-alarm.txt.yml
mrombout: Nice thanks, fixed by adding these rules. I've added two rules where one is the license notice and the other is the license reference. Adding the license notice fixed the test, is the license reference for completeness?

CHANGELOG.rst
mrombout: Definitely not on purpose, I am not sure on `.LICENSE` texts are used in the system. Though, adding the syntax to a `.LICENSE` text does break some tests, I'll have a look at it.

At the moment the key_phrases are added to the index for every rule https://github.com/softsense/scancode-toolkit/blob/issue-2637-allow-license-rules-to-require-the-presence-of-certain-defining-keywords/src/licensedcode/index.py#L334.

src/licensedcode/match.py
mrombout: If the only match does not contain the key phrase wouldn't that indicate a high chance of being a false positive? If so, for my use-case I would rather have no results, but maybe that's not in-line with Scancode philosophy(e.g. not having false negatives)?

tests/licensedcode/test_match.py
pombredanne: > But this feature is not in anymore, is this test still relevant/correct? Since I changed the characters to [[ and ]] (which means nothing), and it didn't break thinking Same for the other similar ones.

Yes it could be irrelevant but no harm done either to keep it: this is minor so your changes are fine.

tests/licensedcode/data/datadriven/lic1/boto-boto3-cw-example-creating-alarm.txt.yml
pombredanne: >  is the license reference for completeness?

yes!

CHANGELOG.rst
pombredanne: License texts are ingested as if they were rules when indexing: https://github.com/nexB/scancode-toolkit/blob/1f7fafdcc74f1110a7812b4ddf84964b4fe67843/src/licensedcode/models.py#L522 so this likely works there too... I checked that there are no dangling {{ braces }} in any .LICENSE so we are fine too

src/licensedcode/match.py
pombredanne: > If the only match does not contain the key phrase wouldn't that indicate a high chance of being a false positive?

yes in the general case, but let's take a practical  example:

1. in query I have  `licensed under the AGPL 3.0` and in rule1 I have `AGPL 3.0`
2. in rule2 I have `licensed under the {{AGPL 3.0 license}}`
3. the query will initially match both rules
4. the match to rule1 may be discarded as contained in the rule2 match
5. the match to rule2 may be discarded as missing a key phrase

I am left with no matches yet `licensed under the AGPL 3.0` was very clearly a license notice.

On the other hand if the match to rule2 is discarded as missing a key phrase early and before the containment filter, then we are at least left with the match to rule1?

mrombout: Okay, that makes sense. But unfortunately I was unable to recreate this scenario in a unit test.

Instead I experimented and moved the key phrase filter up (before the containment filter) and 5 of the datadriven tests I've added broke. Looking into that now. E.g. `tests/licensedcode/test_detection_datadriven1.py::TestLicenseDataDriven1::test_detection_datadriven_lic1_alexa_skills_kit_sdk_for_java_txt` now detects `[apache-2.0, apache-2.0, unknown-license-reference]` as opposed to only `apache-2.0`. 

Also, are you suggesting to skip if there's only one rule like some of the other filters?

```
    if len(matches) < 2:
        return matches
```

tests/licensedcode/data/datadriven/lic1/alexa-skills-kit-sdk-for-java.txt.yml
```diff
@@ -1,2 +1,4 @@
 license_expressions:
   - apache-2.0
+  - apache-2.0
+  - unknown-license-reference
```
mrombout: I'm suspecting that I've added a key phrase to a rule that prevents it from matching, but now also prevents `unknown-license-reference` for being removed as contained.

See which Apache rule it was and see if key phrase can be less strict, or a new rule must be added.

mrombout: `src/licensedcode/data/rules/apache-2.0_151.RULE` would match probably if only it didn't have the word `licensed`. Instead I've added a new rule `apache-2.0_1026.yml` for now.

src/licensedcode/match.py
mrombout: I've moved the key phrase filter up in 2c377b3a6bd69920a7dfb68864d3e7f9bf58901b, to right before the first containment filter. I was unable to reproduce this scenario exactly, but some datadriven tests at least demonstrated the effectiveness (`tests/licensedcode/test_detection_datadriven1.py::TestLicenseDataDriven1::test_detection_datadriven_lic1_alexa_skills_kit_sdk_for_java_txt`)

mrombout: @pombredanne Would the above suffice?

pombredanne (approved):
LGTM and thank you ++ for this!

PR #2775, opened 2021-12-01
Fix azure pipeline vmimage deprecations

AyanSinhaMahapatra:
`macos-10.14` and `vs2017-win2016` vm images are deprecated and
this replaces them with newer images.

See https://devblogs.microsoft.com/devops/hosted-pipelines-image-deprecation/

`macos-10.14` would be deprecated on December 10, 2021
`vs2017-win2016` would be deprecated on March 15, 2022

### Tasks

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -93,9 +93,9 @@ jobs:
 
     - template: etc/ci/azure-posix.yml
       parameters:
-          job_name: mac1014
-          image_name: macos-10.14
-          python_versions: ['3.6', '3.7', '3.8', '3.9']
+          job_name: mac11
+          image_name: macOS-11
+          python_versions: ['3.7', '3.8', '3.9']
           python_architecture: x64
           test_suites:
               all: bin/pytest -n 2 -vvs tests/scancode/test_cli.py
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -123,9 +123,9 @@ jobs:
 
     - template: etc/ci/azure-win.yml
       parameters:
-          job_name: win2016
```
pombredanne, line 126: IMHO going from 2016 to 2022 is too big of a version jump at once..... what is the oldest version available after win2016? use this one instead

azure-pipelines.yml
```diff
@@ -93,9 +93,9 @@ jobs:
 
     - template: etc/ci/azure-posix.yml
       parameters:
-          job_name: mac1014
```
pombredanne, line 96: IMHO going from 10.14 to 11 is too big of a version jump at once..... what is the oldest version available after 10.14? use this one instead

pombredanne (changes requested):
Thanks! I think we should keep also the oldest supported versions first and it's OK to also add newest 11 and 2022 versions like you did... but we cannot only support the latest and greatest.

AyanSinhaMahapatra:
I looked into the images available [here](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#software) and [here](https://github.com/actions/virtual-environments/tree/main/images)

In windows we already did the tests on `windows-2019`, it is just below the `windows-2022` test. These are the only two images listed above.

In MacOS, there are only three, 10.14, 10.15 and 11, of which 10.14 will be deprecated and removed soon. So replaced those tests too.

Also note that I removed python 3.6 in the newer images, as they are not supported and not present so were generating [errors](https://dev.azure.com/nexB/scancode-toolkit/_build/results?buildId=5144&view=results). 

Do you also mean the ordering on the tests in the azure-pipelines.yml file? Like the  `windows-2019` should be before the `windows-2022` test?

pombredanne (approved):
> Do you also mean the ordering on the tests in the azure-pipelines.yml file? Like the windows-2019 should be before the windows-2022 test?

No, we are all good as it is ... thank you ++... merging!

PR #2797, opened 2022-01-15
Fixing OSI identifier for BSD-3-Clause; see also SPDX license metadata

karsten-klein:
Just something rather minor. Perhaps a good example to learn how to contribute to Scancode-Toolkit. We found that linking to OSI using the BSD-3 as identifier does not work. The "official id" is BSD-3-Clause.

You may want to check out: 
https://opensource.org/licenses/BSD-3
https://opensource.org/licenses/BSD-3-Clause

Regards,
Karsten


pombredanne (approved):
Karsten: thank you ++

Issue #2761, opened 2021-11-19
packagecode: gradle nebula dependency lock parser


tardyp:
## Short Description

gradle is a build tool for the ajva ecosystem especially popular in android apps.

nebula gradle dependency lock allows developer to lock their gradle dependencies and sub dependencies to specific versions
packagecode should be able to parse this simple information 

## Possible Labels

- new feature
- package scan

## Select Category

-  [ ] Enhancement
-  [ ] Add License/Copyright
-  [x] Scan Feature
-  [ ] Packaging
-  [ ] Documentation
-  [ ] Expand Support
-  [ ] Other

## **Describe the Update**

lock format is documented here:

https://github.com/nebula-plugins/gradle-dependency-lock-plugin/wiki/Usage#lock-file-format

It is a simple json format that clearly shows the packages and versions.
Should be quite easy to parse

## **How This Feature will help you/your organization**

This will have using scancode to extract our gradle dependencies.

## **Can you help with this Feature**

Renault will probably contribute this feature in the next month

pombredanne:
@tardyp note that I have done quite a bit of research on how to parse gradle builds at least the Groovy kind, and we could likely consider the Kotlin kind too

tardyp:
@pombredanne That particular request is about nebula kind of package locks.

nebula looks like the poetry of gradle. Cool kids use it instead of the default gradle pinning method.
I don't think it is really about kotlin, but just as Poetry.lock is another file format as requirement.txt or Pipfile.lock.

Big advantage I see is that it is a lockfile and there is no need to run the gradle toolchain to extract the info. It is just a json file.

pombredanne:
@tardyp FYI @JonoYang is contributing some support for gradle in #2822

pombredanne:
I think we should also support first the standard Gradle lockfile: https://docs.gradle.org/current/userguide/dependency_locking.html

- Names: `gradle.lockfile` and buildscript-gradle.lockfile`
- Content: This is an ini or properties-like file:

> Each line still represents a single dependency in the group:artifact:version notation
> It then lists all configurations that contain the given dependency
> Module and configurations are ordered alphabetically, to ease diffs
> The last line of the file lists all empty configurations, that is configurations known to have no dependencies



```properties 
# This is a Gradle generated file for dependency locking.
# Manual edits can break the build and are not advised.
# This file is expected to be part of source control.
[...]

Issue #793, opened 2017-10-06
Collect package information from a .gradle script


pombredanne:
This is a package manifest of sorts, equivalent to a pom.xml 
See https://docs.gradle.org/4.2.1/dsl/

pombredanne:
See the PR in https://github.com/nexB/scancode-toolkit/pull/2822
We still need support for kotlin build scripts too.

pombredanne:
See also https://github.com/nexB/scancode-toolkit/issues/2761

pombredanne:
For a kotlin example, see https://github.com/oss-review-toolkit/ort/pull/5118/files#diff-33aaad2b3b1af4604de8953a12f27fd248b6d4ab52d1d26ec5aff807123d2715

PR #2822, opened 2022-01-30
Detect dependencies from build.gradle files

pombredanne:
* Create parser for dependency info from build.gradle files

### DIFF
--- a/src/packagedcode/build.py
+++ b/src/packagedcode/build.py
@@ -9,19 +9,11 @@
 
 from collections import defaultdict
 import ast
-import io
-import json
 import logging
 import os
 
-from packageurl import PackageURL
-from pygments import highlight
-from pygments.formatter import Formatter
-from pygments.lexers import GroovyLexer
-from pygments.token import Token
 import attr
 
-from commoncode import filetype
 from commoncode import fileutils
 from packagedcode import models
 from packagedcode.utils import combine_expressions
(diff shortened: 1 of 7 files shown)


src/packagedcode/build.py
```diff
@@ -9,9 +9,16 @@
...
 import ast
+import io
+import json
 import logging
 import os
 
+from packageurl import PackageURL
+from pygments import highlight
+from pygments.formatter import Formatter
+from pygments.lexers import GroovyLexer
```
tardyp: interresting use of pygments :)

tardyp (commented):
Thanks for this. I thinks its going to be useful. I can foresee some problems with this:

- we need tests with some example files, to see what works exactly. I can provide some twisted examples from my twisted kotlin dev collegues :)

- build.gradle can be either groovy or kotlin.
- those are programmation langage so developers are factorizing deps, so for me it just unreliable to just parse the file, you have to execute it.
- gradle is supporting transiant dependencies. So build.gradle only contains rank 1 deps.
- ort, whitesource do run gradle to work. this is painful as you have to provision your build container with your build toolchain and the scanner toolchain

For those reason I advise my devs to use lockfiles like nebula.
Not all of them have made the switch so this will certainly be useful.

src/packagedcode/build.py
pombredanne: @tardyp the idea of https://github.com/nexB/pygmars/ is to be used in tandem with pygments to create quickly some lightweight parsers for programming languages. 

This is used for instance to lex bash in https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/bashlex.py and parse it in https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/bashparse.py
In this case this is used to enable lightweight Bash parameter extension to parse the Apkbuild scripts from Alpine Linux (to collect metadata) in https://github.com/nexB/scancode-toolkit/blob/9ed2cb4e78a8b1b138bcb201f50203b43098febb/src/packagedcode/alpine.py#L161

The same approach will be used whenever there is a need to parse code to extract structured data as this is easier to read and maintain than a bunch of regex.  The  upcoming Yocto/Bitbake support likely can benefit from some of it too... and we should be able to parse also Autotools scripts that contain interesting metadata.

pombredanne: Note that this is not perfect... and still only static. Next up will be to enable dependency resolution on the deps collected (which may require some live network access ... but this will be an option with proxy support for airgapped envt. ;) )

JonoYang:
@tardyp Thanks for the review! I've got a script going that uses [pygmars](https://github.com/nexB/pygmars/) to roughly glean what I can from build.gradle files. I'll bring the code over.

pombredanne:
@tardyp Some more thoughts in reply to your insightful comments:

> Thanks for this. I thinks its going to be useful. I can foresee some problems with this:
> 
>     * we need tests with some example files, to see what works exactly. I can provide some twisted examples from my twisted kotlin dev collegues :)

This would be great if you can attach these.  

>     * build.gradle can be either groovy or kotlin.

This is may work also with kotlin, but the devils is in the details.

>     * those are programmation langage so developers are factorizing deps, so for me it just unreliable to just parse the file, you have to execute it.

yeah, that's the problem with these executable scripts....

>     * gradle is supporting transiant dependencies. So build.gradle only contains rank 1 deps. 
>     * ort, whitesource do run gradle to work. this is painful as you have to provision your build container with your build toolchain and the scanner toolchain

So the plan here would be to:
1. enable resolving deps WITHOUT running gradle (or in general a package manager) using DependentCode, univers and purl/vers

This is not going to be a perfect solution in all cases.

2. also run tools using gradle minimally with containers (possibly borrowing or reusing code from ORT) to collect a full dependency tree

3. process also the lockfiles... 


> 
> For those reason I advise my devs to use lockfiles like nebula. Not all of them have made the switch so this will certainly be useful.

which would be super useful to have as a parser and having some test files would be super useful too. Do you think nebula is the most common one? Any  other common lockfile format otherwise?

pombredanne (commented):
LGTM!

pombredanne:
See also https://github.com/nexB/scancode-toolkit/issues/2761

pombredanne:
I am keeping https://github.com/nexB/scancode-toolkit/issues/793#issuecomment-1055739373 open until we have explicit tests for Kotlin

tardyp:
> which would be super useful to have as a parser and having some test files would be super useful too. Do you think nebula is the most common one? Any other common lockfile format otherwise?

FWIW, we do have some nebula support code, and we do need to find some slot in our agenda to upstream it :-}

for the rest, sounds like a plan!

pombredanne:
> FWIW, we do have some nebula support code, and we do need to find some slot in our agenda to upstream it :-}

Awesome... I look forward to it!

Issue #2693, opened 2021-09-06
Files model updates: add `for_packages` attribute


pombredanne:
Files model updates: add `for_packages` attribute which is a list of { purl + UUID, e.g. an instance of a package} this file is for. This replaces "installed_files" experiments.

AyanSinhaMahapatra:
Addressed in #2825 . Closing this!

Issue #2692, opened 2021-09-06
Package model updates: Add `package_manifest_paths` attribute


pombredanne:
Package model updates: Add `package_manifest_paths` attribute that is a list of codebase paths for each `package_manifest` file that is contributing to a top-level Package instance (not the absolute locations)

AyanSinhaMahapatra:
Addressed in #2825 , closing this!

Issue #2652, opened 2021-08-20
Clarify the structure of packagedcode.models.Package subclasses


pombredanne:
This structure and how detection/recognition happens is not always obvious.
Also there are possible challenges when "rehydrating" a package plain data to a Package object.

We would need a doc and some refactoring as having several package manifests extensions for a single Package subclass makes things complicated.

Some redesign needed too in light of tracking package instances vs. manifests.

pombredanne:
This has been completed. We now have a simpler structure:
- one class for each DatafileHandler, e.g. a manifest format parser and handler
- a single class for PackageData to keep track of parsed manifest data at the file level
- a single class for Package and Dependency to keep track of assembled manifest data at the top level
Closing this now.

Issue #2098, opened 2020-07-01
Report Packages at the codebase-level


JonoYang:
We should make the package consolidation logic from the consolidation plugin a default function of the Package scanning option. The consolidation plugin would then be focused on files that are not part of a package so we can perform logical groupings on them.

Some changes that would have to change on the Package model/Package scanning process:

- Remove root_path from Package. There is no universal root path for all Packages. 
- Tag a resource if it is part of one or more Package during a Package scan. This would be similar to the `consolidated_to` field, where it would be a list of purls. Possible name for this field is `for_packages`.
- Return detected Packages in a new top level codebase attribute. This would be a set of all detected packages from a codebase.
- Add primary license expression/copyright/holders to Package models, which is populated from top-level key package files (manifests, etc.)
- Add secondary license expression/copyrights/holders to Package. This is populated from the detected license expressions/copyrights of Package resources, excluding top-level key files.

--------------------------------------
Here is an updated design:

The key elements are to:

- report `packages` as top-level. The data structure is the same as the one at the file level but will be the merged data from possibly several manifests and lock files.
- track which files are part of a given package instance

### Package model updates

- [x] #2691
- [x] #2692
- [x] #2747
  
### Files model updates
- [x] #2693
- [x] #2694

This could look this way:
```
packages (aka. package instances)
  package
    package_manifest_paths
    ... data...

files:
   package_manifests: [...] (formerly named packages)
   for_packages: [ list of 
     { package_url: pkg:foo/bar@12, package_instance: UUID}
   ]
```

### For later:

- [ ] sub-packages/embedded could be 1) listed directly as `packages`  of their own 2) related to their parent (or the parent related to them)
- [ ] we could also track the `file_paths` under each top level package instance
- [ ] get all packages to have actual files
- [x] migrate system packages (Alpine, RPM, Debian, Windows) to new model

pombredanne:
Something to consider if we were to track packages at the top level:
- should we track each instance of a package (and its files separately)
- what happens if different package instances have slightly metadata?

pombredanne:
@JonoYang I reckon this was never merged. And we will need this but IMHO we should go one more step:


1. the things we detect today are incorrectly reported as `packages` when they really are package manifests
2. some packages may have multiple manifests (e.g. manifest proper and a lock file). For instance .gemspec/Gemfile/Gemfile.lock or package.json/package-lock.json/yarn.lock or setup.py/setup.cfg/requirements.txt ... etc.

[...]

PR #2825, opened 2022-01-31
Add Package Instances #2691

### LINKED ISSUE #2691: Package model updates: Add new `package_instance`
Package model updates: Add new `package_instance` attribute used to track unique instances of each packages (say you have two identical wheels in the same code tree);use a UUID for this `package_instance` attribute.

AyanSinhaMahapatra:
This PR adds the PackageInstance class and functions to group package
manifests and package data as top level package instances.

Existing package data are ported to this new approach.

Reference: https://github.com/nexB/scancode-toolkit/issues/2098 
Reference: https://github.com/nexB/scancode-toolkit/issues/2691 
Reference: https://github.com/nexB/scancode-toolkit/issues/2692
Reference: https://github.com/nexB/scancode-toolkit/issues/2693
Reference: https://github.com/nexB/scancode-toolkit/issues/2843 
Reference: https://github.com/nexB/scancode-toolkit/issues/2652 

### Tasks


tests/packagedcode/test_package_instance.py
```diff
@@ -0,0 +1,41 @@
...
+
+class TestPackageInstance(PackageTester):
+    test_data_dir = os.path.join(os.path.dirname(__file__), 'data')
+
+    def test_package_instance_scan_python(self):
+        test_dir = self.get_test_loc('instance/pypi')
+        result_file = self.get_temp_file('json')
+        expected_file = self.get_test_loc('instance/python-package-instance-expected.json')
+        run_scan_click(['--package', '--strip-root', '--processes', '-1', test_dir, '--json', result_file])
+        check_json_scan(expected_file, result_file, regen=True)
```
JonoYang: `regen` should be set to `False`

src/packagedcode/__init__.py
```diff
@@ -110,28 +110,36 @@
...
-            raise Exception(msg)
-        else:
-            seen_types[manifest.package_manifest_type] = manifest
+
+PACKAGE_INSTANCES_BY_TYPE = {
+    cls.default_type: cls
+    for cls in PACKAGE_INSTANCE_TYPES
+}
+
+def check_package_manifest_classes():
```
JonoYang: Is `check_package_manifest_classes()` called anywhere? Previously, the code in the function was outside of a function. My assumption was that whenever the `packagedcode` module is imported, this code that checks `PACKAGE_MANIFESTS_BY_TYPE` and `PACKAGE_MANIFEST_TYPES` would always run and give us an exception whenever a package was not properly set.

src/packagedcode/__init__.py
```diff
@@ -110,28 +110,36 @@
...
-                  .format(manifest.package_manifest_type, manifest.__name__, seen.package_manifest_type, seen.__name__,))
-            raise Exception(msg)
-        else:
-            seen_types[manifest.package_manifest_type] = manifest
+
+PACKAGE_INSTANCES_BY_TYPE = {
+    cls.default_type: cls
+    for cls in PACKAGE_INSTANCE_TYPES
+}
+
```
JonoYang, line 150: Style nitpick: there should be an extra new line here.

src/packagedcode/models.py
```diff
@@ -765,6 +765,251 @@ def recognize(cls, location):
...
+        Returns a mapping of package data attributes and corresponding values.
+        """
+        mapping = self.to_dict()
+
+        # Removes PackageInstance specific attributes
+        for attribute in ('package_uuid', 'package_manifest_paths', 'files'):
+            mapping.pop(attribute)
+        
+        # Remove attributes which are BasePackage functions
+        for attribute in ('repository_homepage_url', 'repository_download_url', 'api_data_url'):
```
JonoYang, line 832: I have no strong feelings one way or the other about this, but I am curious as to why `repository_homepage_url`, `repository_download_url`, and `api_data_url` are removed from the package data that is to be returned.

JonoYang (commented):
@AyanSinhaMahapatra Initial work on the `PackageInstance` model looks good. I just have some minor things I pointed out.

tests/packagedcode/test_package_instance.py
AyanSinhaMahapatra: I was having some problems with the `package_uuid` field as it had a UUID that was different for each test run. So I forgot to make this False, now this is resolved. I've also modified the `check_json_scan` to accept a new argument `remove_package_uuid` which removes the UUID qualifier from the `package_uuid` pURL string.

src/packagedcode/__init__.py
AyanSinhaMahapatra: See [here](https://github.com/nexB/scancode-toolkit/pull/2748#discussion_r761111416) this was supposed to be moved into a test instead, but it was not completed. Now I've pushed the proper tests.

src/packagedcode/__init__.py
AyanSinhaMahapatra: Added, thanks!

AyanSinhaMahapatra: Added, thanks!

src/packagedcode/models.py
AyanSinhaMahapatra: So this is a helper function used in the merging of package_manifest data into a package_instance [here](https://github.com/nexB/scancode-toolkit/blob/2098-top-level-packages/src/packagedcode/models.py#L897). These are basically fields which are not set as values in a package data, but obtained from functions like [this](https://github.com/nexB/scancode-toolkit/blob/2098-top-level-packages/src/packagedcode/models.py#L320) in the `to_dict()`. So we don't need to merge these data from the mappings.

pombredanne:
re: https://github.com/nexB/scancode-toolkit/pull/2825#discussion_r800437754

> I was having some problems with the package_uuid field as it had a UUID that was different for each test run. So I forgot to make this False, now this is resolved. I've also modified the check_json_scan to accept a new argument remove_package_uuid which removes the UUID qualifier from the package_uuid pURL string.


There are three ways to handle these cases:

1. ignore it as you have done but you need to have at least one test case that would not ignore the UUID to ensure this is generated correctly.
2. use mocks in tests so that you have a stable mocked UUID created and do not need to ignore it
3. add something like an extra `_test=True` argument to the tested function(s) that is called only in tests and that creates a static UUID that is well known for testing.

The more generic solution is 2. ... use mock.

AyanSinhaMahapatra:
@pombredanne This is ready!

pombredanne (approved):
LGTM. Thanks! merging now.

Issue #1765, opened 2019-10-16
Investigate a way to avoid elf/dwarf parser platform dependencies


licodeli:
### Description
Let's find a good way to handle the OS platform dependencies for elf and dwarf parsers.

One possible solution is:
https://github.com/eliben/pyelftools

pombredanne:
The newer compiledcode plugin now uses pyelftools ... See https://github.com/nexB/scancode-plugins/blob/53067ec8adc19e4809912a3c31eb73d668b918c2/binary-analysis/scancode-compiledcode/setup.py#L46
Closing now!

PR #2834, opened 2022-02-02
Bundle Compiledcode experimental as a plugin

pombredanne:
See also https://github.com/nexB/scancode-toolkit/issues/1765


pombredanne:
We are doing things differently now with various inspectors such as the elf-inspector

PR #2914, opened 2022-04-11
Update summary plugin

JonoYang:
This PR further overhauls the summary plugin in summarycode. The previous summary plugin returned a count of all the detected license expressions, holders, programming languages, and other scan attributes. We are now evolving that, where we have scancode-toolkit automatically deduce the correct origin info for the codebase it scans using the summary plugin.

This PR also updates commoncode to version 30.1.0.

The summary codebase attribute created by the summary plugin now has these fields:

`declared_license_expression`
  - Primary license expression for a scanned codebase
 
`license_clarity_score`
  - The score of how well a codebase has declared its origin information

`declared_holder`
  - Primary holder of the scanned codebase
  
`primary_language`
  - The main programming language of the scanned codebase 
 
`other_license_expressions`
  - Other detected license expressions from the scanned codebase, excluding the `declared_license_expression`
 
`other_holders`
  - Other detected copyright holders from the scanned codebase, excluding the `declared_holder`
 
`other_languages`
  - Other detected programming languages, excluding `primary_language`.


JonoYang:
The remaining pieces left are to figure out what to do when multiple package data is detected at the root of the codebase we are scanning. Currently, we are under the assumption that the codebase we are scanning is a package itself. However, some packages have multiple package manifests at the same root directory. We have some code that picks which package data to use based on the number of code files that are of the same language for those packages, but it is not a reliable heuristic.

JonoYang:
To handle the case of multiple package data present at a root of a directory, we've decided to just combine all the package data and return it as the declared origin information.

pombredanne (changes requested):
Looking great, yet I do not think this is worth updating the attribute name to tallies in the cluecode tests. Changing just the code should be enough IMHO.  What do you think?

JonoYang:
@pombredanne In the cluecode tests, I've renamed the `tallies` attribute back to `summary`. I'm still failing a test for the CycloneDX output plugin and I'm not sure why it's doing that. I've opened a ticket for that #2917

pombredanne (changes requested):
Looking great and avoiding the copyright tests YAML means we have more reasonable churn.
Some expected JSON still need fixing though and IMHO we should not report thing without copyrights or holder in tallies but only things that have a value

PR #2909, opened 2022-04-05
Add system packages support in the new packages model

AyanSinhaMahapatra:
### Tasks

* [x] Reviewed [contribution guidelines](https://github.com/nexB/scancode-toolkit/blob/develop/CONTRIBUTING.rst)
* [x] PR is descriptively titled 📑 and links the original issue above 🔗
* [x] Tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR
  Run [tests](https://scancode-toolkit.readthedocs.io/en/latest/contribute/contrib_dev.html#running-tests) locally to check for errors. 
* [x] Commits are in uniquely-named feature branch and has no merge conflicts 📁

PR #2920, opened 2022-04-22
Dedupe declared holders detected from package data

JonoYang:
This PR dedupes holders in the summary plugin when we collect them from package data


JonoYang:
Closing this as the fix is present in #2909

PR #2932, opened 2022-04-29
Fix package/dependency creation bugs

AyanSinhaMahapatra:
### Tasks


src/packagedcode/pypi.py
```diff
@@ -71,6 +71,14 @@ class BasePypiHandler(models.DatafileHandler):
     def compute_normalized_license(cls, package):
         return compute_normalized_license(package.declared_license)
 
+    @classmethod
```
pombredanne: What do you gain from overriding these two methods with the exact same code that's in the super class?

pombredanne (approved):
LGTM!
I just have a small comment wrt. https://github.com/nexB/scancode-toolkit/pull/2932/files#r861907921 where subclassing is not needed especially since you call super() which will travel up the class hierarchy and MRO alright

src/packagedcode/pypi.py
AyanSinhaMahapatra: If in pypi the `PythonEditableInstallationPkgInfoFile.assign_package_to_resources` is overridden to call `cls.assign_package_to_parent_tree` then, the `assign_package_to_parent_tree` again calls the `assign_package_to_resources` function which has now been overridden to call the same. This causes the recursion.

AyanSinhaMahapatra:
@pombredanne updated that.

pombredanne:
LGTM. Thanks! merging!

Issue #2945, opened 2022-05-06
RFC: categorize codebase files to prioritize license-related risk analysis


johnmhoran:
A codebase can have thousands or more files of a vast range of types, from those that are clearly code (`.py`, `.cpp`, `.so`) to those that clearly are not code (`.txt`, `.xlsx`), and many that are somewhere in between (`.jsp`, `.html`, `.erb`).  Our goal: automate the identification of which files are more or less important in the risk analysis process so that the more important files can be analyzed first, and those that need not be analyzed can be identified and omitted from analysis.

To accomplish this goal, we intend to add a feature that will (1) apply a set of rules -- definitions, really -- to a subset of each file's attributes and (2) add three fields to the ScanCode Toolkit scan output metadata that identify the file by category and sub-category and rank each file for license-risk-related importance, i.e., analysis priority.

**Relevant file attributes**

Define each of the rules in terms of the contents of one or more of these attributes:

- `extension`
- `name`
- `mime_type`
- `file_type`
- `programming_language`

**Categorization fields**

- `analysis_priority` ==> a scale from 1-3, with 1 being most relevant for license-risk-analysis purposes
- `file_category` ==> e.g., `archive`, `binary`, `source`, `manifest`, `doc`, `media`, `script`
- `file_subcategory` ==> e.g., `c++`, `python`, `make`, `json`, `license`, `audio`, `data`

**Output format**

We anticipate that this file categorization data will be available as three new fields in all of ScanCode Toolkit's output formats, e.g., `.json`, `.xlsx` et al.

PR #2958, opened 2022-05-13
[WIP] Add new file categorization plugin #2945

johnmhoran:
This PR add a new file categorization plugin.


johnmhoran:
This PR is related to issue #2945.

src/summarycode/file_cat.py
```diff
@@ -0,0 +1,1265 @@
...
+# SPDX-License-Identifier: Apache-2.0
+# See http://www.apache.org/licenses/LICENSE-2.0 for the license text.
+# See https://github.com/nexB/scancode-toolkit for support or download.
+# See https://aboutcode.org for more information about nexB OSS projects.
+#
+
+from commoncode.datautils import String
+from plugincode.post_scan import PostScanPlugin
+from plugincode.post_scan import post_scan_impl
+from commoncode.cliutils import PluggableCommandLineOption
```
pombredanne, line 13: Please sort your imports

src/summarycode/file_cat.py
```diff
@@ -0,0 +1,1265 @@
...
+from commoncode.cliutils import POST_SCAN_GROUP
+
+"""
+Categorize files.
```
pombredanne, line 17: Could you expand the documentation to explain the workings of this module at a high level?

src/summarycode/file_cat.py
```diff
@@ -0,0 +1,1265 @@
...
+            ("file_category", String(help="File category.")),
+            ("file_subcategory", String(help="File subcategory.")),
+        ]
+    )
+
+    sort_order = 50
+
+    options = [
+        PluggableCommandLineOption(
+            ("--file-cat",),
```
pombredanne, line 69: We need a more explicit, non-abbreviated command line option, like `--categorize-file` or `--file-category` to stay consistent with other options.

src/summarycode/file_cat.py
```diff
@@ -0,0 +1,1265 @@
...
+    analysis_priority = None
+    file_category = None
+    file_subcategory = None
+    category_notes = None
+    rule_applied = None
+
+    @classmethod
+    def categorize(cls, resource):
+        """
+        Return True if this Categorizer applies to this resource or False or None
```
pombredanne, line 106:
```suggestion
        Return True if this Categorizer category applies to this ``resource``.
```
But what is ``resource`` ? a path string? or a location? some object? we need to be precise there.

src/summarycode/file_cat.py
```diff
@@ -0,0 +1,1265 @@
...
+            category = categorize_resource(resource)
+            if not category:
+                continue
+            resource.analysis_priority = category.analysis_priority
+            resource.file_category = category.file_category
+            resource.file_subcategory = category.file_subcategory
+            resource.save(codebase)
+
+
+class Categorizer:
```
pombredanne, line 95: Add some docstring to document what is this class used for and how to extend it
```suggestion
class Categorizer:
    """
    A categorizer can categorize a file Resource based on its path, name,
    extension and other attributes such as file or mime type.
    Subclasses must override the ``categorize`` class method.
    """
```

src/summarycode/file_cat.py
```diff
@@ -0,0 +1,1265 @@
...
+    order = 0
```
pombredanne, line 96: Each of these attributes needs to be documented to explain what they mean.

src/summarycode/file_cat.py
```diff
@@ -0,0 +1,1265 @@
...
+
+def logger_debug(*args):
+    pass
+
+
+if TRACE:
+    import logging
+    import click
+
+    class ClickHandler(logging.Handler):
```
pombredanne, line 32: Nice finding, I did not knew we could do this.

src/summarycode/file_cat.py
```diff
@@ -0,0 +1,1265 @@
...
+    order = 10
+    analysis_priority = "1"
+    file_category = "archive"
+    file_subcategory = "debian"
+    category_notes = "special type of archive"
+    rule_applied = "ArchiveDebian"
+
+    @classmethod
+    def categorize(cls, resource):
+        return extension_in(resource.extension, [".deb"])
```
pombredanne, line 134: Since "extension in" is a common check made in several classes we could simplify this by creating a class variable for "extensions" and have a default "categorize" for an ExtensionCategorizer subclass of Categorizer?

src/summarycode/file_cat.py
```diff
@@ -0,0 +1,1265 @@
...
+class BinaryElfKo(Categorizer):
+    order = 10
+    analysis_priority = "1"
+    file_category = "binary"
+    file_subcategory = "elf-ko"
+    rule_applied = "BinaryElfKo"
+
+    @classmethod
+    def categorize(cls, resource):
+        return resource.mime_type in ["application/x-object"] and extension_startswith(
```
pombredanne, line 221: Why startswith here? Do we have examples of LKMs that do not have a full .ko extension?

src/summarycode/file_cat.py
```diff
@@ -0,0 +1,1265 @@
...
+            resource.extension, (".ko")
+        )
+
+
+class BinaryElfO(Categorizer):
+    order = 10
+    analysis_priority = "1"
+    file_category = "binary"
+    file_subcategory = "elf-o"
```
pombredanne, line 230: o is short. What about elf-object instead?

src/summarycode/file_cat.py
```diff
@@ -0,0 +1,1265 @@
...
+                continue
+            resource.analysis_priority = category.analysis_priority
+            resource.file_category = category.file_category
+            resource.file_subcategory = category.file_subcategory
+            resource.save(codebase)
+
+
+class Categorizer:
+    order = 0
+    analysis_priority = None
```
pombredanne, line 97: Could this be named importance or prominence?
analysis_priority seems to imply this is only use to drive some analysis?

src/summarycode/file_cat.py
```diff
@@ -0,0 +1,1265 @@
...
+    def categorize(cls, resource):
+        return extension_in(resource.extension, [".apk", ".aar"])
+
+
+class ArchiveDebian(Categorizer):
+    order = 10
+    analysis_priority = "1"
+    file_category = "archive"
+    file_subcategory = "debian"
+    category_notes = "special type of archive"
```
pombredanne, line 129: why not just call this notes?

src/summarycode/file_cat.py
```diff
@@ -0,0 +1,1265 @@
...
+            resource.file_category = category.file_category
+            resource.file_subcategory = category.file_subcategory
+            resource.save(codebase)
+
+
+class Categorizer:
+    order = 0
+    analysis_priority = None
+    file_category = None
+    file_subcategory = None
```
pombredanne, line 99: We should always enforce these to be lowercase: we have a mix or capital case or lowercase subcategories at the moment.

src/summarycode/file_cat.py
```diff
@@ -0,0 +1,1265 @@
...
+    order = 0
+    analysis_priority = "3"
+    file_category = "build"
+    file_subcategory = "Bazel"
+    rule_applied = "BuildBazel"
+
+    @classmethod
+    def categorize(cls, resource):
+        return extension_in(resource.extension, [".bzl"]) or (
+            name_in(resource.name, ["build.bazel"]) and resource.type == "file"
```
pombredanne, line 301: is resource_type == file needed here? Are these not all about files?

src/summarycode/file_cat.py
```diff
@@ -0,0 +1,1265 @@
...
+    order = 10
+    analysis_priority = "1"
+    file_category = "source"
+    file_subcategory = "assembler"
+    category_notes = "upper case only"
+    rule_applied = "SourceAssembler"
+
+    @classmethod
+    def categorize(cls, resource):
+        return extension_uppercase_in(resource.extension, [".S"])
```
pombredanne, line 869: Why this upper case?

pombredanne (changes requested):
Thanks! See my nits inline.

Also the test file  `jarDependImg.png`  likely comes from http://svn.apache.org/repos/asf/xmlgraphics/batik/branches/webapi/documentation-sources/resources/images/jarDependImg.png under an Apache license that we did not document and is also too big for a test file. Can you instead create an extra small PNG that is not from a third-party?

Issue #2878, opened 2022-02-25
RFC: a plan for false positive license detection


pombredanne:
### Context

We are reporting too many false positive licenses. We need to fix this!

### Problem

There are several false cases, yet they boil down to these types:

1. False detection of very short and weak license detection rules detected exactly such as:
    - a URL or a project name such as a URL to a well known AGPL-licensed which is not always a sign of AGPL as in https://github.com/nexB/scancode-toolkit/issues/2877 
    - the detection of the word `GPL` in a binary https://github.com/nexB/scancode-toolkit/issues/2874
    - the detection of longer `may not be modified` in https://github.com/nexB/scancode-toolkit/issues/2865

2. Detection of a license text or notice fragment which is too weak to represent a bona fide license detection alone.

3. Detection of longer unknown license references such as
    - a "license introduction" (as in "This is licensed under....") that may be noisy when followed by a bona fide license notice or text.
    - a license reference to the license in a file (as in "See file COPYING for license") where we can follow the reference

4. Lack of proper detection of a structured license tag found in a package manifest which is returned as an unknown license 

5. When fragments of the same license are detected with only copyrights added in between as in https://github.com/nexB/scancode-toolkit/issues/2859

6. When sequence of SPDX licenses id are found in license detection tools

7. Please add yours!

### Solution elements

We could treat and report separately mere clues such as this one: they could be an interesting insight in some cases, but alone they are too weak to be considered a license detection

The upcoming two-step process where license matches are grouped in a license detection is another way to consider. We could detect patterns of license matches that could be resolved in a detection. For instance a license intro followed by a license notice.

The scancode-analyzer heuristics and ML-based detection of false positive is another way

porsche-rishisaxena:
Hi Philippe,

In reference to our collective ORT community meeting, we touch base on the false positive license detection 2 weeks ago on version v30.1.0 where Porsche AG OSO also consolidated a report of false-positive cases. Please find attached the report for your kind reference and review.

[report_false_positives.xlsx](https://github.com/nexB/scancode-toolkit/files/8151669/report_false_positives.xlsx)

CC: @sschuberth

PatteSI:
Thank you for taking action here. 
I will now have a deeper look into our false positive findings as well.
EDIT: I forgot to mention that everything I mention below was found using scanCode 30.1.0
[...]

PR #2961, opened 2022-05-17
Combine license matches in new LicenseDetection

AyanSinhaMahapatra:
See #2878 

### Tasks


src/licensedcode/cache.py
```diff
@@ -138,6 +138,40 @@ def load_or_build(
             raise
 
 
+class MockCacheFromTestRules(LicenseCache):
```
pombredanne: Should this be code in a tests utility rather than here?

src/licensedcode/cache.py
```diff
@@ -353,6 +387,14 @@ def get_index(force=False, index_all_languages=False):
     return get_cache(force=force, index_all_languages=index_all_languages).index
 
 
+def get_index_with_test_rules(test_rules_dir):
```
pombredanne: Same comment as above: why not moving this to tests?

src/licensedcode/data/rules/lead-in_unknown_2.yml
```diff
@@ -1,3 +1,3 @@
 license_expression: unknown-license-reference
-is_license_notice: yes
+is_license_intro: yes
```
pombredanne: Good catch!

src/scancode/api.py
```diff
@@ -129,18 +129,11 @@ def get_urls(location, threshold=50, **kwargs):
     return dict(urls=results)
 
 
-SPDX_LICENSE_URL = 'https://spdx.org/licenses/{}'
```
pombredanne: If you remove these from the api module. you need to document this in the CHANGELOG or provide some way to get this from API?

src/licensedcode/detection.py
```diff
@@ -24,34 +27,97 @@
...
     def logger_debug(*args):
-        return prn(' '.join(isinstance(a, str) and a or repr(a) for a in args))
+        return logger.debug(' '.join(isinstance(a, str) and a or repr(a) for a in args))
+
+
+SPDX_LICENSE_URL = 'https://spdx.org/licenses/{}'
+DEJACODE_LICENSE_URL = 'https://enterprise.dejacode.com/urn/urn:dje:license:{}'
+SCANCODE_LICENSEDB_URL = 'https://scancode-licensedb.aboutcode.org/{}'
+
+# All values of match_coverage less than this value are taken as
```
pombredanne: less of higher?

src/licensedcode/detection.py
```diff
@@ -24,34 +27,97 @@
...
+
+# Values of match_coverage less than this are taken as `imperfect-match-coverage` cases
+IMPERFECT_MATCH_COVERAGE_THR = 95
+
+# Values of match_coverage less than this are reported as `license_clues` matches
+CLUES_MATCH_COVERAGE_THR = 60
+
+# How many Lines in between has to be present for two matches being of a different group
+# (i.e. and therefore, different rule)
+LINES_THRESHOLD = 4
```
pombredanne: Should this be synced with https://github.com/nexB/scancode-toolkit/blob/19d77e99c0c37270752849a1f27bb3ec9ad96c72/src/licensedcode/query.py#L196 and should be a single constant for sanity?

src/licensedcode/detection.py
```diff
@@ -24,34 +27,97 @@
...
+
+# Threshold Values of start line and rule length for a match to likely be a false positive
+# (more than the start_line threshold and less than the rule_length threshold)
+FALSE_POSITIVE_START_LINE_THRESHOLD = 1000
```
pombredanne, line 81: This may require a bit more explanation

src/licensedcode/detection.py
```diff
@@ -24,34 +27,97 @@
...
+FALSE_POSITIVE_RULE_LENGTH_THRESHOLD = 3
+
+# Whether to Use the NLP BERT Models
+USE_LICENSE_CASE_BERT_MODEL = False
+USE_FALSE_POSITIVE_BERT_MODEL = False
+
+ISSUE_CASES_VERSION = 0.1
```
pombredanne: What is this?

src/licensedcode/detection.py
```diff
@@ -24,34 +27,97 @@
...
+# (i.e. and therefore, different rule)
+LINES_THRESHOLD = 4
+
+# Threshold Values of start line and rule length for a match to likely be a false positive
+# (more than the start_line threshold and less than the rule_length threshold)
+FALSE_POSITIVE_START_LINE_THRESHOLD = 1000
+FALSE_POSITIVE_RULE_LENGTH_THRESHOLD = 3
+
+# Whether to Use the NLP BERT Models
+USE_LICENSE_CASE_BERT_MODEL = False
```
pombredanne: Could these be instead used if present and contributed by an optional plugin?

src/licensedcode/detection.py
```diff
@@ -63,44 +129,13 @@ class LicenseDetection:
...
 
-    primary_spdx_license_expression = attr.ib(
-        repr=False,
-        default=None,
-        metadata=dict(
-            help='License expression string for this license detection'
-            'using the SPDX license expression syntax and SPDX license ids.')
-    )
-
     spdx_license_expression = attr.ib(
```
pombredanne: This likely should be a computed field.

src/licensedcode/detection.py
```diff
@@ -118,12 +153,90 @@ class LicenseDetection:
         )
     )
 
+    matches = attr.ib(
+        #repr=matches_compact_repr,
```
pombredanne:
```suggestion
```

src/licensedcode/detection.py
```diff
@@ -118,12 +153,90 @@ class LicenseDetection:
...
+        metadata=dict(
+            help='List of license matches combined in this detection.'
+        )
+    )
+
+    @classmethod
+    def from_matches(cls, matches):
+        """
+        Return a LicenseDetection created out of `matches` list of LicenseMatch
+        objects after computing the license_expression by analyzing and combining
```
pombredanne:
```suggestion
```

src/licensedcode/detection.py
```diff
@@ -118,12 +153,90 @@ class LicenseDetection:
...
+        matches for the different cases.
```
pombredanne:
```suggestion
```

src/licensedcode/detection.py
```diff
@@ -118,12 +153,90 @@ class LicenseDetection:
...
+        default=attr.Factory(list),
+        metadata=dict(
+            help='List of license matches combined in this detection.'
+        )
+    )
+
+    @classmethod
+    def from_matches(cls, matches):
+        """
+        Return a LicenseDetection created out of `matches` list of LicenseMatch
```
pombredanne: Less is more here: let's not expose too much of the sausage making in the docstring.
```suggestion
        Return a LicenseDetection created out of `matches` list of LicenseMatch.
```

src/licensedcode/detection.py
```diff
@@ -118,12 +153,90 @@ class LicenseDetection:
...
+        objects after computing the license_expression by analyzing and combining
+        matches for the different cases.
+        """
+        reasons, license_expression = get_detected_license_expression(
+            matches=matches,
+            analysis=analyze_detection(matches),
```
pombredanne: What if the analysis is not conducive to creating a proper detection?

src/licensedcode/detection.py
```diff
@@ -118,12 +153,90 @@ class LicenseDetection:
...
+        )
+
+        if license_expression == None:
```
pombredanne, line 216:
```suggestion
        if license_expression is None:
```

src/licensedcode/detection.py
```diff
@@ -118,12 +153,90 @@ class LicenseDetection:
...
+            return cls(matches=matches)
```
pombredanne: So we end up having a detection that contains no expression, and is therefore inconclusive?

src/licensedcode/detection.py
```diff
@@ -118,12 +153,90 @@ class LicenseDetection:
...
         return (
             isinstance(other, LicenseDetection)
             and self.matches == other.matches
         )
 
+    @property
+    def query(self):
+        # A LicenseDetection will always be created with matches
+        if self.matches:
+            # All the matches in a file or in a LicenseDetection point to the
```
pombredanne: Are you sure this is guaranteed? I would feel more comfy with an `assert` to protect against this and validate we have indeed only one.

src/licensedcode/detection.py
```diff
@@ -118,12 +153,90 @@ class LicenseDetection:
...
     def __eq__(self, other):
         return (
             isinstance(other, LicenseDetection)
             and self.matches == other.matches
         )
 
+    @property
+    def query(self):
+        # A LicenseDetection will always be created with matches
+        if self.matches:
```
pombredanne: Can a detection ever be create from empty matches?

src/licensedcode/detection.py
```diff
@@ -118,12 +153,90 @@ class LicenseDetection:
         )
     )
 
+    matches = attr.ib(
+        #repr=matches_compact_repr,
+        default=attr.Factory(list),
```
pombredanne, line 171: You could add a validator to ensure this is not empty

src/licensedcode/detection.py
```diff
@@ -118,12 +153,90 @@ class LicenseDetection:
...
+
+        spdx_license_expression = build_spdx_license_expression(
+            str(license_expression),
+            licensing=get_cache().licensing
+        )
+
+        return cls(
+            matches=matches,
+            license_expression=str(license_expression),
+            spdx_license_expression=str(spdx_license_expression),
```
pombredanne: IMHO we should remove `spdx_license_expression` and make this a computed  property
This will make the code simpler all around.

src/licensedcode/detection.py
```diff
@@ -118,12 +153,90 @@ class LicenseDetection:
...
+    def query(self):
+        # A LicenseDetection will always be created with matches
+        if self.matches:
+            # All the matches in a file or in a LicenseDetection point to the
+            # same query
+            return self.matches[0].query
+    
+    @property
+    def qspans(self):
+        return [match.qspan for match in self.matches]
```
pombredanne, line 244: Would not you rather have a single merged span here?

src/licensedcode/detection.py
```diff
@@ -118,12 +153,90 @@ class LicenseDetection:
...
+        for license_match in self.original_licenses:
+            identifier = (license_match.rule_identifier, license_match.coverage(),)
+            data.append(identifier)
+
+        return tuple(data)
+    
+    @property
+    def identifier_with_text(self):
+        """
+        This is an identifier for a issue, which is an unknown license intro,
```
pombredanne: `unknown license intro` ? what is this doing here? I am not sure I get why you mention this

src/licensedcode/detection.py
```diff
@@ -118,12 +153,90 @@ class LicenseDetection:
...
+            # All the matches in a file or in a LicenseDetection point to the
+            # same query
+            return self.matches[0].query
+    
+    @property
+    def qspans(self):
+        return [match.qspan for match in self.matches]
+
+    @property
+    def identifier(self):
```
pombredanne: Where are these identifiers used and what is your intention? Is this just to create something unique? or to create some to check that two detections are the same?
If the later, then you should rather have a function that checks equivalence explicitly rather than having indirection with an identifier.

src/licensedcode/detection.py
```diff
@@ -118,12 +153,90 @@ class LicenseDetection:
...
+        """
+        This is an identifier for a license detection, based on it's underlying
+        license matches.
+        """
+        data = []
+        for license_match in self.original_licenses:
+            identifier = (license_match.rule_identifier, license_match.coverage(),)
```
pombredanne: IMHO coverage is not a good elements of identification. Instead the rule_identifier + matched ispan would be best if you are looking for some kind of content-based id... But this would still be missing that some extra unknown or stop words may exist on the qspan-side within and inside the range of ispan... so this needs a bit more detailing.

src/licensedcode/detection.py
```diff
@@ -155,7 +268,7 @@ def relevance(self):
         This is computed as the relevance of the sum of the underlying matches
         rule length.
         """
-        return compute_relevance(self.rules_len)
+        return compute_relevance(self.rules_length)
```
pombredanne: Are you sure? `self.rules_length` is a  function 
```suggestion
        return compute_relevance(self.rules_length())
```

src/licensedcode/detection.py
```diff
@@ -263,6 +370,278 @@ def matched_text(
...
+
+def _licenses_data_from_match(
+    match,
+    include_text=False,
+    license_text_diagnostics=False,
+    license_url_template=SCANCODE_LICENSEDB_URL,
+):
+    """
+    Return a list of "licenses" scan data built from a license match.
+    Used directly only internally for testing.
```
pombredanne: May be this is best in the tests/  then?

src/licensedcode/detection.py
```diff
@@ -263,6 +370,278 @@ def matched_text(
...
+        matched_rule['matched_length'] = match.len()
+        matched_rule['match_coverage'] = match.coverage()
+        matched_rule['rule_relevance'] = match.rule.relevance
+        # FIXME: for sanity this should always be included?????
+        if include_text:
+            result['matched_text'] = matched_text
+    return detected_licenses
+
+
+def matches_compact_repr(matches):
```
pombredanne: This seems not used anywhere

src/licensedcode/detection.py
```diff
@@ -263,6 +370,278 @@ def matched_text(
...
+    """
+    Return a string representing a list of license matches in a compact way.
+    """
+    return ', '.join(m.rule.identifier for m in matches)
+
+
+def is_correct_detection(license_matches):
+    """
+    Return True if all the license matches in a file-region are correct
```
pombredanne: Do not expose too much of the internal implementation in your docstring,
```suggestion
    Return True if all the matches in ` license_matches` are correct e.g., highly accurate and not ambiguous.
```

src/licensedcode/detection.py
```diff
@@ -263,6 +370,278 @@ def matched_text(
...
+    license detections, as they are either SPDX license tags, or the file content has
```
pombredanne:
```suggestion
    license detections, as they are either SPDX license tags, or the file content has
```

src/licensedcode/detection.py
```diff
@@ -263,6 +370,278 @@ def matched_text(
...
+    a exact match with a license hash.
```
pombredanne:
```suggestion
    a exact match with a license hash.
```

src/licensedcode/detection.py
```diff
@@ -263,6 +370,278 @@ def matched_text(
...
+
+    :param license_matches: list
+        List of LicenseMatch.
+    :param threshold: int
+        A `match_coverage` threshold value in between 0-100
+    """
+    coverage_values = (
+        license_match.coverage() for license_match in license_matches
+    )
+    return any(coverage_value < threshold for coverage_value in coverage_values)
```
pombredanne: IMHO you can conflate the two lines without loosing redability
```suggestion
    return any(license_match.coverage() < threshold for for license_match in license_matches)
```

src/licensedcode/detection.py
```diff
@@ -263,6 +370,278 @@ def matched_text(
...
+
+
+def calculate_query_coverage_coefficient(license_match):
+    """
+    Calculates a `query_coverage_coefficient` value for that match. For a match:
+    1. If this value is 0, i.e. `score`==`match_coverage`*`rule_Relevance`, then
+       there are no extra words in that license match.
+    2. If this value is a +ve number, i.e. `score`!=`match_coverage`*`rule_Relevance`,
```
pombredanne: `+ve`?

pombredanne: Also can you give an example?

src/licensedcode/detection.py
```diff
@@ -263,6 +370,278 @@ def matched_text(
...
+    score_coverage_relevance = (
+        license_match.coverage() * license_match.rule.relevance
+    ) / 100
+
+    return score_coverage_relevance - license_match.score()
+
+
+def is_extra_words(license_matches):
+    """
+    Return True if any of the license matches in a file-region has extra words. Having
```
pombredanne: Let's not talk about file region since this is not a concept here in this function
```suggestion
    Return True if any of the license matches has extra words. Having
```

src/licensedcode/detection.py
```diff
@@ -263,6 +370,278 @@ def matched_text(
...
+    :param matched_license: LicenseMatch.
+    """
+    score_coverage_relevance = (
+        license_match.coverage() * license_match.rule.relevance
+    ) / 100
+
+    return score_coverage_relevance - license_match.score()
+
+
+def is_extra_words(license_matches):
```
pombredanne:
```suggestion
def has_extra_words(license_matches):
```

src/licensedcode/detection.py
```diff
@@ -263,6 +370,278 @@ def matched_text(
...
+        List of LicenseMatch.
+    """
+    return any(match.rule.has_unknown for match in license_matches)
+
+
+def is_unknown_intro(license_match):
+    """
+    Return True if all the license matches is an unknown intro LicenseMatch.
+
+    :param license_matches: list
```
pombredanne: I see you started using this style of docstring... here this is incorrect: this is a not  list . Also this feels superfluous in this simple case and a few other cases where the meaning of arguments is obvious.

src/licensedcode/detection.py
```diff
@@ -263,6 +370,278 @@ def matched_text(
...
+        List of LicenseMatch.
+    """
+    return (
+        license_match.rule.has_unknown and
+        license_match.rule.is_license_intro
+    )
+
+
+def is_license_clues(license_matches):
+    """
```
pombredanne, line 607: Is this instead a method of a LicenseDetection?

src/licensedcode/detection.py
```diff
@@ -295,19 +673,154 @@ def is_license_intro(license_match):
...
-    TAG_FOLLOWED_BY_REF = 'tag-followed-by-ref'
-    UNKNOWN_FOLLOWED_BY_MATCH = 'unknown-ref-followed-by-match'
-    UNVERSIONED_FOLLOWED_BY_VERSIONED = 'un-versioned-followed-by-versioned'
+def is_license_reference_local_file(license_match):
+    """
+    Return True if `license_match` LicenseMatch dict has a
+    non-empty `referenced_filename`, i.e. contains a license
+    reference to a local file.
+    """
+    if license_match['matched_rule']['referenced_filenames']:
```
pombredanne:
```suggestion
    return bool(license_match['matched_rule']['referenced_filenames'])
```

src/licensedcode/detection.py
```diff
@@ -295,19 +673,154 @@ def is_license_intro(license_match):
...
+def combine_license_references(license_matches):
+    """
+    Return a filtered ``license_matches`` list of LicenseMatch objects removing
+    references to local files with licenses.
+
+    A common source of unknown license detections in license references to
+    a local file like "see license in LICENSE.md". In these cases we can
+    remove the unknown reference with the license detection in the file.
+    """
+    return [match for match in license_matches if not is_license_reference_local_file(match)]
```
pombredanne: IMHO inlining is good enough for a single use
```suggestion
    return [match for match in license_matches if not license_match['matched_rule']['referenced_filenames']]
```

src/licensedcode/detection.py
```diff
@@ -295,19 +673,154 @@ def is_license_intro(license_match):
...
+    non-empty `referenced_filename`, i.e. contains a license
+    reference to a local file.
+    """
+    if license_match['matched_rule']['referenced_filenames']:
+        return True
+
+    return False
+
+
+def combine_license_references(license_matches):
```
pombredanne: This is not combining, but only filtering

src/licensedcode/detection.py
```diff
@@ -295,19 +673,154 @@ def is_license_intro(license_match):
...
+    """
+    Return a filtered ``license_matches`` list of LicenseMatch objects removing
+    references to local files with licenses.
+
+    A common source of unknown license detections in license references to
+    a local file like "see license in LICENSE.md". In these cases we can
```
pombredanne: But what is "these cases" here? We are saying too much or too litlle. IMHO too much.

src/licensedcode/detection.py
```diff
@@ -295,19 +673,154 @@ def is_license_intro(license_match):
...
+
+    :param license_matches: list
+        List of LicenseMatch.
+    """
+    # Case where all matches have `matcher` as `1-hash` or `4-spdx-id`
+    if is_correct_detection(license_matches):
+        return "correct-license-detection"
+
+    elif has_unknown_intro_before_detection(license_matches):
+        return "unknown-intro-before-detection"
```
pombredanne: You need to create constants for these

src/licensedcode/detection.py
```diff
@@ -295,19 +673,154 @@ def is_license_intro(license_match):
...
+    A common source of unknown license detections in license references to
+    a local file like "see license in LICENSE.md". In these cases we can
+    remove the unknown reference with the license detection in the file.
+    """
+    return [match for match in license_matches if not is_license_reference_local_file(match)]
+
+
+def get_detected_license_expression(matches, analysis):
+    """
+    Get `reasons` and `combined_license_expression` after combining
```
pombredanne:
```suggestion
    Return a tuple of (reasons, combined_expression) by combining a `matches` list of LicenseMatch objects using an `analysis` code string.
```

src/licensedcode/detection.py
```diff
@@ -295,19 +673,154 @@ def is_license_intro(license_match):
...
+    `matches` list of LicenseMatch objects.
+    """
+    matches_for_expression = None
+    reasons = []
+
+    if analysis == "unknown-intro-before-detection":
```
pombredanne: Use constants

src/licensedcode/detection.py
```diff
@@ -295,19 +673,154 @@ def is_license_intro(license_match):
...
+        reasons.append(CombinationReason.UNKNOWN_INTRO_FOLLOWED_BY_MATCH.value)
+    elif analysis == "unknown-file-reference-local":
+        matches_for_expression = combine_license_references(matches)
+        reasons.append(CombinationReason.UNKNOWN_REFERENCE_TO_LOCAL_FILE.value)
+    else:
+        matches_for_expression = matches
+
+    if isinstance(matches[0], dict):
+        combined_expression = combine_expressions(
+            expressions=[match['matched_rule']['license_expression'] for match in matches_for_expression]
```
pombredanne: Should you rather use deserialized real LicenseMatch objects throughout and rules recreated from the scan data?
This would make the whole code cleaner rather than using dicts and key lookups.

src/licensedcode/detection.py
```diff
@@ -295,19 +673,154 @@ def is_license_intro(license_match):
...
+
+
+def get_matches_from_detections(license_detections):
+    """
+    Return a `license_matches` list of LicenseMatch dicts from a
+    `license_detections` list of LicenseDetection dicts.
+    """
+    license_matches = []
+    for detection in license_detections:
+        license_matches.extend(detection["matches"])
```
pombredanne, line 952: matches should be an attribute and we should not carry around a bunch of dicts IMHO

pombredanne (changes requested):
Thanks! that's quite a piece of work! :bow:  I provided some comments, but I did not review everything just yet. Of note you should use LicenseMatch and objects everywhere and not carry around dictionaries of scan data.
Instead convert these at the boundaries as early as possible deserialized to objects.

tests/licensedcode/data/plugin_license/license-expression/scan.expected.json
```diff
@@ -1,78 +1,62 @@
...
-        "spdx_license_list_version": "3.16",
-        "files_count": 2
-      }
-    }
-  ],
   "files": [
     {
       "path": "apache-1.0.txt",
       "type": "file",
       "licenses": [
```
pombredanne: IMHO rename this to `license_detections` for sanity and avoid too much user confusion.

tests/licensedcode/data/plugin_license/license-expression/scan.expected.json
```diff
@@ -1,78 +1,62 @@
...
-            "is_license_intro": false,
-            "has_unknown": false,
-            "matcher": "2-aho",
-            "rule_length": 368,
-            "matched_length": 368,
-            "match_coverage": 100.0,
-            "rule_relevance": 100
-          }
+          "license_expression": "apache-1.0",
+          "spdx_license_expression": "None",
```
pombredanne: Never None for sure.

tests/licensedcode/data/plugin_license/license-expression/scan.expected.json
```diff
@@ -1,78 +1,62 @@
...
+          "combination_reasons": [],
```
pombredanne: I have no idea what this could mean. Try to find a better more obvious name for this attribute. Also why is this empty?

tests/licensedcode/data/plugin_license/license-expression/scan.expected.json
```diff
@@ -1,78 +1,62 @@
...
+          "matches": [
+            {
+              "key": "apache-1.0",
```
pombredanne: A match is surely not for a key but for an expression (that's a problem in the current licenses output) also we do not want to repeat forever details of licenses which should be pushed out in the --license-reference . Also since a match is for a single rule, this would be best to streamline this rather than having a nested matched_rule. Along these lines, reporting matched_rule reference details only once could help too may be with a new --license-rule-reference that would report the details once

tests/licensedcode/data/plugin_license/license-expression/scan.expected.json
```diff
@@ -1,78 +1,62 @@
...
+                "matched_length": 368,
+                "match_coverage": 100.0,
+                "rule_relevance": 100
+              }
+            }
+          ]
         }
       ],
+      "license_clues": [],
       "license_expressions": [
```
pombredanne: We cannot double dip: if you have detections, this must go.

tests/licensedcode/data/plugin_license/license-expression/scan.expected.json
```diff
@@ -1,78 +1,62 @@
...
         "apache-1.0"
       ],
+      "spdx_license_expressions": [
```
pombredanne: Same as before: this likely needs to go away,  and never  "None"

pombredanne (changes requested):
Here is some feedback on the format. We need to streamline this to the essence and avoid repeating details to remove noise when possible.

src/licensedcode/detection.py
AyanSinhaMahapatra: This would make the code at process_codebase much more complicated tho, as we use have to serialize and de-serialize all the detections (and matches in them?)

src/licensedcode/detection.py
AyanSinhaMahapatra: Yes, modified.

src/licensedcode/detection.py
AyanSinhaMahapatra: Adding.

src/licensedcode/detection.py
AyanSinhaMahapatra: Creating Enum like the combination reasons.

src/licensedcode/detection.py
AyanSinhaMahapatra: Removing.

src/licensedcode/detection.py
AyanSinhaMahapatra: Changing this to filtering.

src/licensedcode/detection.py
AyanSinhaMahapatra: We are having all these functions in case they are reused?

src/licensedcode/detection.py
AyanSinhaMahapatra: Yes.

src/licensedcode/detection.py
AyanSinhaMahapatra: Yes, removing this style and updating the docstrings to reflect the updates.

src/licensedcode/detection.py
AyanSinhaMahapatra: Modifying.

src/licensedcode/detection.py
AyanSinhaMahapatra: Yes.

src/licensedcode/detection.py
AyanSinhaMahapatra: Yes this was present earlier in the code. Removing this.

src/licensedcode/detection.py
AyanSinhaMahapatra: This was being used in the serialization actually earlier too, still being used. Will update  the docstring.

src/licensedcode/detection.py
AyanSinhaMahapatra: This won't happen, so I added another check on creation?

src/licensedcode/detection.py
AyanSinhaMahapatra: No, this will never happen.

src/licensedcode/detection.py
AyanSinhaMahapatra: Yes, putting an assert.

src/licensedcode/detection.py
AyanSinhaMahapatra: Yes, this is added to `license_clues` as discussed above.

src/licensedcode/detection.py
AyanSinhaMahapatra: Added.

src/licensedcode/detection.py
AyanSinhaMahapatra: The license-expression is `None` and matches for this part is added to the `license_clues` instead of `license_detections`

src/licensedcode/detection.py
AyanSinhaMahapatra: Added.

src/licensedcode/detection.py
AyanSinhaMahapatra: Yes.

src/licensedcode/detection.py
AyanSinhaMahapatra: Modified.

src/licensedcode/detection.py
AyanSinhaMahapatra: Removed.

src/licensedcode/detection.py
AyanSinhaMahapatra: This is computed from the `license_expression`, but as this is also returned as an attribute in the LicenseDetection level we need to keep it here?

src/licensedcode/detection.py
AyanSinhaMahapatra: Removing for now, this shouldn't have been brought in.

src/licensedcode/detection.py
AyanSinhaMahapatra: This should not have been moved in from the Analyzer, removing.

src/licensedcode/detection.py
AyanSinhaMahapatra: Added.

src/licensedcode/detection.py
AyanSinhaMahapatra: Yes, creating the constant there and importing instead.

src/licensedcode/detection.py
AyanSinhaMahapatra: Adding more explanation docs.

src/scancode/api.py
AyanSinhaMahapatra: Moving these back into the API and importing from there.

src/licensedcode/cache.py
AyanSinhaMahapatra: Yes, moving this.

src/licensedcode/cache.py
AyanSinhaMahapatra: Yes, moving this.

src/licensedcode/detection.py
AyanSinhaMahapatra: Yes, thanks. Updating this.

src/licensedcode/detection.py
AyanSinhaMahapatra: Actually we would then again merge across all detections and then compute the percentage license text. Isn't it better do this once?

src/licensedcode/detection.py
AyanSinhaMahapatra: Should not be mentioned. Basically in cases of small unknown matches, the matched text is required to create a unique identifier, otherwise the rule identifier and the coverage is enough in most cases. But as we discussed we need something absolute so this will change anyway.

src/licensedcode/detection.py
AyanSinhaMahapatra: Same as https://github.com/nexB/scancode-toolkit/pull/2961#discussion_r877480348

src/licensedcode/detection.py
AyanSinhaMahapatra: Modified.

src/licensedcode/detection.py
AyanSinhaMahapatra: Updated. Thanks!

src/licensedcode/detection.py
AyanSinhaMahapatra: Updated and see https://github.com/nexB/scancode-analyzer/blob/main/tests/test_license_analyzer.py#L124 and https://github.com/nexB/scancode-analyzer/blob/main/tests/data/analyzer/analyzer_is_extra_words_true_one.json for an example.

src/licensedcode/detection.py
AyanSinhaMahapatra: > create something unique

This yes. As the code for finding unique part doesn't check between two detections, rather gets unique ones from a list of identifiers.

tests/licensedcode/data/plugin_license/license-expression/scan.expected.json
AyanSinhaMahapatra: This is never None, that part is fixed. 

Do you mean we should only have `spdx_license_expressions` at resource level and not have individual `spdx_license_expression` for each detection? That could make sense yes.

tests/licensedcode/data/plugin_license/license-expression/scan.expected.json
AyanSinhaMahapatra: Not sure what this means, `license_detections` and `license_clues` are two seperate lists that are added btw. One is a list of LicenseDetection and another is a list of LicenseMatch for which we couldn't create detections.

tests/licensedcode/data/plugin_license/license-expression/scan.expected.json
AyanSinhaMahapatra: Yes, this is now pushed, though it is WIP.

tests/licensedcode/data/plugin_license/license-expression/scan.expected.json
AyanSinhaMahapatra: Should this be populated even if we don't do any combination in the LicenseDetection level?

Not keeping this empty would mean adding `not-combined` at all the `combination_reasons` which didn't have one. 

Yes, this could be named better.

tests/licensedcode/data/plugin_license/license-expression/scan.expected.json
AyanSinhaMahapatra: This has been fixed.

tests/licensedcode/data/plugin_license/license-expression/scan.expected.json
AyanSinhaMahapatra: This makes sense. Will update this.

AyanSinhaMahapatra:
@pombredanne @JonoYang This is ready to review!

JonoYang:
@AyanSinhaMahapatra I'll need a little more time to review this, but I think it's looking good so far. You have some failing CI tests, maybe some other test expectations need to be updated?

src/licensedcode/detection.py
```diff
@@ -400,60 +400,67 @@ def licenses_data_from_match(
     SCANCODE_LICENSE_TEXT_URL = SCANCODE_BASE_URL + '/{}.LICENSE'
     SCANCODE_LICENSE_DATA_URL = SCANCODE_BASE_URL + '/{}.yml'
 
-    detected_licenses = []
+    result = {}
```
pombredanne: The docstring is now incorrect.
Also I wonder if using an object rather than a dict would be cleaner?

src/licensedcode/detection.py
```diff
@@ -95,6 +96,18 @@ class CombinationReason(Enum):
...
 
 
+@attr.s
+class FileRegion:
+    """
+    A file has one or more file-regions, which are separate regions of the file
+    containing some license information (separated by code/text/others in between),
+    and identified by a start line and an end line.
+    """
+    path = attr.ib(type=str)
```
pombredanne, line 140: Is this a relative path or an absolute location?

src/licensedcode/detection.py
```diff
@@ -162,6 +183,19 @@ def from_matches(cls, matches):
             spdx_license_expression=str(spdx_license_expression),
             combination_reasons=reasons,
         )
+    
+    @classmethod
+    def from_mapping(cls, detection):
```
pombredanne: How is this different from calling `LicenseDetection(**detection)` ?

src/licensedcode/detection.py
```diff
@@ -181,36 +215,65 @@ def query(self):
...
+        )
+
     @property
     def identifier(self):
         """
         This is an identifier for a license detection, based on it's underlying
         license matches.
+
+        This is not guaranteed to be unique for a detection, as certain small and
+        unknown matches could have the same value for this identifier, but will
```
pombredanne: You should also use the ispan Span on the matched rule side. Together with rule_identifier this is guaranteed to uniquely identify a match in abstraction of where it was matched. And the coverage may not be as useful.... rather the score would be better as this would take into account unknown words that may exist in the matched range. So I would rather use something such as: `((license_match.rule_identifier, license_match.ispan, license_match.score(),)`

src/licensedcode/detection.py
```diff
@@ -373,6 +436,7 @@ def to_dict(
 
         detection = attr.asdict(self)
         detection["matches"] = data_matches
+        _file_region = detection.pop('file_region')
```
pombredanne: attr.asdict can filter attributes directly.... IMHO no need to pop. 

Also:
- why the line 424 above with ` matches = self.matches` ... I am not sure this is needed.
- could the function ` licenses_data_from_match(` better be a LicenseMatch method?

src/licensedcode/plugin_licenses_reference.py
```diff
@@ -55,30 +63,187 @@ def is_enabled(self, licenses_reference, **kwargs):
...
-
-        db = get_licenses_db()
-        for key in sorted(license_keys):
-            license_details = db[key].to_dict(
-                include_ignorables=False,
-                include_text=True,
+
+            # Get license_expressions from both package and license detections
+            license_licexps = getattr(resource, 'license_expressions', []) or []
+            package_data = getattr(resource, 'package_data', []) or []
```
pombredanne, line 88: You are missing out the top level packages IMHO

src/licensedcode/plugin_licenses_reference.py
```diff
@@ -55,30 +63,187 @@ def is_enabled(self, licenses_reference, **kwargs):
         return licenses_reference
 
     def process_codebase(self, codebase, licenses_reference, **kwargs):
-        from licensedcode.cache import get_licenses_db
-        licensing = Licensing()
-
-        license_keys = set()
+        """
+        Get Licenses and LicenseDB data from all license detections in a codebase level list
```
pombredanne: I am confused about this method and it feels like you may be trying to do too much at once? Also you may be collecting data multiple time, once for each duplicate?

src/licensedcode/plugin_licenses_reference.py
```diff
@@ -55,30 +63,187 @@ def is_enabled(self, licenses_reference, **kwargs):
...
+        license_references.append(
+            db[key].to_dict(include_ignorables=False, include_text=True)
+        )
+
+    return license_references
+
+
+def get_licensedb_references(license_db_data):
+    """
+    """
```
pombredanne, line 150: Either put a docstring or not, but not an empty one

src/licensedcode/detection.py
```diff
@@ -181,36 +215,65 @@ def query(self):
...
+
+        This is not guaranteed to be unique for a detection, as certain small and
+        unknown matches could have the same value for this identifier, but will
+        be unique for most other cases.
         """
         data = []
-        for license_match in self.original_licenses:
-            identifier = (license_match.rule_identifier, license_match.coverage(),)
+        for match in self.matches:
+            identifier = (match['licensedb_identifier'], match['match_coverage'],)
```
pombredanne: What is this new licensedb_identifier?

pombredanne (changes requested):
Thanks! Here is a first batch of review comments.

pombredanne (commented):
Note I reviewed commit-by-commit starting with the oldest and that's unlikely the better approach. I am restarting reviewing looking at the final results instead

src/licensedcode/models.py
```diff
@@ -571,7 +571,7 @@ def validate(licenses, verbose=False, no_dupe_urls=False):
                 error(f'Unknown language: {lic.language}')
 
             if lic.is_unknown:
-                if not 'unknown' in lic.key:
+                if not 'unknown' in lic.key and not 'undetected' in lic.key:
```
pombredanne: IMHO we should introduce soon after this a is_something flag for these licenses rather than to special case their name.

src/licensedcode/plugin_license.py
```diff
@@ -11,37 +11,49 @@
...
 from commoncode.cliutils import SCAN_GROUP
+from commoncode.fileutils import file_name
+
+from licensedcode.cache import get_cache
+from licensedcode.cache import build_spdx_license_expression
+from licensedcode.detection import SCANCODE_LICENSEDB_URL
+from licensedcode.detection import get_detected_license_expression
+from licensedcode.detection import get_matches_from_detections
+from licensedcode.detection import DetectionCategory
 from commoncode.resource import clean_path
```
pombredanne: This likely would benefit from sorting

src/licensedcode/plugin_licenses_reference.py
```diff
@@ -39,7 +43,10 @@ class LicensesReference(PostScanPlugin):
     """
     Add a reference list of all licenses data and text.
     """
-    codebase_attributes = dict(licenses_reference=attr.ib(default=attr.Factory(list)))
+    codebase_attributes = dict(
+        license_references=attr.ib(default=attr.Factory(list)),
+        licensedb_references=attr.ib(default=attr.Factory(list))
```
pombredanne: I would like to discuss this licensedb_references further ... I am not sure I get it entirely.

pombredanne (changes requested):

src/packagedcode/models.py
```diff
@@ -1130,6 +1136,33 @@ def create_default_package_data(cls, **kwargs):
...
         )
 
+    @classmethod
+    def populate_license_fields(cls, package_data):
+        """
+        Run license detection on extracted_license_statement for a PackageData
+        object, and add the declared_license_expression (and the spdx expression)
+        and corresponding LicenseDetection data.
+        """
+        if not package_data.declared_license_expression and package_data.extracted_license_statement:
```
pombredanne, line 1246: What about inverting and splitting the ifs for readability?
```suggestion
        if package_data.declared_license_expression:
            return
        if not package_data.extracted_license_statement:
            return
```

pombredanne (changes requested):
Here is another comment. I think we are also missing a CHANGELOG entry (for the next release)

src/licensedcode/detection.py
AyanSinhaMahapatra: Yes, making this LicenseMatch method instead. Also using filters now, thanks this is much better.

src/licensedcode/detection.py
AyanSinhaMahapatra: Yes, removed this.

src/licensedcode/detection.py
AyanSinhaMahapatra: This is the same as `resource.path` so it depends on CLI options full/strip root. Not the absolute location, as that is `resource.location`.

src/licensedcode/plugin_licenses_reference.py
AyanSinhaMahapatra: Added this and others.

src/licensedcode/detection.py
AyanSinhaMahapatra: This has been reverted back to the original `rule_identifier` as discussed.

src/licensedcode/plugin_licenses_reference.py
AyanSinhaMahapatra: Here there are two levels of de-referencing. One is licenses with a unique license-key, the another is a matched rule/license with a unique rule identifier.

src/packagedcode/models.py
AyanSinhaMahapatra: Yes, added this.

src/licensedcode/plugin_licenses_reference.py
AyanSinhaMahapatra: We do this too now, and take a set out of the list of license_expression so we only have the unique ones.

src/licensedcode/plugin_licenses_reference.py
AyanSinhaMahapatra: We have to go at every occurrence and remove those as we are de referencing these, also we are taking a set at last so we don't have any duplicates. As we are taking expressions from multiple places, we are performing the set operation at last, and there won't be any duplicates at last.

src/licensedcode/detection.py
AyanSinhaMahapatra: Here, we don't have the objects as this is after the `process_codebase` step and we have only mappings. We have to then recreate objects here from the mappings. Will go ahead with this.

src/licensedcode/detection.py
AyanSinhaMahapatra: Updated the docstring. As we'll be converting to mappings anyway and there are a combination of calling functions and attributes and some other processing I'm not sure if we should use an object, what do you suggest here? Looking into this more.

src/licensedcode/plugin_license.py
AyanSinhaMahapatra: Done.

src/licensedcode/models.py
AyanSinhaMahapatra: We didn't add this after all, so resolving this.

AyanSinhaMahapatra:
@pombredanne All green!

docs/source/explanations/license-detection-reference.rst
```diff
@@ -0,0 +1,194 @@
...
+  references.
+
+
+LicenseMatch Result Data
+------------------------
+
+LicenseMatch data currently is based on a ``license key`` instead of being based
+on an ``license-expression``.
+
+So if there is a ``mit and apache-2.0`` license expression detected from a single
```
bwjohnson-ss, line 292: Is case important with license expression operands? Does `AND` usually get capitalized?

AyanSinhaMahapatra: Usually no, we use upper case actually in our output license expressions but https://github.com/nexB/license-expression/ can parse these expression irrespective of upper/lower case.

AyanSinhaMahapatra:
@pombredanne this is all ready!

CHANGELOG.rst
```diff
@@ -2,9 +2,67 @@ Changelog
...
+----------------------------
+
+Important API changes:
+~~~~~~~~~~~~~~~~~~~~~~
+
+- The data structure of the JSON output has changed for licenses at resource level:
+
+  - We now return ``license_detections`` information at the manifest file-level
+  rather than ``licenses``. This three data attributes: ``license_expression``,
+  ``detection_log`` and ``matches``. Here ``matches`` is similar to previous
```
pombredanne:
```suggestion
  ``detection_log`` and ``matches``. Here the ``matches`` attribute is similar to previous
```

CHANGELOG.rst
```diff
@@ -2,9 +2,67 @@ Changelog
...
+  ``licenses`` with some additional changes.
+
+  - A new attribute ``license_clues`` was added, which has license matches same
+    as the ``matches`` field in ``license_detections``. This has license matches
+    which are mere clues and not proper detections.
```
pombredanne:
```suggestion
    that are mere license detection clues and not proper, conclusive license detections.
```

pombredanne (approved):

Issue #2967, opened 2022-05-19
Rename precise_license_detection field to identification_precision in license clarity scoring plugin


JonoYang:
The `precise_license_detection` field from the license clarity scoring plugin should be called `identification_precision`.

JonoYang:
This has been merged into develop.

PR #2968, opened 2022-05-19
Rename precise_license_detection field #2967

JonoYang:
This PR renames the field `precise_license_detection` from the license scoring plugin to `identification_precision`, which is the original intended name for the field.


AyanSinhaMahapatra (approved):
Looks great!

But since we are changing the output format, we need to update it according to this https://scancode-toolkit.readthedocs.io/en/latest/misc/versioning.html btw, the format is [here](https://github.com/nexB/scancode-toolkit/blob/develop/src/scancode_config.py#L87) and add a changelog :P

JonoYang:
@AyanSinhaMahapatra Thanks! I knew I was missing something

JonoYang:
@AyanSinhaMahapatra @pombredanne 

I've updated the changelog with a list of the the new fields that we are using to replace the old field with in the license clarity score and summarizer plugins. Should I also bump the json output version here as well, or are we going to do that when we finalize the 31.0.0 release?

pombredanne (approved):
LGTM! Thanks!

Issue #2971, opened 2022-05-19
Cannot detect system installed packages on distroless rootfs using --system-package option


JonoYang:
I downloaded a distroless docker image from `docker://gcr.io/distroless/base`, extracted the layers, and run scancode on them using the `--system-package` option. However, scancode did not return any system installed packages.

PR #2973, opened 2022-05-20
Convert package data dict to PackageData #2971

JonoYang:
This PR addresses the issue where scancode encountered errors when scanning for system packages from a distroless docker root filesystem. The issue was that the package data on the resource, which is in a dictionary, was being passed into `Package.update()` when `Package.update()` was expecting a `PackageData` object instead.

### DIFF
--- a/src/packagedcode/debian.py
+++ b/src/packagedcode/debian.py
@@ -402,10 +402,11 @@ class DebianDistrolessInstalledDatabaseHandler(models.DatafileHandler):
                 continue
 
             for pkgdt in res.package_data:
+                pkgdt = models.PackageData.from_dict(pkgdt)
                 package.update(
                     package_data=pkgdt,
                     datafile_path=res.path,
-            )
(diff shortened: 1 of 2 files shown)


src/packagedcode/debian.py
```diff
@@ -402,10 +402,11 @@ def assemble(cls, package_data, resource, codebase):
                 continue
 
             for pkgdt in res.package_data:
+                pkgdt = models.PackageData.from_dict(pkgdt)
```
pombredanne: I wonder if this could be done directly in package.update where we could accept `package_data` as a mapping or as an object and convert to object if needed?

JonoYang: Good point, I've updated `Package.update()` to convert the package data to a `PackageData` object if the incoming package data is in a dictionary.

pombredanne (approved):
LGTM!
Thank :)

Issue #2972, opened 2022-05-19
Consider detected copyrights when determining a declared holder from a package manifest in summary plugin


JonoYang:
When scanning the package `atheris v 2.0.11` (https://github.com/google/atheris/archive/refs/tags/2.0.11.tar.gz) using the `--summary` plugin, the `declared_holder` value in the scan summary is `Bitshift`, which is the author of the package. This was determined from the parsed package data from the `setup.py` file of `atheris`. However, the `setup.py` contains a comment that is a copyright statement with the actual copyright holders. The summary plugin should be updated to also consider copyrights detected by the copyright scanner. This value should take precedence over authors.

JonoYang:
It also may not behoove us to use the package authors as a copyright holder when we do not detect an explicit copyright statement from package data.

JonoYang:
@DennisClark @tdruez @pombredanne 

When removing the code that assigns the author or other detected parties from a Package as the declared holder, I noticed that the tallies plugins does some sort of normalization on the detected holders from Resources in the codebase. The majority of the files have `Google LLC` as the copyright holder, but looking at the summary, only `Google, Inc.` shows up as the declared holder. This is done so we are able to group different forms of the same holders together. For example, from https://github.com/nexB/scancode-toolkit/blob/2972-summary-consider-copyrights/src/summarycode/copyright_tallies.py#L487, we normalize `google`, `google llc`, and `google inc` as `Google, Inc.`.  

Should we remove this normalization of holders to a canonical form? Normalizing and grouping the related holders together helps with getting a good count of how many times a particular holder shows up, especially when there are many different forms of copyright statements for that holder. However, it can become confusing when someone wants to verify the summary results and they cannot find the declared holder in files because the detected holder value was changed.

DennisClark:
@JonoYang   I am not convinced that using an author value for Holder when there is no copyright detected is a good thing, although I don't feel strongly about it. However, I vaguely recall some community discussion on this topic, where someone strongly asserted that author is NOT equivalent to copyright, so there is definitely a case for not using it at all for a Holder. 

As far as "normalizing" the holder goes, it is a nice feature if we can still point back to the original somehow.

JonoYang:
@DennisClark 

[...]

PR #2974, opened 2022-05-21
Consider only copyrights in summry #2972

JonoYang:
I removed the logic that uses author or maintainers as the declared holder if no explicit copyright was detected from package data. However, I am still not sure about how I feel the `tally` functions normalize copyright holders to a different value for the sake of grouping as it can be misleading. 

Reference: https://github.com/nexB/scancode-toolkit/issues/2972


pombredanne (approved):
LGTM!
Thanks

PR #2979, opened 2022-06-01
Add support for external licenses in scans

### LINKED ISSUE #480: Add support for "extra", e.g. private or local licenses
At times, someone may have some special notices that may not be worthy of landing in the main scancode licenses dataset. These could be your own proprietary notices or some composite license that you like to have named as such or it could be some advanced fixes that you maintain locally and you do not want yet to upgrade to a newer develop version of scancode. Regardless of the why this is needed, it would be great to have an option to reference an extra, local directory of license texts and rules that would be included in the license detection perimeter.

This could simply take the form of a new command line option such as `--extra-licenses` or `--additional-licenses` that would point to a directory with the same structure as the license data set e.g with a `licenses` and a `rules` subdirectory and when using this option, these extra licenses would validated and loaded in the index the same way that the standard ones are.

Some internal caveats to consider is the license index creation and  cache invalidation: 

should there be a separate index built or not? and if so then a separate cache might be needed. 
If a single index is used, then the cache (and its cache hash) should be built considering these extra licenses files (in addition to the actual scancode source code files and its license and rule files) which means that if you run scancode once without extra, an index cache is built. 

If you run with --extra that cache would be invalidated and a new cache built considering the extra. And if you run scancode again without extra or with a different extra, the cache would be again invalidated and rebuilt. This is not a major issue but may come with small perf penalty when switching usage between with or without extra or using different extra.

KevinJi22:
This is for the 2022 Google Summer of Code. This PR contains all the work that needs to be done for the project.

This adds `--additional-license-directory` as a command line option in license reindexing. This allows users to specify paths to directories of licenses and rules they'd like to use during license detection, but would not like to add to the ScanCode database of licenses. First, the user must run Scancode with the `--reindex-licenses` option and they can use `--additional-license-directory` to add directories of licenses or rules to the index.

This involves adding a new option in `licensedcode/plugin_license.py`, and this option is used as a parameter in `scancode/api.py`. In this approach, the licenses and rules contained in these additional directories are combined with the existing licenses and rules in the ScanCode database to produce a single index. The code for this is found in `licensedcode/cache.py` and the helper methods for loading these licenses and rules are found in `licensedcode/models.py`. 

This commit also includes unit tests to verify that license detection succeeds with additional directories and installed licenses/rules. Part of the setup for the unit test and future tests involves creating a new directory in `tests/licensedcode/data` that contains sample external licenses used in the unit tests.

Fixes #480

### Tasks


src/licensedcode/models.py
```diff
@@ -768,6 +769,47 @@ def get_rules(
     licenses_as_rules = build_rules_from_licenses(licenses_db)
     return chain(licenses_as_rules, rules)
 
+def get_license_dirs(
+    additional_dirs,
+):
+    return [f"{path}/licenses" for path in additional_dirs]
```
JonoYang: Put a docstring here explaining the function.

src/licensedcode/models.py
```diff
@@ -768,6 +769,47 @@ def get_rules(
...
+
+def get_rule_dirs(
+    additional_dirs,
+):
+    return [f"{path}/rules" for path in additional_dirs]
```
JonoYang: Like above, this function should have a docstring

src/licensedcode/models.py
```diff
@@ -768,6 +769,47 @@ def get_rules(
...
+
+def load_licenses_from_multiple_dirs(
+    license_directories,
+    with_deprecated=False,
+):
+    combined_licenses = {}
```
JonoYang: Add docstring

src/licensedcode/models.py
```diff
@@ -768,6 +769,47 @@ def get_rules(
...
+        licenses = load_licenses(licenses_data_dir=license_dir, with_deprecated=False)
+        # this syntax for merging is described here: https://stackoverflow.com/a/26853961
+        combined_licenses = {**combined_licenses, **licenses}
+    return combined_licenses
+
+def get_rules_from_multiple_dirs(
+    licenses_db,
+    rule_directories,
+):
+    # TODO: error handling in case additional_directories is empty
```
JonoYang: Add docstring

tests/licensedcode/test_plugin_license.py
```diff
@@ -242,6 +242,24 @@ def test_reindex_licenses_works():
     run_scan_click(['--reindex-licenses'])
 
 
+def test_detection_with_external_license_directories():
+    test_dir = test_env.get_test_loc('plugin_license/external_licenses/scan', copy=True)
+    external_license_dir = f"{os.path.join(os.path.dirname(__file__), 'data')}/example_external_licenses/"
```
JonoYang: You can use `test_env.get_testloc()` to get your test directory location since we already set up that stuff on line 22.
```suggestion
    external_license_dir = test_env.get_test_loc('example_external_licenses/example1')
```

src/licensedcode/plugin_license.py
```diff
@@ -167,7 +176,7 @@ def setup(self, **kwargs):
         loaded index.
         """
         from licensedcode.cache import populate_cache
-        populate_cache()
+        populate_cache(additional_directories=kwargs.get('additional_directories'))
```
JonoYang: I think its cleaner to have the `kwargs.get()` call set as a variable first
```suggestion
        additional_directories = kwargs.get('additional_directories')
        populate_cache(additional_directories=additional_directories)
```

src/licensedcode/models.py
```diff
@@ -768,6 +769,47 @@ def get_rules(
...
+    # TODO: comment
+    if rule_directories:
+        rules_list = []
+        for rules_dir in rule_directories:
+            rules_list.append(list(load_rules(
```
JonoYang: I think it's better to separate some of this out to their own variables for readability. 
```suggestion
            r = list(load_rules(rules_data_dir=rules_dir)
            combined_rules.append(r)
```

src/licensedcode/models.py
```diff
@@ -768,6 +769,47 @@ def get_rules(
...
+    return combined_licenses
+
+def get_rules_from_multiple_dirs(
+    licenses_db,
+    rule_directories,
+):
+    # TODO: error handling in case additional_directories is empty
+    # TODO: comment
+    if rule_directories:
+        rules_list = []
```
JonoYang: Maybe call this `combined_rules`. We generally want to avoid using the data type in the variable name.

JonoYang (commented):
@KevinJi22 I think this is looking good so far! I left some notes about docstrings and style.

pombredanne:
There is something in your changes that make the tests fail...
I do not think you can just use multiple cached directories. You should use a single cache and then validate, then combine multiple input a single cached index.

src/licensedcode/cache.py
```diff
@@ -151,19 +164,35 @@ def build_index(
...
+    from licensedcode.models import load_licenses_from_multiple_dirs
     from licensedcode.legalese import common_license_words
 
     licenses_data_dir = licenses_data_dir or ldd
     rules_data_dir = rules_data_dir or rdd
 
-    licenses_db = licenses_db or load_licenses(licenses_data_dir=licenses_data_dir)
-    rules = get_rules(licenses_db=licenses_db, rules_data_dir=rules_data_dir)
+    if not licenses_db:
+        if additional_directories:
```
KevinJi22, line 183: @pombredanne I'm actually combining all the licenses and rules in the input directories into a single index here. In `load_licenses_from_multiple_dirs`, all the licenses (including the ones in the ScanCode database) are combined into a single license database, and the same is done with all the rules. I'll update this code with a comment so that this is more clear. # # # #

KevinJi22:
@pombredanne I'm actually combining all the licenses and rules in the input directories into a single index here. In load_licenses_from_multiple_dirs, all the licenses (including the ones in the ScanCode database) are combined into a single license database, and the same is done with all the rules. The _CACHED_DIRECTORIES variable is used to determine whether a new index needs to be built. For example, if we've created an index with additional directories A and B, but now we run a scan with additional directories B and C, we'd need to recreate the combined index.

I'll update this code with comments so that this is more clear.



https://github.com/nexB/scancode-toolkit/blob/b1cf4aa16386e447b024b2fd3f5386251560628d/src/licensedcode/cache.py#L182-L195

src/licensedcode/models.py
```diff
@@ -768,6 +768,63 @@ def get_rules(
     licenses_as_rules = build_rules_from_licenses(licenses_db)
     return chain(licenses_as_rules, rules)
```
JonoYang, line 791: Nitpick: include another new line here. There should be two new lines between functions, as per the PEP 8 style guide.

tests/licensedcode/test_plugin_license.py
```diff
@@ -242,6 +242,41 @@ def test_reindex_licenses_works():
...
+        '--verbose',
+        '-dir', example1_dir,
+        '-dir', example2_dir,
+        '--json', result_file,
+        test_dir,
+    ]
+    run_scan_click(args)
+    test_loc = test_env.get_test_loc('plugin_license/external_licenses/scan_multiple.expected.json')
+    check_json_scan(test_loc, result_file, regen=REGEN_TEST_FIXTURES)
+
```
JonoYang, line 244: Add another new line here.

tests/licensedcode/test_plugin_license.py
```diff
@@ -242,6 +242,41 @@ def test_reindex_licenses_works():
     run_scan_click(['--reindex-licenses'])
```
JonoYang: I would consider adding the `@pytest.mark.scanslow` tag here since this test takes a relatively long time to run due to the license cache being built.

tests/licensedcode/test_plugin_license.py
```diff
@@ -242,6 +242,41 @@ def test_reindex_licenses_works():
...
+        '--verbose',
+        '-dir', example1_dir,
+        '--json', result_file,
+        test_dir,
+    ]
+    run_scan_click(args)
+    test_loc = test_env.get_test_loc('plugin_license/external_licenses/scan.expected.json')
+    check_json_scan(test_loc, result_file, regen=REGEN_TEST_FIXTURES)
+
+
```
JonoYang: Like above, consider adding the `@pytest.mark.scanslow` tag.

JonoYang (commented):
I've added a few nitpicks about formatting. I have an issue trying out the new `-dir` option. I am running `scancode -l -dir tests/licensedcode/data/example_external_licenses/example1 --json-pp ~/Desktop/out.json /home/jono/src/scancode-toolkit-kevinji22/tests/licensedcode/data/plugin_license/external_licenses/scan/`, however I am not seeing the custom license being detected in the scan results.

KevinJi22:
Thanks for the feedback! This bug was actually due to how `os.path.join()` works, since the path that you passed in wasn't an absolute path, so the paths were just concatenated in the function `build_rule_from_license()` in models.py. This meant that the external license wasn't being loaded in as a rule. I added some code in `get_license_dirs()` that converts the input path to an absolute path, which will hopefully prevent this from happening in the future.

This means that running the scan with the same parameters should work this time.

JonoYang (commented):
Looks good! After talking to @pombredanne, the next thing we want on top is to be able to install custom licenses using wheels. Something similar to the way we handle providing binaries to scancode via plugins: https://github.com/nexB/scancode-plugins/tree/main/builtins

We also merged the fix for the PyPI test in develop, so be sure to rebase your branch.

JonoYang:
@KevinJi22 Don't feel compelled to amend your commit and force push it every time you make a change. You can create as many commits as you want on this PR. :slightly_smiling_face:

KevinJi22:
@JonoYang @pombredanne this newest commit adds basic functionality for using installed licenses in scans. It assumes all packages of licenses start with "licenses" (we can change this later) and it assumes a certain directory structure. The only way to test this is manually. I added a directory containing an example package in `tests/licensedcode/data/example_external_licenses/licenses_to_install1`, and you can test it like so:

1. Run `pip install tests/licensedcode/data/example_external_licenses/licenses_to_install1` to install the licenses
2. Run `./scancode -l --json-pp ../temp.json /home/kevin/code/scancode-toolkit/tests/licensedcode/data/plugin_license/installed_licenses/scan`
3. In `temp.json`, you should see the license Example Installed License 1 being detected. You can also run Scancode before installing the license if you want to verify that the code actually works.

src/licensedcode/models.py
```diff
@@ -772,6 +773,81 @@ def get_rules(
...
+    additional_dirs,
+):
+    """
+    Takes in a list of additional directories specified during license detection
+    and produces a list of all the subdirectories containing rule files.
+    """
+    return [f"{str(Path(path).absolute())}/rules" for path in additional_dirs]
+
+
+def get_paths_to_installed_licenses():
```
JonoYang: Add a docstring for this function.

src/licensedcode/models.py
```diff
@@ -772,6 +773,81 @@ def get_rules(
...
+    from importlib_metadata import entry_points
+    from plugincode.location_provider import get_location
+    installed_plugins = entry_points(group='scancode_location_provider')
+    paths = []
+    for plugin in installed_plugins:
+        if plugin.name.startswith('licenses'):
```
JonoYang: I think it would be better to extract the custom license plugin prefix (`'licenses'`) to a variable after line 39, something along the lines of `EXTERNAL_LICENSE_PLUGIN_PREFIX = 'licenses'`, then use that variable here. It would make it easier to see and change what the external license plugin prefix is.

JonoYang (changes requested):
@KevinJi22 

I was able to run the commands you listed to test the `get_paths_to_installed_licenses` function you added. I have some minor things I commented on about it. 

As for creating this into a proper test, a possibility could be to create example external license plugin wheel and upload it to PyPI and then add the example external license plugin as a dependency under the `testing` extra_requires section in the `setup.cfg`. Afterwards, you can create a test to see if the custom license was detected.

@pombredanne What do you think about this? ^

pombredanne:
@KevinJi22 let me review all this in details.... a couple comments:

1. the DCO check is failing... there is likely a commit to amend

2. for the tests, I suggest that this could be a CI-only test. You could add a job in https://github.com/nexB/scancode-toolkit/blob/develop/azure-pipelines.yml or a GH workflow that would exclusively test your feature roughly this way:

2.1. first start by creating a directory in the tests data that would contain the complete and minimal working plugin that contributes some new licenses
2.2. then create a test job that installs scancode normally and then as an extra setup runs pip to install your plugin from that directory (no need to have a wheel or else publish it)
2.3. finally create a test_xxx.py scripts and with a test that assumes that your plugin is installed and only tests the plugin and run that as part of the job

Beyond this there are a few things that would be great to refine and get a well rounded feature:
3. some documentation including pointing to the test sample plugin from 2. 

4. expand the feature to include not only licenses but also 
4.1. license detection rules for these new licenses (the rules for the core licenses would be best kept with thee core licenses, location-wise)  
4.2. possibly also license detection tests such that it becomes easy to have tests for the new licenses and possible rules

5. with the addition of new licenses, there is likely a need to better report issues with the licenses metadata (and also rules metadata with 4.1) ... there are checks done when an index is created and checks done in the tests... See
  - https://github.com/nexB/scancode-toolkit/blob/a61fe17165fef37fb58a99c7b6a0267093ee33ff/tests/licensedcode/test_detection_validate.py#L50
  - https://github.com/nexB/scancode-toolkit/blob/a61fe17165fef37fb58a99c7b6a0267093ee33ff/tests/licensedcode/test_models.py#L108
  - https://github.com/nexB/scancode-toolkit/blob/a61fe17165fef37fb58a99c7b6a0267093ee33ff/src/licensedcode/models.py#L509
  - https://github.com/nexB/scancode-toolkit/blob/a61fe17165fef37fb58a99c7b6a0267093ee33ff/src/licensedcode/models.py#L794


These tests are not aware of plugins of course.... I wonder if some of then may be worthy of integration in the pre or post indexing checks.... as otherwise we may have some lingering issues in the plugins.... And of course if you move of the checks to indexing time, then there needs to be overall a better reporting of any issues to make the message actionable.

KevinJi22:
@pombredanne just wanted clarify about this point:
> These tests are not aware of plugins of course.... I wonder if some of then may be worthy of integration in the pre or post indexing checks.... as otherwise we may have some lingering issues in the plugins....

Are you saying that we might want to do some of the same checks that are done in the tests in the source code? And these checks would happen when the index is created (possibly with external licenses)?

docs/source/cli-reference/basic-options.rst
```diff
@@ -9,6 +9,61 @@
...
+    You can also include multiple directories like so::
+
+        scancode -clpieu --json-pp output.json samples -dir /home/user/external_licenses/external1 -dir /home/user/external_licenses/external2
+
+    If you want to continue running scans with ``/home/user/external_licenses/external1`` and ``/home/user/external_licenses/external2``,
+    you can omit the ``-dir`` option in subsequent scans and they will still be included. ::
+
+        scancode -clpieu --json-pp output.json samples
+
+    However, if you wanted to run a scan with a new set of directories, such as ``home/user/external_licenses/external1``
```
JonoYang: Just wondering: if we were to run: 

`scancode -clpieu --json-pp output.json samples -dir /home/user/external_licenses/external1 -dir /home/user/external_licenses/external2`

then afterwards we wanted to add the custom license `home/user/external_licenses/external3`, would we need to run the command:

`scancode -clpieu --json-pp output.json samples -dir /home/user/external_licenses/external1 -dir /home/user/external_licenses/external2 -dir /home/user/external_licenses/external3`

or is it enough to just call 

`scancode -clpieu --json-pp output.json samples -dir /home/user/external_licenses/external3` 

since we included `/home/user/external_licenses/external1` and `/home/user/external_licenses/external2` before?

azure-pipelines.yml
```diff
@@ -159,6 +160,11 @@ jobs:
           test_suites:
               all: venv\Scripts\pytest -n 2 -vvs tests\scancode\test_cli.py
 
+    - template: etc/ci/azure-installed-external-license.yml
```
JonoYang: Run this test on the Windows and MacOS CI platforms as well

JonoYang (approved):
@KevinJi22 The code looks good, I had a question about the external license path caching.

docs/source/cli-reference/basic-options.rst
KevinJi22: In this case, we'd have to run 
`
scancode -clpieu --json-pp output.json samples -dir /home/user/external_licenses/external1 -dir /home/user/external_licenses/external2 -dir /home/user/external_licenses/external3`

Right now the code checks if the directories passed in from the command line are a subset of the directories included in the cache, and if they aren't, it recreates the index with the input directories. So in this case, initially, `external1` and `external2` would be in the index.

 If we ran 
`scancode -clpieu --json-pp output.json samples -dir /home/user/external_licenses/external1 -dir /home/user/external_licenses/external2 -dir /home/user/external_licenses/external3
`, then `external1`, `external2`, and `external3` would be in the index.

If we ran `scancode -clpieu --json-pp output.json samples -dir /home/user/external_licenses/external3`, only `external3` would be in the index.

I implemented it this way since there isn't a way to "remove" licenses from the index, so I thought that if we always kept on adding licenses to the index each time someone used `-dir`, the index could get really large. Do you think I should implement it such that they always stay in the index?

JonoYang: I think what you did makes sense. Thanks for clarifying!

KevinJi22:
@pombredanne @JonoYang for `5. with the addition of new licenses, there is likely a need to better report issues with the licenses metadata (and also rules metadata with 4.1) ... there are checks done when an index is created and checks done in the tests...`, I have currently not implemented it as a CLI option since the checks will only be run when an index is created. If we are loading the index in from the cached file, the checks also will not run, which seems like the behavior we want. However, I can always add a command line option for maximum user flexibility.

src/licensedcode/models.py
```diff
@@ -772,6 +775,159 @@ def get_rules(
...
+    )
+    if errors:
+        message = ['Errors while validating licenses:']
+        for key, msgs in errors.items():
+            message.append('')
+            message.append(f'License: {key}')
+            for msg in msgs:
+                message.append(f'  {msg!r}')
+        raise InvalidLicense('\n'.join(message))
+
```
JonoYang, line 915: Extra new line here

src/licensedcode/plugin_license.py
```diff
@@ -139,6 +140,15 @@ class LicenseScanner(ScanPlugin):
...
+            ('-dir', '--additional_directories'),
+            required_options=['license'],
+            multiple=True,
+            type=click.Path(exists=True, readable=True, path_type=str),
+            help='Include additional directories for license detection.',
+            help_group=SCAN_OPTIONS_GROUP,
+        ),
+
         PluggableCommandLineOption(
             ('--reindex-licenses',),
```
JonoYang: Just thinking out loud: should reindex licenses also index custom licenses?

@pombredanne

JonoYang (commented):
@KevinJi22 Looks solid. I left some minor comments

JonoYang:
@KevinJi22 

After talking to @pombredanne, we'd like to modify the CLI option for external licenses. Currently, we have `-dir` and `--additional_directories` as CLI options, which we would like to change. 

We want to remove `-dir` as an option because:

1. We have no other single-dash option that is more than a single letter (e.g. `-c`, `-i`, etc.)
2. Single-dash options are for options that are more commonly used, we don't expect this option to be used on the same frequency as `-i`

We want to rename `--additional_directories` to `--external_license_directory` because:

- `--additional_directories` is too vague and something more explicit should be used to immediately convey what the option does / is for

pombredanne:
Actually `--external-license-directory` would be better for consistency as we never use an underscore in a command line option. Note that Click will translate this to an undescore-separated Python variable name alright

docs/source/rst_snippets/basic_options.rst
```diff
@@ -55,6 +55,13 @@ documenting a program's options. For example:
 
 .. include::  /rst_snippets/note_snippets/basic_clpieu.rst
 
+--external-license-directory PATH
+
+          Include paths to directories containing additional licenses and rules to use
+          in license detection. This can be used multiple times for multiple directories.
+
+          Sub-Option of - ``--license``
```
pombredanne: I am not convinced that this should be related to `--license` at all. Instead this is something that should require the `--reindex-licenses` option and should be used for reindexing and not for an actual scan. Otherwise this is confusing.

etc/ci/azure-posix-installed-external-license.yml
```diff
@@ -0,0 +1,38 @@
+parameters:
```
pombredanne: Any reason to create a pipeline just for this rather than just reuse the existing ones?
`venv/bin/python3 -m pip install tests/licensedcode/data/example_external_licenses/licenses_to_install1` could very much run as part of the first line of the tests IMHO

src/licensedcode/cache.py
```diff
@@ -75,12 +78,17 @@ def load_or_build(
         idx_cache_dir = os.path.join(licensedcode_cache_dir, LICENSE_INDEX_DIR)
         create_dir(idx_cache_dir)
         cache_file = os.path.join(idx_cache_dir, LICENSE_INDEX_FILENAME)
+        cached_directories_file = os.path.join(idx_cache_dir, CACHED_DIRECTORIES_FILENAME)
```
pombredanne: It looks like you are persisting the list of extra license directories... Can you elaborate on why would you need this?

docs/source/cli-reference/basic-options.rst
```diff
@@ -9,6 +9,61 @@
 
 ----
 
+``--external-license-directory`` Options
```
pombredanne: Everywhere in your code, you seem to use the term "additional" rather than external. Any reason for not using additional everywhere? It feels to describe better what this is about.

src/licensedcode/cache.py
```diff
@@ -106,13 +117,22 @@ def load_or_build(
                 # Here, the cache is either stale or non-existing: we need to
                 # rebuild all cached data (e.g. mostly the index) and cache it
 
-                licenses_db = load_licenses(licenses_data_dir=licenses_data_dir)
+                if additional_directories:
+                    additional_license_dirs = get_license_dirs(additional_dirs=additional_directories)
+                    validate_additional_license_data(additional_license_dirs)
+                    combined_directories = [licenses_data_dir] + additional_license_dirs
+                    licenses_db = load_licenses_from_multiple_dirs(license_directories=combined_directories)
```
pombredanne: I wonder if it could be simpler to always consider multiple directories and not treat "licenses_data_dir" as special?

src/licensedcode/cache.py
```diff
@@ -325,9 +386,42 @@ def populate_cache(force=False, index_all_languages=False):
             index_all_languages=index_all_languages,
             # used for testing only
             timeout=LICENSE_INDEX_LOCK_TIMEOUT,
+            additional_directories=additional_directories,
         )
 
 
+def need_cache_rebuild(additional_directories):
```
pombredanne: I do not think we should need any cache rebuild logic at all. Either we request an explicit rebuilt of the index with additional licenses or we do not. 

Rebuilding based on some cache invalidation is too complex to work correctly all the times and just checking directory names is not enough: the content inside could have changed for instance.

FWIW, there used to be a fairly sophisticated license cache invalidation logic in the past https://github.com/nexB/scancode-toolkit/blob/aadd743c5f789dc372ef75ed4745829c56cf37a9/src/licensedcode/cache.py#L205
I dropped this as this this was slow and incorrect.

src/licensedcode/models.py
```diff
@@ -73,6 +74,8 @@
 
 CATEGORIES = FOSS_CATEGORIES | OTHER_CATEGORIES
 
+# prefix for name of all externally installed license plugins
+EXTERNAL_LICENSE_PLUGIN_PREFIX = 'licenses'
```
pombredanne: I am not convinced that using a "prefix" is the best design.... Could we do better?
What about simply adding a new extension point that would be derived of the location provider? The location providers are supposed to be unique for a given location. In your cases, this does not hold and you can have multiple paths.

src/licensedcode/models.py
```diff
@@ -772,6 +775,160 @@ def get_rules(
...
+def get_paths_to_installed_licenses_and_rules():
+    """
+    Returns a list of paths to externally packaged licenses or rules that the user has
+    installed. Gets a list of all of these licenses (installed as plugins) and
+    then gets the plugins containing licenses by checking that their names start
+    with a common prefix.
+    """
+    from importlib_metadata import entry_points
+    from plugincode.location_provider import get_location
+    installed_plugins = entry_points(group='scancode_location_provider')
```
pombredanne: IMHO this is the sign that you need a new "scancode_additional_licenses_provider" of sorts and should not rely on scancode_location_provider (but can base yours on it alright).

src/licensedcode/models.py
```diff
@@ -772,6 +775,160 @@ def get_rules(
...
+            paths.append(get_location(location_key))
+    return paths
+
+
+def load_licenses_from_multiple_dirs(
+    license_directories,
+    with_deprecated=False,
+):
+    """
+    Takes in a list of directories containing additional licenses to use in
```
pombredanne: As a general style rule, we prefer:
1. imperative style
2. starting with the outcome

so here, this would become:
```
Return a mapping of {key: License} built from a list of ``license_directories``.
```
... which raises a few extra questions: ```license_directories``: are these absolute locations? say so.

src/licensedcode/models.py
```diff
@@ -772,6 +775,160 @@ def get_rules(
...
+    license detection and combines all the licenses into the same mapping.
+    """
+    combined_licenses = {}
+    for license_dir in license_directories:
+        licenses = load_licenses(licenses_data_dir=license_dir, with_deprecated=False)
+        # this syntax for merging is described here: https://stackoverflow.com/a/26853961
+        combined_licenses = {**combined_licenses, **licenses}
```
pombredanne: Are you sure this is what you want there? IMHO a more and explicit common syntax would be: `combined_licenses.update(licenses)` and avoid recreating a whole new dict at each iteration.

Also what this would mean is that:
1. there can be multiple time the same license key
2. whichever comes last in the loading order wins

IMHO this becomes rather important as this means that additional licenses can override existing licenses. We should prohibit this: I think this is best as otherwise the loading order is undefined and we end up with really byzantine weirdness. You should ensure that the keys are unique and fail loudly with a list of duplicate keys if this happens.

src/licensedcode/models.py
```diff
@@ -772,6 +775,160 @@ def get_rules(
...
+            message.append(f'License: {key}')
+            for msg in msgs:
+                message.append(f'  {msg!r}')
+        raise InvalidLicense('\n'.join(message))
+
+
+def _ignorable_clue_error(rule):
+    """
+    Helper method to validate a single rule's ignorable clues.
+    Returns a pair of the result and expected ignorable clues if
```
pombredanne: Always use imperative style in docstrings.

src/licensedcode/plugin_license.py
```diff
@@ -18,6 +18,7 @@
 from commoncode.resource import clean_path
 from plugincode.scan import ScanPlugin
 from plugincode.scan import scan_impl
+import click
```
pombredanne: This should be right after the `import attr` in the import order (and eventually we should use isort or usort to automate this)

src/licensedcode/plugin_license.py
```diff
@@ -139,6 +140,15 @@ class LicenseScanner(ScanPlugin):
             help_group=SCAN_OPTIONS_GROUP,
         ),
 
+        PluggableCommandLineOption(
+            ('--external-license-directory',),
+            required_options=['license'],
```
pombredanne: this should be `reindex_licenses` instead

src/licensedcode/plugin_license.py
```diff
@@ -139,6 +140,15 @@ class LicenseScanner(ScanPlugin):
...
+            multiple=True,
+            type=click.Path(exists=True, readable=True, path_type=str),
+            help='Include additional directories for license detection.',
```
pombredanne: You use additional here again. So IMHO additional is the right term to use throughout, not external.
Also I would add a `metavar='DIR'` and the help could be refined.... What about:
`Include this directory with additional custom licenses and license rules in the license detection index.`

src/licensedcode/plugin_license.py
```diff
@@ -139,6 +140,15 @@ class LicenseScanner(ScanPlugin):
             help_group=SCAN_OPTIONS_GROUP,
         ),
 
+        PluggableCommandLineOption(
+            ('--external-license-directory',),
+            required_options=['license'],
+            multiple=True,
+            type=click.Path(exists=True, readable=True, path_type=str),
```
pombredanne:
```suggestion
            type=click.Path(exists=True, readable=True, file_okay=False, resolve_path=True, path_type=str),
```

src/licensedcode/plugin_license.py
pombredanne: @JonoYang yes

src/licensedcode/plugin_license.py
```diff
@@ -167,7 +177,8 @@ def setup(self, **kwargs):
         loaded index.
         """
         from licensedcode.cache import populate_cache
-        populate_cache()
+        additional_directories = kwargs.get('external_license_directory')
+        populate_cache(additional_directories=additional_directories)
```
pombredanne: IMHO additional_directories should be used only when reindexing and not in general.

src/licensedcode/plugin_license.py
```diff
@@ -176,6 +187,7 @@ def get_scanner(
         license_text_diagnostics=False,
         license_url_template=SCANCODE_LICENSEDB_URL,
         unknown_licenses=False,
+        additional_directories=None,
```
pombredanne: I am not sure this would be needed further down. Instead the license index has these details

src/scancode/api.py
```diff
@@ -142,6 +142,7 @@ def get_licenses(
     license_url_template=SCANCODE_LICENSEDB_URL,
     unknown_licenses=False,
     deadline=sys.maxsize,
+    additional_directories=None,
```
pombredanne: IMHO this is not needed. As explained earlier, I think that considering additional directories should  be something that's only done when re-indexing and not when scanning. Asking for additional_directories should be an error when scanning.
So much so that I would think as the `--reindex-licenses` option pretty much as if it were a separate command entirely.

src/scancode/api.py
```diff
@@ -168,7 +169,7 @@ def get_licenses(
     from licensedcode import cache
     from licensedcode.spans import Span
 
-    idx = cache.get_index()
+    idx = cache.get_index(additional_directories=additional_directories)
```
pombredanne: The index should be:
- not existing in the cache in which case we could either rebuild or we could simply error out asking for a reindex.
- existing in the cache and used as-is with whatever directories was used to create it....but passing down additional_directories should never be needed.

src/scancode/api.py
```diff
@@ -252,6 +253,7 @@ def _licenses_data_from_match(
         result['homepage_url'] = lic.homepage_url
         result['text_url'] = lic.text_urls[0] if lic.text_urls else ''
         result['reference_url'] = license_url_template.format(lic.key)
```
pombredanne: All these URLs should be empty for additional licenses for now. Later we could provide a way to include them in the data files.

tests/licensedcode/data/example_external_licenses/licenses_to_install1/gpl-1.0.LICENSE
```diff
@@ -0,0 +1,254 @@
+GNU GENERAL PUBLIC LICENSE
```
pombredanne: Using something shorter and with a smaller file size for an example plugin may be better? This is going to be checked out 1000's of times ;)

tests/licensedcode/data/example_external_licenses/licenses_to_install1/setup.py
```diff
@@ -0,0 +1,50 @@
...
+        'Development Status :: 5 - Production/Stable',
+        'Intended Audience :: Developers',
+        'Topic :: Utilities',
+    ],
+    keywords=[
+        'open source', 'scancode_licenses',
+    ],
+    install_requires=[
+        'scancode-toolkit',
+        'wheel',
```
pombredanne: Why would you need wheel and pytest as runtime requirements?

tests/licensedcode/data/plugin_license/external_licenses/scan.expected.json
```diff
@@ -0,0 +1,81 @@
...
+        "--license": true,
+        "--strip-root": true
+      },
+      "notice": "Generated with ScanCode and provided on an \"AS IS\" BASIS, WITHOUT WARRANTIES\nOR CONDITIONS OF ANY KIND, either express or implied. No content created from\nScanCode should be considered or used as legal advice. Consult an Attorney\nfor any legal advice.\nScanCode is a free software code scanning tool from nexB Inc. and others.\nVisit https://github.com/nexB/scancode-toolkit/ for support and download.",
+      "output_format_version": "2.0.0",
+      "message": null,
+      "errors": [],
+      "warnings": [],
+      "extra_data": {
+        "system_environment": {
```
pombredanne: IMHO we should include the list of paths to additional licenses directories (either plain dirs or plugin-contributed) in the output as extra data, otherwise there is no way to know something non-standard was used.

pombredanne (changes requested):
Thanks! Here is my review at last. We can discuss the design further at your convenience.

docs/source/cli-reference/basic-options.rst
KevinJi22: The reason I used "external" was in response to one of your earlier comments:

> Actually --external-license-directory would be better for consistency as we never use an underscore in a command line option. Note that Click will translate this to an undescore-separated Python variable name alright

but I can change it to something like `--additional-license-directory`?

src/licensedcode/cache.py
KevinJi22: This was originally used for determining cache invalidation but now that we're changing the approach to explicitly force reindexing this can be deleted.

src/scancode/api.py
pombredanne: @KevinJi22 you wrote:

> also, for this, I don't think there's a way to identify if a license matches an external license vs. a license already in the licenseDB, right? So I'm not sure how we can set these fields to empty for additional licenses

Good point... may be we need to add some flag or data source field of sorts to the License and Rule? something that would indicate the provenance?
Some food for thoughts: a minimal approach could be an "is_builtin" field added to the License and Rule and that would be set automatically to False for additions and True for builtins?

KevinJi22: This would work as an approach but then I think we'd have to separate out loading the built in licenses/rules and the additional ones, which is a little less clean but I think that's the only way to do it. 

Basically, going back on this idea and treating licenses_data_dir as special:
> I wonder if it could be simpler to always consider multiple directories and not treat "licenses_data_dir" as special?

pombredanne:
@KevinJi22 I would guess this is now ready for review? and no longer a draft/WIP?

KevinJi22:
@pombredanne yup! Just edited the title to reflect that

AyanSinhaMahapatra:
@KevinJi22 we (@pombredanne  and me) added the following changes:

CLI Options:

* various reindexing operations are no longer CLI options
* we have a seperate script scancode-reindex-licenses which does that
* only one additional directory of licenses/rules can be added, i.e. this is not a multiple option

Tests:

* Organize tests data under one single folder.
* Organize test functions under one single file.
* Keep three tests, one for additional directory, one for
  additional plugin, and one for combined testing.
* Edit the CI file accordingly.

Misc:

* code refactoring at `get_rules_from_multiple_dirs`, `get_rules_from_multiple_dirs` etc
* external license directories/plugin info is returned in scan header

Remaining work:

* update how license data (links and other) is returned in `api.py`
* also report new attribute `is_builtin` 
@pombredanne ^ this will create a lot of test file updates, I was thinking I'll do this update in the main LicenseDetection PR as there's lot's of change in license data reporting already. 

Otherwise tests are all green and this is ready AFAIK.
Please review. :bow:

AyanSinhaMahapatra:
> Remaining work:
>    * update how license data (links and other) is returned in api.py
>    * also report new attribute is_builtin
>  @pombredanne ^ this will create a lot of test file updates, I was thinking I'll do this update in the main LicenseDetection PR as
>  there's lot's of change in license data reporting already.

Modified the code to add `is_builtin` flags to license match data when there is an additional license directory or additional license plugins installed. So this work is not remaining anymore.

Ready to review @pombredanne!

Was also wondering if there could be a `--only-builtin` CLI option in `scancode-reindex-licenses` to reindex the license freshly without having to delete and reconfigure the environment, but not sure how to uninstall plugins from the python script. @pombredanne is this doable?

azure-pipelines.yml
```diff
@@ -32,8 +32,7 @@ jobs:
                   --ignore=tests/licensedcode/test_detection_datadriven1.py \
                   --ignore=tests/licensedcode/test_detection_datadriven2.py \
                   --ignore=tests/licensedcode/test_detection_datadriven3.py \
-                  --ignore=tests/licensedcode/test_detection_datadriven4.py \
-                  tests/licensedcode
+                  --ignore=tests/licensedcode/test_detection_datadriven4.py
```
pombredanne: Why removing this? this will be a regression.

```suggestion
                  --ignore=tests/licensedcode/test_detection_datadriven4.py \
                  --ignore=tests/licensedcode/test_additional_license.py \
                   tests/licensedcode
```

docs/source/cli-reference/core-options.rst
```diff
@@ -83,6 +83,58 @@ Comparing Progress Message Options
     ..
         [ToDo] Research and Write Better
 
+``--additional-license-directory`` Option
```
pombredanne: This whole section likely needs to be updated now that we changed the CLI UX.

docs/source/how-to-guides/install_new_license_plugin.rst
```diff
@@ -0,0 +1,139 @@
+.. _install_new_license_plugin:
+
+How to Install External License Plugins to Use in License Detection
```
pombredanne: This likely needs love with the new CLI UX.

docs/source/rst_snippets/core_options.rst
```diff
@@ -14,6 +14,11 @@ All "Core" Scan Options
 --reindex-licenses       Force a check and possible reindexing of the
                          cached license index.
 
+--additional-license-directory PATH
```
pombredanne: Now obsolete

docs/source/rst_snippets/note_snippets/core_indep.rst
```diff
@@ -1,3 +1,4 @@
 .. note::
 
-    All the Core Options are independent options, i.e. They don't depend on other options.
+    All the Core Options except for ``--additional-license-directory`` are independent options, i.e.
```
pombredanne: Obsolete

src/licensedcode/cache.py
```diff
@@ -6,6 +6,7 @@
 # See https://github.com/nexB/scancode-toolkit for support or download.
 # See https://aboutcode.org for more information about nexB OSS projects.
 #
+import click
```
pombredanne: This should be moved further down... this is not stdlib

src/licensedcode/cache.py
```diff
@@ -66,11 +70,15 @@ def load_or_build(
         On the side, we load cached or build license db, SPDX symbols and other
         license-related data structures.
 
-        - If the cache exists, it is returned unless corrupted or ``force`` is True.
+        - If the cache exists, it is returned unless corrupted, ``force`` is True, or if we pass in additional
```
pombredanne: If we merely pass an  additional directory, this should not implicitly trigger a forced rebuild.

pombredanne:
> Was also wondering if there could be a --only-builtin CLI option in scancode-reindex-licenses to reindex the license freshly without having to delete and reconfigure the environment, but not sure how to uninstall plugins from the python script. @pombredanne is this doable?

Sure, but remember that each new added option may be a sign of our failure to automatically do the right thing without options :)

src/licensedcode/cache.py
```diff
@@ -47,9 +49,12 @@ class LicenseCache:
     licensing = attribute(help='Licensing object')
     spdx_symbols = attribute(help='mapping of LicenseSymbol objects by SPDX key')
     unknown_spdx_symbol = attribute(help='LicenseSymbol object')
+    additional_license_directory = attribute(help='Path to an additional license directory used in the license detection')
```
pombredanne, line 52: I am not sure keeping these directories make sense. They cannot really be used at runtime as the additional license directory may not exist anymore and the plugins could have been uninstalled. In a way now that we keep all the license and rule texts and data in the index, these directories do not matter anymore: we could dump the data if needed and we just need to know if a license is built-in or not: this could be a flag on the License or this could just be a separate index set that lists built-in license keys?

src/licensedcode/models.py
```diff
@@ -436,8 +444,6 @@ def write(location, byte_string):
         if text:
             write(self.text_file(licenses_data_dir=licenses_data_dir), text.encode('utf-8'))
 
-        return self
```
pombredanne: Why removing this?

pombredanne (changes requested):
Here some extra feedback. I think we should refactor this a bit more such that indexing licenses and rules just takes a list of directories. Everything else should IMHO just use the things already loaded. And we can simplify the detection of builtins based on a list of builtin licenses... This would help keep the code tidier IMHO. @AyanSinhaMahapatra @KevinJi22 let's chat at your convenience

src/licensedcode/models.py
AyanSinhaMahapatra: Added this back, this shouldn't be removed, used in scripts.

src/licensedcode/cache.py
AyanSinhaMahapatra: This isn't the current behaviour, updating the docstring.

src/licensedcode/cache.py
AyanSinhaMahapatra: This is outdated, couldn't find it.

docs/source/rst_snippets/note_snippets/core_indep.rst
AyanSinhaMahapatra: Removing this.

docs/source/rst_snippets/core_options.rst
AyanSinhaMahapatra: Yes! Removing this.

docs/source/how-to-guides/install_new_license_plugin.rst
AyanSinhaMahapatra: Yes, already added doc changes, see latest commit.

docs/source/cli-reference/core-options.rst
AyanSinhaMahapatra: Yes, moving this to a new page for `other commands`

azure-pipelines.yml
AyanSinhaMahapatra: Yes, adding this back.

AyanSinhaMahapatra: Added this back.

src/licensedcode/cache.py
pombredanne: Actually you are right!

pombredanne:
> Here some extra feedback. I think we should refactor this a bit more such that indexing licenses and rules just takes a list of directories. Everything else should IMHO just use the things already loaded. And we can simplify the detection of builtins based on a list of builtin licenses... This would help keep the code tidier IMHO. @AyanSinhaMahapatra @KevinJi22 let's chat at your convenience

I created https://github.com/nexB/scancode-toolkit/issues/3137 as a follow up!

pombredanne (approved):

CHANGELOG.rst
```diff
@@ -46,14 +46,17 @@ License detection:
...
   return match details once for each matched license expression rather than
   once for each license in a matched expression. There is a new top-level
   "license_references" attribute that contains the data details for each
   detected license only once. This data can contain the reference license text
   as an option.
 
+- We can now detect licenses using custom license texts and license rules.
+  These can be provided as a one off in a directory or packaged as a plugin
+  for consistent reuse and deployment.
+
```
pombredanne, line 59:
```suggestion

- There is a new "scancode-reindex-licenses" command that replace the
  "scancode --reindex-licenses" command line option which has been
  removed. This new command supports simpler reindexing using custom
  license texts and license rules contributed by plugins or stored in an
  additional directory. 
```

pombredanne (approved):
All good to go. Merging

PR #3042, opened 2022-08-04
Update DatafileHandler default methods

JonoYang:
This PR updates `DatafileHandler.assemble()` and `DatafileHandler.assemble_from_many()` to yield Packages, Dependencies, and Resources before associating Packages to Resources using a `package_adder`. This is to help with using packagedcode Package handlers in scancode.io, where a Package needs to be created in scancode.io before it can be associated with a Resource.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -60,6 +60,13 @@ Important API changes:
   under the ``venv`` subdirectory. You mus be aware of this if you use ScanCode
   from a git clone
 
+- ``DatafileHandler.assemble()``, ``DatafileHandler.assemble_from_many()``, and
+  the other ``.assemble()``` methods from the other Package handlers from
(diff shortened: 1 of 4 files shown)


JonoYang:
On second thought, it makes more sense to sort the assembly items after the fact in scancode.io than it is to mandate every `.assembly()` method yields Packages, Dependencies, and Resources in that order.

JonoYang:
@pombredanne 

I've updated the rest of the `assemble()` methods to yield Packages before other assembly items. I tried sorting the output of the `assemble()` method in my code on scancode.io, but it doesn't seem that there is a way around reordering the logic in the different `assemble()` methods when we want to create a Package before associating it to a Resource.

pombredanne (approved):
LGTM!
Thanks
merging.

PR #3062, opened 2022-08-22
Fix issues in `PythonInstalledWheelMetadataFile.assign_package_to_resources()`

JonoYang:
This PR fixes a few issues in `PythonInstalledWheelMetadataFile.assign_package_to_resources()`:

- `PythonInstalledWheelMetadataFile.assign_package_to_resources()` would not correctly find the `site-packages` directory, as it called `.parent()` too many times on the datafile Resource.
- The variable `path_ref` was not being assigned correctly and `PythonInstalledWheelMetadataFile.assign_package_to_resources()` would fail during Package assembly.

### DIFF
--- a/src/packagedcode/models.py
+++ b/src/packagedcode/models.py
@@ -805,9 +805,12 @@ def compute_normalized_license(declared_license, expression_symbols=None):
 
 def add_to_package(package_uid, resource, codebase):
     """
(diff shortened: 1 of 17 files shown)


pombredanne (approved):
LGTM!
merging. Thank you ++

Issue #3072, opened 2022-08-30
Repeated package and dependency results when scanning extracted rubygem


JonoYang:
I extracted a rubygem (https://rubygems.org/downloads/awesome_print-1.8.0.gem) and I scanned it with scancode-toolkit and I got repeated Package and Dependency results. The package `awesome_print` was reported 20 times, and each dependency for `awesome_print` was reported 20 times. 

In `GemfileLockHandler.parse()` it appears that we yield the package and dependencies too many times.

PR #3090, opened 2022-09-02
Process Gemfile.lock processing #3072

JonoYang:
This PR addresses the issue in #3072, where (in a scancode scan) the main gem that the gemfile.lock is for has been reported the same number of times as there are dependencies in the gemfile.lock and that the entire set of dependencies have been repeated the same amount of times.

The changes I've made in this PR was to update the GemfileParser to look for the gem that was created for the contents in the `PATH` section of the gemfile.lock and save it to a new attribute on `GemfileParser` named `primary_gem`.

In `GemfileLockHandler.parse()`, if we have a primary gem from a gemfile.lock, we use that primary gem info as the base of the package data and we consider all other gems from that lockfile as a dependency of the primary gem. If there is no primary gem, then we collect all the dependencies and return them in a package with no name.


pombredanne (approved):
LGTM!
Thanks... merging!

Issue #3083, opened 2022-09-01
No dependency results when scanning celery-5.2.7.tar.gz


JonoYang:
Originally reported at https://github.com/nexB/scancode.io/issues/516

I've downloaded, extracted, and run a package scan on https://files.pythonhosted.org/packages/ce/21/41a0028f6d610987c0839250357c1a00f351790b8a448c2eb323caa719ac/celery-5.2.7.tar.gz but no dependency results show up in the top-level dependencies field.

I think this is due to the pypi package parser getting results from the `celery-5.2.7/PKG-INFO` file at the root of the codebase, but it cannot find any of the dependency versions and other info because the `PKG-INFO` file is not in an `.egg-info` directory. When running the package scanning debug flags, I see that we get the dependencies when we scan `celery-5.2.7/celery.egg-info/PKG-INFO`.

JonoYang:
Fix merged in #3091

PR #3091, opened 2022-09-03
Prefer using PKG-INFO from .egg-info in assemble #3083

JonoYang:
This PR addresses #3083, where we have no dependencies returned when we scanned an installable Python codebase. This issue is cause by `BaseExtractedPythonLayout.assemble()`, where it would use the package data from the PKG-INFO file from the root of a Python codebase rather than from the PKG-INFO file from the .egg-info directory (also located in the root of a Python codebase). 

We want to use the PKG-INFO file from the .egg-info directory because the package data from that contains the dependency information.


tests/packagedcode/data/pypi/unpacked_sdist/prefer-egg-info-pkg-info/celery-expected.json
```diff
@@ -0,0 +1,8105 @@
...
+              "is_runtime": true,
+              "is_optional": false,
+              "is_resolved": false,
+              "resolved_package": {},
+              "extra_data": {}
+            },
+            {
+              "purl": "pkg:pypi/azure-storage-blob@12.9.0",
+              "extracted_requirement": "azure-storage-blob==12.9.0; extra == \"azureblockblob\"",
+              "scope": "azurebl
```
pombredanne: We should have an extracted requirement of "msgpack"

pombredanne: We should have instead:
```suggestion
              "extracted_requirement": "couchbase>=3.0.0",
```

tests/packagedcode/data/pypi/unpacked_sdist/prefer-egg-info-pkg-info/celery-expected.json
```diff
@@ -0,0 +1,8105 @@
...
+      "for_package_uid": "pkg:pypi/celery@5.2.7?uuid=fixed-uid-done-for-testing-5642512d1758",
+      "datafile_path": "celery/celery.egg-info/PKG-INFO",
+      "datasource_id": "pypi_editable_egg_pkginfo"
+    },
+    {
+      "purl": "pkg:pypi/pylibmc",
+      "extracted_requirement": "pylibmc; platform_system != \"Windows\" and extra == \"memcache\"",
+      "scope": "memcache",
+      "is_runtime": true,
+      "is_optional": false,
```
pombredanne: IMHO we should be true for "extras" e.g. not an "install_requires" and therefore not in "install" scope:
```suggestion
      "is_optional": true,
```

pombredanne (changes requested):
Thanks!
There are a few nit that likely are leftover from before... we should better track the extracted requirements and treat PyPI extras as optional

JonoYang:
I've updated the code to return the dependency name in the extracted_requirements and to properly set the is_optional field on the dependencies listed in extra_requires.

Tests are passing, so I will merge this.

Issue #1243, opened 2018-11-01
Collect data from Yocto/bitbake .bb manifest files


pombredanne:
The commonly contain decent license and origin information.

pombredanne:
See also https://gist.github.com/craigez/df6bcd275440ee0047d020511a3d4532#bitbake-recipes-and-yocto-distributions by @craigez

craigez:
I have some little scripts that dump declared licenses, origins, etc... let me know if that would be useful.

pombredanne:
@craigez that would be great! Even if this is not in Python that will be useful to create a new manifest parser in https://github.com/nexB/scancode-toolkit/tree/develop/src/packagedcode

craigez:
I actually insert it into the bitbake process to dump the information. This requires updating the configuration to include this build step.

pombredanne:
@craigez that's fine I guess, especially since this is Python code. I am also curious on how you plugin some code in the bitbake process BTW ... as this can be useful for tracing, licensing scanning, etc.

craigez:
@pombredanne  So, here's an example of the type of script I've used before to extract data for one-off investigations. It was based on the existing spdx.bbclass that uses fossology, but this just dumps out information without scanning: https://gist.github.com/craigez/607ec343a7a08f3b465c7307377ae84a

And here's a hacky way to get it to run: https://gist.github.com/craigez/066285753b69139b9d98b0fdd17ae56a

You can add other bitbake fields like 'license', etc... to get that information into the CSV

Mark and I were also looking at how to incorporate scanning, etc... https://github.com/mcharleb/meta-srcmap (this is also how you properly add a layer...)

pombredanne:
@craigez Thanks you++ for taking the time to put these together ... But this means this is used while actually running a build, correct?

That would be more of a job for TraceCode then IMHO.

ScanCode only does  static analysis and manifest parsing... so I was more thinking about something that can collect information from .bb files found in a codebase at rest.

craigez:
@pombredanne I wanted to do something like that too, but I couldn't find a pre-existing python library to parse the files. It really needs to interpret them though anyway, as files include other files and there are variables defined during the build, etc... So it was easier/quicker for my use case to just add a build step and then use bitbake to run it (without compilation, linking steps, etc...)

pombredanne:
Some extra pointers for some minimal implementation of a static analysis:
[...]

PR #3092, opened 2022-09-04
Add support for bitbake recipes #1243

pombredanne:
This PR adds basic support bitbake recipes treated as package data. 

### Tasks

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,6 +17,7 @@ cryptography==42.0.5
 debian-inspector==31.1.0
 dockerfile-parse==1.2.0
 dparse2==0.7.0
+deprecated == 1.2.18
 extractcode==31.0.0
 extractcode-7z==16.5.210531
 extractcode-libarchive==3.5.1.210531
@@ -40,6 +41,7 @@ lxml==5.1.0
 MarkupSafe==2.1.5
 more-itertools==8.13.0
 normality==2.3.3
+oelint_parser==8.1.0
 packageurl-python==0.15.0
 packaging==24.1
 packvers==21.5
@@ -61,6 +63,7 @@ pyparsing==3.0.9
 pytz==2022.1
 PyYAML==6.0.1
 rdflib==6.2.0
+regex == 2024.11.6
 requests==2.31.0
 saneyaml==0.6.0
 semantic-version==2.8.5
--- a/src/packagedcode/bitbake.py
+++ b/src/packagedcode/bitbake.py
@@ -1,0 +1,194 @@
+#
+# Copyright (c) nexB Inc. and others. All rights reserved.
+# ScanCode is a trademark of nexB Inc.
+# SPDX-License-Identifier: Apache-2.0
+# See http://www.apache.org/licenses/LICENSE-2.0 for the license text.
+# See https://github.com/nexB/scancode-toolkit for support or download.
+# See https://aboutcode.org for more information about nexB OSS projects.
+#
+
+import logging
+import re
+
+from oelint_parser.cls_stash import Stash
+from packageurl import PackageURL
+
+from packagedcode import models
+
+TRACE = True
+
+logger = logging.getLogger(__name__)
+
+if TRACE:
+    import sys
+    logging.basicConfig(stream=sys.stdout)
+    logger.setLevel(logging.DEBUG)
+
+
+class BitbakeBbManifestHandler(models.DatafileHandler):
+    datasource_id = 'bitbake_bb_recipe'
+    # note that there are .bbappend, .bbclass and bitbake.conf files.
+    path_patterns = ('*.bb',)
+    default_package_type = 'bitbake'
+    description = 'BitBake bb recipe manifest'
+    documentation_url = 'https://docs.yoctoproject.org/bitbake/bitbake-user-manual/bitbake-user-manual-metadata.html'
+
+    @classmethod
+    def parse(cls, location):
+
+        oestash = Stash(quiet=True)
+
+        # add any bitbake like-file
+        # TODO: may be we should handle the bbclass and bbappend here?
+        oestash.AddFile(location)
+
+        # Resolves proper cross file dependencies
+        oestash.Finalize()
+
+        # collect all variables of interest.
+        # TODO: we should not get list values. Instead plain strings
+        data = {
+            k: ' '.join(v) if isinstance(v, (list, tuple)) else v
+            for k, v in oestash.ExpandVar(filename=location).items()
+            if v
+        }
+        name = data.get('PN')
+        version = data.get('PV')
+        description = data.get('DESCRIPTION')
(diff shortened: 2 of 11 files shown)


pombredanne:
@chinyeungli @priv-kweihmann you review is welcomed!
@priv-kweihmann I would like to know in particular if I am using oelint-parser correctly in https://github.com/nexB/scancode-toolkit/pull/3092/files#diff-af7c94424fa6a057497534cd0f8f26d906dfb225318819c29c919c8905493dc2R37

requirements.txt
```diff
@@ -39,6 +39,7 @@ lxml==4.9.1
 MarkupSafe==2.1.1
 more-itertools==8.13.0
 normality==2.3.3
+oelint_parser==2.6.0
```
priv-kweihmann: I think current version is 2.8.0 - so this might need an update

src/packagedcode/bitbake.py
```diff
@@ -0,0 +1,192 @@
...
+    documentation_url = 'https://docs.yoctoproject.org/bitbake/bitbake-user-manual/bitbake-user-manual-metadata.html'
+
+    @classmethod
+    def parse(cls, location):
+
+        oestash = Stash(quiet=True)
+
+        # add any bitbake like-file
+        # TODO: may be we should handle the bbclass and bbappend here?
+        oestash.AddFile(location)
```
priv-kweihmann, line 43: bbclass files are handled automatically (if being within the same layer) - but yeah, bbappends have to be added manually

src/packagedcode/bitbake.py
```diff
@@ -0,0 +1,192 @@
...
+        data = {
+            k: ' '.join(v) if isinstance(v, (list, tuple)) else v
+            for k, v in oestash.ExpandVar(filename=location).items()
+            if v
+        }
+        name = data.get('PN')
+        version = data.get('PV')
+        description = data.get('DESCRIPTION')
+        homepage_url = data.get('HOMEPAGE')
+        download_url = data.get('PREMIRRORS')
```
priv-kweihmann, line 59: Why premirrors? Normally one would take this kind of information from SRC_URI

src/packagedcode/bitbake.py
```diff
@@ -0,0 +1,192 @@
...
+        declared_license = data.get('LICENSE')
+
+        # The item.VarName for SRC_URI[*] from the parser are SRC_URI
+        # Therefore, I cannot differentiate md5, sha1, or src file location reference
+        # See: https://github.com/priv-kweihmann/oelint-parser/issues/3
+        sha1 = data.get('SRC_URI[sha1sum]')
+        md5 = data.get('SRC_URI[md5sum]')
```
priv-kweihmann, line 66: md5sum is pretty outdated a while now - people have been encouraged to drop the support for md5 for years now - sha256 on the other hand is mandatory for most of the file based downloads, so this one should be used

src/packagedcode/bitbake.py
```diff
@@ -0,0 +1,192 @@
...
+        sha256 = data.get('SRC_URI[sha256sum]')
+        sha512 = data.get('SRC_URI[sha512sum]')
+
+        dependencies = []
+
+        # Build deps: this is a list of plain BB recipes names
+        # https://docs.yoctoproject.org/bitbake/bitbake-user-manual/bitbake-user-manual-ref-variables.html#term-DEPENDS
+        build_deps = data.get('DEPENDS', '').split()
```
priv-kweihmann, line 74: DEPENDS just refers to buildtime dependencies - I think what you really want is a mix of RDEPENDS:* and DEPENDS - while RDEPENDS (the packages needed at runtime are more of importance) - sometimes DEPENDS contains elements that will be only used at buildtime and leave no trace within the shipped binaries, so from a license point of view they can be safely ignored

tests/packagedcode/data/bitbake/initramfs-live-install-testfs_1.0.bb
```diff
@@ -0,0 +1,17 @@
+SUMMARY = "Live image install script with a second rootfs/kernel"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COREBASE}/meta/COPYING.MIT;md5=3da9cfbcb788c80a0384361b4de20420"
+SRC_URI = "file://init-install-testfs.sh"
+DEPENDS = "do_this and that"
+
+RDEPENDS_${PN} = "grub (== 12.23) parted e2fsprogs-mke2fs"
```
priv-kweihmann, line 7: I think the test should use a recipe that is a bit more complex than this one here - for me https://git.yoctoproject.org/poky/plain/meta/recipes-graphics/cairo/cairo_1.16.0.bb was always a good example

src/packagedcode/bitbake.py
```diff
@@ -0,0 +1,192 @@
...
+            k: ' '.join(v) if isinstance(v, (list, tuple)) else v
+            for k, v in oestash.ExpandVar(filename=location).items()
+            if v
+        }
+        name = data.get('PN')
+        version = data.get('PV')
+        description = data.get('DESCRIPTION')
+        homepage_url = data.get('HOMEPAGE')
+        download_url = data.get('PREMIRRORS')
+        declared_license = data.get('LICENSE')
```
priv-kweihmann: Please keep in mind that LICENSE settings can also be package specific (fine example can be found here https://git.yoctoproject.org/poky/plain/meta/recipes-graphics/cairo/cairo_1.16.0.bb) - so it would be better to iterate over all variable items of LICENSE and then pick the best matching, where plain LICENSE should always be the fall back

src/packagedcode/bitbake.py
priv-kweihmann: I personally would say that license wise DEPENDS doesn't play any role and should not be considered

jloehel:
@pombredanne I am a bit curious how would you handle the empty SRC_URI [1] for the gcc builds? There are also a couple of recipes which have just "local" files and no upstream download URL or where the a couple of files get constructed on the fly. Like mender-artifact-info.bb. I would extract also the download_url from the SRC_URI. The SRC_URI array should include also the patches. In case of git download URLs the SRCREV is also helpful.

I made the experience that some older recipes SRC_URIs are broken. In example for mesa 22.0.3 (https://mesa.freedesktop.org/archive/mesa-22.0.3.tar.xz The redirect to https://archive.mesa3d.org/older-versions/ is not working correctly) and chrony 4.2 (https://download.tuxfamily.org/chrony/chrony-4.2.tar.gz tuxfamily does not work anymore). How will you handle broken download_urls?

It would be nice to extract also LIC_FILES_CHKSUM? Sometimes it includes also license snippets with bline and eline.

[1] https://github.com/openembedded/openembedded-core/commit/0df9d45e0be59e55e585e6d25dedbf0fc55c490c

PR #3111, opened 2022-09-19
Add package_releases in packagedata model

TG1999:
Signed-off-by: Tushar Goel <tushar.goel.dav@gmail.com>

Add package_releases in packagedata model to have more than one downloadable url per package


TG1999:
@AyanSinhaMahapatra @pombredanne @JonoYang please have a look on these model changes, please give your suggestion on this and do let me know if these model changes are useful in anyway. We can remove all the fields from the PackageData model that will be shifted to PackageRelease after we migrate these fields to PackageRelease model.

src/packagedcode/models.py
```diff
@@ -529,6 +529,55 @@ def update(self, other):
         return self
 
 
+wheel = "wheel"
+sdist = "sdist"
+
+RELEASE_TYPES = (
+    None,
+    wheel,
```
JonoYang, line 537: It feels strange to have package-specific release types in the base models.py file.

tests/packagedcode/data/alpine/alpine-container-layer.tar.xz-get-installed-expected.json
```diff
@@ -32,7 +33,7 @@
     "code_view_url": null,
     "vcs_url": "git+http://git.alpinelinux.org/aports/commit/?id=cb70ca5c6d6db0399d2dd09189c5d57827bce5cd",
     "copyright": null,
-    "license_expression": "gpl-2.0",
+    "license_expression": "unknown",
```
JonoYang, line 36: The expected license expression should not be changed.

You may need to run `scancode --reindex-licenses` before regenerating the test results.

JonoYang (changes requested):
@TG1999 

Where will the package_releases field be used? Is it something that scancode toolkit will populate or is it going to be used by another library?

Also, your regenerated test files have issues with some of the expected license_expressions. License expressions that were detected properly as gpl-2.0, gpl-3.0, lgpl-2.1, etc. are now being reported as `unknown`.

pombredanne:
@TG1999 IMHO we should close this and instead have an issue to revisit the topic of packages vs. download URLs.

TG1999:
Added issue https://github.com/nexB/scancode-toolkit/issues/3188

PR #3117, opened 2022-09-26
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/scancode-release.yml
+++ b/.github/workflows/scancode-release.yml
@@ -17,9 +17,13 @@ on:
     tags:
       - "v*.*.*"
 
+permissions: {}
 jobs:
 
   build_scancode_for_pypi:
+    permissions:
+      contents: read # to fetch code (actions/checkout)
+
     name: Build PyPI archives
     runs-on: ubuntu-20.04
 
@@ -67,6 +71,9 @@ jobs:
 
 
   build_scancode_for_release_linux:
+    permissions:
+      contents: read # to fetch code (actions/checkout)
+
     name: Build Release for linux
     runs-on: ubuntu-20.04
     needs:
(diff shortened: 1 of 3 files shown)


pombredanne:
@sashashura Thanks... this looks rather useful!

.github/workflows/scancode-release.yml
```diff
@@ -422,6 +453,9 @@ jobs:
           for %%F in (dist/*.zip) do python etc/release/scancode_release_tests.py dist/%%F
 
   publish_to_gh_release:
+    permissions:
+      contents: write # to create GitHub release (softprops/action-gh-release)
```
pombredanne, line 457: Can you elaborate on what this means?

sashashura: The job requires contents write permission because of the softprops/action-gh-release action in order to create a release. Unfortunately GitHub doesn't have a separate pemission type so it falls into contents write category

AyanSinhaMahapatra:
LGTM! Thanks @sashashura

PR #3121, opened 2022-10-06
fix(nuget): add nuget dependency extract

hellhound-abu:
Fixes #0000

### Tasks

### DIFF
--- a/src/packagedcode/nuget.py
+++ b/src/packagedcode/nuget.py
@@ -27,6 +27,29 @@ def get_dependencies(nuspec):
     dependencies = []
     try:
         if "dependencies" in nuspec:
+            if "group" in nuspec["dependencies"]:
+                for group in nuspec["dependencies"]["group"]:
+                    if "dependency" in group:
+                        for dependency in group["dependency"]:
+                            #dependencies.append(dependency)
+                            dpurl = models.PackageURL(
+                            type='nuget',
+                            namespace=None,
+                            name=dependency.get("@id"),
+                            version=dependency.get("@version"),
+                            qualifiers=None
+                        )
+                            dep_pack = models.DependentPackage(
+                            purl=str(dpurl),
+                            extracted_requirement=dependency.get("@version"),
+                            scope="dependency",
+                            is_runtime=False,
+                            is_optional=False,
+                            is_resolved=True,
+                        )
+
+                            dependencies.append(dep_pack)
+                            
             if "dependency" in nuspec.get("dependencies"):
                 if "@id" and "@version" in nuspec.get("dependencies").get("dependency"):
                     dpurl = models.PackageURL(
--- a/src/packagedcode/nuget.py
+++ b/src/packagedcode/nuget.py
@@ -185,7 +185,7 @@ class NugetNuspecHandler(models.DatafileHandler):
         # Deprecated and not a license expression, just a URL
         elif 'licenseUrl' in nuspec:
             extracted_license_statement = nuspec.get('licenseUrl')
-
+ 
         yield models.PackageData(
             datasource_id=cls.datasource_id,
             type=cls.default_package_type,
@@ -194,17 +194,10 @@ class NugetNuspecHandler(models.DatafileHandler):
             description=description or None,
             homepage_url=nuspec.get('projectUrl') or None,
             parties=parties,
-            extracted_license_statement=extracted_license_statement,
             dependencies=get_dependencies(nuspec),
-            # FIXME: license has evolved and is now SPDX...
-            declared_license=nuspec.get('licenseUrl') or None,
+            extracted_license_statement=extracted_license_statement,
             copyright=nuspec.get('copyright') or None,
             vcs_url=vcs_url,
             **urls,
         )
 
-        if not package_data.license_expression and package_data.declared_license:
-            package_data.license_expression = cls.compute_normalized_license(
-                package_data)
-
-        yield package_data
--- a/src/packagedcode/nuget.py
+++ b/src/packagedcode/nuget.py
@@ -12,10 +12,69 @@ import xmltodict
 from packagedcode import models
 from packagedcode.utils import build_description
 
+
 """
 Handle NuGet packages and their manifests.
 """
-# TODO: add dependencies
+
+# TODO: Add section to handle dependencies of dependencies
+
+
+def get_dependencies(nuspec):
+    """
+    Return a list of Dependent package objects found in a NuGet `nuspec` object.
+    """
+    dependencies = []
+    try:
+        if "dependencies" in nuspec:
+            if "dependency" in nuspec.get("dependencies"):
+                if "@id" and "@version" in nuspec.get("dependencies").get("dependency"):
+                    dpurl = models.PackageURL(
+                        type='nuget',
+                        namespace=None,
+                        name=nuspec.get("dependencies").get(
+                            "dependency").get("@id"),
+                        version=nuspec.get("dependencies").get(
+                            "dependency").get("@version"),
+                        qualifiers=None
+                    )
+                    dep_pack = models.DependentPackage(
+                        purl=str(dpurl),
+                        extracted_requirement=nuspec.get("dependencies").get(
+                            "dependency").get("@version"),
+                        scope="dependency",
+                        is_runtime=False,
+                        is_optional=False,
+                        is_resolved=True,
+                    )
+
+                    dependencies.append(dep_pack)
+
+                else:
+                    for dependency in nuspec.get("dependencies").get("dependency"):
+                        dpurl = models.PackageURL(
+                            type='nuget',
+                            namespace=None,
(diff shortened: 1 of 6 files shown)


AyanSinhaMahapatra:
Thanks! Could you fix the test failures at `tests/packagedcode/test_nuget.py` with:

* `pytest -vvs tests/packagedcode/test_nuget.py` and then
* `SCANCODE_REGEN_TEST_FIXTURES=yes pytest -vvs tests/packagedcode/test_nuget.py --lf`

Also please add your signoff. :bow:

pombredanne:
@hellhound-abu gentle ping.. we would like to merge this... but we cannot unless you update your PR at least with a DCO signoff.

pombredanne:
@6mile ping you seem to be from the same group, may be you can help there?

6mile:
> @6mile ping you seem to be from the same group, may be you can help there?

I've let him know.  Cheers.

jaitaiwan:
Hey folks, Dan from SecureStack here. Abu is on holidays so won’t be able to contribute till 2 weeks from now. I can fix the failing tests if that’s helpful but that won’t help get a DCO from him.

Do you have a way around this or content to wait?

pombredanne:
@jaitaiwan Thank you ++ for your reply! 
No rush,  we can wait alright!
We are just making sure we do not have too many PRs that could go stale from lack of TLC especially like this one which is valuable addition.
 
Whatever you guys are doing on SecureStack looks great BTW! :+1:

pombredanne:
@hellhound-abu gentle ping ... you may be back now?

hellhound-abu:
Updated the branch with necessary changes. If we can merge it whenever possible.

Thanks for waiting.

src/packagedcode/nuget.py
```diff
@@ -12,10 +12,92 @@
 from packagedcode import models
 from packagedcode.utils import build_description
 
+
 """
 Handle NuGet packages and their manifests.
 """
-# TODO: add dependencies
+
+# TODO: Add section to handle dependencies of dependencies
```
pombredanne, line 20: For this you may want to create a separate issue... TODO comments (that I tend to abuse) eventually get lost.
Note also that for a full, dynamic dependency resolution of nugets. there is this new project at https://github.com/nexB/nuget-inspector

src/packagedcode/nuget.py
```diff
@@ -12,10 +12,92 @@
...
+            if "group" in nuspec["dependencies"]:
+                for group in nuspec["dependencies"]["group"]:
+                    if "dependency" in group:
+                        for dependency in group["dependency"]:
+                            #dependencies.append(dependency)
+                            dpurl = models.PackageURL(
+                            type='nuget',
+                            namespace=None,
+                            name=dependency.get("@id"),
+                            version=dependency.get("@version"),
```
pombredanne, line 39: You can only put a version here if this is not a range so IMHO you would have to test for this.

src/packagedcode/nuget.py
```diff
@@ -12,10 +12,92 @@
...
+                            qualifiers=None
+                        )
+                            dep_pack = models.DependentPackage(
+                            purl=str(dpurl),
+                            extracted_requirement=dependency.get("@version"),
+                            scope="dependency",
```
pombredanne, line 45: Would not each dependency group become a scope here?

src/packagedcode/nuget.py
```diff
@@ -12,10 +12,92 @@
...
+                            is_runtime=False,
+                            is_optional=False,
+                            is_resolved=True,
```
pombredanne, line 48: A dependency is resolved if an only if we have a concrete version and not a range. Here the version is optional AFAIK and could be a range?

src/packagedcode/nuget.py
```diff
@@ -12,10 +12,92 @@
...
+                        )
+
+                            dependencies.append(dep_pack)
+                            
+            if "dependency" in nuspec.get("dependencies"):
```
pombredanne, line 53: I would prefer something like this, and then reuse the `nuget_dependency` var below:
```suggestion
            nuget_dependencies = nuspec.get("dependencies") or {}
            nuget_dependency = nuget_dependencies.get("dependency")
            if nuget_dependency:
```

src/packagedcode/nuget.py
```diff
@@ -12,10 +12,92 @@
...
+                        name=nuspec.get("dependencies").get(
+                            "dependency").get("@id"),
+                        version=nuspec.get("dependencies").get(
+                            "dependency").get("@version"),
+                        qualifiers=None
+                    )
+                    dep_pack = models.DependentPackage(
+                        purl=str(dpurl),
+                        extracted_requirement=nuspec.get("dependencies").get(
+                            "dependency").get("@version"),
```
pombredanne, line 67: Factor version here and above in a variable.

src/packagedcode/nuget.py
```diff
@@ -12,10 +12,92 @@
...
+                        scope="dependency",
+                        is_runtime=False,
+                        is_optional=False,
+                        is_resolved=True,
```
pombredanne, line 71: Same comment as above... resolve is True IFF we have a concrete pinned, non-empty version

src/packagedcode/nuget.py
```diff
@@ -12,10 +12,92 @@
...
+                            is_runtime=False,
+                            is_optional=False,
+                            is_resolved=True,
+                        )
+                        dependencies.append(dep_pack)
+
+        return dependencies
+
+    except Exception as e:
+        print(e)
```
pombredanne, line 99: Please do not print exceptions. In general your should avoid to wrap you code in a try/except altogether as this can never be debugged this way. Let the exception bubble up and we will catch it where needed, and worse case at the top level.

src/packagedcode/nuget.py
```diff
@@ -62,14 +144,15 @@ def parse(cls, location):
 
         # Summary: A short description of the package for UI display. If omitted, a
         # truncated version of description is used.
-        description = build_description(nuspec.get('summary') , nuspec.get('description'))
+        description = build_description(
```
pombredanne, line 147: Why wrapping this on tow lines? I am not sure this helps with readability.

tests/packagedcode/data/nuget/bootstrap.nuspec.json.expected
```diff
@@ -1,104 +1,115 @@
...
-    "file_references": [],
-    "extra_data": {},
-    "dependencies": [],
-    "repository_homepage_url": "https://www.nuget.org/packages/bootstrap/4.0.0-alpha",
-    "repository_download_url": "https://www.nuget.org/api/v2/package/bootstrap/4.0.0-alpha",
-    "api_data_url": "https://api.nuget.org/v3/registration3/bootstrap/4.0.0-alpha.json",
-    "datasource_id": "nuget_nupsec",
-    "purl": "pkg:nuget/bootstrap@4.0.0-alpha"
-  }
+          "type": "nuget",
```
pombredanne, line 3: Please avoid changing the spacing if possible as this makes reviewing the diff difficult

pombredanne (changes requested):
@hellhound-abu Thank you ++ for the updates! I posted a few review comments inline for your consideration. There is also one failing test.

src/packagedcode/nuget.py
jaitaiwan: I say the same thing about TODO's internally 😂 That nuget-inspector proejct looks promising.

src/packagedcode/nuget.py
jaitaiwan: Out of curiosity, any reason for `or {}` over `get("...", {})`?

pombredanne: @jaitaiwan re:

> Out of curiosity, any reason for or {} over get("...", {})?

This is an idiom I like to use to make sure that I am getting a proper empty dict in all cases even if the get returns None or some emptyish value.
Getting an empty dict when you expect a dict enable chaining the next call without requiring an extra check for empties.

```Python
>>> repr({1: None, 2: "", 3: {"bar": "baz"}}.get(1, {}))
'None'
>>> repr({1: None, 2: "", 3: {"bar": "baz"}}.get(1, {}) or {})
'{}'
>>> repr({1: None, 2: "", 3: {"bar": "baz"}}.get(1) or {})
'{}'

```

jaitaiwan: Fair call. thanks!

pombredanne:
@hellhound-abu Happy new year! gentle ping

pombredanne:
@AyanSinhaMahapatra since there is no reply for now, let's move this to the next milestone

AyanSinhaMahapatra:
@pombredanne moved this to v32.1

pombredanne:
I have applied the fixes in https://github.com/nexB/scancode-toolkit/pull/3206 on top of your commits. So I am losing this and will merge this other branch!
Thanks!

PR #3135, opened 2022-10-25
Separate Package parsing functions

JonoYang:
This PR updates some code in packagecode for use outside of scancode toolkit. Currently, our package handlers return a generator of PackageData objects when we parse a manifest file. I would like to use the code to get the package object itself, rather than the generator. I've updated the code for the freebsd, haxe, maven, and npm parsers such that the parse method functionality is separated: the parse function for the freebsd, haxe, maven, and npm parsers now yields from another function named `_parse`, which returns a single PackageData object for a given manifest.


pombredanne (approved):
LGTM. Thanks! Merging

PR #3150, opened 2022-11-18
Release 32.0.0rc1 prep

AyanSinhaMahapatra:
### Tasks

* [x] Reviewed [contribution guidelines](https://github.com/nexB/scancode-toolkit/blob/develop/CONTRIBUTING.rst)
* [x] PR is descriptively titled 📑 and links the original issue above 🔗
* [x] Tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR
  Run [tests](https://scancode-toolkit.readthedocs.io/en/latest/contribute/contrib_dev.html#running-tests) locally to check for errors. 
* [x] Commits are in uniquely-named feature branch and has no merge conflicts 📁

PR #3154, opened 2022-11-21
More License Detection changes

AyanSinhaMahapatra:
* Creates a top-level `licenses` attribute with unique license detections
* makes the licenses-reference default, and instead adds a --no-licenses-reference option 

### Tasks


tests/cluecode/data/plugin_email_url/urls-threshold.expected.json
```diff
@@ -1,4 +1,6 @@
 {
+  "license_references": [],
+  "rule_references": [],
```
pombredanne: Rule references is not an obvious, non-ambiguous name for a top level attributes. It would have to be `license_rule_references` or in any case something that very explicitly is about licenses

pombredanne: Also for non-license CLI tests use the --no-license-references option this way the expectation will not change

tests/licensedcode/data/plugin_license/text/scan.expected.json
```diff
@@ -21,69 +183,16 @@
...
-                }
-              ]
+              "rule_url": null
             }
           ]
         }
       ],
       "license_clues": [],
       "percentage_of_license_text": 100.0,
+      "for_licenses": [
```
pombredanne: what is a licenses in this context? is this a license detection? if so we have to be explicit ... and also I wonder if we could come with an id that's a little more friendly than a UUID? may be an expression cleaned with the `commoncode.text.python_safe_name` function and padding a dash and the UUID? And if this is about license detections, then IMHO the top level attribute should be also `license_detections` and not a plain licenses

tests/licensedcode/data/plugin_license/text/scan.expected.json
```diff
@@ -107,39 +216,16 @@
...
-                  "owner": "Free Software Foundation (FSF)",
-                  "homepage_url": "http://www.gnu.org/prep/maintain/html_node/License-Notices-for-Other-Files.html",
-                  "text_url": "",
-                  "reference_url": "https://scancode-licensedb.aboutcode.org/fsf-ap",
-                  "scancode_url": "https://github.com/nexB/scancode-toolkit/tree/develop/src/licensedcode/data/licenses/fsf-ap.LICENSE",
-                  "spdx_license_key": "FSFAP",
-                  "spdx_url": "https://spdx.org/licenses/FSFAP"
-                }
-              ]
+              "rule_url": "https://github.com/nexB/scancode-toolkit/tree/develop/src/licensedcode/data/licenses/fsf-ap.LICENSE"
```
pombredanne: To have reasonably stable URLs will need to update the models to ensure that we never reuse existing rule ids and never rename rules either.... we need to move to a soft-delete mode like with licenses... or could we publish the rules on the license DB too?

tests/licensedcode/data/plugin_license/text/scan.expected.json
```diff
@@ -1,4 +1,166 @@
...
+      "rule_identifier": "spdx-license-identifier: gpl-2.0 WITH linux-syscall-exception-gpl OR linux-openib",
+      "referenced_filenames": [],
+      "is_license_text": false,
+      "is_license_notice": false,
+      "is_license_reference": false,
+      "is_license_tag": true,
+      "is_license_intro": false,
+      "rule_length": 13,
+      "rule_relevance": 100,
+      "matched_text": "SPDX-License-Identifier: ((GPL-2.0 WITH Linux-syscall-note) OR Linux-Openib) */"
```
pombredanne: Is this really it? is this the full rule text?... IMHO the matched text is NOT something related top a rule reference, this is an attribute of a license match, e.g., at the intersection between a rule and a scanned file.

tests/licensedcode/data/plugin_license/text/scan.expected.json
```diff
@@ -1,4 +1,166 @@
...
+      "text": "NOTE! This copyright does *not* cover user programs that use kernel\nservices by normal system calls - this is merely considered normal use\nof the kernel, and does *not* fall under the heading of \"derived work\".\nAlso note that the GPL below is copyrighted by the Free Software\nFoundation, but the instance of code that it refers to (the Linux\nkernel) is copyrighted by me and others who actually wrote it.\n\nAlso note that the only valid version of the GPL as far as the kernel\nis concerned is _this_ particular version of the license (ie v2, not\nv2.2 or v3.x or whatever), unless explicitly otherwise stated.\n\nLinus Torvalds"
+    }
+  ],
+  "rule_references": [
+    {
+      "license_expression": "gpl-2.0 WITH linux-syscall-exception-gpl OR linux-openib",
```
pombredanne: IMHO here we should just dump the rule, as is. The order should be that of https://github.com/nexB/scancode-toolkit/blob/a410f6ebb93a9b22343d7737897b61f3cd2a6ac3/src/licensedcode/models.py#L1773 .... but that order is not correct and should be updated to align exactly with the order of a dump, and the text should come last.

pombredanne (changes requested):
Thanks!
See some comments inline for your review.

tests/cluecode/data/plugin_email_url/urls-threshold.expected.json
AyanSinhaMahapatra: Yes changing it to `license_rule_references`, that's much better.

tests/licensedcode/data/plugin_license/text/scan.expected.json
AyanSinhaMahapatra: Implementing this.

tests/licensedcode/data/plugin_license/text/scan.expected.json
AyanSinhaMahapatra: Yes! we should ensure we don't change rule names (will open an issue for that?), we also have this issue for rules in licenseDB: https://github.com/nexB/scancode-licensedb/issues/3

tests/licensedcode/data/plugin_license/text/scan.expected.json
AyanSinhaMahapatra: yes, this belongs in the match details. Moving it there.

tests/licensedcode/data/plugin_license/text/scan.expected.json
AyanSinhaMahapatra: addressed. Closing this.

AyanSinhaMahapatra:
#3150 now has all the changes from this PR.

tests/licensedcode/data/plugin_license/sqlite/sqlite.expected.json
```diff
@@ -1,4 +1,60 @@
...
+  "license_references": [
+    {
+      "key": "blessing",
+      "short_name": "SQLite Blessing",
+      "name": "SQLite Blessing",
+      "category": "Public Domain",
+      "owner": "SQLite",
+      "homepage_url": "https://sqlite.org/src/artifact/df5091916dbb40e6",
+      "is_builtin": true,
+      "spdx_license_key": "blessing",
```
pombredanne, line 34: I wonder if we should not force including all the attributes even if empty? That's been the standard for how we report things in SCTK so far

pombredanne: This can wait for later

pombredanne (approved):
LGTM! merging!

tests/licensedcode/data/plugin_license/sqlite/sqlite.expected.json
AyanSinhaMahapatra: @pombredanne do you mean the `is_builtin` flag? That is the only attribute we are adding whenever there are extra licenses installed/added to the index, our previous discussion was we add other attributes too based on CLI options/other optional elements. But apart from this all other attributes are always added to the reference. Note that we have more attributes added on tge release prep branch.

pombredanne: I meant all the attributes that are empty. For instance in https://github.com/nexB/scancode-toolkit/blob/f1cc917e002d4269869ffefa47acb1bdfb3d2d56/src/licensedcode/models.py#L393 we do not report these

pombredanne: I entered https://github.com/nexB/scancode-toolkit/issues/3182

Issue #2738, opened 2021-10-14
Add a command line option to dump the license data


fviernau:
The OSS Review Toolkit needs to have all license IDs along with the respective license text for the ScanCode version it uses.
Currently it extracts it from the license data directory, see e.g. [1] [2]. So it depends on the directory structure as well as on the file formats of ScanCode license data. A ScanCode command line option for dumping

1. licenseIDs
2. corresponding texts
3. corresponding license category (not needed now, only maybe nice to have IMO to enable using the categories within ORT by implementing a conversion to [3])

..to a JSON? file would help to remove that mentioned dependency.

[1] https://github.com/oss-review-toolkit/ort/blob/093a510aa8545cc61cc172d7a7a274f56df87d84/spdx-utils/build.gradle.kts#L62-L71
[2] https://github.com/oss-review-toolkit/ort/blob/093a510aa8545cc61cc172d7a7a274f56df87d84/spdx-utils/build.gradle.kts#L238-L343
[3] https://github.com/oss-review-toolkit/ort/blob/093a510aa8545cc61cc172d7a7a274f56df87d84/model/src/main/kotlin/licenses/LicenseClassifications.kt

AyanSinhaMahapatra:
There are two options:

 - We have a command line option (I was thinking something like  --get-license-data folder_path) in SCTK to:
Just generate the .json/.yml files used to generate the LicenseDB pages and have validation tests (this way we have the sort of API for license data and other tools like ORT can consume and not read the actual License files)
- Do everything in option 1. plus we actually generate the pages (.html files and the index.html) too locally (this would mean adding almost everything in the repository scancode-licensedb into scancode-toolkit, history preserved of course) i.e. move the code into a licensedcode file and the template/static also there.

@pombredanne :

IMHO there are  different things
- SCTK shall be able to generate a dump that would include JSON/YML and HTML  (with more or less flexbility TBD, but a single option for now is fine). The output should be exactly like the  scancode-licensedb output
- The  scancode-licensedb repo will stay as is, minus the generation code and would instead use the the CLI option in 1. to create itself (and the history stay in this repo entirely)

AyanSinhaMahapatra:
This is in develop now, and will be released in a release-candidate soon, @fviernau can you try it out. More details in the PR linked above ^ .

pombredanne:
@AyanSinhaMahapatra I think this is all done now? correct? If so please close.

AyanSinhaMahapatra:
[...]

PR #3197, opened 2023-01-11
Update license db generation

AyanSinhaMahapatra:
Reference: #2738 

* Removes the CLI option `--dump-license-data`
* add a new console script `scancode-license-data-dump`
* Removes is_builtin attribute from details
* Removes the text attribute as there already was a license_text displayed so this wasn't necessary

### Tasks


tests/licensedcode/data/license_db/license_dump/bison-exception-2.0.json
```diff
@@ -4,10 +4,8 @@
   "name": "Bison exception to GPL 2.0 or later",
   "category": "Copyleft Limited",
   "owner": "Free Software Foundation (FSF)",
-  "is_builtin": true,
```
pombredanne, line 7: Why removing this?

src/licensedcode/license_db.py
```diff
@@ -132,7 +136,7 @@ def generate_details(output_path, environment, licenses, test=False):
         base_context_mapping = base_context
     license_details_template = environment.get_template("license_details.html")
     for lic in licenses.values():
-        license_data = lic.to_dict(include_text=True)
+        license_data = lic.to_dict(include_text=False, include_builtin=False)
```
pombredanne: Why skipping "is_builtin"?
If you dump the licenses would not the builtin and other licenses be included?

src/licensedcode/license_db.py
```diff
@@ -200,19 +204,38 @@ def generate(
...
+    click.secho(f'Dumping license data to: {path}', err=True)
+    count = generate(build_location=path)
     click.secho(f'Done dumping #{count} licenses.', err=True)
-    ctx.exit(0)
+
+
+@click.command(name='scancode-license-data')
+@click.option(
+    '--path',
+    type=click.Path(exists=False, readable=True, file_okay=False, resolve_path=True, path_type=str),
```
pombredanne:
```suggestion
    type=click.Path(exists=False, writable=True, file_okay=False, resolve_path=True, path_type=str),
```

src/licensedcode/license_db.py
```diff
@@ -200,19 +204,38 @@ def generate(
...
+    help_group=MISC_GROUP,
+    cls=PluggableCommandLineOption,
+)
+@click.help_option('-h', '--help')
+def dump_scancode_license_data(
+    path,
+    *args,
+    **kwargs,
+):
+    """Reindex scancode licenses and exit"""
```
pombredanne: Are you sure this is a correct docstring?

pombredanne (changes requested):
Some nit pickings for your review.
Thanks!

tests/licensedcode/data/license_db/license_dump/bison-exception-2.0.json
AyanSinhaMahapatra: Because all licenses in the licenseDB is builtin by default, so this is redundant information, at least in this scope.

src/licensedcode/license_db.py
AyanSinhaMahapatra: Atleast in our case, we don't have any licenses that isn't builtin. The builtin is defined as licenses that aren't a part of licensedb. And this was just adding the `is_builtin` flag and this information is redundant.

But if we are thinking about other users we probably should just make including other licenses with builtin licenses when dumping.

So we can do two things:

1. We can add a command line option flag to facilitate our case, i.e. not showing the is_builtin flag, as everything hosted in licenseDB is builtin license by definition of the flag?
2. We can do the same thing we do in our license reference plugin, we check if there are non-builtin licenses installed, and if so then only we use the is_builtin flag.

src/licensedcode/license_db.py
AyanSinhaMahapatra: Thanks, adding this.

src/licensedcode/license_db.py
AyanSinhaMahapatra: Oops, missed this. Fixing.

src/licensedcode/license_db.py
AyanSinhaMahapatra: I think 2. is the one we want. Less command line options, less confusion. Just doing what should be done?

pombredanne: 2. works, even though from an API standpoint having fields present in some cases and not other case is not the best case. Go with 2.

AyanSinhaMahapatra: > even though from an API standpoint having fields present in some cases and not other case is not the best case

We discussed this in the case of license-references, and we concluded that we also do this everywhere in scancode-toolkit where attributes are dependent on what CLI options are being used, just in this case it is the state of license index and CLI commands used in past.

AyanSinhaMahapatra: I've already updated the PR with 2. btw, ready for review.

pombredanne (approved):
LGTM!

PR #3218, opened 2023-01-23
Add required phrase to JSR rule

bennati:
Avoids ambiguity with other rules

Signed-off-by: Bennati, Stefano <stefano.bennati@here.com>


pombredanne (approved):
Thanks! LGTM.

zhernovs:
Hi @pombredanne
Is there any chance to include this into some 31.* patch? 
We would love to use newer version of scancode (currently on 30.1.0) together with ORT. But [ORT still doesn't support new output format version](https://github.com/oss-review-toolkit/ort/issues/6617) that you have in 32.* On the other hand we can't use 31.* without a fix from this commit. Perhaps another patch version (31.2.5) won't hurt.

PR #3220, opened 2023-01-23
Get valid yaml output

### LINKED ISSUE #3219: scancode license scan produces invalid yaml
### Description

> Please leave a brief description of the bug or feature request:

scancode license scan now has license references by default, and as license texts
often has whitespace to the left, this leads to invalid yaml output when `--license`
option is used.

### How To Reproduce

> Tell us how to reproduce the issue. 

From scancode latest develop:

```
./configure --clean && ./configure --dev
scancode -clipeu --classify --summary samples --yaml output.yaml
python
> import saneyaml
> path = "path/to/output.yaml"
>  data = saneyaml.load(open(path).read())
```

This results in:
```
  File "<stdin>", line 1, in <module>
  File "/home/ayansinha/nexB/write_access/scancode-toolkit/venv/lib/python3.8/site-packages/saneyaml.py", line 62, in load
    return yaml.load(s, Loader=loader)
  File "/home/ayansinha/nexB/write_access/scancode-toolkit/venv/lib/python3.8/site-packages/yaml/__init__.py", line 81, in load
    return loader.get_single_data()
  File "/home/ayansinha/nexB/write_access/scancode-toolkit/venv/lib/python3.8/site-packages/yaml/constructor.py", line 49, in get_single_data
    node = self.get_single_node()
  File "yaml/_yaml.pyx", line 673, in yaml._yaml.CParser.get_single_node
  File "yaml/_yaml.pyx", line 687, in yaml._yaml.CParser._compose_document
  File "yaml/_yaml.pyx", line 731, in yaml._yaml.CParser._compose_node
  File "yaml/_yaml.pyx", line 845, in yaml._yaml.CParser._compose_mapping_node
  File "yaml/_yaml.pyx", line 729, in yaml._yaml.CParser._compose_node
  File "yaml/_yaml.pyx", line 806, in yaml._yaml.CParser._compose_sequence_node
  File "yaml/_yaml.pyx", line 731, in yaml._yaml.CParser._compose_node
  File "yaml/_yaml.pyx", line 847, in yaml._yaml.CParser._compose_mapping_node
  File "yaml/_yaml.pyx", line 860, in yaml._yaml.CParser._parse_next_event
yaml.parser.ParserError: while parsing a block mapping
  in "<unicode string>", line 561, column 9
did not find expected key
  in "<unicode string>", line 600, column 40
```

### System configuration

* What OS are you running on? (Windows/MacOS/Linux) ubuntu 20.04
* What version of scancode-toolkit was used to generate the scan file? 32.0.0rc1
* What installation method was used to install/run scancode? (pip/source download/other) source checkout

AyanSinhaMahapatra:
Reference: https://github.com/nexB/saneyaml/issues/9

Fixes #3219 

* Make license texts YAML safe
* Modify License.dump() to dump yaml safe text
* Add tests to verify yaml output validity
* Add checks to License.validate() for yaml validity of text
 
### Tasks


pombredanne:
We need to find a better way in saneyaml IMHO

AyanSinhaMahapatra:
@pombredanne License texts were producing invalid YAML files when there were
different indentations in the text along with empty newlines (only when both was
present, and not just empty newlines), and to make these texts YAML safe we add
a single space to empty newlines to license texts with different indentation levels.

So we update only those which were producing invalid YAML, so we are updating only a small number of files, when compared to updating all files with empty newlines.

JonoYang (changes requested):
@AyanSinhaMahapatra Looks good! I think there should be a new entry in the changelog for this fix. 

On another note, I will also have to be careful not to remove those spaces on the newlines when I open the license files you edited, as I have a vscode option enabled that removes spaces from newlines on save.

src/licensedcode/models.py
```diff
@@ -658,6 +662,12 @@ def validate(licenses, verbose=False, no_dupe_urls=False):
             # local text consistency
             text = lic.text
 
+            data = {"text": text}
+            # We are testing whether we can dump as yaml and load from yaml
+            # without failing (i.e. whether the text is yaml safe) 
+            yaml_string = yaml_dump(data, indent=4)
+            loaded_yaml = yaml_load(yaml_string)
```
pombredanne: You should:
- test the load() also with the plain yaml library load.
- validate that the reloaded yaml is the same as the `data` in both cases, at least for the text, since that's where the problem exists.
- capture the exception and report these as errors the same way other errors are reported

src/licensedcode/models.py
```diff
@@ -737,6 +747,17 @@ def validate(licenses, verbose=False, no_dupe_urls=False):
         return errors, warnings, infos
 
 
+def get_yaml_safe_text(text):
+
+    data = {"text": text}
```
pombredanne, line 769: IMHO you could do simpler and more efficient as the yaml load/dump is expensive?
What about this instead, which should also be included first in saneyaml IMHO?
On the other hand you may be fine with your approach since this expensive dump/load is guaranteed to be always correct AND is rarely used, only when we dump license or rules... not too often
```python
def pad_empty_lines(text):
   """
   Return the text with empty lines padded with a space
   to be YAML-safe.
   >>> text = "\n\nfoo\n\n\ndada\n \n\n"
   >>> expected = " \n \nfoo\n \n \ndada\n \n \n"
   >>> assert pad_empty_lines(text) == expected
   """
   ends_with_new_lines = text.endswith("\n")
   text = "\n".join([line or " " for line in text.splitlines(False)])
   if ends_with_new_lines:
       text += "\n"
   return text

```

src/licensedcode/models.py
```diff
@@ -479,6 +481,8 @@ def dump(self, licenses_data_dir):
         """
         metadata = self.to_dict(include_builtin=False)
         content = self.text
+        # We add a space to empty newlines to make this text YAML safe
+        content = get_yaml_safe_text(content)
```
pombredanne, line 485: What about rules? there must be a good bunch of these that are also incorrect?

pombredanne (changes requested):
Thanks! see some comments inline for your consideration.

pombredanne:
You also have a failiing test: "FAILED tests/formattedcode/test_output_yaml.py::test_scan_produces_valid_yaml"

src/licensedcode/models.py
AyanSinhaMahapatra: For some reason rules don't fail. This could be from a whitespace stripping operation applied to the left of the text previously (the rules don't seem to have first line indented)

src/licensedcode/models.py
AyanSinhaMahapatra: We are doing more than this and hence it is expensive. The YAML load was not failing just because of the empty newlines but: both empty newlines and different indentation levels (?) being present. So we only pad newlines with space on failing load(). Also considering other options here.

src/licensedcode/models.py
AyanSinhaMahapatra: 1. If we test also for load() with plain library we are getting [scancode.lockfile.LockTimeout error](https://dev.azure.com/nexB/scancode-toolkit/_build/results?buildId=10460&view=logs&j=56739ecf-8d40-52c7-9da4-d79b306f05c9&t=c0917742-18c4-555f-7c19-b4b94cb323a4&l=13642) but the tests pass so this works allright.
2. validate that the reloaded yaml is the same as the data in both cases: this fails because `saneyaml.dump()` strips adds/newlines, I looked a bit into saneyaml but it is not obvious how to fix this.
3. capture the exception and report these as errors: Done!

src/licensedcode/models.py
AyanSinhaMahapatra: Added tests for this too.

src/licensedcode/models.py
AyanSinhaMahapatra: Fixed these by moving the validate outside the index file lock, and also making sure we don't run any of the `thorough` load/dump checks while creating the index. But perform these expensive operations only on the test side.

AyanSinhaMahapatra: Tracking the point 2. here in another issue separately as this is a saneyaml issue: https://github.com/nexB/saneyaml/issues/10

AyanSinhaMahapatra:
@pombredanne all green, ready for your review again!

src/licensedcode/cache.py
```diff
@@ -130,32 +130,32 @@ def load_or_build(
 
         lock_file = os.path.join(scancode_cache_dir, LICENSE_LOCKFILE_NAME)
 
+        additional_directories = []
```
pombredanne: I reviewed this more and I think we have these states:
1. index loading of a prebuilt index: we should not do any validation on rules and licenses. This is what happens on https://github.com/nexB/scancode-toolkit/pull/3220/files#diff-51085e58e72e83ba456876b9fd68b26c1e343417de07bb82a29c3e49ce4c7d45R112 alright

2.1. index building with validation: there we MUST do the validation inside a lock. Otherwise we would have a possible scan that starts on 5 processes and there is no index. Each of the 5 processes would do the validation separately. There you need to increase the LICENSE_INDEX_LOCK_TIMEOUT value in https://github.com/nexB/scancode-toolkit/pull/3220/files#diff-51085e58e72e83ba456876b9fd68b26c1e343417de07bb82a29c3e49ce4c7d45R32 this is 4 minutes for now. We should increase to 6 minutes

2.2 index building and validation separated: in this alternative approach we would separate validation in a new CLI command, and index building WOULD not require validation and may fail, and we would have to assume that validation is done on request beforehand

AyanSinhaMahapatra: > there we MUST do the validation inside a lock. Otherwise we would have a possible scan that starts on 5 processes and there is no index. Each of the 5 processes would do the validation separately.

Ah! I understand now, thanks for the explanation @pombredanne , I've reverted back to validating inside the lock, we still don't validate the YAML load/dump here anyway so this should be fine. 

> index building and validation separated:

Seems more complicated, confusing and would require more changes. I like the process we have now much more. Unless this has some good benifits it seems we should continue using our current approach?

license-yaml-update.ipynb
```diff
@@ -0,0 +1,661 @@
+{
```
pombredanne: I think this was committed by mistake?

src/licensedcode/models.py
```diff
@@ -655,8 +662,22 @@ def validate(licenses, verbose=False, no_dupe_urls=False):
...
-            # local text consistency
             text = lic.text
+            if thorough:
+                # local text consistency
+                data = {"text": text}
+                # We are testing whether we can dump as yaml and load from yaml
+                # without failing (i.e. whether the text is yaml safe)
+                # Using saneyaml
+                try:
+                    yaml_string = saneyaml_dump(data, indent=4)
```
pombredanne: Why do you use indent 4 as opposed to the default?

src/licensedcode/models.py
```diff
@@ -655,8 +662,22 @@ def validate(licenses, verbose=False, no_dupe_urls=False):
...
+                    loaded_yaml = saneyaml_load(yaml_string)
+                except Exception:
+                    errors['GLOBAL'].append(
+                        f'Error invalid YAML text at: {lic.key}, failed during saneyaml.load()'
```
pombredanne, line 674: Can you find a more explicit error message that would hint on what to do to fix this?

src/licensedcode/models.py
```diff
@@ -737,6 +758,17 @@ def validate(licenses, verbose=False, no_dupe_urls=False):
         return errors, warnings, infos
 
 
+def get_yaml_safe_text(text):
+
```
pombredanne, line 764: Add some docstring that explains what this does and why, as this is a key item IMHO

src/licensedcode/models.py
```diff
@@ -805,6 +837,7 @@ def get_rules(
     licenses_data_dir=licenses_data_dir,
     rules_data_dir=rules_data_dir,
     validate=False,
+    validate_thorough=False,
```
pombredanne, line 846: Is there any case where it makes sense NOT to validate thorough now that we have an increased timeout? Otherwise explain when we do thorough vs. non thorough

license-yaml-update.ipynb
AyanSinhaMahapatra: :sweat: my bad, must have been not careful during amending.

src/licensedcode/models.py
AyanSinhaMahapatra: removing this, unintended.

src/licensedcode/models.py
AyanSinhaMahapatra: Done! Also added a new option ``--load-dump`` to the ``scancode-reindex-licenses`` script to do this.

src/licensedcode/models.py
AyanSinhaMahapatra: Added, thanks!

src/licensedcode/models.py
AyanSinhaMahapatra: It does not make sense to do this expensive validation process every time we reindex/create index as this increases the time. We are instead choosing to do this only while testing and this should be sufficient to prevent having license texts that would produce invalid YAML.

AyanSinhaMahapatra:
@pombredanne addressed comments, ready for review again!

docs/source/cli-reference/other-commands.rst
```diff
@@ -100,3 +100,11 @@ reindex the licenses without these licenses from the additional plugins.
 
 Rebuild the license index including texts all languages (and not only
 English) and exit. This is an EXPERIMENTAL option.
+
+
+``--load-dump`` Option
```
pombredanne, line 105: :heart:

pombredanne (approved):
LGTM! :+1: Thanks

PR #3252, opened 2023-02-14
Update doc to reference attrib in AbcTK

chinyeungli:
Only updated the documentation to reference attribution generation from a ScanCode scan with AboutCode Toolkit.

Signed-off-by: Chin Yeung Li <tli@nexb.com>


pombredanne (approved):
Thanks!

PR #3254, opened 2023-02-15
Add required phrase rules automatically

AyanSinhaMahapatra:
Reference: #2637 #2878 

### Tasks


src/licensedcode/models.py
```diff
@@ -2844,83 +2896,39 @@ def find_rule_base_location(name_prefix, rules_directory=rules_data_dir):
...
+    from a dictionary of rules by their identifier.
+    """
+    rules_by_expression = defaultdict(list)
+
+    for rule in rules_by_identifier.values():
+        # Only return rules with license_expression (i.e. skip false positives)
+        if rule.license_expression:
+            rules_by_expression[rule.license_expression].append(rule)
+
+    return OrderedDict(sorted(rules_by_expression.items()))
```
pombredanne: You can use a plain dict to the same effect.

src/licensedcode/models.py
```diff
@@ -1428,13 +1425,33 @@ class BasicRule:
             'Mutually exclusive from any is_license_* flag')
     )
 
+    is_required_phrase = attr.ib(
+        default=False,
+        repr=False,
+        metadata=dict(
+            help='True if this is rule text is a required phrase '
```
pombredanne:
```suggestion
            help='True if this is rule text is a required phrase. '
```

src/licensedcode/models.py
```diff
@@ -1428,13 +1425,33 @@ class BasicRule:
...
+            'A required phrase is often a part of another larger rule text '
+            'but is an essential section of the rule text which must be '
+            'present in the case of partial matches, otherwise the match '
+            'will be a false positive and misleading. ')
```
pombredanne:
```suggestion
            'will be a false positive and misleading.')
```

src/licensedcode/models.py
```diff
@@ -1428,13 +1425,33 @@ class BasicRule:
...
+    )
+
+    skip_creating_required_phrases = attr.ib(
+        default=False,
+        repr=False,
+        metadata=dict(
+            help='True if this rule needs to be skipped while creating '
+            'required phrase rules. Required phrase rules are created out '
+            'of other rule texts which have marked required phrases. ')
```
pombredanne:
```suggestion
            'of other rule texts which have marked required phrases.')
```

src/licensedcode/models.py
```diff
@@ -1796,22 +1813,27 @@ def has_unknown(self):
...
+    @property
+    def license_flags(self):
+        return (
             self.is_license_notice,
             self.is_license_text,
             self.is_license_reference,
             self.is_license_tag,
             self.is_license_intro,
             self.is_license_clue,
+            self.is_required_phrase,
```
pombredanne: IMHO this is not a license flag

src/licensedcode/models.py
```diff
@@ -1796,22 +1813,27 @@ def has_unknown(self):
...
 
-    def validate(self, licensing=None, thorough=False):
-        """
-        Validate this rule using the provided ``licensing`` Licensing and yield
-        one error message for each type of error detected.
-        """
-        is_false_positive = self.is_false_positive
-
-        license_flags = (
+    @property
```
pombredanne, line 1817: If is_required_phrase is not a flag, then the update to validate may no longer be needed?

src/licensedcode/models.py
```diff
@@ -1428,13 +1425,33 @@ class BasicRule:
...
+        repr=False,
+        metadata=dict(
+            help='True if this is rule text is a required phrase '
+            'A required phrase is often a part of another larger rule text '
+            'but is an essential section of the rule text which must be '
+            'present in the case of partial matches, otherwise the match '
+            'will be a false positive and misleading. ')
+    )
+
+    skip_creating_required_phrases = attr.ib(
```
pombredanne: What about instead: skip_collecting_required_phrases

src/licensedcode/tokenize.py
```diff
@@ -113,6 +115,117 @@ def key_phrase_tokenizer(text, stopwords=STOPWORDS):
...
+    combined_text = ''
+    
+    for token_tuple in token_tuples:
+        _value, token = token_tuple
+        combined_text += token
+        
+    return combined_text
+
+
+def add_key_phrase_markers(text, key_phrase_span):
```
pombredanne: Use only one of key_phrases or required_phrases

pombredanne (commented):
Here are some comments for your consideration.

src/licensedcode/models.py
AyanSinhaMahapatra: Thanks, resolved.

src/licensedcode/models.py
AyanSinhaMahapatra: Added, thanks!

src/licensedcode/models.py
AyanSinhaMahapatra: Added, thanks!

src/licensedcode/models.py
AyanSinhaMahapatra: Added, thanks!

src/licensedcode/models.py
AyanSinhaMahapatra: Removed, thanks!

src/licensedcode/models.py
AyanSinhaMahapatra: Used elsewhere to make things compact, keeping this.

src/licensedcode/models.py
AyanSinhaMahapatra: This is better, added this, thanks!

src/licensedcode/tokenize.py
AyanSinhaMahapatra: Renaming `key phrase` -> `required phrase` everywhere as this is more verbose.

src/licensedcode/tokenize.py
```diff
@@ -103,17 +105,133 @@ def key_phrase_tokenizer(text, stopwords=STOPWORDS):
...
+    """
+    Given a string `text` and a Span object `required_phrase_span`, add required phrase
+    markers to the `text` around the tokens which the span represents, while
+    being mindful of whitespace and stopwords.
+    """
+    tokens_tuples_without_markers = list(matched_query_text_tokenizer(text))
+    tokens_tuples_with_markers = []
+    token_index = 0
+
+    for token_tuple in tokens_tuples_without_markers:
```
pombredanne: What do you gain from an intermediate list?
What about this?
```suggestion
    for token_tuple in matched_query_text_tokenizer(text):
```

src/licensedcode/tokenize.py
```diff
@@ -103,17 +105,133 @@ def key_phrase_tokenizer(text, stopwords=STOPWORDS):
...
+    Returns a string `combined_text` combining token tuples from the list `token_tuples`,
+    which are token tuples created by the tokenizer functions.
+    """
+    combined_text = ''
+    
+    for token_tuple in token_tuples:
+        _value, token = token_tuple
+        combined_text += token
+        
+    return combined_text
```
pombredanne: What about this, without intermediate accumulation of strings?
```suggestion
    return ''.join(token for _, token in token_tuples)
```

src/licensedcode/tokenize.py
```diff
@@ -103,17 +105,133 @@ def key_phrase_tokenizer(text, stopwords=STOPWORDS):
...
+
+    The list of spans `old_required_phrase_spans` contains all the spans of required
+    phrases or ignorables already present in a rule text, and the other list of spans
+    `new_required_phrase_spans` contains the proposed new required phrases.
+    """
+    if not old_required_phrase_spans:
+        return new_required_phrase_spans
+    
+    for new_span in new_required_phrase_spans:
+        if not any(
```
pombredanne: This is a double negative... what about:
```suggestion
        if any(old_span.overlap(new_span) for old_span in old_required_phrase_spans:
            continue
```
or 
```suggestion
        if not any(old_span.overlap(new_span) for old_span in old_required_phrase_spans:
            yield ...
```

src/licensedcode/models.py
```diff
@@ -1796,24 +1814,37 @@ def has_unknown(self):
...
+
+    @property
+    def license_flags(self):
+        return {
+            license_flag_name: getattr(self, license_flag_name)
+            for license_flag_name in self.license_flag_names
+        }
+
+    @property
+    def license_flag_values(self):
```
pombredanne: since this is used only twice, is this needed?

src/licensedcode/models.py
```diff
@@ -1866,6 +1897,12 @@ def validate(self, licensing=None, thorough=False):
             if not check_is_list_of_strings(self.referenced_filenames):
                 yield 'referenced_filenames must be a list of strings'
 
+            if (self.is_license_clue or self.is_license_intro) and self.is_required_phrase:
```
pombredanne, line 1896: When is this an issue?

src/licensedcode/models.py
```diff
@@ -1796,24 +1814,37 @@ def has_unknown(self):
...
-            self.is_license_reference,
-            self.is_license_tag,
-            self.is_license_intro,
-            self.is_license_clue,
-        )
-
-        has_license_flags = any(license_flags)
-        has_many_license_flags = len([l for l in license_flags if l]) != 1
+        has_license_flags = any(self.license_flag_values)
+        has_many_license_flags = len([l for l in self.license_flag_values if l]) > 1
```
pombredanne: This is minor, but what about this if you remove the self.license_flag_values method above?
```suggestion
        license_flags_count = sum(self.license_flag.values()
        has_many_license_flags = license_flags_count > 1
        has_license_flags = boolean(license_flags_count)

```

src/licensedcode/models.py
```diff
@@ -2260,6 +2291,8 @@ def dump(self, rules_data_dir):
         rule_file = self.rule_file(rules_data_dir=rules_data_dir)
 
         metadata = self.to_dict()
+        if kwargs:
```
pombredanne, line 2292: Can you explain in a comment this is used only in debug and possibly prefix kwargs with an underscore?

src/licensedcode/models.py
```diff
@@ -2291,6 +2324,9 @@ def load(self, rule_file, with_checks=True):
             raise e
 
         known_attributes = set(attr.fields_dict(self.__class__))
+        # This is an attirbute used to debug marking required phrases, and is not needed
```
pombredanne:
```suggestion
        # This is an attribute used to debug marking required phrases, and is not needed
```

src/licensedcode/models.py
```diff
@@ -1428,13 +1424,35 @@ class BasicRule:
...
+    is_required_phrase = attr.ib(
+        default=False,
+        repr=False,
+        metadata=dict(
+            help='True if this is rule text is a required phrase. '
+            'A required phrase is often a part of another larger rule text '
+            'but is an essential section of the rule text which must be '
+            'present in the case of partial matches, otherwise the match '
+            'will be a false positive and misleading.')
+    )
```
pombredanne, line 1436: I wonder if this would not be best computed an index time ... this is True if a rule text is entirely a required phrase, this would mean one less stored flag.

src/licensedcode/models.py
```diff
@@ -1428,13 +1424,35 @@ class BasicRule:
             'Mutually exclusive from any is_license_* flag')
     )
 
+    is_required_phrase = attr.ib(
+        default=False,
+        repr=False,
+        metadata=dict(
+            help='True if this is rule text is a required phrase. '
```
pombredanne, line 1431: I am still unsure about this field... is this when the whole rule text is a required phrase? Because looking at this again now, I wonder if this would not be simpler to generate the flag when we index a rule based on this fact.

pombredanne (commented):
Some minor nits and a question on the is_required_phrase flag for your consideration!

src/licensedcode/models.py
AyanSinhaMahapatra: There are a couple reasons why we are doing it this way, explicitly with a flag, instead of computing these on the fly:

1. We are not generating these required phrase rules while loading the index from the rules, this is instead an one time operation and a thing that only runs on tests or new rule additions, where we go thorough the rules collect required phrases and mark them in other rules (while also saving them, by creating a new file, or already marking an existing rule with the flag). If we start generating this on the fly while creating the index we are making the index creation step more expensive (we now only collect required phrases from a single rule, but if we do this we have to also add them as rules to the main index) and it will take even more time.
2. Today the process to create required phrase rules is not fully automated at first, we need to review all the newly created required phrase rules and then mark some origin rules with `skip_collecting_required_phrases: yes` so that some rule-specific required phrases are not created as required phrases. This process will become much more abstract and non-intuitive, and would be much more difficult to debug.
3. The license detection process update will also use these new required phrase rules. In cases where we don't have good matches, we will try to do a license detection just filtering required phrase rules from all the matches, and this would involve looking at all matches which has `is_required_phrase` as True, therefore making this flag necessary.

src/licensedcode/models.py
AyanSinhaMahapatra: This is to prevent marking of license clues or license intro rules as required phrases as by definition these two types of rules are either non-specific intros or weak clues as rules, which cannot be considered required phrases for any specific license expressions.

src/licensedcode/models.py
pombredanne: OK.Thanks!

pombredanne (approved):
Looking fine!

AyanSinhaMahapatra:
Closing this as we now have https://github.com/aboutcode-org/scancode-toolkit/pull/3924

PR #3286, opened 2023-03-08
Improve License Detection reporting

### LINKED ISSUE #3266: License detection_log:  position and "- not-combined" values are eventually confusing
The detection log should be positioned at the bottom not at the top in the JSON/YAML outputs
But the value "not-combined" is puzzling and not obvious. It is further more cofusing when looking this kind of results:
```
    -   identifier: mit_and_gpl_2_0-75d8c570-6ad2-0cbe-e6b5-432c94d552a9
        license_expression: mit AND gpl-2.0
        count: 1
        detection_log:
            - not-combined
        matches:
            -   score: '100.0'
                start_line: 5
                end_line: 5
                matched_length: 3
                match_coverage: '100.0'
                matcher: 2-aho
                license_expression: mit
                rule_identifier: mit_172.RULE
                rule_url: https://github.com/nexB/scancode-toolkit/tree/develop/src/licensedcode/data/rules/mit_172.RULE
            -   score: '100.0'
                start_line: 7
                end_line: 7
                matched_length: 2
                match_coverage: '100.0'
                matcher: 2-aho
                license_expression: gpl-2.0
                rule_identifier: gpl-2.0_818.RULE
                rule_url: https://github.com/nexB/scancode-toolkit/tree/develop/src/licensedcode/data/rules/gpl-2.0_818.RULE
```
Here two licenses matches have been obviously combined in a single expression in the resulting license detection. Yet this is reported as "not combined"

### LINKED ISSUE #3269: RFC: Consider reintroducing the --license-reference option
After some good usage of the latest RC, we should IMHO consider reintroducing the --license-reference option and NOT always include this data in all scans. With ` --license-reference` both rules and license reference would be included

AyanSinhaMahapatra:
Fixes #3266 
Fixes #3269 

This PR improves the License Detection output with important and cosmetic changes:

- Remove 'not-combined' from the detection log because it is possibly confusing #3266
- Move detection log after matches in the JSON output #3266
- Add rule relevance to the match details to JSON output
- Rename count to detection count in the JSON output
- Show detection_log only when the  --license-text-diagnostics option is requested
- Make license references a command line option, do not include them by default #3269
- Update test expectations accordingly

### Tasks


tests/licensedcode/data/additional_licenses/additional_license_combined_test.expected.json
```diff
@@ -28,6 +26,7 @@
           "matcher": "2-aho",
           "license_expression": "example-installed-2",
           "rule_identifier": "example-installed-2.LICENSE",
+          "rule_relevance": 100,
```
pombredanne: IMHO we should also inline the rule length here, as the combo of 
"score", "matched_length", "match_coverage", "matcher", "rule_relevance" and "rule_length" are what matters to diagnostic detection issues (at least that's what I use)

tests/licensedcode/data/additional_licenses/additional_license_combined_test.expected.json
```diff
@@ -386,9 +388,6 @@
       "license_detections": [
         {
           "license_expression": "example-installed-1 AND example-installed-2 AND example1 AND example2 AND apache-2.0",
-          "detection_log": [
-            "not-combined"
-          ],
```
pombredanne, line 391: Even if empty, "detection_log" should be there always.

pombredanne (changes requested):
LGTM overall. Thanks!
Let's have a chat at your convenience though as I think we can do better wrt. inlining or not inlining license matches.

tests/licensedcode/data/additional_licenses/additional_license_combined_test.expected.json
AyanSinhaMahapatra: Refer https://github.com/nexB/scancode-toolkit/issues/3266#issuecomment-1443672136, we have moved detection log to only be there with `--license-text-diagnostics` as we don't want this always.

tests/licensedcode/data/additional_licenses/additional_license_combined_test.expected.json
AyanSinhaMahapatra: The initial idea was, everything that is detection related (example: score, match_coverage) stays here in the match level, everything that is a rule attribute (example: rule_relevance, rule_length) is moved to the license references. Now there are two options:

1.  add a CLI option `inline` which would have all these details, which would facilitate detection diagnostics. 

2. Not have this at all, and we add all the details in the `--review` options, and we have the details available with `--license-reference` anyway.

I think we want to move towards workbench/scancode.io license text to diagnose licence detection issues, and this makes the case where we want to have these details inlined by default a very tiny minority, not sure we want to move away from our general direction to support this. 

What do yoh think @pombredanne?

tests/licensedcode/data/additional_licenses/additional_license_combined_test.expected.json
AyanSinhaMahapatra: Well in the comment it's `--license-diagnostics` but it's weird to have both `--license-diagnostics` and `--license-text-diagnostics` both. We could have the former instead of the later btw, it's probably better (shorter and more general).

tests/licensedcode/data/additional_licenses/additional_license_combined_test.expected.json
AyanSinhaMahapatra: After discussion with @pombredanne and @DennisClark, it is concluded that we need to have:

1. lean top level license detections (no match details)
2. resource/package level details have the matches, and also the detection_identifier pointing to the top level unique detection, and not in a separate parallel list.

tests/licensedcode/data/additional_licenses/additional_license_combined_test.expected.json
AyanSinhaMahapatra: From @pombredanne : we might want to have both, as there could be two different use cases here:
1. having usable matched_text for generating notices, with the detection_diagnostics available.
2. having matched text diagnostics to see what words were matched.

AyanSinhaMahapatra: Done, we have a follow-up at https://github.com/nexB/scancode-toolkit/issues/3288

tests/licensedcode/data/additional_licenses/additional_license_combined_test.expected.json
AyanSinhaMahapatra: This is added, done!

AyanSinhaMahapatra:
@pombredanne This is ready for review!

tests/licensedcode/data/licenses_reference_reporting/license-reference-works-with-clues.expected.json
```diff
@@ -3,430 +3,80 @@
...
-          "matched_length": 35,
-          "match_coverage": 100.0,
-          "matcher": "2-aho",
-          "license_expression": "python",
-          "rule_identifier": "python_not_not-a-license_269.RULE",
-          "rule_url": "https://github.com/nexB/scancode-toolkit/tree/develop/src/licensedcode/data/rules/python_not_not-a-license_269.RULE"
-        }
-      ]
+      "detection_count": 1,
+      "detection_log": []
```
pombredanne, line 7: I think you may want this removed from top level unless requested.

pombredanne (changes requested):
This is weird... as there should be no detection_log at the top

pombredanne (commented):
LGTM otherwise and ready to merge!

tests/licensedcode/data/licenses_reference_reporting/license-reference-works-with-clues.expected.json
AyanSinhaMahapatra: @pombredanne false alarm :P This was added because we had `--license-diagnostics` enabled here. 
See for example:

-  a test with this disabled (i.e. not used) [test file](https://github.com/nexB/scancode-toolkit/blob/license-detection-follow-up/tests/licensedcode/test_licenses_reference.py#L32)  [result file](https://github.com/nexB/scancode-toolkit/blob/license-detection-follow-up/tests/licensedcode/data/licenses_reference_reporting/scan-with-reference.expected.json#L83) 

-  a test with this enabled [test_file](https://github.com/nexB/scancode-toolkit/blob/license-detection-follow-up/tests/licensedcode/test_plugin_license_detection.py#L66) [result_file](https://github.com/nexB/scancode-toolkit/blob/license-detection-follow-up/tests/licensedcode/data/plugin_license/unknown_intro/scan-unknown-intro-dual-license.expected.json#L73)

AyanSinhaMahapatra:
Thanks @pombredanne, merging this!

Issue #2912, opened 2022-04-09
Get more Python package data when scanning setup.cfg


JonoYang:
In Python packages, the `setup.py` file can contain origin info for a package. We can determine what the license, copyright holder, homepage and download URLs, as the package owner/maintainer declared it. However, some Python packages (like scancode-toolkit and other nexB projects) declare all of the package info and whatnot in `setup.cfg`, then let setuptools handle everything else. It does not seem that we get license or holder information when we do a Package scan on a `setup.cfg`, we only get the declared dependencies.

It would be convenient to be able to get more Package information from `setup.cfg` when scanning Python packages.

PR #3302, opened 2023-03-27
Add copyright holder field to PackageData model 

### LINKED ISSUE #3290: Add Copyright Holder information for Packages
This is to support https://github.com/nexB/scancode.io/issues/631
This feature consist in:
- adding a new field to the package model named "holder" (or "primary holder" or "copyright holder"
- add code that will process the package "copyright" when present and extract the holder from it. There is code for this likely in https://github.com/nexB/scancode-toolkit/blob/22109c4db81ebbf256babd0ff51cd97c11cb3c1d/src/summarycode/copyright_tallies.py#L225 that's used to computed summary holders and/or in cluecode.py to run a copyright detection on the package value and get the holder out of it. It should be canonicalized IMHO

keshav-space:
Fixes #3290

### Tasks


src/packagedcode/models.py
```diff
@@ -711,8 +711,34 @@ class PackageData(IdentifiablePackageData):
         repr=True,
     )
 
+    holder= String(
```
AyanSinhaMahapatra: It makes more sense to have this just after copyrights:
src/packagedcode/models.py#L612.

We would also need to regen the test files also.

src/packagedcode/models.py
```diff
@@ -711,8 +711,34 @@ class PackageData(IdentifiablePackageData):
...
+            holder_list = list(
+                detector.detect(
+                    numbered_lines,
+                    include_copyrights=False,
+                    include_holders=True,
+                    include_authors=False,
+                )
+            )
+            if holder_list:
+                self.holder = "\n".join(
```
AyanSinhaMahapatra: We are joining copyright holders with newlines, would it make more sense to have this as a list? 
Also note that copyrights also is a single string, and if we make holders a list, we should have the same there too.

This would be moving towards better package data like we did for licenses. Moving from just a expression string to objects with details. (Though we don't need any more details for copyrights holders, as start/end line don't apply here like files.)

AyanSinhaMahapatra (changes requested):
Thanks @keshav-space see comments for changes.

Also not that we only have copyrights detected for debian packages only, I could not see any issues for detecting copyrights for all other supported package manifests, apart from this https://github.com/nexB/scancode-toolkit/issues/2912, we should look into this and have issues to add copyright/holder support for all packages.

@pombredanne @JonoYang what fo you think?

src/packagedcode/models.py
keshav-space: I chatted with @pombredanne, and he suggested that we stick to the current string format because shifting to a list will create a whole lot of other issues, such as how to serialize those lists..

tests/packagedcode/data/alpine/apkbuild-problems/alpine14/main/sudo/APKBUILD-expected.json
```diff
@@ -86,5 +86,6 @@
   "repository_download_url": null,
   "api_data_url": null,
   "datasource_id": "alpine_apkbuild",
+  "holder": null,
```
AyanSinhaMahapatra: The `holder` attribute has not gone up with `copyright` as these tests don't fail for attribute order? We should ideally force regen anything with package scan on as otherwise we would have these changes pop up later when we regen for other cases and this would be more churn IMHO

AyanSinhaMahapatra (changes requested):
See comments for your consideration, thanks!

tests/packagedcode/data/alpine/apkbuild-problems/alpine14/main/sudo/APKBUILD-expected.json
keshav-space: Done

tests/summarycode/data/tallies/full_tallies/tallies_key_files-details.expected.json-lines
```diff
@@ -20,11 +20,11 @@
         "warnings": [],
         "extra_data": {
           "system_environment": {
-            "operating_system": "linux",
```
AyanSinhaMahapatra, line 23: Minor nitpick for next time, we can avoid  this churn on force regen by either removing the header if not necessary for test purposes, or just not commit these changes if possible. Don't have to change anything for now.

AyanSinhaMahapatra (approved):
LGTM! Thanks ++ @keshav-space 

We need a changelog entry for this and we're ready to merge.

keshav-space:
> We need a changelog entry for this and we're ready to merge.

@AyanSinhaMahapatra Changelog for this will go under [v32.0.0](https://github.com/nexB/scancode-toolkit/blob/develop/CHANGELOG.rst#v3200-next-roadmap) ?

AyanSinhaMahapatra:
@keshav-space yes! No need to bump any version since we're still doing release candidates.

JonoYang (approved):
@keshav-space Looks good, thanks!

AyanSinhaMahapatra:
@keshav-space Thanks++
Merging!

PR #3315, opened 2023-04-12
Add `get_top_level_resources()` to `DatafileHandler` class 

JonoYang:
This PR adds ~~`get_key_files()`~~ `get_top_level_resources()` to the `DatafileHandler` class. This method yields the resources that are considered key files of a package. Currently, `MavenPomXmlHandler` is the only DatafileHandler that has ~~`get_key_files()`~~ `get_top_level_resources()` implemented. The file classification plugin was made into a post-scan plugin, now that it has to wait for package data to be populated before working.

This change allows us to run the scan summary plugin on an extracted source JAR in addition to a standard codebase.


JonoYang:
@pombredanne 

I'd like your opinion on how the logic of determining the key files of a package should happen. Currently for the Maven POM xml DatafileHandler returns the Resources from the `META-INF` directory relative to where a detected POM.xml is. (https://github.com/nexB/scancode-toolkit/blob/package_key_files/src/packagedcode/maven.py#L156). In the file classification plugin, we then determine whether or not the files yielded by `get_key_files()` are key_files (https://github.com/nexB/scancode-toolkit/blob/package_key_files/src/summarycode/classify_plugin.py#L137). 

The logic here is strange:

- `DatafileHandler.get_key_files()` doesn't return key files, but rather top-level files. In the current setup, it should really be called `get_top_level_files()`. 
- I can update `get_key_files()` to return the known key files (pom.xml, etc.) and manifests, but there may be more that we miss.
- I could call `set_classification_flags` (https://github.com/nexB/scancode-toolkit/blob/package_key_files/src/summarycode/classify.py#L110) in `get_key_files()` to check whether or not a resource is a key file before yielding it.

pombredanne:
@JonoYang you wrote:

> The logic here is strange:
> 
> - DatafileHandler.get_key_files() doesn't return key files, but rather top-level files. In the current setup, it should really be called get_top_level_files().

In https://github.com/nexB/scancode-toolkit/blob/9e15906220dabbd7efd3372cde69dba317bbbfaf/src/packagedcode/maven.py#L156 you return the children of the "META-INF" dir from what I can see. These would be the key files of a JAR alright to me, or is there something I do not see there?  

> - I can update get_key_files() to return the known key files (pom.xml, etc.) and manifests, but there may be more that we miss.

That's ok: returning not too much and expend later if needed is better IMHO. I have seen whol JS npms stuffed under META-INF in the past.

> - I could call set_classification_flags (https://github.com/nexB/scancode-toolkit/blob/package_key_files/src/summarycode/classify.py#L110) in get_key_files() to check whether or not a resource is a key file before yielding it.

This sounds reasonable to me, typically the key files from the root of a source repo and up being copied to the META-INF dir at build time.

JonoYang:
@pombredanne 

> you return the children of the "META-INF" dir from what I can see. These would be the key files of a JAR alright to me, or is there something I do not see there?

Every file in the META-INF directory would be a top-level file, but not every one of those file would be a key file. I was just thinking that the method name was misleading because of that.

Edit: I ended up renaming `get_key_files()` to `get_top_level_resources()` just to be correct.

src/packagedcode/maven.py
```diff
@@ -152,6 +152,20 @@ def assign_package_to_resources(cls, package, resource, codebase, package_adder)
             package_adder=package_adder
         )
 
+    @classmethod
+    def get_top_level_resources(cls, manifest_resource, codebase):
+        """
+        Yield Resources that are top-level based on a JAR's directory structure
+        """
+        if 'META-INF' in manifest_resource.path:
```
AyanSinhaMahapatra, line 160: Just a question: should we also include key files at the root of a jar on some cases?

From #3334 
> The key files of a JAR are typically under META-INF, or OSGI-INF and more rarely at the root of a JAR.

`OSGI-INF` in some cases has `bundle*.properties` for languages and .xml files, are these to be considered too?

src/packagedcode/models.py
```diff
@@ -1304,6 +1304,13 @@ def populate_license_fields(cls, package_data):
         if package_data.extracted_license_statement and not isinstance(package_data.extracted_license_statement, str):
             package_data.extracted_license_statement = repr(package_data.extracted_license_statement)
 
+    @classmethod
+    def get_top_level_resources(cls, manifest_resource, codebase):
+        """
+        Yield Resources that are considered top-level for a Package type.
+        """
+        pass
```
AyanSinhaMahapatra, line 1312: Should we have some reasonable default here, which for source archives of most ecosystems is files that are actually on the top level (and then this would be overridden like in the case of Maven jars, the function you added)? 
Or should this be in a different function which can be called in individual `get_top_level_resources` functions and in this way we don't have any reasonable defaults (and thus more accurate atleast, support for key files for various datafile handlers can be added later).

What would be better in the summary plugin context?


I was also looking to use these functions for getting key files (I will add another filter of `get_legal_or_readme()` function as used currently like [here](https://github.com/nexB/scancode-toolkit/blob/package_key_files/src/packagedcode/licensing.py#L348) to get selective key files which usually has license info from all the top level files) in package license detection post processing:

If we don't have license detections in extracted license statements from a package manifest:

we currently get sibling files which are legalese, and add license detections present there to the package license detections and populate other license fields from there. Like [here](https://github.com/nexB/scancode-toolkit/blob/package_key_files/src/packagedcode/plugin_package.py#L192) 

Does this make sense to use `get_top_level_resources` functions added in this PR for these?

AyanSinhaMahapatra (approved):
Looks great, thanks! I've asked for your advice on some questions above for using this in package license detection too, see comments for those

src/packagedcode/maven.py
JonoYang: @AyanSinhaMahapatra We could add that, but I'd be more inclined to explicitly add those files if they provide some sort of origin info. I'd also like to see a few variations of JARs that have key files not in the META-INF directory. In the code, I consider anything in the `META-INF` directory to be top-level for the sake of simplicity as I do not know all of the variations of files that can be present in the root of a JAR.

src/packagedcode/models.py
JonoYang: @AyanSinhaMahapatra 

> What would be better in the summary plugin context?

Currently, the only piece of code calling the new `get_top_level_resources` method is in the file classification plugin, where when we encounter a Resource with package_data, we "reconstitute" the Package from the package data and then run `get_top_level_resources` (https://github.com/nexB/scancode-toolkit/blob/6a561109b3fc8362d3b47d7bf2fda34a6d6ea1eb/src/summarycode/classify_plugin.py#L136)

If we scan a codebase with no package information in it, then the file classification plugin works as before. I haven't checked this for myself, but I would think that the same behavior as the default behavior for `get_top_level_resources` like what you are described in your first paragraph.

If you wanted to use the `get_top_level_resources` method in other places, like for the parts you're working on, then I think that having a default implementation makes sense. My intentions for not implementing a default behavior was that I wanted to not change the way the file classification plugin works, except in the case of a JAR.

JonoYang: @AyanSinhaMahapatra Though I'd like to get this idea merged in first, then we can implement `get_top_level_resources` for more packages

JonoYang:
@AyanSinhaMahapatra 

I rebased this branch onto the latest changes from develop, but I am failing this test for generating yaml output: https://dev.azure.com/nexB/scancode-toolkit/_build/results?buildId=10643&view=logs&j=1412dbfa-421a-5a40-8ddf-7073fc46aa19&t=13c90161-2109-5957-f563-a96fefc7d869&l=7692

I tried to regenerated the test files, but I get this error:
```
___________________________________________________________________ test_scan_produces_valid_yaml ____________________________________________________________________

    def test_scan_produces_valid_yaml():
        test_dir = test_env.get_test_loc('yaml/package-and-licenses')
        result_file = "/home/jono/asdf.txt"#test_env.get_temp_file('json')
        args = [
            '-clipeu', '--license-text', '--classify', '--summary', '--license-references',
            '--license-clarity-score', test_dir, '--yaml', result_file]
        run_scan_click(args)
        expected = test_env.get_test_loc('yaml/package-and-licenses-expected.yaml')
>       check_yaml_scan(expected, result_file, regen=REGEN_TEST_FIXTURES)

tests/formattedcode/test_output_yaml.py:63: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/formattedcode/test_output_yaml.py:83: in check_yaml_scan
    results = load_yaml_results(
tests/formattedcode/test_output_yaml.py:121: in load_yaml_results
    scan_results = saneyaml.load(scan_results)
venv/lib/python3.10/site-packages/saneyaml.py:62: in load
    return yaml.load(s, Loader=loader)
venv/lib/python3.10/site-packages/yaml/__init__.py:81: in load
    return loader.get_single_data()
venv/lib/python3.10/site-packages/yaml/constructor.py:49: in get_single_data
    node = self.get_single_node()
yaml/_yaml.pyx:673: in yaml._yaml.CParser.get_single_node
    ???
yaml/_yaml.pyx:687: in yaml._yaml.CParser._compose_document
    ???
yaml/_yaml.pyx:731: in yaml._yaml.CParser._compose_node
    ???
yaml/_yaml.pyx:845: in yaml._yaml.CParser._compose_mapping_node
    ???
yaml/_yaml.pyx:729: in yaml._yaml.CParser._compose_node
    ???
yaml/_yaml.pyx:806: in yaml._yaml.CParser._compose_sequence_node
    ???
yaml/_yaml.pyx:731: in yaml._yaml.CParser._compose_node
    ???
yaml/_yaml.pyx:847: in yaml._yaml.CParser._compose_mapping_node
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   yaml.parser.ParserError: while parsing a block mapping
E     in "<unicode string>", line 141, column 9
E   did not find expected key
E     in "<unicode string>", line 180, column 40

yaml/_yaml.pyx:860: ParserError
```

tests/formattedcode/data/yaml/package-and-licenses-expected.yaml
```diff
@@ -33,34 +33,7 @@ headers:
         platform_version: '#67-Ubuntu SMP Mon Mar 13 14:22:10 UTC 2023'
         python_version: "3.8.10 (default, Mar 13 2023, 10:26:41) \n[GCC 9.4.0]"
       spdx_license_list_version: '3.20'
-      additional_license_directory: /home/ayansinha/nexB/write_access/scancode-extra/tests/licensedcode/data/additional_licenses/additional_dir
-      additional_license_plugins:
-        - /home/ayansinha/nexB/write_access/scancode-extra/venv/lib/python3.8/site-packages/licenses_to_install1
-        - /home/ayansinha/nexB/write_access/scancode-extra/venv/lib/python3.8/site-packages/licenses_to_install2
       files_count: 4
-summary:
```
AyanSinhaMahapatra, line 41: The summary should probably appear first in the JSON before Packages/Deps/Licenses IMHO the previous order made a lot of sense, what do you think?

AyanSinhaMahapatra:
@JonoYang weird that on the CI and on my local, it only gave an AssertionError like [here](https://dev.azure.com/nexB/scancode-toolkit/_build/results?buildId=10643&view=logs&j=1412dbfa-421a-5a40-8ddf-7073fc46aa19&t=13c90161-2109-5957-f563-a96fefc7d869&l=6379), so I pushed the fix and it's all passing now. I have a comment above on the order of the `Summary` in JSON though. 

This is weird that there's a YAMLError on your local though :thinking:

tests/formattedcode/data/yaml/package-and-licenses-expected.yaml
JonoYang: @AyanSinhaMahapatra 

I think this is an effect of having the summary and/or file classification plugin run after the package post-scan step, now that file classification has been updated to handle resources with package data in them (for get_top_level_resources). 

@pombredanne 

Do you know of a way to order the top-level codebase attributes without changing the plugin order?

JonoYang:
@AyanSinhaMahapatra Thanks for updating the yaml test results! Merging.

PR #3397, opened 2023-05-12
3396 update get license detections and expression

JonoYang:
This PR introduces the fix by @AyanSinhaMahapatra in `PackageData.get_license_detections_and_expression()` where a default value for `default_relation_license` is set when a `PackageData` object does not have a `datasource_id` value.

I've also pinned Sphinx to 5.3.0 to avoid the docs test failures.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -136,7 +136,7 @@ testing =
     pytest-rerunfailures
 
 docs =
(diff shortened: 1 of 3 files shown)


AyanSinhaMahapatra (approved):
LGTM! Thanks @JonoYang

Issue #3393, opened 2023-05-09
Bug in npm_api_url


JonoYang:
npm_api_url from packagedcode.npm has a bug where if a namespace is passed in, the version value is set to an empty string, even if a version value has been passed in via arguments.

AyanSinhaMahapatra:
@JonoYang this should be fixed now? Can I close this?
Thanks!

PR #3398, opened 2023-05-12
Do not set version to empty string in npm_api_url #3393

JonoYang:
This PR updates `npm.npm_api_ur()` to allow us to properly create npm API URLs for namespaced packages with versions. Previously, if a namespace value was passed into `npm.npm_api_ur()`, the version value would automatically be set to an empty string.


AyanSinhaMahapatra (approved):
LGTM, thanks @JonoYang 

There are some test failures at https://dev.azure.com/nexB/scancode-toolkit/_build/results?buildId=10746&view=logs&j=1412dbfa-421a-5a40-8ddf-7073fc46aa19, a regen should fix these allright?

PR #3447, opened 2023-07-05
Update maven detections

AyanSinhaMahapatra:
Update maven package and license detections:

* Fix MANIFEST.MF parsing
* Add package assembly for both pom.xml and MANIFEST.MF
* implement functionality to do package specific license detection
* run license detection on whole extracted license statement for maven
* Add rules for license detection fixes
* Fix detection for pom.properties file
* Fix --todo plugin for package detection
* Fix false positive case for bare word rules
* Fix issue of unknown license detection in package manifest

### Tasks

* [x] Updated CHANGELOG.rst


src/packagedcode/maven.py
```diff
@@ -1171,6 +1172,63 @@ def _parse(
...
+                try_as_expression=try_as_expression,
+                approximate=approximate,
+                expression_symbols=expression_symbols,
+            )
+        
+        new_extracted_license = extracted_license.copy()
+        
+        for license_entry in new_extracted_license:
+            license_entry.pop("distribution")
+            if not license_entry.get("name", None):
```
pombredanne: Nit: None is already the default for a get

pombredanne: Also what about this instead:
`{key: value for key, value in license_entry.items() if value and key != 'distribution'}`

src/packagedcode/jar_manifest.py
```diff
@@ -278,7 +296,7 @@ def dget(s):
     package['homepage_url'] = dget('Implementation-URL') or dget('Implementation-Url')
 
     # Bundle-DocURL: http://logging.apache.org/log4j/1.2
-    package['documentation_url'] = dget('Bundle-DocURL')
+    #package['documentation_url'] = dget('Bundle-DocURL')
```
JonoYang: Should this line be removed if it's being commented out?

src/packagedcode/jar_manifest.py
```diff
@@ -369,7 +387,7 @@ def dget(s):
     # Deps:
     # Require-Bundle
 
-    package['notes'] = dget('Comment')
+    #package['notes'] = dget('Comment')
```
JonoYang: See comment on line 299

src/packagedcode/maven.py
AyanSinhaMahapatra: Done, thanks!

AyanSinhaMahapatra: > Also what about this instead:
> {key: value for key, value in license_entry.items() if value and key != 'distribution'}

Actually we are doing a `pop()` so we actually not have empty key/value pairs in the original extracted license statement. Which is correct since we are creating these key/value pairs irrespective of them being present in the first place.

src/packagedcode/jar_manifest.py
```diff
@@ -369,10 +339,61 @@ def dget(s):
...
+def get_datasource_id(package_type):
+    """
+    Get the corresponding `datasource_id` for the given
+    `package_type`. This is a seperate function to avoid
+    cyclic imports.
+    """
+    from packagedcode.maven import JavaJarManifestHandler
+    from packagedcode.maven import JavaOSGiManifestHandler
+
+    if package_type == 'maven':
```
JonoYang:
```suggestion
    if package_type in ['maven', 'jar']:
        return JavaJarManifestHandler.datasource_id
```

Minor change just to avoid repetition.

JonoYang (changes requested):
@AyanSinhaMahapatra Looks good, I just highlighted some small issues.

src/packagedcode/jar_manifest.py
AyanSinhaMahapatra: Adding these in the `extra_data` field instead, thanks.

src/packagedcode/jar_manifest.py
AyanSinhaMahapatra: Addressed, thanks!

src/packagedcode/jar_manifest.py
AyanSinhaMahapatra: Added this, thanks!

tests/summarycode/data/classify/with_package_data.expected.json
```diff
@@ -500,15 +542,64 @@
...
+          "license_detections": [],
+          "other_license_expression": null,
+          "other_license_expression_spdx": null,
+          "other_license_detections": [],
+          "extracted_license_statement": null,
+          "notice_text": null,
+          "source_packages": [],
+          "file_references": [],
+          "extra_data": {
+            "pom_properties": {
```
AyanSinhaMahapatra: Aren't these the namespace/name/version? Why are we having these in the extra-data instead of having these in the proper attributes (and then using these in the assembly)?

AyanSinhaMahapatra: I'm adding these in the main package attributes instead, so we have the correct PURL for these

src/licensedcode/data/rules/bsd-3-clause-no-nuclear-warranty_and_jj2000_2.RULE
```diff
@@ -0,0 +1,10 @@
+---
+license_expression: bsd-3-clause-no-nuclear-warranty AND jj2000
+is_license_tag: yes
+minimum_coverage: 95
+---
+
+name: {{JJ2000}}
+url: LICENSE-JJ2000.txt
+name: {{BSD 3-clause License w/nuclear disclaimer}}
```
pombredanne: could it better to split these in two rules?

src/licensedcode/data/rules/json_17.RULE
```diff
@@ -0,0 +1,32 @@
+---
+license_expression: json
+is_license_tag: yes
```
pombredanne:
```suggestion
is_license_text: yes
```

src/licensedcode/data/rules/lgpl-2.1_and_apache-2.0_1.RULE
```diff
@@ -0,0 +1,12 @@
...
+license_expression: lgpl-2.1 AND apache-2.0
+is_license_tag: yes
+ignorable_urls:
+    - http://www.apache.org/licenses/LICENSE-2.0.txt
+    - http://www.gnu.org/licenses/licenses.html
+---
+
+name: LGPL, version 2.1
+url: http://www.gnu.org/licenses/licenses.html
+name: Apache License v2.0
```
pombredanne: Would it be better to split this in two?

src/licensedcode/data/rules/lgpl-2.1_and_apache-2.0_2.RULE
```diff
@@ -0,0 +1,12 @@
...
+license_expression: lgpl-2.1 AND apache-2.0
+is_license_tag: yes
+ignorable_urls:
+    - http://www.apache.org/licenses/
+    - http://www.gnu.org/licenses/licenses.html
+---
+
+name: LGPL, version 2.1
+url: http://www.gnu.org/licenses/licenses.html
+name: ASL, version 2
```
pombredanne: Would it be better to split this in two?

src/licensedcode/detection.py
```diff
@@ -729,6 +749,27 @@ def collect_license_detections(codebase, include_license_clues=True):
     return all_license_detections
 
 
+
+def verify_package_license_expression(license_detection_mappings, license_expression):
+
```
pombredanne: What is this method doing and returning? Some doc would help.

src/licensedcode/detection.py
```diff
@@ -978,9 +1019,13 @@ def is_false_positive(license_matches, package_license=False):
         match_rule_length == 1
         for match_rule_length in match_rule_length_values
     )
-
-    is_gpl_bare = all(
-        'gpl_bare' in license_match.rule.identifier
+    bare_rules = ['gpl_bare', 'freeware_bare', 'public-domain_bare']
+    is_bare_rule = all(
+        any([
+            True
```
pombredanne: If you use this, prefer this code .... but IMHO we should have a flag on the Rule model instead
```suggestion
            bare_rule in license_match.rule.identifier
```

src/licensedcode/detection.py
```diff
@@ -978,9 +1019,13 @@ def is_false_positive(license_matches, package_license=False):
...
+            for bare_rule in bare_rules
+            if bare_rule in license_match.rule.identifier
```
pombredanne:
```suggestion
```

src/packagedcode/__init__.py
```diff
@@ -253,3 +253,12 @@ def get_package_handler(package_data):
         raise UnknownPackageDatasource(package_data)
     return ppc
 
+
+PACKAGE_DATA_CLASS_WITH_CUSTOM_LICENSE_DETECTION = [
+    maven.MavenPackageData
+]
+
+PACKAGE_DATA_CLASS_BY_DATASOURCE_ID = {
```
pombredanne, line 257: Create the data at once statically rather that doing this work on each import
```suggestion
PACKAGE_DATA_CLASS_BY_DATASOURCE_ID = {maven.MavenPackageData.datasource_id: maven.MavenPackageData}
```

src/packagedcode/jar_manifest.py
```diff
@@ -369,10 +339,67 @@ def dget(s):
     # Deps:
     # Require-Bundle
 
-    package['notes'] = dget('Comment')
+    comment = dget('Comment')
+
+    if comment or doc_url:
+        package["extra_data"] = {}
+        if comment:
+            package["extra_data"]['notes'] = comment
```
pombredanne, line 347: we need to find a better way than using extra_data.

src/packagedcode/jar_manifest.py
```diff
@@ -278,7 +248,7 @@ def dget(s):
     package['homepage_url'] = dget('Implementation-URL') or dget('Implementation-Url')
 
     # Bundle-DocURL: http://logging.apache.org/log4j/1.2
-    package['documentation_url'] = dget('Bundle-DocURL')
+    doc_url = dget('Bundle-DocURL')
```
pombredanne, line 251: I get what you are doing, but we are losing data here.

src/packagedcode/jar_manifest.py
```diff
@@ -369,10 +339,67 @@ def dget(s):
...
+        package["extra_data"] = {}
+        if comment:
+            package["extra_data"]['notes'] = comment
+        if doc_url:
+            package["extra_data"]['documentation_url'] = doc_url
+
     return package
 
 
+def parse_scm_connection(scm_connection):
```
pombredanne, line 354: @TG1999 ping ... is this related or relevant to your purldb work?

src/packagedcode/jar_manifest.py
```diff
@@ -369,10 +339,67 @@ def dget(s):
...
+    `package_type`. This is a seperate function to avoid
+    cyclic imports.
+    """
+    from packagedcode.maven import JavaJarManifestHandler
+    from packagedcode.maven import JavaOSGiManifestHandler
+
+    if package_type in ['maven', 'jar']:
+        return JavaJarManifestHandler.datasource_id
+    elif package_type == 'osgi':
+        return JavaOSGiManifestHandler.datasource_id
```
pombredanne, line 400: What if a maven package also has OSGi thing?s

src/packagedcode/licensing.py
```diff
@@ -764,22 +767,33 @@ def get_license_detections_and_expression(
...
         return detection_data, license_expression
 
-    license_detections = get_normalized_license_detections(
-        extracted_license=extracted_license_statement,
-        try_as_expression=try_as_expression,
-        approximate=approximate,
-        expression_symbols=expression_symbols,
-    )
+    if datasource_id in PACKAGE_DATA_CLASS_BY_DATASOURCE_ID:
+        package_data_class = PACKAGE_DATA_CLASS_BY_DATASOURCE_ID.get(datasource_id, None)
```
pombredanne:
```suggestion
        package_data_class = PACKAGE_DATA_CLASS_BY_DATASOURCE_ID.get(datasource_id)
```

pombredanne (changes requested):
Please see some nits for your consideration

src/packagedcode/jar_manifest.py
AyanSinhaMahapatra: We are returning this `package` as-is to be used for creating a PackageData class, see https://github.com/nexB/scancode-toolkit/blob/update-maven-detections/src/packagedcode/maven.py#L138 with `models.PackageData(**manifest,)`. We don't have a `documentation_url` or a `notes` field in PackageData, and this is why the Manifest parsing used to fail and not return anything. I'm just ensuring we return whatever we can, instead of crashing and not returning anything. And since I'm having this in `extra_data`, we're not really losing any data IMHO

src/packagedcode/jar_manifest.py
AyanSinhaMahapatra: See https://github.com/nexB/scancode-toolkit/pull/3447#discussion_r1262231378. We might want to add fields to the PackageData class then. And we might handle this separately, AFAIK this data is not really used anywhere and this is just a fix to make sure MANIFEST parsing returns everything else, and not crash.

src/packagedcode/jar_manifest.py
AyanSinhaMahapatra: We can improve the parsing and handle more cases allright, but let's do this separately through another issue, and if you have an example of that case please provide the same. But this was also a fix, and I'm only making sure this works without crashing here.

src/packagedcode/licensing.py
AyanSinhaMahapatra: Done, thanks!

src/packagedcode/__init__.py
AyanSinhaMahapatra: This would be better, doing this instead, thanks!

src/licensedcode/detection.py
AyanSinhaMahapatra: Docstring added, thanks!

src/licensedcode/detection.py
AyanSinhaMahapatra: Yeah this could be a computed flag in the Rule object allright, will implement this seperately. Thanks!
For now adding the suggested change.

src/licensedcode/detection.py
AyanSinhaMahapatra: Added this, thanks!

src/licensedcode/data/rules/lgpl-2.1_and_apache-2.0_1.RULE
AyanSinhaMahapatra: Done, thanks!

src/licensedcode/data/rules/lgpl-2.1_and_apache-2.0_2.RULE
AyanSinhaMahapatra: Done, thanks!

src/licensedcode/data/rules/json_17.RULE
AyanSinhaMahapatra: Done, thanks!

src/licensedcode/data/rules/bsd-3-clause-no-nuclear-warranty_and_jj2000_2.RULE
AyanSinhaMahapatra: Done, thanks!

AyanSinhaMahapatra:
@pombredanne @JonoYang addressed all your comments, thanks!

src/licensedcode/data/rules/bsd-3-clause-no-nuclear-warranty_12.RULE
```diff
@@ -0,0 +1,7 @@
+---
+license_expression: bsd-3-clause-no-nuclear-warranty
+is_license_tag: yes
```
pombredanne, line 3: You could add a:

```suggestion
is_license_tag: yes
referenced_filenames:
    - LICENSE-Sun.txt
```

src/licensedcode/data/rules/jj2000_1.RULE
```diff
@@ -0,0 +1,7 @@
+---
+license_expression: jj2000
+is_license_tag: yes
+---
+
+name: {{JJ2000}}
+url: LICENSE-JJ2000.txt
```
pombredanne, line 9: You could add a referenced filenames for this

src/packagedcode/licensing.py
```diff
@@ -764,22 +767,33 @@ def get_license_detections_and_expression(
...
     if not extracted_license_statement:
         return detection_data, license_expression
 
-    license_detections = get_normalized_license_detections(
-        extracted_license=extracted_license_statement,
-        try_as_expression=try_as_expression,
-        approximate=approximate,
-        expression_symbols=expression_symbols,
-    )
+    if datasource_id in PACKAGE_DATA_CLASS_BY_DATASOURCE_ID:
```
pombredanne:
```suggestion
    package_data_class = PACKAGE_DATA_CLASS_BY_DATASOURCE_ID.get(datasource_id)
    if package_data_class:
```

src/packagedcode/licensing.py
```diff
@@ -764,22 +767,33 @@ def get_license_detections_and_expression(
...
+        package_data_class = PACKAGE_DATA_CLASS_BY_DATASOURCE_ID.get(datasource_id)
```
pombredanne:
```suggestion

```

pombredanne (approved):
LGTM with a few extra nitrs that you can fix or not as you like before merging. Thanks!

src/licensedcode/data/rules/bsd-3-clause-no-nuclear-warranty_12.RULE
AyanSinhaMahapatra: Added this thanks!

src/licensedcode/data/rules/jj2000_1.RULE
AyanSinhaMahapatra: Added this thanks!

src/packagedcode/licensing.py
AyanSinhaMahapatra: Added this, thanks!

src/packagedcode/licensing.py
AyanSinhaMahapatra: Added this, thanks!

PR #3451, opened 2023-07-12
Edit `check_rdf_scan` so that SPDX rdf tests don't automatically pass #3448

### LINKED ISSUE #3448: SPDX rdf tests pass automatically
### Description

The SPDX rdf tests currently pass automatically. This is due to [this line](https://github.com/nexB/scancode-toolkit/blob/a15174f31efaf8816e8c9a65c9f85c4beffc0227/tests/formattedcode/test_output_spdx.py#L130) in the code, which basically sets `expected == result` per default.
The bug can be fixed by removing the linked line.

As almost all spdx tests are skipped via the `@pytest.mark.scanslow` annotation, this does not affect most of the tests.
However, if you remove these annotations they will fail with the above fix.

When upgrading the spdx-tools to the new 0.8 version, these test have to be slightly adapted, anyway. I can offer to do this while working on the library upgrade.

### How To Reproduce

Remove [this line](https://github.com/nexB/scancode-toolkit/blob/a15174f31efaf8816e8c9a65c9f85c4beffc0227/tests/formattedcode/test_output_spdx.py#L130) in the code and execute for example [this test](https://github.com/nexB/scancode-toolkit/blob/a15174f31efaf8816e8c9a65c9f85c4beffc0227/tests/formattedcode/test_output_spdx.py#L197) (you have to remove the annotation first).

### System configuration

Ubuntu 22.04.2 LTS, Scancode 32.0.5rc3

armintaenzertng:
Also adapted the `expected.rdf` files from tests that failed thereafter.

Fixes #3448 

### Tasks

* [x] Updated CHANGELOG.rst

### DIFF
--- a/tests/formattedcode/test_output_spdx.py
+++ b/tests/formattedcode/test_output_spdx.py
@@ -127,7 +127,6 @@ def check_rdf_scan(expected_file, result_file, regen=REGEN_TEST_FIXTURES):
     else:
         with io.open(expected_file, encoding='utf-8') as i:
             expected = json.load(i)
-            expected = load_and_clean_rdf(result_file)
 
     assert json.dumps(result, indent=2) == json.dumps(expected, indent=2)
 
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -8,6 +8,7 @@ The following organizations or individuals have contributed to ScanCode:
 - Alex Blekhman @a-tinsmith
 - Alexander Gschrei @agschrei
 - Armijn Hemmel @armijnhemel
+- Armin Tänzer @armintaenzertng
 - Arnaud Jeansen @ajeans
 - Arnav Mandal @arnav-mandal1234
 - Aryan Kenchappagol @aryanxk02
--- a/tests/formattedcode/data/spdx/license_known/expected.rdf
+++ b/tests/formattedcode/data/spdx/license_known/expected.rdf
@@ -59,7 +59,7 @@
             "ns1:checksum": {
               "ns1:Checksum": {
                 "ns1:checksumValue": "2b8b815229aa8a61e483fb4ba0588b8b6c491890",
-                "ns1:algorithm": "SHA1"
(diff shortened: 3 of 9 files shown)


tests/formattedcode/test_output_spdx.py
```diff
@@ -127,7 +127,6 @@ def check_rdf_scan(expected_file, result_file, regen=REGEN_TEST_FIXTURES):
     else:
         with io.open(expected_file, encoding='utf-8') as i:
             expected = json.load(i)
```
AyanSinhaMahapatra, line 129: See [this comment](https://github.com/nexB/scancode-toolkit/issues/3448#issuecomment-1630954027) from @pombredanne , I think this is what we want?

```suggestion
        expected = load_and_clean_rdf(expected_file)
```

The `load_and_clean_rdf` function at https://github.com/nexB/scancode-toolkit/blob/a15174f31efaf8816e8c9a65c9f85c4beffc0227/tests/formattedcode/test_output_spdx.py#L45 does the read and loading, and also gets rid of parts which can be variable across tests.
See https://github.com/nexB/scancode-toolkit/blob/develop/tests/formattedcode/test_output_spdx.py#L29 for the variable stripping part.

AyanSinhaMahapatra (changes requested):
Thanks! See comments below

tests/formattedcode/test_output_spdx.py
armintaenzertng: The "expected" rdf files have already undergone the `load_and_clean_rdf()` process. Especially they are not in xml format anymore, so `load_and_clean_rdf(expected_file)` raises an error, as the function expects rdf-xml.
As can be seen by the green tests, the current code works fine because `expected_file` is already in json format and cleaned.

AyanSinhaMahapatra: Ah I see the issue, my bad. Thanks for the explanation.

AyanSinhaMahapatra (approved):
LGTM! Thanks++

PR #3456, opened 2023-07-17
Upgrade `spdx-tools` to v0.8.1 #3455

### LINKED ISSUE #3455: Upgrade `spdx-tools` to version 0.8
## Short Description
Upgrade `spdx-tools` to v0.8

## Possible Labels

- dependencies

## Select Category

-  [ ] Enhancement
-  [ ] Add License/Copyright
-  [ ] Scan Feature
-  [ ] Packaging
-  [ ] Documentation
-  [x] Expand Support
-  [ ] Other

## **Describe the Update**

Update the `spdx-tools` dependency from `0.7.0rc0` to `0.8.0rc2` or later.

The new 0.8 version of the `spdx-tools` lays the groundwork for the upcoming SPDX-3.0 implementation in python. We therefore expect broad adoption of the 0.8 version.
Tools that use the 0.8 version of the `spdx-tools` currently can't simultaneously use scancode which depends on the 0.7 version (there can't be two different versions of the same library in a single environment).
This issue currently blocks adoption of the `spdx-tools` in tern, for example.

## **Example/Links if Any**

Pending PR in tern that introduces `spdx-tools` in version 0.8: https://github.com/tern-tools/tern/pull/1233

I will open a PR that takes care of the upgrade while preserving the current feature set.

armintaenzertng:
As the `spdx-tools` library has been refactored, this requires some code adaption of the current `output_spdx.py` and regeneration of the test files.
Also, the SPDX license list is now handled by the `license-expression` library, which is why we don't need the corresponding code anymore that adapts this list.

Fixes #3455

### Tasks

* [x] Updated CHANGELOG.rst

### DIFF
--- a/setup-mini.cfg
+++ b/setup-mini.cfg
@@ -105,7 +105,7 @@ install_requires =
     pymaven_patch >= 0.2.8
     requests >= 2.7.0
     saneyaml >= 0.6.0
-    spdx_tools == 0.7.0rc0
+    spdx_tools == 0.8.1
     text_unidecode >= 1.0
     toml >= 0.10.0
     urlpy
--- a/src/formattedcode/output_spdx.py
+++ b/src/formattedcode/output_spdx.py
@@ -9,22 +9,10 @@
 import os
 import sys
 import uuid
+from datetime import datetime
 from io import BytesIO
 from io import StringIO
 
-from spdx.checksum import Checksum
-from spdx.checksum import ChecksumAlgorithm
-from spdx.creationinfo import Tool
-from spdx.document import ExtractedLicense
-from spdx.document import Document
-from spdx.license import License
-from spdx.file import File
-from spdx.package import Package
-from spdx.relationship import Relationship
-from spdx.utils import calc_verif_code
-from spdx.utils import NoAssert
-from spdx.utils import SPDXNone
-from spdx.version import Version
 
 from license_expression import Licensing
 from commoncode.cliutils import OUTPUT_GROUP
@@ -32,6 +20,22 @@ from commoncode.cliutils import PluggableCommandLineOption
 from commoncode.fileutils import file_name
 from commoncode.fileutils import parent_directory
 from commoncode.text import python_safe_name
+from spdx_tools.spdx.model import SpdxNoAssertion
+from spdx_tools.spdx.model import Version
+from spdx_tools.spdx.model import CreationInfo
+from spdx_tools.spdx.model import Actor
+from spdx_tools.spdx.model import ActorType
+from spdx_tools.spdx.model import Document
+from spdx_tools.spdx.model import Package
+from spdx_tools.spdx.model import File
+from spdx_tools.spdx.model import Checksum
+from spdx_tools.spdx.model import ChecksumAlgorithm
+from spdx_tools.spdx.model import ExtractedLicensingInfo
+from spdx_tools.spdx.model import SpdxNone
+from spdx_tools.spdx.model import Relationship
+from spdx_tools.spdx.model import RelationshipType
+from spdx_tools.spdx.spdx_element_utils import calculate_package_verification_code
+
 from formattedcode import FileOptionType
 from licensedcode.detection import get_matches_from_detection_mappings
 from plugincode.output import output_impl
@@ -62,50 +66,6 @@ if TRACE or TRACE_DEEP:
 """
 Output plugins to write scan results in SPDX format.
 """
-
-_spdx_list_is_patched = False
-
-
-def _patch_license_list():
-    """
-    Patch the SPDX Python library license list to match the list of ScanCode
-    known SPDX licenses.
-    """
-    global _spdx_list_is_patched
-    if not _spdx_list_is_patched:
-        from spdx.config import LICENSE_MAP
-        from licensedcode.models import load_licenses
-        licenses = load_licenses(with_deprecated=True)
-        spdx_licenses = get_licenses_by_spdx_key(licenses.values())
-        LICENSE_MAP.update(spdx_licenses)
-        _spdx_list_is_patched = True
-
-
-def get_licenses_by_spdx_key(licenses):
-    """
-    Return a mapping of {spdx_key: license object} given a ``license`` sequence
-    of License objects.
-    """
-    spdx_licenses = {}
-    for lic in licenses:
-        if not (lic.spdx_license_key or lic.other_spdx_license_keys):
-            continue
-
-        if lic.spdx_license_key:
-            name = lic.name
-            slk = lic.spdx_license_key
-            spdx_licenses[slk] = name
-            spdx_licenses[name] = slk
-
-            for other_spdx in lic.other_spdx_license_keys:
-                if not (other_spdx and other_spdx.strip()):
-                    continue
-                slk = other_spdx
-                spdx_licenses[slk] = name
-                spdx_licenses[name] = slk
-
-    return spdx_licenses
-
 
 @output_impl
 class SpdxTvOutput(OutputPlugin):
@@ -220,7 +180,7 @@ def write_spdx(
     tool_version,
     notice,
     package_name='',
-    download_location=NoAssert(),
+    download_location=SpdxNoAssertion(),
     as_tagvalue=True,
     spdx_version = (2, 2),
     with_notice_text=False,
@@ -240,7 +200,6 @@ def write_spdx(
     licensing = Licensing()
 
     as_rdf = not as_tagvalue
-    _patch_license_list()
 
     ns_prefix = '_'.join(package_name.lower().split())
     comment = notice + f'\nSPDX License List: {scancode_config.spdx_license_list_version}'
@@ -248,29 +207,36 @@ def write_spdx(
     version_major, version_minor = scancode_config.spdx_license_list_version.split(".")
     spdx_license_list_version = Version(major=version_major, minor=version_minor)
 
-    doc = Document(
-        version=Version(*spdx_version),
-        data_license=License.from_identifier('CC0-1.0'),
-        comment=notice,
-        namespace=f'http://spdx.org/spdxdocs/{ns_prefix}-{uuid.uuid4()}',
-        license_list_version=scancode_config.spdx_license_list_version,
-        name='SPDX Document created by ScanCode Toolkit'
+    tool_name = tool_name or 'ScanCode'
+    creator = Actor(ActorType.TOOL, f'{tool_name} {tool_version}')
+
+    creation_info = CreationInfo(
+        spdx_id="SPDXRef-DOCUMENT",
+        spdx_version=f"SPDX-{spdx_version[0]}.{spdx_version[1]}",
+        data_license='CC0-1.0',
+        document_comment=comment,
+        document_namespace=f'http://spdx.org/spdxdocs/{ns_prefix}-{uuid.uuid4()}',
+        license_list_version=spdx_license_list_version,
+        name='SPDX Document created by ScanCode Toolkit',
+        creators=[creator],
+        created=datetime.now(),
     )
 
-    tool_name = tool_name or 'ScanCode'
-    doc.creation_info.add_creator(Tool(f'{tool_name} {tool_version}'))
-    doc.creation_info.set_created_now()
-    doc.creation_info.license_list_version = spdx_license_list_version
 
     package_id = '001'
-    package = doc.package = Package(
+    package = Package(
         name=package_name,
         download_location=download_location,
         spdx_id=f'SPDXRef-{package_id}',
     )
 
+    doc = Document(
+        creation_info=creation_info,
+        packages=[package],
+    )
+
     # Use a set of unique copyrights for the package.
-    package.cr_text = set()
+    package_copyright_texts = set()
 
     all_files_have_no_license = True
     all_files_have_no_copyright = True
@@ -285,13 +251,18 @@ def write_spdx(
         # Set a relative file name as that is what we want in
         # SPDX output (with explicit leading './').
         name = './' + file_data.get('path')
+
+        if file_data.get('file_type') == 'empty':
+            checksum = Checksum(ChecksumAlgorithm.SHA1, "da39a3ee5e6b4b0d3255bfef95601890afd80709")
+        else:
+            # FIXME: this sets the checksum of a file to the empty string hash if unknown; tracked in https://github.com/nexB/scancode-toolkit/issues/3453
+            checksum = Checksum(ChecksumAlgorithm.SHA1, file_data.get('sha1') or 'da39a3ee5e6b4b0d3255bfef95601890afd80709')
+
         file_entry = File(
             spdx_id=f'SPDXRef-{sid}',
-            name=name)
-        if file_data.get('file_type') == 'empty':
-            file_entry.set_checksum(Checksum(ChecksumAlgorithm.SHA1, "da39a3ee5e6b4b0d3255bfef95601890afd80709"))
-        else:
-            file_entry.set_checksum(Checksum(ChecksumAlgorithm.SHA1, file_data.get('sha1') or ''))
+            name=name,
+            checksums=[checksum]
+        )
 
         file_license_detections = file_data.get('license_detections')
         license_matches = get_matches_from_detection_mappings(file_license_detections)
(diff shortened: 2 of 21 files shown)


src/formattedcode/output_spdx.py
```diff
@@ -9,29 +9,22 @@
...
-from spdx.utils import SPDXNone
-from spdx.version import Version
 
 from license_expression import Licensing
 from commoncode.cliutils import OUTPUT_GROUP
 from commoncode.cliutils import PluggableCommandLineOption
 from commoncode.fileutils import file_name
 from commoncode.fileutils import parent_directory
 from commoncode.text import python_safe_name
+from spdx_tools.spdx.model import SpdxNoAssertion, Version, CreationInfo, Actor, ActorType, Document, Package, File, \
```
pombredanne: If you do not mind, our convention is one import per line, no multiline imports.

src/formattedcode/output_spdx.py
```diff
@@ -312,63 +272,63 @@ def write_spdx(
...
-                        spdx_license = ExtractedLicense(spdx_id)
-                        spdx_license.name = file_license.short_name
+                    spdx_license = licensing.parse(spdx_id)
+
+                    if is_license_ref:
+                        text = match.get('matched_text')
                         # FIXME: replace this with the licensedb URL
                         comment = (
                             f'See details at https://github.com/nexB/scancode-toolkit'
                             f'/blob/develop/src/licensedcode/data/licenses/{license_key}.yml\n'
```
pombredanne:
```suggestion
                            f'/blob/develop/src/licensedcode/data/licenses/{license_key}.LICENSE\n'
```

src/formattedcode/output_spdx.py
```diff
@@ -312,63 +272,63 @@ def write_spdx(
...
 
-                    if not is_license_ref:
-                        spdx_license = License.from_identifier(spdx_id)
-                    else:
-                        spdx_license = ExtractedLicense(spdx_id)
-                        spdx_license.name = file_license.short_name
+                    spdx_license = licensing.parse(spdx_id)
+
+                    if is_license_ref:
+                        text = match.get('matched_text')
```
pombredanne, line 289: I think we would want the matched text in most cases, unless this is matched to a full license text for a standard SPDX license. For all other cases (notices, license_ref, etc.) we should get this.

src/formattedcode/output_spdx.py
```diff
@@ -312,63 +272,63 @@ def write_spdx(
...
+                    package.license_info_from_files += [spdx_license]
 
         elif license_matches is None:
             all_files_have_no_license = False
-            file_entry.add_lics(NoAssert())
+            file_entry.license_info_in_file += [SpdxNoAssertion()]
 
         else:
-            file_entry.add_lics(SPDXNone())
+            file_entry.license_info_in_file += [SpdxNone()]
```
pombredanne: Why is this `license_info_in_file` and not `license_info_in_files` or some plural since this is a list of things?

src/formattedcode/output_spdx.py
```diff
@@ -312,63 +272,63 @@ def write_spdx(
...
+                            license_name=file_license.short_name,
+                            comment=comment,
+                        )
+                        doc.extracted_licensing_info += [extracted_license]
 
                     # Add licenses in the order they appear in the file. Maintaining
                     # the order might be useful for provenance purposes.
-                    file_entry.add_lics(spdx_license)
-                    package.add_lics_from_file(spdx_license)
+                    file_entry.license_info_in_file += [spdx_license]
```
pombredanne: Why not use `.append()` here and in other places? Creating and then adding a list to an existing list feels too much.

src/formattedcode/output_spdx.py
```diff
@@ -379,33 +339,29 @@ def write_spdx(
...
-    if len(package.cr_text) == 0:
+    if not package_copyright_texts:
         if all_files_have_no_copyright:
-            package.cr_text = SPDXNone()
+            package.copyright_text = SpdxNone()
         else:
-            package.cr_text = NoAssert()
+            package.copyright_text = SpdxNoAssertion()
     else:
         # Create a text of alphabetically sorted copyright
```
pombredanne, line 369: IMHO we should not use a set and instead preserve the original order of copyrights: it matters.

tests/formattedcode/data/spdx/license_known/expected.tv
```diff
@@ -1,35 +1,36 @@
-# Document Information
+## Document Information
```
pombredanne, line 1: Why using a double ## pound?

tests/formattedcode/data/spdx/license_known/expected.tv
```diff
@@ -1,35 +1,36 @@
...
 SPDXVersion: SPDX-2.2
 DataLicense: CC0-1.0
-DocumentNamespace: http://spdx.org/spdxdocs/scan
```
pombredanne, line 4: Why change the ordering of the fields?

tests/formattedcode/data/spdx/license_known/expected.tv
```diff
@@ -1,35 +1,36 @@
...
-DocumentName: SPDX Document created by ScanCode Toolkit
-LicenseListVersion: 3.20
```
pombredanne, line 6: Why removing this version? this sounds a big change?

tests/formattedcode/data/spdx/license_known/expected.tv
```diff
@@ -1,35 +1,36 @@
...
+Visit https://github.com/nexB/scancode-toolkit/ for support and download.
+SPDX License List: 3.21</text>
+## Creation Information
+## Package Information
 PackageName: scan
 SPDXID: SPDXRef-001
 PackageDownloadLocation: NOASSERTION
+FilesAnalyzed: True
 PackageVerificationCode: 517c26f04b5a3edaaa0c17bf1a6571953209e209
-PackageLicenseDeclared: NOASSERTION
```
pombredanne, line 20: Why not keeping the original ordering? Declared before Concluded is the right natural ordering we have until now?

tests/formattedcode/data/spdx/license_known/expected.tv
```diff
@@ -1,35 +1,36 @@
...
 PackageLicenseConcluded: NOASSERTION
 PackageLicenseInfoFromFiles: Apache-2.0
 PackageLicenseInfoFromFiles: CC0-1.0
+PackageLicenseDeclared: NOASSERTION
 PackageCopyrightText: NONE
-# File
+## File Information
```
pombredanne, line 26: Why the extra # and the trailing "Information"?

tests/formattedcode/data/spdx/license_known/expected.rdf
```diff
@@ -1,117 +1,105 @@
...
+      },
+      "spdx:relationship": [
         {
-          "ns1:Relationship": {
-            "ns1:relationshipType": {
+          "spdx:Relationship": {
+            "spdx:relationshipType": {
               "@rdf:resource": "http://spdx.org/rdf/terms#relationshipType_contains"
             },
-            "ns1:spdxElementId": "SPDXRef-001",
```
pombredanne, line 97: Where have the Ids gone? I thought these were must have data?

tests/formattedcode/data/spdx/license_ref/expected_with_text.tv
```diff
@@ -85,10 +83,10 @@ License for ScanCode datasets
 ScanCode includes datasets (e.g. for license detection) that are dedicated
 to the Public Domain using the Creative Commons CC0 1.0 Universal (CC0 1.0)
 Public Domain Dedication: http://creativecommons.org/publicdomain/zero/1.0/</text>
-LicenseID: LicenseRef-scancode-other-permissive
```
pombredanne, line 88: Why change the order? Id, name then comment is the proper order here.

tests/formattedcode/data/spdx/license_ref/expected_with_text.tv
```diff
@@ -85,10 +83,10 @@ License for ScanCode datasets
...
-LicenseName: Other Permissive Licenses
-LicenseComment: <text>See details at https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/other-permissive.yml
+LicenseName: ScanCode acknowledgment
+LicenseComment: <text>See details at https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/scancode-acknowledgment.yml
 </text>
+LicenseID: LicenseRef-scancode-other-permissive
```
pombredanne, line 89: "scancode-acknowledgment" is not "other-permissive" so there is likely a bug here.

tests/formattedcode/data/spdx/license_ref/expected_with_text.tv
```diff
@@ -131,10 +129,10 @@ License for ScanCode datasets
 ScanCode includes datasets (e.g. for license detection) that are dedicated
 to the Public Domain using the Creative Commons CC0 1.0 Universal (CC0 1.0)
 Public Domain Dedication: http://creativecommons.org/publicdomain/zero/1.0/</text>
-LicenseID: LicenseRef-scancode-scancode-acknowledgment
-LicenseName: ScanCode generated data acknowledgment
-LicenseComment: <text>See details at https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/scancode-acknowledgment.yml
+LicenseName: Other Permissive Licenses
+LicenseComment: <text>See details at https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/other-permissive.yml
 </text>
+LicenseID: LicenseRef-scancode-other-copyleft
```
pombredanne, line 135: Same nit as above

tests/formattedcode/data/spdx/license_ref/expected_with_text.tv
```diff
@@ -176,4 +174,7 @@ License for ScanCode datasets
 =============================
 ScanCode includes datasets (e.g. for license detection) that are dedicated
 to the Public Domain using the Creative Commons CC0 1.0 Universal (CC0 1.0)
-Public Domain Dedication: http://creativecommons.org/publicdomain/zero/1.0/</text>
\ No newline at end of file
+Public Domain Dedication: http://creativecommons.org/publicdomain/zero/1.0/</text>
+LicenseName: Other Copyleft Licenses
+LicenseComment: <text>See details at https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/other-copyleft.yml
```
pombredanne: Same as above. This is incorrect

tests/formattedcode/data/spdx/unicode/expected.tv
```diff
@@ -75,4 +73,7 @@ ExtractedText: <text> * SOFTWARE LICENSE
 * ON ANY THEORY OF LIABILITY, INCLUDING, BUT NOT LIMITED TO, CONTRACT, STRICT
 * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT
 * OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
-* DAMAGE.</text>
\ No newline at end of file
+* DAMAGE.</text>
+LicenseName: Agere BSD
```
pombredanne, line 77: Let's not reorder these

pombredanne (changes requested):
Thanks!
See some comments for your review.

setup-mini.cfg
```diff
@@ -105,7 +105,7 @@ install_requires =
     pymaven_patch >= 0.2.8
     requests >= 2.7.0
     saneyaml >= 0.6.0
-    spdx_tools == 0.7.0rc0
+    spdx_tools == 0.8.0rc2
```
AyanSinhaMahapatra: Thanks for the PR.
I had a question on this being a release-candidate.

The readme says the following:
```
The main features of v0.8 are:

    experimental support for the upcoming SPDX v3 specification (note, however, that support is neither complete nor stable at this point, as the spec is still evolving)
```

SPDX v3 is not out yet and the spec could change/evolve more from this point, and
it's admittedly `neither complete nor stable`, shouldn't we wait for a stable release of 
`spdx-tools` then?
(which I'm assuming would be after SPDX v3 would be out officially) 
Does this PR change the spec version of the SPDX output produced?

armintaenzertng: Thanks for voicing your concerns, @AyanSinhaMahapatra, but I can alleviate them! :)
This PR does not introduce any SPDX v3 features into scancode and makes no use of the experimental/unstable SPDX-3 code.
We will release 0.8.0 soon and I'll update the dependency then. The 0.8.0 version is stable for SPDX v2, which is the only part of the library used in scancode.

So, to answer your last question:
No, this PR does not change the spec version of the SPDX output produced.

tests/formattedcode/data/spdx/license_known/expected.tv
armintaenzertng: Regarding this and similar questions about the change of output:
During the refactoring of the python-tools we had to rewrite most of the serializers. For this, we followed the practices employed by the [official SPDX spec examples](https://github.com/spdx/spdx-spec/blob/development/v2.3.1/examples/SPDXTagExample-v2.3.spdx), especially the ordering of information and formatting of comments (which use double ## and "Information" in most places).

src/formattedcode/output_spdx.py
armintaenzertng: This is the license information found in a single file. Information has no plural. It is also the official name in serializations, see [here](https://spdx.github.io/spdx-spec/v2.3/file-information/#86-license-information-in-file-field).

tests/formattedcode/data/spdx/license_known/expected.tv
armintaenzertng: The LicenseListVersion is getting "cleaned" in the process of regenerating test files, see [here](https://github.com/nexB/scancode-toolkit/blob/200fb10eb22218c36f68e37f2f9dc86342adaa35/tests/formattedcode/test_output_spdx.py#L151).
 I suppose the previous test file has not been regenerated since the cleaning code was introduced.

tests/formattedcode/data/spdx/license_known/expected.rdf
armintaenzertng: There was a bug in the previous spdx-tools version that did not include the correct namespace in the SpdxIds. Now, as the document namespace includes a generated UUID, it can't be used for direct comparison.  Therefore I had to introduce the code [here](https://github.com/nexB/scancode-toolkit/pull/3456/files/26f26c5f036f0e3468d5552788216a55f042907b#diff-5d7d79946ca5faabcbfa18aead7e79acc42cb26fd70a254dd0142231d1d211e6R34-R36) to strip the namespace from the data.
Still, this will at least check that all SpdxIds have the same document namespace as they wouldn't be stripped otherwise.

armintaenzertng: Sorry, I misunderstood your comment.
The "SPDXRef-001" is not missing, it [can be found a good deal further up](https://github.com/nexB/scancode-toolkit/pull/3456/files/26f26c5f036f0e3468d5552788216a55f042907b?diff=split&w=0#diff-bcb5f143ba2af18266072e7c8004418d08e8b57400f643fa8c623cafde1e4430R17) with the package that the relationship belongs to. This is similar to the [RDF relationship example in the spec](https://spdx.github.io/spdx-spec/v2.3/relationships-between-SPDX-elements/#1113-examples).

tests/formattedcode/data/spdx/license_ref/expected_with_text.tv
armintaenzertng: The order in the [official spec example](https://github.com/spdx/spdx-spec/blob/development/v2.3.1/examples/SPDXTagExample-v2.3.spdx#L292) and implemented in the serializer is: Id, extracted text, name, cross reference, comment.
So, not different from your proposal?

tests/formattedcode/data/spdx/license_ref/expected_with_text.tv
armintaenzertng: The order is: Id first, then name after that. So the name that belongs to this Id is "Other Permissive License".
We should probably include an empty line before the Id to better separate the different licenses.

tests/formattedcode/data/spdx/license_ref/expected_with_text.tv
armintaenzertng: As you reference the LicenseComment here, I'm not sure I know what you mean.
In case you meant the LicenseName, that corresponds to the Id "LicenseRef-scancode-other-copyleft", which seems fine to me.

tests/formattedcode/data/spdx/license_ref/expected_with_text.tv
armintaenzertng: Ah, actually, the empty lines get stripped during the cleaning step.

src/formattedcode/output_spdx.py
armintaenzertng: Sure, I fixed that.

src/formattedcode/output_spdx.py
armintaenzertng: done

src/formattedcode/output_spdx.py
armintaenzertng: The append function circumvented our previous type checking, so I was in the habit of avoiding that. Shouldn't be an issue anymore, I changed all occurrences to `.append()`.

src/formattedcode/output_spdx.py
armintaenzertng: As this issue was not introduced by this PR, I opened #3457 for it.

src/formattedcode/output_spdx.py
armintaenzertng: Correct me if I'm wrong, but isn't this what happens here? If we have a LicenseRef-, i.e. not an SPDX license, we enter the if clause and get the matched text.
This behavior is the same as in the previous implementation, so if you think this is a problem, I'd like to open a separate issue for this as this is not really part of this PR.

setup-mini.cfg
AyanSinhaMahapatra: Thanks for the clarifications @armintaenzertng :smile:

mjherzog: @armintaenzertng Would it be possible to correct the README for v0.08 to reflect what you explained?

armintaenzertng: The README was updated to hopefully make clear that the SPDX3 code is not used in production yet.

armintaenzertng:
@pombredanne, I believe I addressed all your comments above (apart from the reordering, for which we will continue to adhere to the specification examples) and implemented them or opened an issue if that problem already existed prior to this PR.
Also, we have officially released spdx-tools v0.8.0, I updated the version in this PR.

The CI currently fails for `macos12` only. The failing test `test_scan_does_scan_php_composer` succeeds for all other platforms and also locally for me. This seems like an issue that's unrelated to this PR.

AyanSinhaMahapatra:
> we have officially released spdx-tools v0.8.0, I updated the version in this PR.

Thanks @armintaenzertng !

> The CI currently fails for macos12 only. The failing test test_scan_does_scan_php_composer succeeds for all other platforms and also locally for me. This seems like an issue that's unrelated to this PR.

These are flukes we see sometimes, I've reran the tests and all green now.

maxhbr:
Hey @pombredanne and @AyanSinhaMahapatra , just was checking in, if I can help to get it merged. Are there still open concerns?

src/formattedcode/output_spdx.py
```diff
@@ -63,50 +67,6 @@ def logger_debug(*args):
 Output plugins to write scan results in SPDX format.
 """
 
-_spdx_list_is_patched = False
```
pombredanne, line 66: Why did you remove this?

pombredanne (changes requested):

src/formattedcode/output_spdx.py
maxhbr: The `LICENSE_MAP` was dropped in https://github.com/spdx/tools-python/commit/cb6be81c9f0705120f3565989a7e7d5fc585e129, during the refactoring. The interface used for this patching is no longer available. As the spdx-tools are now dependent on https://github.com/nexB/license-expression, maybe there is another way of patching?

src/formattedcode/output_spdx.py
pombredanne: Ack. Thanks for the clarification

src/formattedcode/output_spdx.py
pombredanne: if this is a collection, I would rather have a plural as a convention irrespective of how things may be named in SPDX.

src/formattedcode/output_spdx.py
pombredanne: @maxhbr Thanks for the clarification.... we need to find a way as ScanCode cannot rely on anything else but the licenses it knows about in a given version, i.e. the licenses must be the ones available in ScanCode toolkit not these provided by any other library (not even license-expression)

pombredanne (changes requested):

src/formattedcode/output_spdx.py
armintaenzertng: @pombredanne I believe this is not an issue. The only place in the old/current code that uses the `LICENSE_MAP` is [this](https://github.com/nexB/scancode-toolkit/blob/5dfcfcbbcda9a5bcd70a1fcf887babf661a87a03/src/formattedcode/output_spdx.py#L316), i.e. the construction of a `License` object. In the new 0.8 of the python-tools we don't have `License` objects anymore. Instead, every license is just a `LicenseExpression` object which carries no additional information about SPDX or otherwise for the purpose of serialization that we discuss here.
The logic that determines the correct SPDX license identifier happens [in these lines](https://github.com/armintaenzertng/scancode-toolkit/blob/update-spdx-tools-version/src/formattedcode/output_spdx.py#L278-L284) which are entirely ScanCode-internal. The only external non-trivial thing happening after that is [this line](https://github.com/armintaenzertng/scancode-toolkit/blob/4918b04efaf4ffa0b9c446787baa0029ce00093a/src/formattedcode/output_spdx.py#L286), which just converts the `spdx_id` string to a `LicenseExpression` object, but without any "license list logic" behind it, especially as the `licensing` is defined via `Licensing()` and not via `get_spdx_licensing()`. That `LicenseExpression` is in this case no more than a glorified string that gets added to the lists of its file/package and subsequently serialized to a simple string in the final output.

Please correct me if I'm wrong, but I see no need to keep the code fragment in question, as we make no use of ScanCode-external license lists.

maxhbr:
@pombredanne , does https://github.com/nexB/scancode-toolkit/pull/3456#discussion_r1300071128 resolve the issue?

armintaenzertng:
Just a small update: I rebased the code on the current state of the repo and updated the spdx-tools dependency to 0.8.1 (the update has no impact on the code in this PR).

armintaenzertng:
Hi @AyanSinhaMahapatra,
could you please have a look at https://github.com/nexB/scancode-toolkit/pull/3456#discussion_r1300071128 and see if that solves the open issue with this PR?

maxhbr:
hey, any updates?

tests/formattedcode/data/spdx/license_known/expected.tv
pombredanne: It does not make any sense: ## makes the file bigger for no good reason. The fact that spec examples use this weird way does not make it part of the spec FWIW. The spec states that a # is a comment not that you should use a double ##. Now this is not critical so I can live with this change but this feels like cargo culting.

Also the logical order from most significant to least significant is namespace then name. This is BTW also the order of the spec examples you reference https://github.com/spdx/spdx-spec/blob/development/v2.3.1/examples/SPDXTagExample-v2.3.spdx#L3 

I do not care too much for the change so I can live with it  but there is no rationale for this change that I can understand.

pombredanne (approved):
Thanks... I have added a new nit for your review wrt. the field ordering problem but I am goimng to merge this as-is now.

PR #3479, opened 2023-08-01
fix: allow fedora based packages

### LINKED ISSUE #831: System package scan for fedora distro does not work
I've tried to run a docker and rootfs scan of a fedora base and it does not appear to detect installed rpms.

To replicate, simply spin up a scancode.io portal and start a scan of fedora server with docker://fedora:38. It detects pypi packes but none of the installed rpms.

philcali:
Fixes https://github.com/nexB/scancode.io/issues/831

### Tasks

* [ ] Updated CHANGELOG.rst


src/packagedcode/rpm.py
```diff
@@ -231,8 +231,7 @@ class RpmInstalledNdbDatabaseHandler(BaseRpmInstalledDatabaseHandler):
 class RpmInstalledSqliteDatabaseHandler(BaseRpmInstalledDatabaseHandler):
     # used by newer RHEL/CentOS/Fedora
     datasource_id = 'rpm_installed_database_sqlite'
-    path_patterns = ('*var/lib/rpm/rpmdb.sqlite',)
-    filetypes = ('berkeley',)
+    path_patterns = ('*var/lib/rpm/rpmdb.sqlite', '*usr/lib/sysimage/rpm/rpmdb.sqlite',)
```
pombredanne: We could also relax it even more like in:
```suggestion
    path_patterns = ('*rpm/rpmdb.sqlite',)
```

pombredanne (changes requested):
Thanks! This is exactly what is needed! :heart: 
Do you mind adding your DCO signoff?

Also we need some tests to avoid a regression.
It does not have to be a full end-to-end test as it would be enough to "recognize" an A SQLite db ... may something like one of these tests? https://github.com/nexB/scancode-toolkit/blob/develop/tests/packagedcode/test_recognize.py that could use a super small rpmdb (eventually trimmed down to get only one or two RPMs)

Another test we would need is an end-to-end test from the CLI using the --system-package option,

src/packagedcode/rpm.py
philcali: I like that. I'll make the change. Also. I have a pending change to support rpm-ostree variants too.

pombredanne: Thanks!

philcali: I created a unit test, but I noted that unrelated tests seem to fail (unsure about those). Also, I was not sure about the `--system-package` testing. Happy to author, but would need a pointer to the starting thread.

tests/packagedcode/test_recognize.py
```diff
@@ -196,3 +196,9 @@ def test_recognize_go_sum(self):
         packages = recognize_package_data(test_file)
         assert packages
         assert isinstance(packages[0], models.PackageData)
+
+    def test_recognize_rpmdb_sqlite(self):
+        test_file = self.get_test_loc('rpm/rpmdb.sqlite')
```
pombredanne, line 201: I wonder if we could get a much smaller test file for this? we only need a few packages not everything.

pombredanne (changes requested):
LGTM! Let's try to find a smaller test SQLite DB and let's merge.

philcali:
Updated the DB! Thanks!

AyanSinhaMahapatra (approved):
LGTM!
Thanks for the fix @philcali 
Merging!

PR #3495, opened 2023-08-26
Bump commoncode to v31.0.3

JonoYang:
This PR bumps commoncode to v31.0.3

This version of commoncode fixes a VirtualCodebase creation issue when there is a directory under the root with the same name as the root directory itself. https://github.com/nexB/commoncode/issues/57


AyanSinhaMahapatra (approved):
Thanks++ for the fix @JonoYang, this should have been merged allright before, sorry for missing.

Issue #3541, opened 2023-10-05
Synchronize with SPDX License List 3.22


DennisClark:
SPDX License List Version: 3.22 2023-10-05  has been released at https://spdx.org/licenses/ 

Release details are at https://github.com/spdx/license-list-XML/releases/tag/v3.22 

Please synchronize with this latest release in SCTK and LicenseDB.

AyanSinhaMahapatra:
This is completed, closing!

PR #3554, opened 2023-10-17
Update to spdx 3.22

AyanSinhaMahapatra:
These license additions and license/rule updates are added
automatically using the new spdx-synclic script. Features are:

* Update old licenses on `perfect-detection`
* Add new licenses where we don't have good detection at all
* Deprecate generic rules when we have perfect detections to those,
  as we are adding licenses for those
* In case of multiple detections/matches we are adding a new license and
  adding the detection details to the new file as notes to review there

Also reviewed the license additions and removed notes in resolved
licenses. Added category and owners, deprecate rules as required.

Reference: https://github.com/spdx/license-list-XML/releases/tag/v3.22
Reference: https://github.com/nexB/scancode-toolkit/issues/3541

### Tasks

* [x] Updated CHANGELOG.rst


src/licensedcode/data/licenses/bsd-3-clause-flex.LICENSE
```diff
@@ -0,0 +1,134 @@
+---
+key: bsd-3-clause-flex
+short_name: BSD 3-Clause Flex variant
+category: Permissive
+owner: Regents of the University of California
+name: BSD 3-Clause Flex variant
+notes: |
```
AyanSinhaMahapatra: See https://github.com/nexB/scancode-toolkit/tree/develop/src/licensedcode/data/licenses/flex-2.5.LICENSE which has a slightly different license text, but has rules which are matching.

src/licensedcode/data/licenses/bsd-3-clause-hp.LICENSE
```diff
@@ -0,0 +1,73 @@
+---
+key: bsd-3-clause-hp
+short_name: Hewlett-Packard BSD variant license
+name: Hewlett-Packard BSD variant license
+category: Permissive
+owner: HP - Hewlett Packard
+notes: |
```
AyanSinhaMahapatra: bsd-3-clause-hp.LICENSE:

This just has HP over a bsd-new?

AyanSinhaMahapatra (commented):
Things remaining:

- [x] Add also HPND-Pbmplus and HP-1989 licenses.

for reference:
HPND-Pbmplus:

https://github.com/spdx/license-list-data/blob/main/text/HPND-Pbmplus.txt
https://github.com/spdx/license-list-data/blob/main/text/xlock.txt
See https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/rules/libpbm_1.RULE
and https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/libpbm.LICENSE

HP-1989:
https://github.com/spdx/license-list-data/blob/main/text/HP-1989.txt
https://github.com/spdx/license-list-data/blob/main/text/HP-1986.txt
See https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/osf-1990.LICENSE
and https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/rules/osf-1990_3.RULE
and https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/rules/osf-1990_6.RULE

- [x] Review more comments above and remove notes with detection details.
- [x] Remove copyrights
- [x] Add more license rules for these licenses.
- [x] Review

src/licensedcode/data/licenses/bsd-3-clause-flex.LICENSE
DennisClark: @AyanSinhaMahapatra   Rather than create a new scancode license, I think that we should update the existing scancode flex-2.5 license to match this new SPDX license: 
1. replace the license text in the scancode licensedb
2. update the spdx_license_key to BSD-3-Clause-flex to replace the current LicenseRef

I am aware that in the license text we normally avoid copyrights and references to specific organizations and code projects, but I think that is appropriate for this license, which we would call a component_license.

src/licensedcode/data/licenses/bsd-3-clause-hp.LICENSE
DennisClark: @AyanSinhaMahapatra  your point is valid, I think, although there might be other minor differences, but please go ahead and create this one, since we really need to stay in synch with the SPDX list.  Be sure to include the HP references in the text.

DennisClark (changes requested):
see my notes regarding the flex license.  Everything else looks good.

DennisClark:
@AyanSinhaMahapatra  Regarding HPND-Pbmplus,  I think we need to update 
https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/libpbm.LICENSE
to change the spdx_license_key from xlock to HPND-Pbmplus
and make sure that the texts are the same
and
we need to create a new scancode license for xlock to correspond with the spdx xlock 

In both cases, I think the RULE specs need to be rather aggressive regarding required text to identify a good detection.

I hope that this is similar to what you were thinking?

DennisClark:
@AyanSinhaMahapatra   Regarding HP-1989, I think we need to update 
https://github.com/nexB/scancode-toolkit/blob/develop/src/licensedcode/data/licenses/osf-1990.LICENSE 
to change the spdx_license_key from the current licenseref to  HP-1989  
but leave the scancode license text the way it is now (with the various company names genericized).  

and 

we need to create a new license in scancode to correspond with HP-1986  
using a scancode license key of hp-1986

AyanSinhaMahapatra:
@DennisClark regarding Regarding HP-1989 what you said makes sense, 
as we had already `spdx_license_key: HP-1989` for `key: osf-1990` (this was added in SPDX 3.20)
> but leave the scancode license text the way it is now (with the various company names genericized).
This makes sense, I'll update this.

But in case of HPND-Pbmplus, we already had the spdx license key as `spdx_license_key: xlock` for `key: libpbm` (this was added in spdx 3.20 too) so if we `change the spdx_license_key from xlock to HPND-Pbmplus` we will make a change in the SPDX license key (in the other case above we are not changing). As we preserve old spdx license keys in `other_spdx_license_keys` when we change these keys, here the spdx key xlock would then be present at two places, the `other_spdx_license_keys` of the old license and at the `spdx_license_key` of the new license which would be inconsistent.

If we instead change the license texts so that we have `key: libpbm` and `spdx_license_key: xlock` and then the new license added is `key: hpnd-pbmplus` and `spdx_license_key: HPND-Pbmplus` would this be incorrect? We will also change the license text to ke consistent with spdx in this case. So basically the issue is we either have to change license texts or change the old key ->  spdx_key relationship, and which one is less problematic? What do you think?

DennisClark:
@AyanSinhaMahapatra  on further reflection, your remarks starting with "If we instead change the license texts ... "  make more sense; please go ahead and do that.   

The good news in all this is that these licenses are extremely rare, so I doubt that we'll be causing much trouble with this solution, if any.

AyanSinhaMahapatra:
All green! Merging.

PR #3555, opened 2023-10-20
npm: support aliases in yarn lock v1

schischi:
ScanCode does not handle aliases in `yarn.lock v1` file.
Aliases have the form: `<alias-package>@npm:<package>`
More info about aliases: https://classic.yarnpkg.com/lang/en/docs/cli/add/#toc-yarn-add-alias

The current code incorrectly parses aliases and end up throwing an exception, resulting in empty results.
A single alias in a yarn.lock file is gonna make the parser return 0 packages.

This patch adds the logic to handle them by simply dropping the alias part, and just keeping the package part.

Test plan: updated the unit tests.

Signed-off-by: Adrien Schildknecht <adrs@fb.com>


pombredanne (approved):
Thank you ++... LGTM!

pombredanne:
@schischi do you know if these aliases also show up in yarn v2?

AyanSinhaMahapatra:
@pombredanne it does not seem like we have this in yarn v2 as there is no mention of aliases in the doc for yarn v2 at https://yarnpkg.com/cli/add similar to https://classic.yarnpkg.com/lang/en/docs/cli/add/#toc-yarn-add-alias, which is the doc for yarn v1.

PR #3602, opened 2023-11-21
Support cargo workspaces

### LINKED ISSUE #3598: Do not crash with Cargo workspaces
We need to support correctly workspaces:
https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html

> A workspace is a set of packages that share the same Cargo.lock and output directory. 

Here is an example: https://github.com/tauri-apps/tauri/blob/dev/Cargo.toml

AyanSinhaMahapatra:
Fixes #3598 

### Tasks


src/packagedcode/cargo.py
```diff
@@ -82,22 +163,23 @@ def parse(cls, location):
...
+    # Fields in PackageData model: Fields in cargo
+    "homepage_url": "homepage",
+    "vcs_url": "repository",
+    "keywords": "categories",
+    "extracted_license_statement": "license",
+    # These are fields carried over to avoid re-detection of licenses
+    "license_detections": "license_detections",
+    "declared_license_expression": "declared_license_expression",
+    "declared_license_expression_spdx": "declared_license_expression_spdx",
+}
```
JonoYang, line 182: Formatting: add an extra new line here

JonoYang (commented):
@AyanSinhaMahapatra Looks good to me, I left a comment about adding a newline

src/packagedcode/cargo.py
```diff
@@ -31,11 +106,14 @@ class CargoTomlHandler(models.DatafileHandler):
...
     def parse(cls, location):
         package_data = toml.load(location, _dict=dict)
-
         core_package_data = package_data.get('package', {})
+        workspace = package_data.get('workspace', {})
 
         name = core_package_data.get('name')
         version = core_package_data.get('version')
+        if isinstance(version, dict) and "workspace" in version:
+            version = "workspace"
```
pombredanne: is this really the case? workspace is used as version?

pombredanne (commented):
LGTM.... with a nit wrt. using "workspace" as a placeholder value?
Could we use something better, may be using extra data instead for fields that should be copied back from the workspace in a lower package?

Also would the top level workspace exist as a package of its own or not?

src/packagedcode/cargo.py
AyanSinhaMahapatra: Yes! See https://github.com/nexB/scancode-toolkit/blob/ff84e1cc04903c6bbb9d1eaae3794fe61a718594/tests/packagedcode/data/cargo/cargo_toml/tauri-examples/Cargo.toml 

And you're right this is weird to put this in the version (or other fields, as if they are not overwritten by the workspace values for some bug, they remian), using extra_data instead here.

AyanSinhaMahapatra:
> Could we use something better, may be using extra data instead for fields that should be copied back from the workspace in a lower package?

Yeah this is much better and correct, doing this now.

> Also would the top level workspace exist as a package of its own or not?

So I did consider doing this, but the top-level workspace usually does not have any purl fields there, only some extra data wrt source/licenses, see https://github.com/nexB/scancode-toolkit/blob/support-cargo-workspaces/tests/packagedcode/data/cargo/cargo-with-workspace/Cargo.toml#L12 for example. So for now we are not adding top-level packages here, but we can revisit this if we get better examples. But if there are other manifests of other flavours, we'll be adding packages for those.

AyanSinhaMahapatra:
I've removed the improved package assembly part from this PR and pushed this as a separate branch at: https://github.com/nexB/scancode-toolkit/tree/update-package-assembly, since this was not ready to merge. Otherwise this was reviewed and approved, so merging!

PR #3607, opened 2023-11-23
nuget: parse .csproj and packages.config files

### LINKED ISSUE #2392: Add support for extra visual studio and Nuget "manifests"
There are several files we could squeeze data from:
- packages.config https://docs.microsoft.com/en-us/nuget/reference/packages-config
- https://docs.microsoft.com/en-us/nuget/consume-packages/migrate-packages-config-to-package-reference
- https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-properties
- https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build
- vbproj, csproj, fsproj VS "solutions" files

schischi:
Add support for 2 manifest files:
  - packages.config: XML file holding the list of package dependencies;
  - *.csproj: describe a project with dependencies (NuGet package references);

The manifest files do not contain a lot of info, but enough to build a package-url.

fixes #2392


pombredanne:
@schischi Thanks you ++
This is needed and long overdue alright!
There also is another tool -- if .NET and C# are your poison of choice -- at https://github.com/nexB/nuget-inspector which is written in C#  and is designed to also resolve .NET dependencies. It handles many if not all current and legacy manifests formats including all the *.*proj variants, packages.config and various more recent JSON files that dotnet, msbuild and Visual Studio dabble with.  See https://github.com/nexB/nuget-inspector/blob/main/src/nuget-inspector/ProjectScanner.cs for some entry point.

It is much more involved and bigger than a plain parser as it uses the native NuGet and .NET libraries for better or worse.

The inspector also has an extensive test suite which may come handy for the staic analysis here.

eeisegn:
Is there a plan to merge this into the mainline?

AyanSinhaMahapatra (changes requested):
@schischi thanks a lot! And sorry about the super late reply. :sweat_smile: 

Parsing of these manifests are actually supported at https://github.com/aboutcode-org/nuget-inspector (along with dependency resolution and getting package data for those), but it probably makes sense to have something minimal here too as nuget inspector would be harder to integrate in SCTK context. This is why we were not sure about merging this PR here as some efforts were ongoing and we want to avoid duplicate functionality as much as possible. But IMHO it makes sense to add this here otherwise. 


Some updates are required before we merge this:
1. we want to create a single package from a `.csproj` file (we get the package name from the filename: `*.csproj`) and populate dependencies too if they are present
2. we want to create similarly a single package with no purl fields and only dependencies from `packages.config` files
3. If we have a `.csproj` file and a `packages.config` file in the same directory we want to perform package assembly to merge the package data from these files (so we will get the name from the former and dependencies from the later for example)
4. If the dependencies are pinned we want to also add a skinny package_data object at the `resolved_package` attribute inside each dependency

References:
1. We have lots of examples at https://github.com/aboutcode-org/nuget-inspector/blob/main/tests/data/ both individual files and pairs of these two manifests.
2. See also https://github.com/aboutcode-org/scancode-toolkit/pull/3825 which we added for https://github.com/aboutcode-org/scancode.io/issues/1263, these are fully resolved dependencies, and we basically want to have the same structure here.

@eeisegn thanks for the ping!

PR #3628, opened 2023-12-27
Fix-up an accidental use of SPDX's `WITH` operator in a rule

fviernau:
Adjust a rule to return `bsd-new AND facebook-patent-rights-2 OR gpl-2.0` instead of `bsd-new WITH facebook-patent-rights-2 OR gpl-2.0`.

See individual commits.


AyanSinhaMahapatra (approved):
Thanks @fviernau!
LGTM! Merging.

PR #3650, opened 2024-02-07
Support conan in packagedcode

### LINKED ISSUE #2388: Add support for conan.io C/C++ packages
### Description
Conan is a C/C++ package manager and we should be able to collect metadata from it. It happens also to be coded in Python
See https://github.com/conan-io 

@marco-m-pix4d ping

keshav-space:
- Fixes #2388 

### Tasks


src/packagedcode/conan.py
```diff
@@ -0,0 +1,281 @@
...
+    def visit_Assign(self, node):
+        if not node.targets or not isinstance(node.targets[0], ast.Name):
+            return
+        if not node.value or not (
+            isinstance(node.value, ast.Constant) or isinstance(node.value, ast.Tuple)
+        ):
+            return
+        variable_name = node.targets[0].id
+        values = node.value
+        if variable_name == "name":
```
AyanSinhaMahapatra: Maybe we can have a dict mapper for the value mappings and then use setattr() instead of this long if-elif statement.

src/packagedcode/conan.py
```diff
@@ -0,0 +1,281 @@
...
+                    type=cls.default_package_type,
+                    primary_language=cls.default_primary_language,
+                    namespace=None,
+                    version=version,
+                    download_url=url,
+                    sha256=sha256,
+                )
+
+    @classmethod
+    def assemble(
```
AyanSinhaMahapatra, line 217: Usually we have the assemble function under a base class and then inherit that in all other datafile handlers if these manifests can exist on their own and also together, to cover all the cases, would that also benifit us here?

src/packagedcode/conan.py
```diff
@@ -0,0 +1,281 @@
...
+            cls.assign_package_to_resources(
+                package=package,
+                resource=resource,
+                codebase=codebase,
+                package_adder=package_adder,
+            )
+        yield resource
+
+
+def is_constraint_resolved(constraint):
```
AyanSinhaMahapatra, line 270: These functions would benefit from having docstrings for verbosity

tests/packagedcode/data/conan/recipes/boost/manifest/conanfile.py
```diff
@@ -0,0 +1,1831 @@
+from conan import ConanFile
+from conan.errors import ConanException, ConanInvalidConfiguration
+from conan.tools.apple import is_apple_os, to_apple_arch, XCRun
+from conan.tools.build import build_jobs, check_min_cppstd, cross_building, valid_min_cppstd, supported_cppstd
```
AyanSinhaMahapatra: We can probably remove some unused sections from these huge conanfile.py test files safely without any effect on the parsing.

src/packagedcode/conan.py
```diff
@@ -0,0 +1,281 @@
...
+        for version, source in sources.items():
+            sha256 = source.get("sha256", None)
+            urls = []
+            source_urls = source.get("url")
+            if isinstance(source_urls, str):
+                urls.append(source_urls)
+            elif isinstance(source_urls, list):
+                urls.extend(source_urls)
+
+            for url in urls:
```
AyanSinhaMahapatra: Here all the URLs in one section is for one package, so we should yield one package only? Possibly with the other urls stored somewhere else? Extra data maybe? Two packages having the same sha256 value probably should not have two seperate entries. Even after combining the package data from conandata.yml and conanfile.py we have duplicate package instance entries with same purls, which should not be the case.

Another thing, but not too important probably, as I assume the conandata.yml files are not seen in isolation:
Also if it is clear and is a standard for conan package archive filenames having the name and version, we should probably also try to get the name out of it so we have valid purls? This would not have any impact on assembly as they would be merged allright if the purl is same, but would have purls also in the package data output.

src/packagedcode/conan.py
```diff
@@ -0,0 +1,281 @@
...
+            version=parser.version,
+            description=parser.description,
+            homepage_url=parser.homepage_url,
+            keywords=parser.keywords,
+            extracted_license_statement=parser.license,
+            dependencies=dependencies,
+        )
+
+
+class ConanDataHandler(models.DatafileHandler):
```
AyanSinhaMahapatra, line 175: I saw some different kinds of conandata.yml too like this  https://github.com/Ultimaker/Cura/blob/main/conandata.yml is this seen often? do we support this? Is this a different version of conandata?

AyanSinhaMahapatra (changes requested):
@keshav-space Thanks++
Looking great already, I had some questions and comments for you to consider, see below.

tests/packagedcode/data/plugin/help.txt
```diff
@@ -174,6 +174,20 @@ Package type:  composer
   description:       PHP composer lockfile
   path_patterns:    '*composer.lock'
 --------------------------------------------
```
AyanSinhaMahapatra, line 176: Run also this script: https://github.com/nexB/scancode-toolkit/blob/2388-support-conan/docs/scripts/regen_package_docs.sh to update available package parsers script at `docs/source/reference/available_package_parsers.rst`

src/packagedcode/conan.py
keshav-space: Good idea :)

src/packagedcode/conan.py
keshav-space: conanfile.py can exist on its own, but conandata.yml should not exist in isolation. We need the custom assembly only in the case of conandata.yml

src/packagedcode/conan.py
keshav-space: > Here all the URLs in one section is for one package, so we should yield one package only? Possibly with the other urls stored somewhere else? Extra data maybe? Two packages having the same sha256 value probably should not have two seperate entries. Even after combining the package data from conandata.yml and conanfile.py we have duplicate package instance entries with same purls, which should not be the case.

Yes, you are right. We should not duplicate the package based on different download urls. I discussed this with @pombredanne , and he suggested that we should pick the first download url and revisit this issue(https://github.com/nexB/scancode-toolkit/issues/3654) later on if we feel the need to handle it differently.

> Another thing, but not too important probably, as I assume the conandata.yml files are not seen in isolation:
> Also if it is clear and is a standard for conan package archive filenames having the name and version, we should probably also try to get the name out of it so we have valid purls? This would not have any impact on assembly as they would be merged allright if the purl is same, but would have purls also in the package data output.

It's not standard practice to put the package name in `conandata.yml`. conandata.yml ideally only contains information that is supposed to change or be added over time, such as version, download_url, checksums, and it's unlikely that the name would be changed over time.

src/packagedcode/conan.py
keshav-space: Technically speaking `conandata.yml` can contain arbitrary information, but we're only interested in the part specified here https://docs.conan.io/2/tutorial/creating_packages/handle_sources_in_packages.html#using-the-conandata-yml-file. This section of conandata.yml specifies how to provide the url and sha256 for each version.
Apart from that, developers can specify any other information in this YML file that they might need to automate the `conanfile.py`.

tests/packagedcode/data/plugin/help.txt
keshav-space: Done!

src/packagedcode/conan.py
pombredanne: Which brings to light the limits of a single-file-at-a-time approach. Should we start looking outside of the current file? ScanCode is not design for this at all though for now.

pombredanne (approved):
LGTM! we just need a changelog

src/packagedcode/conan.py
keshav-space: In case of conan, we need to process 2 files to obtain the package data. However, if we encounter a scenario where we have more than 2 files that need to be processed to obtain package data, processing multiple files at once would be helpful. Otherwise, the assembly will become a bit complex.

AyanSinhaMahapatra (approved):
LGTM! Thanks @keshav-space 
Merging!

Issue #3547, opened 2023-10-11
Refine referenced_filenames in license rules


pombredanne:
We should improve license rules "referenced_filenames" support.

Today we can follow the referenced "referenced_filenames" of a matched license rule and assigned the license of referenced file to the matched that referenced it.

For instance, with a match to an `unknown-license-reference` rule with text `see license in license.txt`, a detection would group this match and the match(es) in `license.txt`.

This works in some cases and not at all in some other cases. But in earnest, there are at least two different styles of file references that we are mixing up today:

1. The "See license in LICENSE file" style where the rule is for an unknown license (reference) and the file reference is either a notice or license text. Here we should deference and get the license found in the LICENSE file as the license of the whole detection. 

The license in the referenced file  becomes the new license of this match, overriding its license expression.

2. The "Licensed under the LGPL. See COPYING for details" style where the rule is a notice and the file reference points to a license text or a notice. Here we should merged the license in this rule and the license in the file reference in a smart way. For instance "LGPL" here is for LGPL-2.0 or later, but the COPYING may be for a an LGPL-2.1 and the apparent contradiction should be resolved in the detection. 

The license in the referenced file only supports the license of this match and could either:
- refine and precise the license of the match (like in the case above).
- merely supports a notice. 

For instance a notice may be: "Licensed under the GPLv2 or MIT. See GPL.txt and MIT.txt for details". Here the expression is a choice of `GPL-2.0 OR MIT`. If we combine the `GPL-2.0` in GPL.txt and the `MIT` in MIT.txt we end up, overriding this expression with a a new detection and an expression of `GPL-2.0 AND MIT` which is incorrect..

Another example is the case where we have an LGPL-3.0 notice and where there is a n LGPL and a GPL text where the GPL is only supporting the LGPL.

AyanSinhaMahapatra:
> where there is a n LGPL and a GPL text where the GPL is only supporting the LGPL.

@pombredanne are there other cases of dependent keys like this?

And does other examples of case 2 come to mind where we don't want to use referenced matches?

pombredanne:
I bumped into this again, and referenced_filenames is not always helping with accurate license detection:

When a notice that contains referenced_filenames is correctly detected and non-ambiguous, we should improve detection processing
- for instance, notice to GPL-2.0-or-later points to COPYING with GPL-2.0-only: we should still have a GPL-2.0-or-later
[...]

PR #3681, opened 2024-02-27
Refine referenced filenames #3547

AyanSinhaMahapatra:
This PR refine how we handle "referenced filenames" that we track in license detection rules.
We use improved heuristics to decide if a something detected like "See license in COPYING" will follow the license found in the file "COPYING" or not follow and reuse it in the detection, as suggested in the issue:
- https://github.com/nexB/scancode-toolkit/issues/3547

### Tasks


src/licensedcode/detection.py
```diff
@@ -1356,6 +1375,53 @@ def has_references_to_local_files(license_matches):
     )
 
 
+def use_referenced_license_expression(referenced_license_expression, license_detection, licensing=Licensing()):
```
pombredanne, line 1378: What about this:
```
    Return True if the ``license_detection`` LicenseDetection should include
    the matches represented by the ``referenced_license_expression`` string.
    Return False otherwise.
    This determination is based on multiple factors detailed in comments below. 
    
    Used when we have a ``license_detection`` with a match to a license rule like
    "See license in COPYING" and where the ``referenced_license_expression`` is the
    expression found in the "COPYING" file. This  expression is the combined
    expression for all matches found in "COPYING".

    The actual combination of other matches in ``license_detection`` is the
    responsibility of the caller.
```

src/licensedcode/detection.py
```diff
@@ -1356,6 +1375,53 @@ def has_references_to_local_files(license_matches):
...
+    same_expression = referenced_license_expression == license_detection.license_expression
+    same_license_keys = license_keys == referenced_license_keys
+
+    # If we have the same license keys but not the same license expression then
+    # the reference could merely be pointing to notices, combining which produces
+    # a different expression, and the original detection is correct
+    if same_license_keys and not same_expression:
+        return False
+
+    if len(referenced_license_keys) > 5:
```
pombredanne, line 1427:
```suggestion
    # when there are many license keys in an expression, and there are no
    # unknown or other cases, we cannot safely conclude that we should
    # follow the license in the referenced filenames. This is likely
    # a case where we have larger notices and several combined expressions, 
    if len(referenced_license_keys) > 5:
```

tests/licensedcode/data/plugin_license/ignored_reference/or_and_problem.expected.json
```diff
@@ -0,0 +1,264 @@
...
... software interchange.\n\n  If distribution of object code is made by offering access to copy\nfrom a designated place, then offering equivalent access to copy the\nsource code from the same place satisfies the requirement to\ndistribute the source code, even though third parties are not\ncompelled to copy the source along with the object code.\n\n  5. A program that contains no derivative of any portion of the\nLibrary, but is designed to work with the Library by being compiled or\nlinked with it, is called a \"work that uses the Library\".  Such a\nwork, in isolation, is not a derivative work of the Library, and\ntherefore falls outside the scope of this License.\n\n  However, linking a \"work that uses the Library\" with the Library\ncreates an executable that is a derivative of the Library (because it\ncontains portions of the Library), rather than a \"work that uses the\nlibrary\".  The executable is therefore covered by this License.\nSection 6 states terms for distribution of such executables.\n\n  When a \"work that uses the Library\" uses material from a header file\nthat is part of the Library, the object code for the work may be a\nderivative work of the Library even tho
```
pombredanne, line 161: Here we do not have what I expect:
The notice "lgpl-2.1 OR mpl-1.1" has references to license texts that are full texts and we should end up with "lgpl-2.1 OR mpl-1.1" and not "lgpl-2.1 AND mpl-1.1"

pombredanne (changes requested):
Thanks!
You may want to add a few other test cases... the current case that passes is not correct.

src/licensedcode/detection.py
AyanSinhaMahapatra: Done, thanks!

src/licensedcode/detection.py
AyanSinhaMahapatra: Added, thanks!

tests/licensedcode/data/plugin_license/ignored_reference/or_and_problem.expected.json
AyanSinhaMahapatra: We were doing this incorrectly in the case of multiple referenced files, where we were considering whether to keep the matches individually for each file, and hence the bug. I've update this to consider all the referenced files together, and we're getting `lgpl-2.1 OR mpl-1.1` correctly now.

pombredanne (approved):
Thanks! Go!

PR #3693, opened 2024-03-08
Add support to replace directive in go.mod

### LINKED ISSUE #3492: Support directives in go.mod
## Short Description
We do not support all go.mod features such as replace directives in go.mod files and we should.
See:
- https://github.com/golang/gofrontend/tree/d04b024021bb7dbaa434a6d902bd12beb08e315f/libgo/go/golang.org/x/mod/modfile
- https://go.dev/ref/mod#go-mod-file
- and https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/go_mod.py

And an example see https://github.com/golang/tools/blob/9425c2e70961cb8c2a3b5764658fb70e03d194e4/gopls/go.mod and https://github.com/golang/tools/blob/9425c2e70961cb8c2a3b5764658fb70e03d194e4/go.mod

shravankshenoy:
Fixes #3492 

## About replace directive in go.mod
Replaces the content of a module at a specific version (or all versions) with another module version or with a local directory. Syntax as shown below

`replace module-path [module-version] => replacement-path [replacement-version]`

Detailed info at https://go.dev/doc/modules/gomod-ref

## Changes made
- Added regex to get module-path, module-version, replacement-path and replacement-version from replace directive
- Wrote conditions to handle single line and multi line replace directives
- Reran opencensus test to account for the replace directive
- Added new test using milvus go.mod to test the replace directive functionality
`


src/packagedcode/go_mod.py
```diff
@@ -188,6 +259,7 @@ def parse_gomod(location):
 
     gomods.require = require
     gomods.exclude = exclude
+    gomods.replace = replace
```
AyanSinhaMahapatra, line 262: Also add `replace` as an attribute of `class GoModule`

src/packagedcode/go_mod.py
```diff
@@ -158,6 +168,67 @@ def parse_gomod(location):
...
+                )
+            )
+
+            replace.append(
+                GoModule(
+                    namespace=replacement_namespace,
+                    name=replacement_name,
+                    version=parsed_rep_link.group("replacement_version"),
+                )
+            )
```
AyanSinhaMahapatra, line 228: This is not correct, as we are treating the module to be replaced and the replacement module the same way. I think the first module in the replace directive (the module to be replaced) should be added as a dependency with scope as `exclude` and the second module in the replace directive (the replacement module) should be added with scope as `require`.

@pombredanne what do you think?

tests/packagedcode/data/golang/gomod/milvus/go.mod
```diff
@@ -0,0 +1,255 @@
+module github.com/milvus-io/milvus
```
AyanSinhaMahapatra, line 1: this is quite a large test file, would be nice to truncate this to only have the essential portion required to test the functionality, or find a shorter example. Also add the go.mod files linked in the main issue as tests.

tests/packagedcode/data/golang/gomod/milvus/go.mod
```diff
@@ -0,0 +1,255 @@
...
+	k8s.io/apimachinery v0.28.6 // indirect
+	sigs.k8s.io/yaml v1.3.0 // indirect
+)
+
+replace (
+	github.com/apache/pulsar-client-go => github.com/milvus-io/pulsar-client-go v0.6.10
+	github.com/bketelsen/crypt => github.com/bketelsen/crypt v0.0.4 // Fix security alert for core-os/etcd
+	github.com/expr-lang/expr => github.com/SimFG/expr v0.0.0-20231218130003-94d085776dc5
+	github.com/go-kit/kit => github.com/go-kit/kit v0.1.0
+	github.com/milvus-io/milvus/pkg => ./pkg
```
AyanSinhaMahapatra, line 250: In the case where the replace directive contains a reference to a local directory, we can try to resolve this in the `assemble` step of the datafile handler, if there are package manifests in the local directory with package data for the replacement module. 

1. On the `parse` function side we store the local references in the `extra_data` attribute somehow, with the referenced path.
2. On the assembly side, we try to walk the codebase for this referenced directory.
3. If the referenced directory is present in the codebase, and we have parsed go package data present there, we use that to populate the required dependency list.

AyanSinhaMahapatra (changes requested):
@shravankshenoy thanks for the PR!

I've added a couple of comments for your consideration, to the best of my knowledge (since I'm not too familiar with go build systems), please review and update the PR accordingly, as applicable.

src/packagedcode/go_mod.py
pombredanne: We need to track the origin and renamed module separately when parsing alright.
And then we would need to only report the replaced modules and not both... and there could be trailing comments in replace that are also useful to track

AyanSinhaMahapatra:
@shravankshenoy  gentle ping

PR #3701, opened 2024-03-19
Release prep v32.1.0

AyanSinhaMahapatra:
### Tasks


requirements.txt
```diff
@@ -1,18 +1,19 @@
-attrs==22.1.0
+attrs==23.2.0
 banal==1.0.6
-beautifulsoup4==4.11.1
+beautifulsoup4==4.12.3
 binaryornot==0.4.4
+beartype==0.17.2
```
AyanSinhaMahapatra, line 5: Note that 3 new packages has been added to the requirements list from dependencies of spdx-tools:
- `beartype==0.17.2`
- `semantic-version==2.10.0`
- `uritools==4.0.2`
These are from https://github.com/spdx/tools-python/blob/v0.8.2/pyproject.toml#L27 and have been added sometime in between in spdx-tools v0.8.x
They seem to be harmless and does not have any dependencies of their own, and also maintained well. See their source repos for more info:
- https://github.com/beartype/beartype/
- https://github.com/rbarrois/python-semanticversion
- https://github.com/tkem/uritools/

AyanSinhaMahapatra (commented):
This is now ready to review and release.
See all the release scripts passing at 
https://github.com/AyanSinhaMahapatra/scancode-toolkit/actions/runs/8388278250

requirements.txt
pombredanne: Shrikes. Me does not want more dependencies, especially things that are possibly problematic like the python-semanticversion lib ... we use some of it in univers and it is problematic. Oh well. This may bring in some conflicts. We'll manage.

pombredanne (approved):
Looking fine!
GO!!!

requirements.txt
AyanSinhaMahapatra: We use `semantic-version==2.8.5` at [univers](https://github.com/nexB/univers/blob/main/requirements.txt#L4C1-L4C24) should we use the same here?
From https://github.com/rbarrois/python-semanticversion/blob/master/ChangeLog it does not look too problematic to upgrade to 2.10.0, but I might be wrong.

@pombredanne @keshav-space what do you think? It would be nicer to look into this now so we don't have dependency conflicts that we can't get around later in other places.

AyanSinhaMahapatra: Should I just use the version used in univers, 2.8.5 to avoid issues?
Because I get a couple of errors when running  the univers test suite against semantic-version==2.10.0 :
```
FAILED tests/test_versions.py::test_enhanced_semantic_version - AssertionError: assert EnhancedSemanticVersion('1.2.3-pre.1+build.1') > EnhancedSemanticV...
FAILED tests/test_versions.py::test_semver_version - AssertionError: assert SemverVersion(string='1.2.3+42') > SemverVersion(string='1.2.3+23')
```

pombredanne: This library is problematic and I cannot fathom why it is used in SPDX
I filed https://github.com/spdx/tools-python/issues/802 

On the univers side, see https://github.com/nexB/univers/issues/121
In all cases, pin the same version as used in univers to avoid conflicts.

AyanSinhaMahapatra: Thanks @pombredanne I'll run the release tests once again with this just in case.

AyanSinhaMahapatra: Note that `2.8.5` is also passing the SCTK tests and the release tests:
- https://dev.azure.com/nexB/scancode-toolkit/_build/results?buildId=13149&view=results
- https://github.com/AyanSinhaMahapatra/scancode-toolkit/actions/runs/8391889303

keshav-space: After trying out the `2.10.0` it seems like https://github.com/nexB/univers/issues/121  is still a bug.

```python
>>> from semantic_version import Version
>>> v = Version("1.0.0-beta")
>>> v.next_major()
Version('1.0.0')
>>> v.next_minor()
Version('1.0.0')
>>> v.next_patch()
Version('1.0.0')
```
Also https://github.com/rbarrois/python-semanticversion/issues/132 is still not fixed, although the changelog on python-semanticversion says that this has been fixed.
```python
>>> from semantic_version import Version
>>> v1 = Version("2.0.12+123")
>>> v2 = Version("2.0.12+1")
>>> v3 = Version("2.0.12+321")
>>> v4 = Version("2.0.12+22")
>>> sorted([v1,v2,v3,v4])
[Version('2.0.12+123'), Version('2.0.12+1'), Version('2.0.12+321'), Version('2.0.12+22')]
>>> sorted([v4,v2,v1,v3])
[Version('2.0.12+22'), Version('2.0.12+1'), Version('2.0.12+123'), Version('2.0.12+321')]
```

@AyanSinhaMahapatra IMO we should go with `2.8.5` the same version that we have on `univers`.

AyanSinhaMahapatra: Thanks for looking into this @keshav_space, I'm going ahead with `2.8.5` the same version used in univers

AyanSinhaMahapatra:
Thanks @pombredanne @keshav-space , all green, merging!

PR #3744, opened 2024-04-20
alpine url bug fix and AGPL License version issue

404-geek:


pombredanne:
@404-geek Thanks... can you check the test failures? Also in the future it is usually best to avoid mixing two unrelated changes in the same PR... here the license detection issue is not related to anything about alpine VCS URL handling, AFAIK ;)

404-geek:
> @404-geek Thanks... can you check the test failures? Also in the future it is usually best to avoid mixing two unrelated changes in the same PR... here the license detection issue is not related to anything about alpine VCS URL handling, AFAIK ;)

Hi @pombredanne 

Sorry for mixing up 2 issues in same PR, will take care of this in the future.
I was going through the build failure of readlines of pdf using pdf miner 

and there is regression issue in pdfminer for which it is failing.

    @pytest.mark.xfail(reason='Latest pdfminer.six from 2022 has a regression')
    def test_numbered_text_lines_does_not_fail_on_autocad_test_pdf(self):
        test_file = self.get_test_loc('pdf/AutoCad_Diagram.pdf')
        result = list(numbered_text_lines(test_file))
        assert result == []

should i comment this specific test for now ?

404-geek:
> @404-geek Thanks... can you check the test failures? Also in the future it is usually best to avoid mixing two unrelated changes in the same PR... here the license detection issue is not related to anything about alpine VCS URL handling, AFAIK ;)

Sure @pombredanne will take care of not keeping 2 issues in same PR in the future

I have resolved all the test issues. Can you please check once

src/licensedcode/data/rules/agpl-3.0-plus_or_commercial-license_4.RULE
```diff
@@ -1,5 +1,5 @@
 ---
-license_expression: agpl-3.0-plus OR commercial-license
```
AyanSinhaMahapatra, line 2: While the change is correct note that the rule filenames reflect the license-expression of the rule and a change in the license-expression needs to happen like this:

1. Deprecate the current rule using the `is_deprecated` attribute
2. Reindex the licenses with `scancode-reindex-licenses` (This will remove the deprecated rule from the index so you can add a new one, otherwise two rules with the same text cannot exist in the index)
3. Add a new rule for just `agpl-3.0-plus` and add notes about the origin, that this was previously the src/`agpl-3.0-plus_or_commercial-license_4.RULE` rule. Add the issue reference in the commit. (You can paste the rule content at `etc/scripts/licenses/buildrules-template.txt` and then run `python etc/scripts/licenses/buildrules.py etc/scripts/licenses/buildrules-template.txt` in an activated virtualenv, and this will create the rule automatically, with the appropriate filename)

AyanSinhaMahapatra (changes requested):
Thanks++ @404-geek , we are mostly there.

AyanSinhaMahapatra:
@404-geek gentle ping, we need a small update here.

pombredanne:
@404-geek ping :) I would like to merge this soon enough!

src/licensedcode/data/rules/agpl-3.0-plus_or_commercial-license_4.RULE
404-geek: Hi @AyanSinhaMahapatra  I need some help on the `buildrules-template.txt` for generation.

AyanSinhaMahapatra:
See https://github.com/aboutcode-org/scancode-toolkit/issues/3740#issuecomment-2809544545, will merge the commits thorugh https://github.com/aboutcode-org/scancode-toolkit/pull/4093, closing.

PR #3751, opened 2024-04-26
Fix yarn lock v1 parser to handle aliases better

AyanSinhaMahapatra:
Reference: #3555 

See also https://classic.yarnpkg.com/lang/en/docs/cli/add/#toc-yarn-add-alias, so ignoring these aliases are harmless.


tests/packagedcode/data/npm/yarn-lock/v1-complex2/yarn.lock
```diff
@@ -0,0 +1,109 @@
...
+    semver "^6.3.1"
+
+ansi-styles@^3.2.0, ansi-styles@^3.2.1:
+  version "3.2.1"
+  resolved "https://registry.yarnpkg.com/ansi-styles/-/ansi-styles-3.2.1.tgz#41fbb20243e50b12be0f04b8dedbf07520ce841d"
+  integrity sha512-VT0ZI6kZRdTh8YyJw3SMbYm/u+NqfsAxEpWO0Pf9sq8/e94WxxOpPKx9FR1FlyCtOVDNOQ+8ntlqFxiRc+r5qA==
+  dependencies:
+    color-convert "^1.9.0"
+
+change-case@^5.4.3, change-case@^5.4.4, "latest-change-case@npm:change-case@^5.4.3":
```
AyanSinhaMahapatra, line 22: We were getting a
 ```
AssertionError: Different names for same dependency is not supported: {('', 'latest-change-case@npm:change-case'), ('', 'change-case')}
E           assert 2 == 1
E            +  where 2 = len({('', 'change-case'), ('', 'latest-change-case@npm:change-case')})
```
While trying to parse this yarn.lock file, at [this assert statement](https://github.com/nexB/scancode-toolkit/blob/fix-yarn-lock-v1-parser/src/packagedcode/npm.py#L660).

pombredanne: OK, so this is resolved!

pombredanne (approved):
Thanks !
Please merge.

PR #3829, opened 2024-06-26
Add parser for `swift-show-dependencies.deplock`

keshav-space:
Reference: https://github.com/nexB/scancode.io/issues/1278

### Tasks


src/packagedcode/swift.py
```diff
@@ -47,12 +47,75 @@ def logger_debug(*args):
...
+    @classmethod
+    def _parse(cls, swift_dependency_relation, package_only=False):
+
+        if TRACE:
+            logger_debug(
+                f"SwiftShowDependenciesDepLockHandler: deplock: package: {swift_dependency_relation}"
+            )
+
+        dependencies = get_flatten_dependencies(
+            swift_dependency_relation.get("dependencies")
```
AyanSinhaMahapatra: Let's use named arguments as much as possible.

src/packagedcode/swift.py
```diff
@@ -47,12 +47,75 @@ def logger_debug(*args):
...
+        )
+
+        package_data = dict(
+            datasource_id=cls.datasource_id,
+            type=cls.default_package_type,
+            primary_language=cls.default_primary_language,
+            namespace=None,
```
AyanSinhaMahapatra: You can just remove the line here, this is None by default. Also nice to add a one-line comment that we can't get namespace from just here.

src/packagedcode/swift.py
```diff
@@ -47,12 +47,75 @@ def logger_debug(*args):
...
+        cls, package_data, resource, codebase, package_adder=models.add_to_package
+    ):
+        siblings = resource.siblings(codebase)
+        swift_manifest_resource = [
+            r
+            for r in siblings
+            if r.name in ("Package.swift.json", "Package.swift.deplock")
+        ]
+
+        # Skip the assembly if the Swift manifest is present.
```
AyanSinhaMahapatra, line 100: This is not the correct way to do assembly. See https://github.com/nexB/scancode-toolkit/blob/develop/src/packagedcode/npm.py#L70 or other parsers on creating a base handler class with assembly and all other manifests inheriting from there.
From there look for the main manifest in sibling files from the location, and parse and yield from there accordingly. By this way we assemble from the first package that gets picked up, instead of walking the codebase further for more manifests.
Just a nit, assembly is working correctly otherwise.

tests/packagedcode/test_swift.py
```diff
@@ -101,3 +101,24 @@ def test_package_scan_swift_end_to_end_full_mapboxmaps_resolved_only(self):
         check_json_scan(
             expected_file, result_file, remove_uuid=True, regen=REGEN_TEST_FIXTURES
         )
+
+    def test_package_scan_swift_end_to_end_full_mapboxmaps_resolved_only(self):
+        test_dir = self.get_test_loc("packages/vercelui")
```
AyanSinhaMahapatra, line 106: This is a nit. Could be possibly improved later too.

1. Let's not only add full-scan tests. This is much harder to test and review in general, and also takes up much more space. Instead do individual tests of `parse()` and `is_datafile()` functions (at least one test per parser function). One test should only check one thing that can go wrong. It's fine to have a couple assembly functions later, but using barebones manifests just enough for assembly.
2. Let's truncate test files (like big the ones with dependency trees) to reduce file size for the tests, we have a very large number of tests and this would keep increasing the git checkout size/source size more rapidly that we can manage. Let's either find smaller examples, or reduce the files so that there is just enough to test the features.

AyanSinhaMahapatra (changes requested):
@keshav-space LGTM! Thanks++
This works nicely for package resolution and dependency graph creation on the SCIO side. I have a couple nits for your consideration, this is ready to merge afterwards.
Also merge from latest to resolve conflicts.

AyanSinhaMahapatra:
One more question, I assume we cannot parse `Package.swift` files like this one here: https://github.com/swift-cloud/VercelUI/blob/main/Package.swift right, becuase this is in swift and not parseable in python easily?

keshav-space:
> One more question, I assume we cannot parse `Package.swift` files like this one here: https://github.com/swift-cloud/VercelUI/blob/main/Package.swift right, becuase this is in swift and not parseable in python easily?

@AyanSinhaMahapatra  We are not directly parsing the `Package.swift` file, but we parse the `Package.swift.deplock`, which is a JSON dump of `Package.swift`. In the future, we may add support for natively parsing the `Package.swift` file, and we can use Tree-Sitter (an optional dependency) for creating and parsing the Swift AST.

src/packagedcode/swift.py
keshav-space: Done!

src/packagedcode/swift.py
keshav-space: Done!

src/packagedcode/swift.py
keshav-space: Tracking this here https://github.com/nexB/scancode-toolkit/issues/3837

tests/packagedcode/test_swift.py
keshav-space: @AyanSinhaMahapatra I've truncate the large test file. Will add more individual tests along with the fix for https://github.com/nexB/scancode-toolkit/issues/3837 in next PR.

AyanSinhaMahapatra (approved):
Thanks @keshav-space! Merging!

PR #3959, opened 2024-10-23
Declare ngram variable in select_ngrams

JonoYang:
* This is to avoid an undeclared variable issue when running the last part of the function


JonoYang:
Closing this in favor of another pr

PR #3982, opened 2024-11-09
Support for Python Private Packages Using Private : : Classifier #3968

### LINKED ISSUE #3968: Support Python private packages
See:
- https://github.com/astral-sh/uv/pull/7976#discussion_r1825190088
from @chrisrodrigue

> "Private :: Do Not Upload" is a handy classifier which causes PyPI to reject a package. It's handy for packages that you don't want to accidentally publish, so maybe it could be used in the full example?

alok1304:
Fixes #3968
Changes Implemented:
Added support for the `Private :: Do Not Upload` classifier in the setup configuration.
Updated the `setup.cfg`  file to enable users to designate their packages as private using this classifier.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,6 +23,7 @@ classifiers =
     Programming Language :: Python :: 3.12
     Topic :: Software Development
     Topic :: Utilities
+    Private :: Do Not Upload 
 
 keywords =
     open source


stefan6419846:
How does this solve the linked issue? Your change basically declares `scancode-toolkit` as a private package, which surely is not what the issue has been about.

alok1304:
Hi @stefan6419846 , thanks for the feedback! I see now that my recent commit may have inadvertently declared `scancode-toolkit` as private, which wasn’t the intention. I think here is the goal here is to add support for the` is_private `flag by detecting the `Private ::` classifier in user packages, rather than marking `scancode-toolkit` itself as private.

can you tell how I work on it.

AyanSinhaMahapatra:
@alok1304 you have to update the `pyproject.toml` parser at https://github.com/aboutcode-org/scancode-toolkit/blob/develop/src/packagedcode/pypi.py#L503 to make sure we handle these private package manifests correctly. See also https://github.com/aboutcode-org/scancode-toolkit/pull/3779/commits/f848741989d0d40309fc1220c58b5a9a145c897e to check how we did this for other package ecosystems. Start by adding a test file from https://github.com/aboutcode-org/scancode-toolkit/issues/3968#issuecomment-2455595655, similarly like the tests in https://github.com/aboutcode-org/scancode-toolkit/blob/develop/tests/packagedcode/test_pypi.py#L305

alok1304:
@AyanSinhaMahapatra Thankyou!!

alok1304:
@AyanSinhaMahapatra 
```
name = project_data.get('name')
is_private = False

# Check for privacy classifier
classifiers = project_data.get('classifiers', [])
for classifier in classifiers:
    if 'Private ::' in classifier:
        is_private = True
        break  # Once a private classifier is found, no need to continue checking

# If no classifier or name, package is private by default
if not name:
    is_private = True
```


is this correct??

PR #3987, opened 2024-11-16
added support to parse labels in dockerfile

### LINKED ISSUE #3561: Treat Dockerfile and Containerfile as "non-assembled" package data, collect OCi labels
It would useful as these are package like and have structured data as seen in https://github.com/jitsi/docker-jitsi-meet/blob/stable-8960-1/jibri/Dockerfile#L5

```
LABEL org.opencontainers.image.title="Jitsi Broadcasting Infrastructure (jibri)"
LABEL org.opencontainers.image.description="Components for recording and/or streaming a conference."
LABEL org.opencontainers.image.url="https://github.com/jitsi/jibri"
LABEL org.opencontainers.image.source="https://github.com/jitsi/docker-jitsi-meet"
LABEL org.opencontainers.image.documentation="https://jitsi.github.io/handbook/"
```

See for full spec: https://github.com/opencontainers/image-spec/blob/main/annotations.md

Note that container-inspector already supports parsing Dockerfile.

VarshaUN:
Fixes #3561

### Tasks

Signed-off-by: Varsha U  N <varshaun58@gmail.com>


src/packagedcode/__init__.py
```diff
@@ -97,6 +98,7 @@
     debian.DebianSourcePackageTarballHandler,
 
     distro.EtcOsReleaseHandler,
+    dockerfileLABEL.DockerfileHandler,
```
AyanSinhaMahapatra:
```suggestion
    dockerfile.DockerfileHandler,
```
What about this, we usually don't use uppercase here in file/module names and maybe this name is better?

src/packagedcode/dockerfileLABEL.py
```diff
@@ -0,0 +1,72 @@
...
+# See http://www.apache.org/licenses/LICENSE-2.0 for the license text.
+# See https://github.com/nexB/scancode-toolkit for support or download.
+# See https://aboutcode.org for more information about nexB OSS projects.
+#
+
+
+
+import io
+from pathlib import Path
+from dockerfile_parse import DockerfileParser
```
AyanSinhaMahapatra: https://pypi.org/project/dockerfile-parse/ is nice to use, also note that we use this at container inspector https://github.com/aboutcode-org/container-inspector/blob/main/src/container_inspector/dockerfile.py#L14C8-L14C24 as we parse dockerfiles there too. `container-inspector` is also a scancode dependency, so we have to think about whether to move this code there. But overall I think this is fine here, as packagedata creation/datafile handlers should be in scancode.

I don't see this used anywhere else in scancode-toolkit, we have this in the requirements at https://github.com/aboutcode-org/scancode-toolkit/blob/develop/requirements.txt#L18  to pin the version when we import container-inspector. But we need to add dockerfile-parse here: https://github.com/aboutcode-org/scancode-toolkit/blob/develop/setup.cfg#L73 as this is not constrained at https://github.com/aboutcode-org/container-inspector/blob/main/setup.cfg#L62

src/packagedcode/dockerfileLABEL.py
```diff
@@ -0,0 +1,72 @@
...
+
+import io
+from pathlib import Path
+from dockerfile_parse import DockerfileParser  
+from packagedcode import models
+from packagedcode import utils
+
+
+class DockerfileHandler(models.DatafileHandler):
+    datasource_id = 'dockerfile'
```
AyanSinhaMahapatra:
```suggestion
    datasource_id = 'dockerfile_oci_labels'
```
Here I think more explicit is better, as this is not really about the usual dockerfile contents but only the OCI labels

src/packagedcode/dockerfileLABEL.py
```diff
@@ -0,0 +1,72 @@
...
+    default_package_type = 'docker-image'
+    path_patterns = ('Dockerfile', 'containerfile', '*.dockerfile')  
+    description = 'Dockerfile (OCI) metadata handler'
+    documentation_url = 'https://docs.docker.com/engine/reference/builder/'
```
AyanSinhaMahapatra: Shouldn't this be https://github.com/opencontainers/image-spec/blob/main/annotations.md ? Again this is not really about the usual dockerfile contents but only the OCI labels

src/packagedcode/dockerfileLABEL.py
```diff
@@ -0,0 +1,72 @@
...
+from pathlib import Path
+from dockerfile_parse import DockerfileParser  
+from packagedcode import models
+from packagedcode import utils
+
+
+class DockerfileHandler(models.DatafileHandler):
+    datasource_id = 'dockerfile'
+    default_package_type = 'docker-image'
+    path_patterns = ('Dockerfile', 'containerfile', '*.dockerfile')
```
AyanSinhaMahapatra: We need to test the `is_datafile()` here with examples from here:
* https://github.com/search?q=org.opencontainers.image+path%3A**%2FDockerfile&type=code&ref=advsearch
* https://github.com/search?q=org.opencontainers.image+path%3A**%2F*containerfile*&type=code&ref=advsearch
* https://github.com/search?q=org.opencontainers.image+path%3A**%2F*.dockerfile&type=code&ref=advsearch

Also have these as tests for parsers, choosing from more popular repos/more diverse examples of usage.

See for example other tests like this: https://github.com/aboutcode-org/scancode-toolkit/blob/develop/tests/packagedcode/test_npm.py#L445

src/packagedcode/dockerfileLABEL.py
```diff
@@ -0,0 +1,72 @@
...
+        Extract OCI labels from the Dockerfile using container-inspector.
+        """
+        labels = {}
+        parser = DockerfileParser()
+        parser.parse(dockerfile_path)
+        labels = parser.labels
+        return labels
+
+    @classmethod
+    def assemble(cls, package_data, resource, codebase, package_adder):
```
AyanSinhaMahapatra: I don't think we want to assemble files for these dockerfile  handlers since this is not a proper package instance that you can find in a package ecosystem, maybe we want to use `models.NonAssemblableDatafileHandler` like in https://github.com/aboutcode-org/scancode-toolkit/blob/develop/src/packagedcode/distro.py#L20 and remove this function altogether.

You can see in the issue title too: `"non-assembled" package data`

src/packagedcode/dockerfileLABEL.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        Parse a Dockerfile and yield one or more PackageData objects with OCI labels and metadata.
+        """
+        labels = cls.extract_oci_labels_from_dockerfile(location)
+
+       
+        package_data = {
+            'datasource_id': cls.datasource_id,
+            'type': cls.default_package_type,
+            'name': labels.get('name', 'unknown'),
```
AyanSinhaMahapatra: There is no point having the defaults as `unknown`, it's better to have these as `None` which is what all other parsers do too, if there is no data found, that attribute is just kept empty. This is since we use existence of a value in these attributes for other functions.

src/packagedcode/dockerfileLABEL.py
```diff
@@ -0,0 +1,72 @@
...
+    description = 'Dockerfile (OCI) metadata handler'
+    documentation_url = 'https://docs.docker.com/engine/reference/builder/'
+
+    @classmethod
+    def parse(cls, location, package_only=False):
+        """
+        Parse a Dockerfile and yield one or more PackageData objects with OCI labels and metadata.
+        """
+        labels = cls.extract_oci_labels_from_dockerfile(location)
+
```
AyanSinhaMahapatra: A nit, let's have less space here

AyanSinhaMahapatra (changes requested):
@VarshaUN Thanks++

Looking good, need another round of changes and a lot more tests here for the added functionality, and this should be much better.

VarshaUN:
@AyanSinhaMahapatra I have added everything that you told me. Please review it.

src/packagedcode/dockerfile.py
```diff
@@ -0,0 +1,59 @@
...
+        """
+        Parse a Dockerfile and yield one or more PackageData objects with OCI labels and metadata.
+        """
+        labels = cls.extract_oci_labels_from_dockerfile(location)
+        package_data = {
+            'datasource_id': cls.datasource_id,
+            'type': cls.default_package_type,
+            'name': labels.get('name', 'None'),  
+            'version': labels.get('version', 'None'),  
+            'license_expression': labels.get('license', 'None'),
```
AyanSinhaMahapatra:
```suggestion
            'license_expression': labels.get('license'),
```
`.get()` already returns `None` as default if specified key is not present in the dict.

src/packagedcode/dockerfile.py
```diff
@@ -0,0 +1,59 @@
...
+from packagedcode import models
+from packagedcode import utils
+import fnmatch
+
+
+class DockerfileHandler(models.DatafileHandler):
+    datasource_id = 'dockerfile_oci_labels'
+
+    @classmethod 
+    def is_datafile(cls, path):
```
AyanSinhaMahapatra: You don't need to define this function, as it inherits a base implementation from the base class here: https://github.com/aboutcode-org/scancode-toolkit/blob/develop/src/packagedcode/models.py#L1085 which used the `path_patterns` set by use in the DatafileHandler we define. So you just need to define a `path_patterns` for the DatafileHandler class above: https://github.com/aboutcode-org/scancode-toolkit/blob/467449a11499e1c1e4357742a64ce0c22b7b5198/src/packagedcode/models.py#L1057 see other handlers for examples of path patterns.

src/packagedcode/dockerfile.py
```diff
@@ -0,0 +1,59 @@
...
+
+import io
+from pathlib import Path
+from dockerfile_parse import DockerfileParser  
+from packagedcode import models
+from packagedcode import utils
+import fnmatch
+
+
+class DockerfileHandler(models.DatafileHandler):
```
AyanSinhaMahapatra:
```suggestion
class DockerOCILabelsHandler(models.DatafileHandler):
```

AyanSinhaMahapatra: Also can you inherit from `NonAssemblableDatafileHandler` instead? I had mentioned this also here: https://github.com/aboutcode-org/scancode-toolkit/pull/3987#discussion_r1863394989

tests/packagedcode/test_dockerfile.py
```diff
@@ -0,0 +1,60 @@
...
+        with open(expected_file) as f:
+            return json.load(f)
+
+    def test_is_datafile(self):
+        dockerfiles = [
+            'test.dockerfile',
+            'test.containerfile',
+            'psql.dockerfile'
+        ]
+        for dockerfile in dockerfiles:
```
AyanSinhaMahapatra: This is correct that you are using loops to reduce code, but we can use `pytest.mark.parametrize`
see https://github.com/aboutcode-org/scancode-toolkit/blob/develop/tests/packagedcode/test_cargo.py#L183 for example, this:
- reduces code even further
- creates many different tests, one for each value in the iterable, so tests fail/pass individually for each file in the list
- cleaner and modular

tests/packagedcode/test_dockerfile.py
```diff
@@ -0,0 +1,60 @@
...
+
+from packagedcode import dockerfile
+import pytest
+import os.path
+import json
+from pathlib import Path
+from packagedcode.dockerfile import DockerfileHandler
+
+class TestDockerfileHandler:
+
```
AyanSinhaMahapatra: Can you also run one full scan for one case, like the test here: https://github.com/aboutcode-org/scancode-toolkit/blob/develop/tests/packagedcode/test_cargo.py#L158, one is enough, don't have to do this for all files

tests/packagedcode/data/docker/psql.dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM postgres:15.3
```
AyanSinhaMahapatra: Could you also add the source of this, nice that you added a comment on the other docker files about their source.

tests/packagedcode/test_dockerfile.py
```diff
@@ -0,0 +1,60 @@
...
+            ('psql.dockerfile', 'psql-expected.json')
+        ]
+        for dockerfile, expected in test_files:
+            test_file = self.get_test_loc(f'data/docker/{dockerfile}')
+            expected_loc = self.get_test_loc(f'data/docker/{expected}')
+            packages = list(DockerfileHandler.parse(str(test_file)))
+            expected_packages = self.load_expected(expected_loc)
+            assert packages == expected_packages
+
+    def test_extract_oci_labels_from_dockerfile(self, mocker):
```
AyanSinhaMahapatra: There is an issue here, how is the files generated by this test different from the test functions  `test_parse_dockerfile` above? You are using the same filenames for the test expectations so we are missing a set of test files. 

We want to have two sets of tests distinctly both for `DockerfileHandler.parse` and `DockerfileHandler.extract_oci_labels_from_dockerfile` and this should generate 2 sets of files so 2 docker/container files and total 6 test expectation files. You might want to use `-package.expected.json` and `-expected.json` to differentiate there.

src/packagedcode/dockerfile.py
```diff
@@ -0,0 +1,59 @@
+#
```
AyanSinhaMahapatra: Please rename this file from `dockerfile` to `dockerfile_ocilabels`, seems more appropriate

tests/packagedcode/data/docker/test.containerfile
```diff
@@ -0,0 +1,80 @@
...
+COPY dataplane dataplane 
+COPY tools/udp-test-server tools/udp-test-server
+COPY xtask xtask
+COPY Cargo.toml Cargo.toml
+COPY Cargo.lock Cargo.lock
+COPY .cargo .cargo
+
+# We need to tell bpf-linker where it can find LLVM's shared library file.
+# Ref: https://github.com/aya-rs/rustc-llvm-proxy/blob/cbcb3c6/src/lib.rs#L48
+ENV LD_LIBRARY_PATH="/usr/lib/llvm-$LLVM_VERSION/lib"
```
AyanSinhaMahapatra: it might make sense to trim a little bit the parts of this file which are not useful for it's tests to make it smaller (in a way that doesn't give errors on the parsing). The repo quickly gets large because of large test files, as we have several thousands of these at these point, and it adds up to the repo size.

AyanSinhaMahapatra (changes requested):
Thanks++ @VarshaUN this is looking great! A few more comments for your consideration. Thank you for your patience 

Can you also remove this file which was added with your PR: `tests/packagedcode/__pycache__/test_parse_pyproject_toml.cpython-312-pytest-8.3.3.pyc.26520`?

Please also make sure you look into test failures (many are failing as an effect of this PR) and try to resolve them if they are related to your PR, also merge from develop since it's been a while.

src/packagedcode/__init__.py
```diff
@@ -18,6 +18,7 @@
 from packagedcode import debian
 from packagedcode import debian_copyright
 from packagedcode import distro
+from packagedcode import dockerfile
```
AyanSinhaMahapatra: You need to change imports after you change filenames :) 

Please run a subset of the test suite locally (atleast the part where you made changes) after you make changes, look for failing CI like : https://dev.azure.com/nexB/scancode-toolkit/_build/results?buildId=16107&view=logs&j=1412dbfa-421a-5a40-8ddf-7073fc46aa19&t=13c90161-2109-5957-f563-a96fefc7d869 and make sure we pass all the tests.

tests/packagedcode/data/docker/test-dockerfile/test.dockerfile-package.expected.json
```diff
@@ -0,0 +1,18 @@
+[
```
AyanSinhaMahapatra: This file and the file below have exactly the same data, you are not doing this correctly.

These are the two cases you want to test against:

1. The labels are being parsed correctly
2. The parsed data is being correctly mapped to a PackageData mapping

This should have had the fields of PackageData, mapped correctly, your tests look okay but it does not seem like you've run them or regenerated expectations.

AyanSinhaMahapatra (changes requested):
@VarshaUN thanks for the updates, please fix all test failures. Looks good and ready to merge otherwise.

Dockerfile
pombredanne, line 1: Do not change our dockerfile... instead add that file from k8s as a test file.

VarshaUN: My bad! , I will look into it.Thanks!

PR #4075, opened 2025-01-06
support-python-private-classifier

### LINKED ISSUE #3968: Support Python private packages

VarshaUN:
Fixes #3968 

### Tasks

Signed-off-by : Varsha U N <varshaun58@gmail.com>


src/packagedcode/pypi.py
```diff
@@ -542,8 +544,10 @@ def parse(cls, location, package_only=False):
             keywords=get_keywords(project_data),
             parties=get_pyproject_toml_parties(project_data),
             dependencies=dependencies,
+            is_private=is_private,
             extra_data=extra_data,
-            **urls,
+            homepage_url=urls.get('homepage'),
```
AyanSinhaMahapatra: I don't think the change makes sense here, we should just include the `**urls` like before.

src/packagedcode/pypi.py
```diff
@@ -517,6 +517,8 @@ def parse(cls, location, package_only=False):
         if license_file:
             extra_data['license_file'] = license_file
 
+        classifiers = project_data.get('classifiers', [])
+        is_private = any('Private ::' in classifier for classifier in classifiers)
```
AyanSinhaMahapatra: Can you move these lines up to before https://github.com/aboutcode-org/scancode-toolkit/pull/4075/files#diff-166f21bb085d47341daf98771ee29cd76e2f68ddeb34b8eb981a9d1f3b05e8b9R514, and pass is_private as an argument to the `get_urls()` function too?

Then inside the `get_urls()` function, we need not call the `get_pypi_urls()` function as we do not have this package at pypi if `is_private` is `True`. Because repository_homepage_url, repository_download_url and api_data_url are all pypi.org URLs for public packages.

AyanSinhaMahapatra (changes requested):
@VarshaUN Thanks++ for the PR, this is looking great.
See comments for some small improvements and we can merge after that.
Apologies for the late review. :sweat_smile:

src/packagedcode/pypi.py
VarshaUN: Yes , sure but what if is_private is false , so I just added an if-else in `get_urls`

VarshaUN:
@AyanSinhaMahapatra  I have made the changes you told me , Kindly review it . Thanks!

src/packagedcode/pypi.py
AyanSinhaMahapatra: That is what I meant, we were already handling the is_private False calse, so we needed to handle the private package case correctly here.

src/packagedcode/pypi.py
```diff
@@ -2281,7 +2286,7 @@ def get_pypi_urls(name, version, **kwargs):
     )
 
 
-def get_urls(metainfo, name, version, poetry=False):
+def get_urls(metainfo, name, version,is_private, poetry=False):
```
AyanSinhaMahapatra:
```suggestion
def get_urls(metainfo, name, version, is_private=False, poetry=False):
```

If you don't add defaults to a function you just added, this will fail at all the places this function is being invoked other than in the `PyprojectTomlHandler.parse()` function.

And you can see the test failures at https://dev.azure.com/nexB/scancode-toolkit/_build/results?buildId=15878&view=logs&j=1412dbfa-421a-5a40-8ddf-7073fc46aa19

You should run all the tests for a module you are changing locally before you push any code, making sure they pass. In this case tests/packagedcode/test_pypi.py

src/packagedcode/pypi.py
```diff
@@ -511,6 +511,9 @@ def parse(cls, location, package_only=False):
         description = project_data.get('description') or ''
         description = description.strip()
 
+        classifiers = project_data.get('classifiers', [])
+        is_private = any('Private ::' in classifier for classifier in classifiers)
```
AyanSinhaMahapatra: On further investigation, it seems this is present in other kinds of python package manifests too, like:
- [`setup.py`](https://github.com/search?q=%22Private+%3A%3A%22+path%3A**%2Fsetup.py&type=code&ref=advsearch) : PythonSetupPyHandler
- [`setup.cfg`](https://github.com/search?q=%22Private+%3A%3A%22+path%3A**%2Fsetup.cfg&type=code&ref=advsearch) SetupCfgHandler
- [poetry `pyproject.toml`](https://github.com/search?q=%22Private+%3A%3A%22+%22poetry%22+path%3A**%2Fpyproject.toml&type=code&ref=advsearch) : PoetryPyprojectTomlHandler
- `parse_metadata()` function which is used in multiple parsers (even though these private packages cannot be released, it can still be built and installed so these can be present in metadata files too)

You need to take these two lines out in a function and implement the logic for all these other parsers, similarly like you have done here for pyproject.toml.

Note that other pypi parsers which are lockfiles are not applicable as they don't have these classifiers.

AyanSinhaMahapatra (changes requested):
Thanks++ @VarshaUN!

A couple more related changes are required as private classifiers are used generally in other python package manifests too, not just pyproject.toml files, so please see my comments.

VarshaUN:
Hey @AyanSinhaMahapatra I have moved that two lines of code into a function in utils.py and imported it in pypi.py and used it in handlers as you have mentioned.

src/packagedcode/pypi.py
```diff
@@ -698,6 +701,8 @@ def parse(cls, location, package_only=False):
                 )
                 dependencies.append(dependency.to_dict())
 
+        is_private = is_private_package(poetry_data.get('classifiers', []))
+
```
AyanSinhaMahapatra, line 706: You have added `is_private` but you have not passed it to the `package_data` mapping below. This might be the case elsewhere too, can you check that you've done this for all cases?

It might be best to always add an example of each case of package manifests which are private package and test them with `check_packages_data` to verify we are doing this correctly i.e. the `is_private=True` for test expectation files in each case. If you don't add tests, you're more than likely to miss things and the functionality might not work.

See https://github.com/aboutcode-org/scancode-toolkit/pull/4075/#discussion_r2046764579 for examples to use as test case.

AyanSinhaMahapatra (changes requested):
@VarshaUN thanks++ for the updates to support all cases where this is seen, looking good. Could you add tests to cover all the cases, and this should be ready to merge.

Also please make sure all tests pass, and verify that the test expectation changes are intended and related to this PR: https://dev.azure.com/nexB/scancode-toolkit/_build/results?buildId=16089&view=logs&j=1412dbfa-421a-5a40-8ddf-7073fc46aa19&t=13c90161-2109-5957-f563-a96fefc7d869

src/packagedcode/utils.py
```diff
@@ -304,3 +304,6 @@ def is_simple_path(path):
 
 def is_simple_path_pattern(path):
     return path.endswith('*') and path.count('*') == 1
+
+def is_private_package(classifiers):
```
pombredanne: This should not be in utils, but in the pypi.py

PR #4089, opened 2025-01-13
Update package assembly in conda installations

AyanSinhaMahapatra:
Parse conda metadata JSON manifests and use the package data and files information present to improve conda package assembly.

Reference: https://github.com/aboutcode-org/scancode-toolkit/issues/4083

### Tasks


JonoYang:
@AyanSinhaMahapatra You need to regenerate some package test expectations, but otherwise lgtm!

JonoYang (approved):

azure-pipelines.yml
```diff
@@ -233,8 +233,8 @@ jobs:
 
     - template: etc/ci/azure-posix.yml
       parameters:
-          job_name: macos12_cpython_latest_from_pip
-          image_name: macos-12
+          job_name: macos14_cpython_latest_from_pip
+          image_name: macos-14
```
pombredanne, line 237: Could you bump only to macos-13 for now https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md ?

src/packagedcode/conda.py
```diff
@@ -23,7 +24,284 @@
...
+                    package=package,
+                    resource=resource,
+                    codebase=codebase,
+                    package_adder=package_adder,
+                )
+                yield resource
+
+            return
+
+        # For a conda metadata JSON, try to find the corresponding meta.yaml and
```
pombredanne: For clarity and readability, it may make sense to make the `if CondaMetaYamlHandler.is_datafile(resource.location):` "else" condition explicit, or to add a new `if CondaMetaJsonHandler.is_datafile(resource.location)` .... also I wonder if the code is not too long and too nested? Why not move the bulk of the code to each subclasses? CondaMetaYamlHandler and CondaMetaYamlHandler? It may make the code more readable and require fewer tests

pombredanne (changes requested):
See comments inline. You may also consider refactoring a bit to streamline your class hierarchy?

azure-pipelines.yml
AyanSinhaMahapatra: We already have them, see https://github.com/aboutcode-org/scancode-toolkit/pull/4089/files/20663f45f0cc0743e179117c28914387598e177b#diff-7915b9b726a397ae7ba6af7b9703633d21c031ebf21682f3ee7e6a4ec52837a5L245
Basically we had the following runners: macos-12, macos-13, now we have macos-13, macos-14

src/packagedcode/conda.py
AyanSinhaMahapatra: I don't think we can move the code to each subclasses, as they are needed together to consider all the possible scenarios effectively (both the manifests being present or not present, and for merging package/files data from both manifests). This is effectively also the case in many other major ecosystems where we have to merge/create packages from multiple manifests, using a Basehandler class with a base assembly:
1. npm: https://github.com/aboutcode-org/scancode-toolkit/blob/develop/src/packagedcode/npm.py#L82
2. pypi: https://github.com/aboutcode-org/scancode-toolkit/blob/develop/src/packagedcode/pypi.py#L159
3. maven: https://github.com/aboutcode-org/scancode-toolkit/blob/develop/src/packagedcode/maven.py#L62

This becomes a bit more complex when workspaces are involved or like in this case where the manifests are located in entirely different places and assigning package files/getting to manifests is based on more complex logic of file paths.

It is possible to break the code down further into functions which stay in the same base class though, let me do that.

PR #4100, opened 2025-01-16
add length to progress bar

### LINKED ISSUE #3342: display total number of files being scanned above progress bar
## Short Description

It would be great to display the number of files that will be scanned in total above the progress bar so I can see how many files are still left to be scanned.

## Possible Labels

- new feature

## Select Category

-  [X ] Enhancement
-  [ ] Add License/Copyright
-  [ ] Scan Feature
-  [ ] Packaging
-  [ ] Documentation
-  [ ] Expand Support
-  [ ] Other

## **Describe the Update**

Display the number of files that will be scanned. It should already be known because first an inventory is done.

It will help me save waiting.

## **Possible Solution/Implementation Details**

Display the number of files found during inventory before/during scanning.

I can help test!

xsuchy:
so the progress bar shows real progress

Fixes #3342

### Tasks

The testsuite is broken for me and produce lots of not relevant errors. I will appreciate testing.
* [-] Updated documentation pages (if applicable)

### DIFF
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -1264,6 +1264,8 @@ def scan_codebase(
 
     # NOTE: we never scan directories
     resources = ((r.location, r.path) for r in codebase.walk() if r.is_file)
+    if progress_manager:
+        resources = list(resources)
 
     use_threading = processes >= 0
     runner = partial(
@@ -1299,6 +1301,7 @@ def scan_codebase(
 
         if progress_manager:
             scans = progress_manager(scans)
+            scans.length = len(resources)
             # hack to avoid using a context manager
             if hasattr(scans, '__enter__'):
                 scans.__enter__()
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -108,6 +108,9 @@ This is a patch release with bugfixes and improvements::
   https://github.com/aboutcode-org/scancode-toolkit/pull/4476
 
 
+- Progress bar now shows length. You can estimate the duration now.
+  See https://github.com/aboutcode-org/scancode-toolkit/issues/3342
+
 v32.4.0 - 2025-06-26
 --------------------
 


src/scancode/cli.py
```diff
@@ -1176,10 +1176,11 @@ def run_scanners(
 
     progress_manager = None
     if not quiet:
+        resources = list(((r.location, r.path) for r in codebase.walk() if r.is_file))
```
pombredanne: This will be a problem when we have a codebase with 10 million files, we may need to find a better way? ... otherwise the system will literally be stalled for minutes before displaying a progress bar.

src/scancode/cli.py
```diff
@@ -1176,10 +1176,11 @@ def run_scanners(
...
         echo_func(f'Scan files for: {scan_names} with {processes} process(es)...')
         item_show_func = partial(path_progress_message, verbose=verbose)
         progress_manager = partial(progressmanager,
-            item_show_func=item_show_func,
+            item_show_func=item_show_func, length=len(resources)*2,
```
pombredanne: Why the times 2?

pombredanne (changes requested):
Miroslav and thanks!
There is an issue wrt. to making this work with large codebase. Even a linux kernel tarball may be problematic... we may need to find a better way.

src/scancode/cli.py
xsuchy: Because https://github.com/aboutcode-org/scancode-toolkit/issues/3344 Otherwise the progressbar go to 170/170 and then continue to 340/170. This *2 can be removed when we find when we find where is the count done twice. I have a suspicion, but it needs more testing.

src/scancode/cli.py
xsuchy: For the kernel, the delay is 45 sec until it traverses 200k files. While the scan itself runs several hours.
With -q there is no delay.

xsuchy:
Several things I considered as an option:

1) instead `resources = list(((r.location, r.path) for r in codebase.walk() if r.is_file))` use simpler `resources = list((1 for r in codebase.walk() if r.is_file))` because we just need the count. This is consume less memory, but does not give any measurable time.

2) pass the `resources` to `scan_codebase` so we do not need to count it twice. Hmm. I will do this. PR updated.

xsuchy:
I tried to improve the performance of walk() in https://github.com/aboutcode-org/commoncode/pull/78

armijnhemel:
Looking at the code I don't see that this would fix #3342. I don't see that the number of files that will be processed is displayed.

xsuchy:
> Looking at the code I don't see that this would fix #3342. I don't see that the number of files that will be processed is displayed.

The line with `resources = list(resources)` is the number of files (times two for some reason).

The line with 'scans = progress_manager(scans)` set up progress meter. But because constructor does not have the number of items, it does not show the real progress.

The line `scans.length = len(resources)*2` sets the the lentgs of the queue in the progress meter so it then shows the real progress and number of processed items.

armijnhemel:
> > Looking at the code I don't see that this would fix #3342. I don't see that the number of files that will be processed is displayed.
> 
> The line with `resources = list(resources)` is the number of files (times two for some reason).
> 
> The line with 'scans = progress_manager(scans)` set up progress meter. But because constructor does not have the number of items, it does not show the real progress.
> 
> The line `scans.length = len(resources)*2` sets the the lentgs of the queue in the progress meter so it then shows the real progress and number of processed items.

Yes but in #3342 I don't care about the progress meter, I just want to see the number of files that will be scanned :-)

xsuchy:
Screenshot_20250518_110057

This is how it looks like: You can see the number of files being scanned on the right side of progress bar as $DONE/$TOTAL.

And when I was doing this screenshot, I rebased this PR.

armijnhemel:
> Screenshot_20250518_110057
> 
> This is how it looks like: You can see the number of files being scanned on the right side of progress bar as $DONE/$TOTAL.
> 
> And when I was doing this screenshot, I rebased this PR.

Oh, this would be great. @pombredanne, maybe some time should be spent finding out where the double counting happens? I tried to look for it but I couldn't find it.

xsuchy:
Rebased.

xsuchy:
Can you please consider a new round of review?
With https://github.com/aboutcode-org/commoncode/pull/78 merged, the penalty at the startup is really minimal.

xsuchy:
I removed the `*2` because I found the cause. It is fixed in https://github.com/aboutcode-org/commoncode/pull/100, so the length is now correct.
Please review.

armijnhemel:
> so the progress bar shows real progress
> 
> Fixes #3342

Actually it would not fix that issue, as I am asking for something else. What I am asking for in that issue is to display a total number of files that will be scanned, so I have an indication of how many files in total will be scanned. This PR fixes a different issue.

PR #4130, opened 2025-02-04
Add basic npm overrides parsing to package.json

### LINKED ISSUE #4122: Support npm overrides
These are similar to Go replace.
See https://docs.npmjs.com/cli/v11/configuring-npm/package-json#overrides

> 
> overrides
> 
> If you need to make specific changes to dependencies of your dependencies, for example replacing the version of a dependency with a known security issue, replacing an existing dependency with a fork, or making sure that the same version of a package is used everywhere, then you may add an override.
> 
> Overrides provide a way to replace a package in your dependency tree with another version, or another package entirely. These changes can be scoped as specific or as vague as desired.
> 
> Overrides are only considered in the root package.json file for a project. Overrides in installed dependencies (including workspaces) are not considered in dependency tree resolution. Published packages may dictate their resolutions by pinning dependencies or using an npm-shrinkwrap.json file.

We should design how to handle these, as they do not apply to a detected package, but to other packages in the dependency tree and only in specific conditions.

divizbansal:
Fixes #4122 

**_This PR adds a minimal approach to parse and store overrides in package.json into extra_data, plus tests._**

### Tasks


tests/licensedcode/data/license_db/license_dump/bash-exception-gpl.LICENSE
```diff
@@ -1,20 +1,20 @@
...
-Paragraph 2c of the General Public License.  This is to say, there is
-no requirement for Bash to print a notice when it is started
-interactively in the usual way.  We made this exception because users
-and standards expect shells not to print such messages.  This
-exception applies to any program that serves as a shell and that is
+---
+key: bash-exception-gpl
+short_name: Bash exception to GPL
+name: Bash exception to GPL
+category: Copyleft
```
AyanSinhaMahapatra, line 5: These changes are not related to the issue, please remove these. Probably caused by your IDE settings

tests/packagedcode/data/npm/overrides_test/package.json
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "my-override-example",
+    "version": "1.0.0",
+    "overrides": {
```
AyanSinhaMahapatra, line 4: Can you add real examples seen in the wild? 
See https://github.com/search?q=path%3A*%2Fpackage.json+overrides&type=code

tests/packagedcode/test_npm.py
```diff
@@ -543,6 +543,24 @@ def test_npm_scan_with_private_package_json_and_yarn_lock(self):
             expected_file, result_file, remove_uuid=True, regen=REGEN_TEST_FIXTURES
         )
 
+    def test_parse_npm_package_json_with_overrides(self):
```
AyanSinhaMahapatra, line 546: Tests will look like this: https://github.com/aboutcode-org/scancode-toolkit/blob/56aa1e7209120b56b79012f36762d1920aa9c7aa/tests/packagedcode/test_npm.py#L191

src/packagedcode/npm.py
```diff
@@ -578,6 +578,10 @@ def _parse(cls, json_data, package_only=False):
             if value:
                 extra_data[extra_data_field] = value
 
+        overrides = json_data.get('overrides')
+        if overrides:
+            extra_data['overrides'] = overrides
```
AyanSinhaMahapatra, line 583: This is okay as a start but we need to do a bit more here.
See the go replace directives PR and reviews which is a similar issue: https://github.com/aboutcode-org/scancode-toolkit/pull/3693#discussion_r1527842940

AyanSinhaMahapatra (changes requested):
@divizbansal Thanks++ for the PR. See comments for improvements

PR #4186, opened 2025-03-11
Rename bsd-no-disclaimer-unmodified to *-axis-*

lzcunt:
This license seems to be a variant of the BSD-Axis license, not a variant of the BSD-2-Clause license. This patch fixes it by renaming the license identifier to reflect reality.

### Tasks

* [X] Reviewed [contribution guidelines](https://github.com/nexB/scancode-toolkit/blob/develop/CONTRIBUTING.rst)
* [X] PR is descriptively titled 📑 and links the original issue above 🔗
* [X] Commits are in uniquely-named feature branch and has no merge conflicts 📁

### DIFF
--- a/src/licensedcode/data/licenses/bsd-axis-no-disclaimer-unmodified.LICENSE
+++ b/src/licensedcode/data/licenses/bsd-axis-no-disclaimer-unmodified.LICENSE
@@ -1,12 +1,12 @@
 ---
-key: bsd-no-disclaimer-unmodified
-short_name: BSD-2-Clause no disclaimer Unmod
-name: BSD-2-Clause without Warranty Disclaimer with Unmodified requirement
+key: bsd-axis-no-disclaimer-unmodified
+short_name: BSD-Axis no disclaimer Unmod
+name: BSD-Axis without Warranty Disclaimer with Unmodified requirement
 category: Permissive
 owner: Unspecified
 notes: A rare variant that comes with a slightly different wording and no warranty disclaimer
     and the words "without modifications" applying to the license conditions
-spdx_license_key: LicenseRef-scancode-bsd-no-disclaimer-unmodified
+spdx_license_key: LicenseRef-scancode-bsd-axis-no-disclaimer-unmodified
 ---
 
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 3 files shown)


DennisClark (commented):
We normally do not change the key of a license once it is in use.  I am fine with changing the license name and short name to the suggested values, but we should leave the key as it is.

DennisClark (commented):
we should also leave the spdx_license_key as it currently is.

lzcunt:
What about the file name? Is that safe to change without breaking anything?

DennisClark (commented):
also, we should not change the file name

lzcunt:
Sure then, I'll update this to only touch the name and short name

DennisClark:
@lzcunt   thanks for doing this; your suggested name changes make a lot of sense.

pombredanne:
@lzcunt 

As pointed by @DennisClark 
1. we do not update license name nor change their license key nor SPDX key as a principle
2. we can however do small inconsequential changes or mark as deprecated one license and create a new one, carefully as we still need to keep the old SPDX key in the new one and point to the new one in notes of the old one

Can you elaborate where you found this?

lzcunt:
> Can you elaborate where you found this?

GNU-EFI has weird licensing, which I've been nerdsniped into figuring out. I've opened ncroxon/gnu-efi#68 about this. LICENSE.efilib is just BSD-2-Clause with no effective differences, so it can be added as a rule. Please read the linked issue for information about LICENSE.efilib which is out of scope for this PR. It also has information related to the license in this PR as well.

BSD(-Axis?)-No-Disclaimer-Unmodified is found in a couple arm assembly files and riscv/loongarch assembly files based on the aforementioned files. Upstream doesn't seem to be aware of its existence. The license seems to originate from a `arm_math.c` file that was copied around multiple projects.

While I was looking for an identifier for it, I noticed ScanCode LicenseDB had an exact match which incorrectly labeled it as a variant of the BSD-2-Clause license when it's actually a variant of BSD-3-Clause with the second clause removed. Then I found BSD-Axis-Unmodified on LicenseDB which had the disclaimer so I assumed this was based on the BSD-Axis-Unmodified license.

I'm not a user of ScanCode and I'm only interested in using LicenseDB as a standard-ish registry of LicenseRef- identifiers. I can drop this PR and not use a LicenseRef-scancode- identifier for packaging GNU-EFI, which is how it's usually done in nixpkgs for licenses without standard SPDX identifiers anyway. However, I believe naming the license correctly (or at least not referring to it as BSD-2-Clause when it has the 3rd clause of BSD-3-Clause) would be beneficial for ScanCode as well.

TL;DR: There's two different licenses in question here, one being the one I renamed in this PR and the other is a variant of BSD-2-Clause with a potentially-legally-non-binding-but-has-to-be-reproduced disclaimer about the EFI spec hosted on the original website efilib was also hosted in (which is down) and shouldn't apply to the efilib code itself. So it doesn't put any extra restrictions and therefore *is* BSD-2-Clause. Meanwhile the other license matches BSD-No-Disclaimer-Unmodified which is named wrongly.

IANAL

PR #4194, opened 2025-03-17
Fix and enhance support for different bazel metadata versions

### LINKED ISSUE #4196: Extraction of package data from Bazel Metadata files is too strict
## Short Description
Currently [in the package code](https://github.com/aboutcode-org/scancode-toolkit/blob/e830934bfa8cb61060fa8b60c12945fee93c6c44/src/packagedcode/build.py#L377), we check for one of two versions of .bzl files and convert them into `PackageData` slightly differently. There are 2 issues there:
1. Currently, the check is broken making it much looser than intended
2. If fixed, the check is too tight, because METADATA.bzl does not follow strict standards

I propose to remove the check altogether and always try to extract as much information from the file as possible.

## Possible Labels

- package formats
- package scan
- bug
- easy

## Select Category

-  [x] Enhancement
-  [ ] Add License/Copyright
-  [ ] Scan Feature
-  [ ] Packaging
-  [ ] Documentation
-  [x] Expand Support
-  [ ] Other

## **Describe the Update**
I propose to combine the bodies of the `if`-statements, i.e. we simply check for the presence of the known fields in METADATA.bzl files and map them to their corresponding field in PackageData. This change also makes it easy to add further fields, if needed. 
In fact, I'll also add support of a field `package_url` containing a PURL, which is widely used at my organization.

See #4194 for implementation.

We use METADATA.bzl files for dependency management. When importing packages from Maven, the name usually contains a colon `:`, which means it is not suitable as name field of a `PURL`. This causes ScanCode to error when parsing our .bzl files.  We do include a valid PURL which ScanCode does not recognize currently.

PR is already open: #4194

abraemer:
I noticed that the conditions deciding between the two version for `.bzl` files are somewhat wrong. In Python `'a' and 'b' in some_string` is equal to `('a' and 'b') in some_string` and thus simplifies to `'b' in some_string`. In this instance, I think it shouldn't make a huge difference because fortunately the strings `upstream_address` and `vcs_commit_hash` still somewhat distinguish between versions.

I think just fixing the conditions makes them too restrictive (AFAIK there is no real standard for these METADATA.bzl files), so I combined both paths into one. This ensures that we always extract as much information as possible. 

I also added the option to use information from a `package_url` field which is important to me as we often have some `METADATA.bzl` files which are generated from Maven dependencies and thus the `name` is invalid by PURL specification. I wrote a somewhat realistic test case for this.

### Tasks

Closes #4196

### DIFF
--- a/src/packagedcode/build.py
+++ b/src/packagedcode/build.py
@@ -13,6 +13,7 @@ import ast
 from collections import defaultdict
 
 from commoncode import fileutils
+from packageurl import PackageURL
 
 from licensedcode.cache import build_spdx_license_expression
 from licensedcode.cache import get_cache
@@ -374,52 +375,27 @@ class BuckMetadataBzlHandler(BaseStarlarkManifestHandler):
                 )
             )
 
-        if (
-            'upstream_type' in metadata_fields
-            and 'name' in metadata_fields
-            and 'version' in metadata_fields
-            and 'licenses' in metadata_fields
-            and 'upstream_address' in metadata_fields
-        ):
-            # TODO: Create function that determines package type from download URL,
-            # then create a package of that package type from the metadata info
-            package_data = dict(
-                datasource_id=cls.datasource_id,
-                type=metadata_fields.get('upstream_type', cls.default_package_type),
-                name=metadata_fields.get('name'),
-                version=metadata_fields.get('version'),
-                extracted_license_statement=metadata_fields.get('licenses', []),
-                parties=parties,
-                homepage_url=metadata_fields.get('upstream_address', ''),
-                # TODO: Store 'upstream_hash` somewhere
-            )
-            yield models.PackageData.from_data(package_data, package_only=True)
+        # TODO: Create function that determines package type from download URL,
+        # then create a package of that package type from the metadata info
+        package_data = dict(
+            datasource_id=cls.datasource_id,
+            type=metadata_fields.get('upstream_type', metadata_fields.get('package_type', cls.default_package_type)),
+            name=metadata_fields.get('name'),
+            version=metadata_fields.get('version'),
+            extracted_license_statement=metadata_fields.get('licenses', metadata_fields.get('license_expression')),
+            parties=parties,
+            homepage_url=metadata_fields.get('upstream_address', metadata_fields.get('homepage_url')),
+            download_url=metadata_fields.get('download_url'),
+            vcs_url=metadata_fields.get('vcs_url'),
+            sha1=metadata_fields.get('download_archive_sha1'),
+            # TODO: Store 'upstream_hash` somewhere
+        )
+        if 'vcs_commit_hash' in metadata_fields:
+            package_data["extra_data"] = dict(vcs_commit_hash=metadata_fields['vcs_commit_hash'])
+        if 'package_url' in metadata_fields:
+            package_data.update(PackageURL.from_string(metadata_fields['package_url']).to_dict())
+        yield models.PackageData.from_data(package_data, package_only=True)
 
-        if (
-            'package_type' in metadata_fields
-            and 'name' in metadata_fields
-            and 'version' in metadata_fields
-            and 'license_expression' in metadata_fields
-            and 'homepage_url' in metadata_fields
-            and 'download_url' in metadata_fields
-            and 'vcs_url' in metadata_fields
-            and 'download_archive_sha1' in metadata_fields
-            and 'vcs_commit_hash' in metadata_fields
-        ):
-            package_data = dict(
-                datasource_id=cls.datasource_id,
-                type=metadata_fields.get('package_type', cls.default_package_type),
-                name=metadata_fields.get('name'),
-                version=metadata_fields.get('version'),
-                extracted_license_statement=metadata_fields.get('license_expression', ''),
-                parties=parties,
-                homepage_url=metadata_fields.get('homepage_url', ''),
-                download_url=metadata_fields.get('download_url', ''),
-                vcs_url=metadata_fields.get('vcs_url', ''),
-                sha1=metadata_fields.get('download_archive_sha1', ''),
-                extra_data=dict(vcs_commit_hash=metadata_fields.get('vcs_commit_hash', ''))
-            )
-            yield models.PackageData.from_data(package_data, package_only=True)
 
     @classmethod
     def assign_package_to_resources(cls, package, resource, codebase, package_adder):
--- a/src/packagedcode/build.py
+++ b/src/packagedcode/build.py
@@ -377,23 +377,47 @@ class BuckMetadataBzlHandler(BaseStarlarkManifestHandler):
 
         # TODO: Create function that determines package type from download URL,
         # then create a package of that package type from the metadata info
+        
+        if 'upstream_type' in metadata_fields:
+            package_type = metadata_fields['upstream_type']
+        elif 'package_type' in metadata_fields:
+            package_type = metadata_fields['package_type']
+        else:
+            package_type = cls.default_package_type
+
+        if 'licenses' in metadata_fields:
+            extracted_license_statement = metadata_fields['licenses']
+        else:
+            extracted_license_statement = metadata_fields.get('license_expression')
+
+        if 'upstream_address' in metadata_fields:
+            homepage_url = metadata_fields['upstream_address']
+        else:
+            homepage_url = metadata_fields.get('homepage_url')
+        
+
+        extra_data = {}
+        if 'vcs_commit_hash' in metadata_fields:
+            extra_data['vcs_commit_hash'] = metadata_fields['vcs_commit_hash']
+        if 'upstream_hash' in metadata_fields:
+            extra_data['upstream_hash'] = metadata_fields['upstream_hash']
+
         package_data = dict(
             datasource_id=cls.datasource_id,
-            type=metadata_fields.get('upstream_type', metadata_fields.get('package_type', cls.default_package_type)),
+            type=package_type,
             name=metadata_fields.get('name'),
             version=metadata_fields.get('version'),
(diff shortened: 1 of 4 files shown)


abraemer:
Test failures seem unrelated to my changes

tests/packagedcode/data/build/metadatabzl/with-package-url/METADATA.bzl
```diff
@@ -0,0 +1,12 @@
...
+    "licenses": [
+        "BSD-3-Clause",
+    ],
+    "maintainers": [
+        "oss_foundation",
+    ],
+    "name": "androidx.compose.animation:animation",
+    "upstream_address": "https://developer.android.com/jetpack/androidx/releases/compose-animation#0.0.1",
+    "version": "0.0.1",
+    "package_url" : "pkg:maven/androidx.compose.animation/animation@0.0.1"
```
AyanSinhaMahapatra, line 11: Could you maybe link to some examples of this type of manifests with `package_url` fields and maybe add one of those as tests, it's best to use real world examples probably. Or if you got this from some real example, you can also link that file here.

src/packagedcode/build.py
```diff
@@ -374,54 +375,27 @@ def parse(cls, location, package_only=True):
...
+            type=metadata_fields.get('upstream_type', metadata_fields.get('package_type', cls.default_package_type)),
+            name=metadata_fields.get('name'),
+            version=metadata_fields.get('version'),
+            extracted_license_statement=metadata_fields.get('licenses', metadata_fields.get('license_expression')),
+            parties=parties,
+            homepage_url=metadata_fields.get('upstream_address', metadata_fields.get('homepage_url')),
+            download_url=metadata_fields.get('download_url'),
+            vcs_url=metadata_fields.get('vcs_url'),
+            sha1=metadata_fields.get('download_archive_sha1'),
+            # TODO: Store 'upstream_hash` somewhere
```
AyanSinhaMahapatra: Could you add a line here to store `upstream_hash` in the `extra_data` mapping for this PackageData?

src/packagedcode/build.py
```diff
@@ -374,54 +375,27 @@ def parse(cls, location, package_only=True):
...
+        )
+        if 'vcs_commit_hash' in metadata_fields:
+            package_data["extra_data"] = dict(vcs_commit_hash=metadata_fields['vcs_commit_hash'])
```
AyanSinhaMahapatra: Wouldn't it be better to do a get() rather than checking if this exists in the mapping?

AyanSinhaMahapatra (changes requested):
Thanks a lot @abraemer , a couple nits for your consideration. Looking good otherwise.

src/packagedcode/build.py
abraemer: Done

src/packagedcode/build.py
abraemer: When including the `upstream_hash`, I rewrote this part. Now we always create the `extra_data` dict with all keys and I use `.get` to extract the values.

tests/packagedcode/data/build/metadatabzl/with-package-url/METADATA.bzl
abraemer: Unfortunately, I cannot link you one of our internal files. However my example file is very close to an actual file (the version number is wrong but that's all).

abraemer:
friendly ping since it has been 2 weeks now and I think this ready :)  What do you think @AyanSinhaMahapatra?

src/licensedcode/juliacode.jl
```diff
@@ -0,0 +1,175 @@
...
+        !haskey(set2, k) && continue
+        ret[k] = min(v, set2[k])
+    end
+    return ret
+end
+
+set_high_intersection_filter(set::BitSet, cutoff) = filter(<=(cutoff), set)
+set_high_intersection_filter(set::MSet, cutoff) = filter(pair -> pair.first<=cutoff, set)
+
+function compare_token_sets(qset, iset, len_legalese, min_matched_length_high, min_matched_length; minimum_containment=0, high_resemblance_threshold=0.8)
```
AyanSinhaMahapatra: This file and the code though looks quite interesting, doesn't seem to be related to licensedcode at all and is in julia which we don't use  :P maybe you commited this by mistake? Curious about what you're trying to do here, some kind of post-processing on scancode license detection results?

src/packagedcode/build.py
```diff
@@ -374,54 +375,28 @@ def parse(cls, location, package_only=True):
...
+            type=metadata_fields.get('upstream_type', metadata_fields.get('package_type', cls.default_package_type)),
+            name=metadata_fields.get('name'),
+            version=metadata_fields.get('version'),
+            extracted_license_statement=metadata_fields.get('licenses', metadata_fields.get('license_expression')),
+            parties=parties,
+            homepage_url=metadata_fields.get('upstream_address', metadata_fields.get('homepage_url')),
+            download_url=metadata_fields.get('download_url'),
+            vcs_url=metadata_fields.get('vcs_url'),
+            sha1=metadata_fields.get('download_archive_sha1'),
+            # TODO: Store 'upstream_hash` somewhere
```
AyanSinhaMahapatra: Coudl you also remove the `TODO` line here since you've addressed that? Thanks

src/packagedcode/build.py
```diff
@@ -374,54 +375,28 @@ def parse(cls, location, package_only=True):
...
-                vcs_url=metadata_fields.get('vcs_url', ''),
-                sha1=metadata_fields.get('download_archive_sha1', ''),
-                extra_data=dict(vcs_commit_hash=metadata_fields.get('vcs_commit_hash', ''))
-            )
-            yield models.PackageData.from_data(package_data, package_only=True)
+        # TODO: Create function that determines package type from download URL,
+        # then create a package of that package type from the metadata info
+        package_data = dict(
+            datasource_id=cls.datasource_id,
+            type=metadata_fields.get('upstream_type', metadata_fields.get('package_type', cls.default_package_type)),
```
AyanSinhaMahapatra: Very minor nit: These 2 levels of `.get()` and defaults on fallback are not particularly readable even though it's correct, what do you think about changing this to a if-else block with something like:
```
if 'upstream_type' in metadata_fields:
   type = metadata_fields.get('upstream_type')
elif 'package_type' in metadata_fields:
   type = metadata_fields.get('package_type')
else:
   type = cls.default_package_type
```
Or something else which is more readable like this? Same for `extracted_license_statement` and  `homepage_url` below.
Usually this is why we have variables with more complicated extraction defined before we create the `package_data = dict()` mapping, and only keep simple `.get()` operations with defaults there.

tests/packagedcode/data/build/metadatabzl/with-package-url/METADATA.bzl
AyanSinhaMahapatra: > Unfortunately, I cannot link you one of our internal files.

No worries on that obviously, but can you find some other real examples with `package_url` anywhere on github/elsewhere, or is this relatively new?

AyanSinhaMahapatra (changes requested):
Thanks for the changes @abraemer, this is almost ready with a couple small changes, apologies for the late review :sweat_smile:

src/licensedcode/juliacode.jl
abraemer: Oh wow how did that end up here? :sweat_smile: 
This from an experiment where I rewrote a small, self-contained part in Julia to see whether that speeds things up (it did but only in single-threaded mode so far).
Anyhow, good catch - I removed the file again.

src/packagedcode/build.py
abraemer: I rewrote the nested `.get`s to explicit `if-else` statements. In addition, I also refactored the `extra_data` to follow the other fields.

tests/packagedcode/data/build/metadatabzl/with-package-url/METADATA.bzl
abraemer: I searched on GitHub but did not find any other files that use that field but also I did not find many examples of bzl files at all. So perhaps it is not very common (yet?).
As this doesn't interfere with other things, would you be okay with merging this anyways? After all it just adds another possible data source for bzl-files that has the potential to be useful for others :)

abraemer:
@AyanSinhaMahapatra friendly ping :)

src/packagedcode/build.py
```diff
@@ -374,54 +375,49 @@ def parse(cls, location, package_only=True):
...
+        else:
+            extracted_license_statement = metadata_fields.get('license_expression')
+
+        if 'upstream_address' in metadata_fields:
+            homepage_url = metadata_fields['upstream_address']
+        else:
+            homepage_url = metadata_fields.get('homepage_url')
+        
+        extra_data = dict(
+            vcs_commit_hash=metadata_fields.get('vcs_commit_hash'),
```
AyanSinhaMahapatra: These should be added only if they are present, optionally. Otherwise we have these keys with empty values. See the only remaining test failures because of this: https://dev.azure.com/nexB/scancode-toolkit/_build/results?buildId=15913&view=logs&j=1412dbfa-421a-5a40-8ddf-7073fc46aa19&t=13c90161-2109-5957-f563-a96fefc7d869

AyanSinhaMahapatra (commented):
Thanks++ @abraemer, last nit for your considerations. Ready to merge otherwise. Thanks for your patience.  

Could you also merge from develop again, should get rid of the flaky test failures.

abraemer:
@AyanSinhaMahapatra last nit is fixed and branch is up-to-date with develop. Let's see what CI says :)

AyanSinhaMahapatra:
Thanks++ @abraemer, all green and merging!

Issue #4420, opened 2025-06-04
Improve score when `extra-words` are found in the correct position


alok1304:
Labels: GSoC

msafeerhassan:
Is it still open for contribution or not?

alok1304:
> Is it still open for contribution or not?

Yes, it is open. Actually, I am working on this issue.

PR #4432, opened 2025-06-19
Improve score by supporting `extra_phrase` for extra words in rules

alok1304:
Follow up of:
- https://github.com/aboutcode-org/scancode-toolkit/pull/4424

Add new phrases like `extra_phrase` this is special for `extra-words`. This phrase is represented in the format `[[n]]`, where n indicates the maximum number of `extra-words` allowed at that position in the rule.

If `extra-words` appear at the correct position and their count does not exceed the allowed limit `n`, then the score is increased to `100`.

Reference #4420 

### Tasks

Signed-off-by: Alok Kumar [alokkumarjipura9973@gmail.com](mailto:alokkumarjipura9973@gmail.com)


src/licensedcode/data/rules/bsd-new_578.RULE
```diff
@@ -6,7 +6,9 @@ minimum_coverage: 99
 
 Software License Agreement (BSD License)
 
-Redistribution and use in source and binary forms, with or without
+[[15]]
```
AyanSinhaMahapatra: Could you explain why these two `[[15]]` and `[[4]]` are added here? This is incorrect and problematic to the detection of this rule IMHO

AyanSinhaMahapatra: If this is for allowing copyright statements + `all rights reserved` statements, then this has to be done separately as post-processing as this allows any other large piece of text to be allowed towards the top of the rule/text which might allow/silence other license detetions and is not intended at all.

AyanSinhaMahapatra: Let's add a simpler case as a test of `3-seq` match which does not have copyright statements in it and keep this text example (`tests/licensedcode/data/plugin_license/extra-words/scan-extra-words-3-seq-license.expected.json` rename to `scan-extra-words-with-copyrights-license.expected.json`) as is because we need to process these further later.

src/licensedcode/detection.py
```diff
@@ -1807,6 +1827,10 @@ def analyze_detection(license_matches, package_license=False):
         threshold=IMPERFECT_MATCH_COVERAGE_THR,
     ):
         return DetectionCategory.IMPERFECT_COVERAGE.value
+    
+    # Case where `extra-words` are in the right place
+    elif is_extra_words_at_valid_positions(license_matches=license_matches):
```
AyanSinhaMahapatra: Having extra words and having extra words at valid positions are not two seperate `if`.

The `is_extra_words_at_valid_positions` call needs to be called conditionally inside the `elif has_extra_words()` statement, and the detection log populated on that basis also with an `if-else` inside.

src/licensedcode/detection.py
```diff
@@ -1159,6 +1163,16 @@ def has_low_rule_relevance(license_matches):
     )
 
 
+def is_extra_words_at_valid_positions(license_matches):
+    """
+    Return True if any of the matches in ``license_matches`` List of LicenseMatch
+    has extra words are in the correct place.
+    """
```
AyanSinhaMahapatra: We need to check both a bit explicitly:
1. For all the matches which have extra words, they are in correct location
2. For all the matches which does not have extra words, they are correct detections

AyanSinhaMahapatra: And add a test accordingly

AyanSinhaMahapatra (changes requested):
Thanks @alok1304! Looking much better

See comments for your consideration. I've updated your PR description to mention that this is a follow up PR, since there is important context and reviews in the previous PR, we need to preserve this as required.

src/licensedcode/data/rules/bsd-new_578.RULE
alok1304: Okay, I will add a test for `3-seq` where there is no copyright statements.

alok1304: > Could you explain why these two `[[15]]` and `[[4]]` are added here? This is incorrect and problematic to the detection of this rule IMHO

as of now I remove this part

src/licensedcode/detection.py
alok1304: got it , thanks

src/licensedcode/data/rules/bsd-new_578.RULE
alok1304: > If this is for allowing copyright statements + `all rights reserved` statements, then this has to be done separately as post-processing as this allows any other large piece of text to be allowed towards the top of the rule/text which might allow/silence other license detetions and is not intended at all.

Actually this copyright statements + `all rights reserved` statements before this some rule test is matched that's why I adding `extra-phrase` marker , okay I will remove that part.

src/licensedcode/detection.py
alok1304: > And add a test accordingly

where I should add a test and like how I implement for all `license_matches`

alok1304:
I addet test for `3-seq` where there is no detection of copyrights statements , Ref: https://github.com/xyzzy-022/xyzzy/blob/5a16eb998470241b33ad3caa6a4946d0448a16b6/LEGAL.md?plain=1#L97
this file when we scan we got `extra-words` so I added `extra-phrase` marker in that corresponding matched rule. Such that we can improve the `score`.

src/licensedcode/detection.py
alok1304: Also i added test for `is_extra_words_position_valid` see: https://github.com/aboutcode-org/scancode-toolkit/pull/4432/files#diff-d1520ccce311f8f4d4932ba68589bc23b098f8090a307b0b440edb4a846ae21cR1325

alok1304:
Actually, the previous test case is failing because of `extra-phrase` is removed from rules while loading, after that when scan same rules then matcher is `3-seq` not any one of these`"1-hash", "1-spdx-id", "2-aho"` this is due to `extra-phrase` marker that are present in rules. This test case is failing https://github.com/aboutcode-org/scancode-toolkit/blob/develop/tests/licensedcode/test_detection_validate.py#L101
```
   def check_rule_or_license_can_be_detected_exactly(licensish):
        """
        Check that a rule or license can be detected exactly, either by thyself (or with an exact match
        to any rule for deprecated rules).
        """
        idx = cache.get_index()
        deadline = time() + 20  # ms
        matches = idx.match(query_string=licensish.text, _skip_hash_match=True, deadline=deadline)
        # ensure we can self-detect exactly
        expected = [licensish.identifier]
        results = [m.rule.identifier for m in matches]
    
        if results != expected:
            expected.append(f'file://{licensish.rule_file()}')
            assert results == expected
        icm = is_correct_detection(matches)
        if not icm:
            expected.append(f'file://{licensish.rule_file()}')
>           assert results == expected

E           AssertionError: assert ['bsd-new_newlib3.RULE'] == ['bsd-new_newlib3.RULE',\n 'file:///home/vsts/work/1/s/src/licensedcode/data/rules/bsd-new_newlib3.RULE']
E             Right contains one more item: 'file:///home/vsts/work/1/s/src/licensedcode/data/rules/bsd-new_newlib3.RULE'
E             Full diff:
E               [
E                'bsd-new_newlib3.RULE',
E             -  'file:///home/vsts/work/1/s/src/licensedcode/data/rules/bsd-new_newlib3.RULE',
E               ]
```

 because `is_correct_detection` gives False because it not include `3-seq` in matcher see: https://github.com/aboutcode-org/scancode-toolkit/blob/develop/src/licensedcode/detection.py#L1090
That's why the test case is failing. So, I added `3-seq` see:  https://github.com/aboutcode-org/scancode-toolkit/pull/4432/files#diff-5f6ccc1d8f9540cae715bddaf337246f4c69c48ca596b2086e05b18c541aef4aR1098

Now, all the test cases are passed :)

src/licensedcode/detection.py
```diff
@@ -1087,7 +1095,7 @@ def is_correct_detection(license_matches):
     ]
 
     return (
-        all(matcher in ("1-hash", "1-spdx-id", "2-aho") for matcher in matchers)
+        all(matcher in ("1-hash", "1-spdx-id", "2-aho", "3-seq") for matcher in matchers)
```
alok1304: i add `3-seq` because of `extra-phrase` is removed from rules while loading, after that when we scan same rules, now matcher is `3-seq` not any one of these `"1-hash", "1-spdx-id", "2-aho"` this is due to `extra-phrase `marker that are present in rules. That's why this gives `3-seq`. We already loaded the rules where `extra-phrase` is removed, but not in the actual rule file.

src/licensedcode/detection.py
alok1304: added test

tests/licensedcode/data/plugin_license/extra-words/scan-extra-words-3-seq-license/with-copyrights/LICENSE
```diff
@@ -1,31 +1,31 @@
...
-  written permission.
-
-THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR
-IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
-FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR
-CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
-DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
-DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
-IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT
+Software License Agreement (BSD License)
```
AyanSinhaMahapatra: NO need for this renaming, please keep this `3-seq` test as-is. Same for the test expectation file and test name.

tests/licensedcode/data/plugin_license/extra-words/scan-extra-words-3-seq-license.expected.json
```diff
@@ -1,70 +0,0 @@
-{
```
AyanSinhaMahapatra, line 1: No need to delete and add a new file with some other name, please restore this.

tests/licensedcode/data/plugin_license/extra-words/scan-extra-words-2-aho-license.expected.json
```diff
@@ -6,7 +6,7 @@
       "license_expression_spdx": "BSD-3-Clause",
       "detection_count": 1,
       "detection_log": [
-        "extra-words"
+        "extra-words-permitted-in-rule"
```
AyanSinhaMahapatra, line 9: Test expectations look good, thanks!

src/licensedcode/detection.py
AyanSinhaMahapatra: @alok1304 we cannot add this here, `is_correct_detection` cannot have `"3-seq"` built in, you need to create a different function with different name as this is not the same anymore. Where are you using this or why ahave you added this, because we need to ensure rule texts are matched exactly?

AyanSinhaMahapatra (changes requested):
Thanks @alok1304. looking great overall, see comments for your consideration.

src/licensedcode/detection.py
AyanSinhaMahapatra: We need to create a new function for use at https://github.com/aboutcode-org/scancode-toolkit/pull/4432#issuecomment-3028620324, we cannot put `3-seq` in the old `is_correct_detection` function because that disturbs the detection post-processing.

alok1304: > We need to create a new function for use at [#4432 (comment)](https://github.com/aboutcode-org/scancode-toolkit/pull/4432#issuecomment-3028620324), we cannot put `3-seq` in the old `is_correct_detection` function because that disturbs the detection post-processing.

okay, got it

tests/licensedcode/data/plugin_license/extra-words/scan-extra-words-3-seq-license.expected.json
```diff
@@ -1,70 +1,70 @@
...
...butions in binary form must reproduce the above\r\n  copyright notice, this list of conditions and the\r\n  following disclaimer in the documentation and/or other\r\n  materials provided with the distribution.\r\n\r\n* Neither the name of Kevin Decker nor the names of its\r\n  contributors may be used to endorse or promote products\r\n  derived from this software without specific prior\r\n  written permission.\r\n\r\nTHIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS \"AS IS\" AND ANY EXPRESS OR\r\nIMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND\r\nFITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR\r\nCONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL\r\nDAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,\r\nDATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER\r\nIN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT\r\nOF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.",
```
AyanSinhaMahapatra: This is the test I asked to rename as `scan-extra-words-with-copyrights-license.expected.json`. See https://github.com/aboutcode-org/scancode-toolkit/pull/4432#discussion_r2190319444

The test where we are getting a score of 100 from `3-seq` matches should be`scan-extra-words-3-seq-license.expected.json`, like it was before this PR.
This test file which has copyrights as extra words and the score is still not perfect because of this, we can name as `scan-extra-words-with-copyrights-license.expected.json`.

We need to address improving score in matches with extra words for copyrights/URLs seperately, in another PR.

tests/licensedcode/test_plugin_license_detection.py
```diff
@@ -130,6 +130,24 @@ def test_license_match_extra_words_3_seq():
     check_json_scan(test_loc, result_file, regen=REGEN_TEST_FIXTURES)
 
 
+def test_license_match_extra_words_3_seq_without_copyrights():
```
AyanSinhaMahapatra:
```suggestion
def test_license_match_extra_words_3_seq_with_copyrights():
```
The new test added for the new case should be the one which does not have perfect score yet. The olds tests that we added in https://github.com/aboutcode-org/scancode-toolkit/pull/4402 should remain as-is (only change in the expectation, with score and detection log improvements). The new case should be added in a new test.

tests/licensedcode/test_detection_validate.py
```diff
@@ -99,7 +100,7 @@ def check_rule_or_license_can_be_detected_exactly(licensish):
         assert results == expected
 
     icm = is_correct_detection(matches)
-    if not icm:
+    if not icm and not has_extra_phrases(matches):
```
AyanSinhaMahapatra: In this case, we are ignoring incorrect detections when we have extra phrases in rules. I think it would be better if we process the text to remove extra-word markers above at `idx.match(query_string=licensish.text,` this is cleaner.

AyanSinhaMahapatra (approved):
Thanks @alok1304, looking good!

Two minor changes requested for your consideration:
1. in the test and test filenames we should add the new case as a new test and let the old tests (which we have perfect scores for now) stay as-is
2. In the rule detection tests we should process the text before running license detection instead of ignoring rules with extra word markers.

Asking a review from @pombredanne too as this is ready for him to take a look at.

tests/licensedcode/test_detection_validate.py
alok1304: makes sense, I will do this

tests/licensedcode/test_plugin_license_detection.py
```diff
@@ -126,6 +126,24 @@ def test_license_match_extra_words_3_seq():
...
     run_scan_click(args)
+    test_loc = test_env.get_test_loc('plugin_license/extra-words/scan-extra-words-3-seq-with-copyrights-license.expected.json')
+    check_json_scan(test_loc, result_file, regen=REGEN_TEST_FIXTURES)
+
+
+def test_license_match_extra_words_3_seq_with_copyrights():
+    test_dir = test_env.get_test_loc('plugin_license/extra-words/scan-extra-words-3-seq-license-with-copyright/')
+    result_file = test_env.get_temp_file('json')
+    args = [
+        '--license',
```
AyanSinhaMahapatra, line 138: Can you also use the `--copyright` option here so we have copyright detection in the resource, we have to use this to resolve the issue.

alok1304:
Hey @AyanSinhaMahapatra, this is ready to merge.

src/licensedcode/match.py
```diff
@@ -1071,6 +1075,84 @@ def merge_matches(matches, max_dist=None, trace=TRACE_MERGE):
 # early from the loops: trying to check containment on wildly separated matches
 # does not make sense
 
+def is_extra_words_position_valid(match):
```
pombredanne, line 1078: I wonder if the scoring should not be more subtle. In practice, we can have multiple regions allowing for extra words. And some regions may have OK extra words, and some other may not have that, so this is may not be a binary True or False, with 100 or not 100 but instead a score computation that may be more nuanced?

alok1304: sorry for late reply, actually i am busy in my exams.
yes you are right,   any or one of the extra-words are at wrong place then they simple return 'False' as of now

alok1304: For score computation:
 lets take an eg suppose there are 3 position of extra-words , lets take we got  2 extra-words at two place which is at correct places and 1 extra-words coming at wrong place then I am thinking how we can compute the score? here 
 only 1 extra-words at wrong place.

Issue #2562, opened 2021-06-17
Do not understand the --from-json feature


ididitmyway:
### Description

> Hi there, I am confused about the --from-json freature
>
> Is it a feature to convert one output to another litke from json to json-pp,
> or is it a feature to import the last scan result, so that the scantime itsef could be improved, because it is scipping anything it 
> already had have scanned in the last scan ?

### Select Category

-  [ ] Inconsistency
-  [ ] New Section Request
-  [ ] General Improvement
-  [ ] Typo/Mistakes
-  [x] Other

pombredanne:
The purpose is that instead of using real code directories or files as an input you can take existing JSON scans and process them further.
For instance:
- `--from-json --to-csv foos.csv  foos.json`
Or combine two in one
- `--from-json --to-json combined.json  foo.json bar.json`
Or apply some other options (as long as there is no access to files needed, eg. scan options that access files will not be available,

pombredanne:
The documentation may need some love though!

PR #4467, opened 2025-07-09
Improve documentation

### LINKED ISSUE #2980: CLI Option `processes`, what is a good value?
### Description

The documentation contains very little information about the effect of the  `--processes` parameter. It mentions the default value, but not much more. 

I am struggling to understand how many processes would be a sensible value to improve the speed of the scan? The default is 1, I guess that is a conservative value to ensure Scancode-Toolkit works even in resource restricted environments (e.g. low powered VM).

Should the processes be equal to the number of cores of the CPU? Is there an amount of processes which is just too much? Because not every part of the code can be parallelized.

While it is difficult to give a general recommendation. It would be beneficial to have more information, so that users can find there individual "good" value.

### Link to Documentation Page

https://scancode-toolkit.readthedocs.io/en/stable/cli-reference/list-options.html#all-core-scan-options

### Select Category

-  [ ] Inconsistency
-  [ ] New Section Request
-  [x] General Improvement
-  [ ] Typo/Mistakes
-  [ ] Other

For more information have a look at: #608

chinyeungli:
Fixes 
#2980 Added note on recommended process number for optimal performance
#2562 Enhance description of the `--from-json` option
Removed duplicate content from "All Available Options"

### Tasks

* [] Tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR

The failing tests don't related to the documentation update.

Signed-off-by: Chin Yeung Li <tli@nexb.com>

### DIFF
--- a/docs/source/cli-reference/core-options.rst
+++ b/docs/source/cli-reference/core-options.rst
@@ -84,6 +84,20 @@ Comparing Progress Message Options
     This inputs the scan results from ``sample.json``, runs the post-scan plugin ``--classify`` and
     outputs the results for this scan to ``sample_2.json``.
 
+    You can also convert the input .json file to other output format
+    without using any scan option.
+
+    For instance,::
+
+        scancode --from-json sample.json --yaml sample.yml
+
+    The above command takes ``sample.json`` as input and generates file in .yml formats.
+
+    .. note::
+
+        This only works for options that require only the input JSON file;
+        options that rely on other files or source code will not work.
+
 ----
 
 .. _max_in_memory_option:
--- a/docs/source/cli-reference/core-options.rst
+++ b/docs/source/cli-reference/core-options.rst
@@ -84,14 +84,14 @@ Comparing Progress Message Options
     This inputs the scan results from ``sample.json``, runs the post-scan plugin ``--classify`` and
     outputs the results for this scan to ``sample_2.json``.
 
-    You can also convert the input .json file to other output format
+    You can also convert the input ``.json`` file to other output format
     without using any scan option.
 
     For instance,::
 
         scancode --from-json sample.json --yaml sample.yml
 
-    The above command takes ``sample.json`` as input and generates file in .yml formats.
+    The above command takes ``sample.json`` as input and generates file in ``.yml`` formats.
 
     .. note::
 
--- a/docs/source/cli-reference/list-options.rst
+++ b/docs/source/cli-reference/list-options.rst
@@ -3,7 +3,8 @@
 All Available Options
 =====================
 
-This section contains an exhaustive list of all ScanCode options, arranged in various sections.
+This section contains an exhaustive list of all ScanCode-Toolkit options,
+arranged in various sections.
 The sections are as follows:
 
 - Basic Scan Options
@@ -13,22 +14,9 @@ The sections are as follows:
 - Pre-Scan Options
 - Post-Scan Options
 
-There's also another section for ``extractcode`` options.
-
-The order of the sections and all their options is the same as in the :ref:`cli_help_text`,
-available in the command line.
-
 .. _cli_basic:
 
 .. include::  /rst_snippets/basic_options.rst
-
-----
-
-.. include::  /rst_snippets/extract.rst
-
-----
-
-.. include::  /rst_snippets/scancode-reindex-licenses.rst
 
 ----
 
--- a/docs/source/rst_snippets/core_options.rst
+++ b/docs/source/rst_snippets/core_options.rst
@@ -25,3 +25,12 @@ All "Core" Scan Options
                          including and below the starting point. INTEGER
                          must be positive or zero for no limit.
                          [Default: 0]
+
+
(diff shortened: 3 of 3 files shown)


docs/source/rst_snippets/core_options.rst
```diff
@@ -25,3 +25,12 @@ All "Core" Scan Options
                          including and below the starting point. INTEGER
                          must be positive or zero for no limit.
                          [Default: 0]
+
+
+.. note::
+
+    For optimal performance, it's recommended to set the `-n, --processes`
```
AyanSinhaMahapatra: @chinyeungli this was merged as default with https://github.com/aboutcode-org/scancode-toolkit/pull/4104
See also docs above which was changed:
```
[Default: (number of CPUs)-1]
```
This explanation is nice, but we need to modify this a bit.
We could also say it's safer to use `(number of CPUs)-2` for local machines with other applications (browser/IDEs/chatrooms etc) open, or even less after looking at current CPU usage.

We should probably have a seperate section on performance and explanations, but this can be done a bit more explicitly as a sub-task of https://github.com/aboutcode-org/scancode-toolkit/issues/4069 eith detailed explanations and recommendations on CPU/memory

docs/source/cli-reference/list-options.rst
```diff
@@ -13,25 +14,12 @@ The sections are as follows:
...
-The order of the sections and all their options is the same as in the :ref:`cli_help_text`,
-available in the command line.
-
 .. _cli_basic:
 
 .. include::  /rst_snippets/basic_options.rst
 
 ----
 
-.. include::  /rst_snippets/extract.rst
```
AyanSinhaMahapatra, line 27: I'm not sure we want to remove these sections, they are pretty useful IMHO

docs/source/cli-reference/core-options.rst
```diff
@@ -84,6 +84,20 @@ Comparing Progress Message Options
     This inputs the scan results from ``sample.json``, runs the post-scan plugin ``--classify`` and
     outputs the results for this scan to ``sample_2.json``.
 
+    You can also convert the input .json file to other output format
```
AyanSinhaMahapatra:
```suggestion
    You can also convert the input ``.json`` file to other output format
```

docs/source/cli-reference/core-options.rst
```diff
@@ -84,6 +84,20 @@ Comparing Progress Message Options
...
+    without using any scan option.
+
+    For instance,::
+
+        scancode --from-json sample.json --yaml sample.yml
+
+    The above command takes ``sample.json`` as input and generates file in .yml formats.
```
AyanSinhaMahapatra:
```suggestion
    The above command takes ``sample.json`` as input and generates file in ``.yml`` formats.
```

docs/source/cli-reference/core-options.rst
```diff
@@ -84,6 +84,20 @@ Comparing Progress Message Options
...
+
+    .. note::
+
+        This only works for options that require only the input JSON file;
```
AyanSinhaMahapatra, line 98: Generally the post-scan options (like ``--summary``, ``--todo``) are configured to just operate on the codebase (with top-level licenses/packages etc and all resource level results) and this should work fine with `--from-json`
The core scan options (like ``--license`` ``--info``) and pre-scan options cannot be run here. There are some post-scan options which have required options as dependencies, so those also need to be there in the loaded scan if those options are being specified.

AyanSinhaMahapatra (changes requested):
Thanks @chinyeungli looks good.
See my comments for your consideration

docs/source/cli-reference/list-options.rst
chinyeungli: @AyanSinhaMahapatra I'm removing this from the list-options since extract.rst is already included in other-commands.rst. Having it in both places seem redundant.

docs/source/cli-reference/core-options.rst
chinyeungli: @AyanSinhaMahapatra I am wondering is it true that all the "Pre-Scan" options cannot be run with ``--from-json``?
https://scancode-toolkit.readthedocs.io/en/latest/cli-reference/list-options.html#all-pre-scan-options
It seems to be the "--ignore" and "--include" don't really need the code files to have it work (I didn't look at the source code to verify it). Also, https://scancode-toolkit.readthedocs.io/en/latest/cli-reference/core-options.html#from-json-option , the current ``--from-json` is usung ``--classify`` as a sample.

For the post-scans, I want to verify that if an option's dependencies are missing, the command still executes without errors, it just won’t return results related to the missing dependencies, correct?

docs/source/cli-reference/core-options.rst
chinyeungli: Fixed in https://github.com/aboutcode-org/scancode-toolkit/pull/4467/commits/6c46ab4b62d07e76bfed484db2b12cabbe8788a4

PR #4498, opened 2025-07-30
With requirement for fingerprint, we now need to build pyicu.

### LINKED ISSUE #4497: Dependency on pyicu
### Description

With the recent addition to requiring fingerprints, we also now pull in the requirement for a later version of normality which itself requires pyicu.

This means we now need to build that, so we need to add in the required Linux libs as well as update the Python requirements to address these new versions.

### How To Reproduce

docker build on <develop>

### System configuration

* What OS are you running on? docker

jimjag:
Update to-be-installed requirements as needed.

Fixes #4497

### Tasks

* [ x Updated documentation pages (if applicable)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,7 +22,7 @@ extractcode==31.0.0
 extractcode-7z==16.5.210531
 extractcode-libarchive==3.5.1.210531
 fasteners==0.19
-fingerprints==1.2.3
+fingerprints==1.3.0
 ftfy==6.3.1
 future==1.0.0
 gemfileparser2==0.9.4
@@ -40,7 +40,7 @@ license-expression==30.4.4
 lxml==5.4.0
 MarkupSafe==3.0.2
 more-itertools==10.7.0
-normality==2.6.1
+normality==3.0.1
 packageurl-python==0.17.1
 packaging==25.0
 packvers==21.5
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -22,7 +22,7 @@ extractcode==31.0.0
 extractcode-7z==16.5.210531
 extractcode-libarchive==3.5.1.210531
 fasteners==0.19
-fingerprints==1.2.3
+fingerprints==1.3.0
```
AyanSinhaMahapatra, line 25: We cannot upgrade `fingerprints` just yet, because they have removed support for python3.9, see https://github.com/opensanctions/fingerprints/blob/8eea92ff357e080c6c8bb0807352e7b5f83d9c14/pyproject.toml#L19C1-L19C28
This is not EOL yet: https://endoflife.date/python

Also why the tests are failing because we test and support python3.9.

Note also as mentioned in https://github.com/aboutcode-org/scancode-toolkit/issues/4493#issue-3272342439, fingerprints is now unmaintained.
So we would want to consider switching to rigour in the future.

AyanSinhaMahapatra (commented):
Thanks for the report and the PR @jimjag 
See my comments below on why we cannot update fingerprints just yet, as we want to continue to support python3.9

We can conditionally update `fingerprints` based on python versions like we do for click, attrs : https://github.com/aboutcode-org/scancode-toolkit/blob/d171a74333a5c3980906ccd60cc56fdd764bb2c9/setup.cfg#L73
Could you update the PR accordingly so we can check if the tests pass allright?

PR #4500, opened 2025-07-31
Cython seq

JonoYang:
This PR replaces licensedcode/seq.py with  a Cython implementation based off of CyDifflib and the existing seq.py file.


kash831 (approved):

JonoYang:
The cython code that performs the sequence matching has been moved to https://github.com/aboutcode-org/cyseq and is available from pypi as `cyseq`. We now import this library on linux systems and check to see if it available before importing it.

PR #4589, opened 2025-10-20
Adding false positive rules for Go and pycryptodomex  GPL hits

CsatariGergely:
One for Go, where the GPLv2 licenses is mentioned in the context of GCC and one for pycryptodomex where the GPL license is mentioned in the context of a website.

### Tasks

* [X] Reviewed [contribution guidelines](https://github.com/nexB/scancode-toolkit/blob/develop/CONTRIBUTING.rst)
* [X] PR is descriptively titled 📑 and links the original issue above 🔗
* [X] Commits are in uniquely-named feature branch and has no merge conflicts 📁
* [X] Updated documentation pages (if applicable)
* [X] Updated CHANGELOG.rst (if applicable)


AyanSinhaMahapatra (commented):
@CsatariGergely Thanks a lot for the PR! These reports and license rules to fix detection issues are always much appreciated :heart: 
What about having these rules as license clues instead of false-positives?

IMHO these license texts although they are not directly related to the license-expression for the file it is found in, they could still be important license related clues which could be interesting.

A false positive rule deletes a license text match from the results without any indication, a license clue if detected, is not reflected in the final license-expression for the file, but is present in a seperate file-level section `license_clues` and thus can be reviewed if someone wants to review license related clues.

The license-clue vs false-positive distinction should happen thus in a case by case basis, and IMHO proper false-positives are always texts which are not license related at all, but vaguely resembles license names/texts in some way, and license clues are always somewhat license related, even if they are not directly related to the license of a file.

We should probably better document this somewhere (probably in the documentation section `how to add a license rule`) :sweat_smile: 

What do you think? @CsatariGergely @pombredanne

CsatariGergely:
In my opinion these are clearly false positives and not license clues. Both of these are talking of the licenses of other things than the code under scan. Adding them to license clues would be confusing.

PR #4608, opened 2025-11-18
Parse `.npmrc` files 

### LINKED ISSUE #4494: Parse npmrc
See https://docs.npmjs.com/cli/v11/configuring-npm/npmrc

levi42x:
- Added a new `NpmrcHandler` to enable parsing of `.npmrc` files.  
- This handler extracts key/value pairs (such as registry settings, authentication tokens, and `init.*` defaults) and stores them in `extra_data` on `PackageData`, making npm configuration discoverable through ScanCode outputs.

This PR includes the following updates:
- Added `NpmrcHandler` in `npm.py`
- Added corresponding unit tests in `test_npmrc.py`
- Added `.npmrc` test fixture and its expected result file
- Verified that the basic parsing functionality works as intended

Fixes #4494

### Tasks
- [x] Reviewed contribution guidelines  
- [x] PR is descriptively titled 📑 and links the original issue above 🔗  
- [x] Tests pass locally ✔️  
- [x] Commits are in a uniquely-named feature branch and contain no merge conflicts 📁  
- [ ] Updated documentation pages (if applicable)  
- [ ] Updated `CHANGELOG.rst` (can be added upon request)

> Any feedback, suggestions, or requested changes are very welcome!

Signed-off-by: Shekhar Suman <levi42x@gmail.com>


tests/packagedcode/test_npmrc.py
```diff
@@ -0,0 +1,15 @@
+import os
```
AyanSinhaMahapatra, line 1: No need to create a new file, add the tests in the test_npm.py file

tests/packagedcode/test_npmrc.py
```diff
@@ -0,0 +1,15 @@
...
+
+from packagedcode import npm
+from packages_test_utils import PackageTester
+from scancode_config import REGEN_TEST_FIXTURES
+
+
+class TestNpmrc(PackageTester):
+    test_data_dir = os.path.join(os.path.dirname(__file__), 'data')
+
+    def test_parse_basic_npmrc(self):
```
AyanSinhaMahapatra, line 11: also add a test to check `npm.NpmrcHandler.is_datafile()` works as expected

tests/packagedcode/data/npm/basic/.npmrc
```diff
@@ -0,0 +1,8 @@
+; sample .npmrc for tests
+# a comment line
```
AyanSinhaMahapatra, line 2: Please use real .npmrc files seen in the wild like with: https://github.com/search?q=path%3A*.npmrc&type=code 

This also helps you looks for what kind of data is present and whether we can use these differently rather than just storing in the `extra_data`

src/packagedcode/npm.py
```diff
@@ -500,6 +500,48 @@ def update_workspace_members(cls, workspace_members, codebase):
...
+            for line in lines:
+                line = line.strip()
+                if not line or line.startswith(';') or line.startswith('#'):
+                    continue
+                if '=' not in line:
+                    continue
+                key, value = line.split('=', 1)
+                key = key.strip()
+                value = value.strip()
+                # ignore empty key but allow empty values
```
AyanSinhaMahapatra, line 528: are there cases with empty values that is still useful to keep? can you provide examples?

src/packagedcode/npm.py
```diff
@@ -500,6 +500,48 @@ def update_workspace_members(cls, workspace_members, codebase):
...
+    description = 'npm .npmrc configuration file'
+    documentation_url = 'https://docs.npmjs.com/cli/v11/configuring-npm/npmrc'
+
+    @classmethod
+    def parse(cls, location, package_only=False):
+        """
+        parse [.npmrc] file and store result in key : value pair.
+        convert key : value pair to object and return it.
+        """
+        extra_data = {}
```
AyanSinhaMahapatra, line 517: There are .npmrc examples with license/author info which we need to parse and store properly, this is not just extra data.

Please also research examples found in the wild/docs to see what other fields we can use like this to map useful data to package data fields.

src/packagedcode/npm.py
```diff
@@ -500,6 +500,48 @@ def update_workspace_members(cls, workspace_members, codebase):
         for member in workspace_members:
             member.save(codebase)
 
+class NpmrcHandler(BaseNpmHandler):
+    datasource_id = 'npmrc'
+    path_patterns = ('*/.npmrc',)
+    default_package_type = 'npm'
+    default_primary_language = None
+    description = 'npm .npmrc configuration file'
+    documentation_url = 'https://docs.npmjs.com/cli/v11/configuring-npm/npmrc'
```
AyanSinhaMahapatra, line 509:
```suggestion
    documentation_url = 'https://docs.npmjs.com/cli/configuring-npm/npmrc'
```
this would pin the link to this version and will soon be outdated. I see some other npm doc URLs also have versions, can you also remove those in the PR?

src/packagedcode/npm.py
```diff
@@ -500,6 +500,48 @@ def update_workspace_members(cls, workspace_members, codebase):
         for member in workspace_members:
             member.save(codebase)
 
+class NpmrcHandler(BaseNpmHandler):
```
AyanSinhaMahapatra, line 503: This should be part of the `npm` assembly only if the data present in .npmrc files can be used to updated package data information in other npm manifests. The registry information present could be used to update npm urls in npm packages found alongside these manifests but these URLs are not always usable. And you'd have to modify the `BaseNpmHandler.assemble` function to handle/use these too.

So let's keep this as a subclass of `models.NonAssemblableDatafileHandler` to avoid any assembly from these files and keep things simple for now.

AyanSinhaMahapatra (changes requested):
Thanks @levi42x see comments for your consideration. A bit of research into fields could be nice to see if we can use anything to populate the package-data fields, see https://docs.npmjs.com/cli/v11/using-npm/config#config-settings for all the fields and check out real world usages for these config values.

src/packagedcode/npm.py
levi42x: thanks for pointing it out.
i did not even think about it at that time..
i'm not able to find any empty value (in key value pair) that could be useful.

levi42x: yet i have not seen any empty values case, should i include it to handler or just skip it (example: registry= ) ?

src/packagedcode/npm.py
levi42x: i'll update the handler to map meaningful fields (license, author, homepage, etc.) to ScanCode’s structured package fields.

and other config-only fields (like proxy, cafile, always-auth) in extra_data, since they don’t map to package metadata.

PR #4616, opened 2025-11-26
feat: Add UV package manager support for Python projects

### LINKED ISSUE #4501: Support parsing python UV manifests
We need to support parse python UV based manifests for package and depednencies:

* pyproject.toml with UV specific tags and attributes
* uv.lock lockfile with dependencies and packages data

https://github.com/python-attrs/attrs/blob/main/uv.lock can be used as a test.

https://docs.astral.sh/uv/getting-started/installation/

levi42x:
Fixes #4501

##  Description

This PR adds support for parsing Python UV package manager projects in `scancode-toolkit`.
It enables detection, parsing, and full package assembly for UV-based projects using
`pyproject.toml` and `uv.lock`.

## Changes

- **UvPyprojectTomlHandler**: Parses `pyproject.toml` files containing [tool.uv] sections
- **UvLockHandler**: Parses `uv.lock` files and extracts resolved dependencies
- **BaseUvPythonLayout**: Coordinates UV project assembly and resource handling
- Dependency parsing: Reuses existing PyPI dependency parsing logic to avoid duplication
- Tests: Adds an _end-to-end_ package assembly test covering a complete UV project layout

##  Tasks

---

**Signed-off-by:** Shekhar Suman <levi42x@gmail.com>

### DIFF
--- a/src/packagedcode/pypi.py
+++ b/src/packagedcode/pypi.py
@@ -529,6 +529,13 @@ def is_poetry_pyproject_toml(location):
         return False
 
 
+def is_uv_pyproject_toml(location):
+    with open(location, 'r') as file:
+        if "tool.uv" in file.read():
+           return True
+        return False
+
+
 class BasePoetryPythonLayout(BaseExtractedPythonLayout):
     """
     Base class for poetry python projects.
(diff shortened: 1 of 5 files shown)


src/packagedcode/pypi.py
```diff
@@ -832,6 +842,310 @@ def parse(cls, location, package_only=False):
...
+                package_resource = [r for r in siblings if r.name == 'pyproject.toml']
+                if package_resource:
+                    package_resource = package_resource[0]
+
+        if not package_resource:
+            # we do not have a pyproject.toml
+            yield from yield_dependencies_from_package_resource(resource)
+            return
+
+        if codebase.has_single_resource:
```
AyanSinhaMahapatra: This should be at the beginning of the function

tests/packagedcode/test_pypi.py
```diff
@@ -405,6 +405,30 @@ def test_parse_pyproject_toml_poetry_univers(self):
...
+    def test_is_uv_lock(self):
+        test_file = self.get_test_loc('pypi/uv/attrs-uv.lock')
+        assert pypi.UvLockHandler.is_datafile(test_file)
+
+    def test_parse_uv_lock_attrs(self):
+        test_file = self.get_test_loc('pypi/uv/attrs-uv.lock')
+        package = pypi.UvLockHandler.parse(test_file)
+        expected_loc = self.get_test_loc('pypi/uv/attrs-uv.lock-expected.json')
+        self.check_packages_data(package, expected_loc, regen=REGEN_TEST_FIXTURES)
+
```
AyanSinhaMahapatra: Can you also add an end-to-end example with a `--package` scan (see the test file for examples, add a test similarly) in a uv package manifest layout and show that the package assembly works as expected. You've added an `assemble` function, but this is not tested otherwise.

tests/packagedcode/data/pypi/uv/pyproject-with-uv.toml
```diff
@@ -0,0 +1,23 @@
+[project]
+name = "attrs-example"
```
AyanSinhaMahapatra: Please use real world examples, and include the link

AyanSinhaMahapatra: If you include a file that is large, please shorten it to only keep the relevant parts which you need to test the parser for, as we want to keep filesize and repo size as minimal as possible

AyanSinhaMahapatra: also find a file which has all/almost all the values which could be present, for example this fake example doesn't have licenses, copyrights and other usable package data fields.

src/packagedcode/pypi.py
```diff
@@ -832,6 +842,310 @@ def parse(cls, location, package_only=False):
...
+                is_optional=False,
+                is_direct=False,
+                is_pinned=True,
+                resolved_package=resolved_package.to_dict()
+            )
+            dependencies.append(dependency.to_dict())
+
+        extra_data = {}
+        extra_data['python_version'] = requires_python
+        extra_data['lock_version'] = version
```
AyanSinhaMahapatra:
```suggestion
        extra_data['lock_version'] = lock_version
```
this would be confusing with the proper package version, let's be as descriptive as possible in variable names to improve readability

src/packagedcode/pypi.py
```diff
@@ -832,6 +842,310 @@ def parse(cls, location, package_only=False):
...
+
+        package_data = dict(
+            datasource_id=cls.datasource_id,
```
AyanSinhaMahapatra, line 1106: there are a lot more package data fields which can be parsed and added, please modify to include those. See https://github.com/aboutcode-org/scancode-toolkit/blob/c05a13d87b41cfc629ad4497cf40bcb9c26ee703/src/packagedcode/models.py#L548 for more details for all the fields and refer to other package data parsers too

src/packagedcode/pypi.py
AyanSinhaMahapatra: this is also likely wrong as you're using the same variable for version

src/packagedcode/pypi.py
```diff
@@ -832,6 +842,310 @@ def parse(cls, location, package_only=False):
...
+            type=cls.default_package_type,
+            primary_language='Python',
+            extra_data=extra_data,
+            dependencies=dependencies,
```
AyanSinhaMahapatra, line 1110: we also need to populate URLs with packageURL fields, see https://github.com/aboutcode-org/scancode-toolkit/blob/c05a13d87b41cfc629ad4497cf40bcb9c26ee703/src/packagedcode/pypi.py#L794 for example

src/packagedcode/pypi.py
```diff
@@ -529,6 +529,16 @@ def is_poetry_pyproject_toml(location):
         return False
 
 
+def is_uv_pyproject_toml(location):
+    with open(location, 'r') as file:
+        data = file.read()
```
AyanSinhaMahapatra:
```suggestion
        if "tool.uv" in file.read():
           return True
        
        return False
```

src/packagedcode/pypi.py
```diff
@@ -832,6 +842,310 @@ def parse(cls, location, package_only=False):
         yield models.PackageData.from_data(package_data, package_only)
 
 
+def parse_dependency_requirement(requirement, scope='dependencies', is_runtime=True):
```
AyanSinhaMahapatra, line 842: Could this be replaced by using `get_requires_dependencies`? The processing looks similar.

Or maybe abstract out similar code into functions that both these functions use? As we atleast have some code duplication here.

AyanSinhaMahapatra (changes requested):
Thanks++ @levi42x for the PR, much needed update! ❤️ 

This is a great start and a step in the right direction, but needs some more updates and love. See my comments for your consideration.

tests/packagedcode/test_pypi.py
levi42x: I’ve removed unit tests and applied just one `package scan`.  The file to scan is `tests/packagedcode/test_pypi.py` and the lines to scan are `89` to `95`.

tests/packagedcode/data/pypi/uv/pyproject-with-uv.toml
levi42x: I checked out a few `uv package manager` projects (like :  [a5Chin](https://github.com/a5chin/python-uv/tree/main), [gdamjan](https://github.com/gdamjan/uv-getting-started), [prrao87](https://github.com/prrao87/uv-demo) ) but ended up following the example you linked in #4501 ([python-attrs](https://github.com/python-attrs/attrs/tree/main)).
since it’s more relevant and covers more fields.
I’ve included all the important fields in both`tests/packagedcode/data/pypi/uv/attrs/pyproject.toml` and `tests/packagedcode/data/pypi/uv/attrs/uv.lock`
If there are any other fields you think should be added, please let me know!

src/packagedcode/pypi.py
levi42x: thanks for the suggestion!
i’ve refactored the code so that `parse_dependency_requirement` now shares logic with `get_requires_dependencies`, and any common parts are abstracted out to avoid duplication.
let me know if you see any further improvements!

AyanSinhaMahapatra:
Fixed with https://github.com/aboutcode-org/scancode-toolkit/pull/4980 instead, thanks @levi42x

PR #4641, opened 2025-12-16
Fix: Use WITH operator for license exceptions in SPDX expressions (Issue #4623)

asmitbanik:
Summary
Fixes invalid SPDX license expressions by using the WITH operator instead of AND when combining base licenses with license exceptions.

Problem
ScanCode was producing invalid SPDX expressions when detecting licenses with exceptions:
❌ Before: "GPL-3.0-only AND GCC-exception-3.1"
✅ After:  "GPL-3.0-only WITH GCC-exception-3.1"
According to the SPDX specification, exceptions must always appear on the right side of a WITH expression. Using AND to combine a license with an exception produces an invalid SPDX expression.

Solution
Added a new function combine_expressions_with_exceptions() in detection.py that:
Checks if a license key has is_exception=True in the license database
Combines exceptions with their preceding base license using WITH instead of AND
Falls back to AND for regular license combinations

Changes
Modified: src/licensedcode/detection.py
Added combine_expressions_with_exceptions() function
Updated get_detected_license_expression() to use the new function

Testing
Test: ['gpl-3.0', 'gcc-exception-3.1']Result: gpl-3.0 WITH gcc-exception-3.1 ✓Test: ['gpl-3.0', 'gcc-exception-3.1', 'mit']Result: gpl-3.0 WITH gcc-exception-3.1 AND mit ✓Test: ['gpl-2.0', 'classpath-exception-2.0']Result: gpl-2.0 WITH classpath-exception-2.0 ✓

Impact
This fix applies to all 281 license exceptions in ScanCode's database, ensuring valid SPDX expressions for any license+exception combination.
Related Issues
Fixes #[issue_number] - Invalid SPDX expressions with license exceptions

Checklist:
[✓] Code follows project conventions
[✓] Self-reviewed the code changes
[✓] Tested with multiple exception types (GCC, Classpath, etc.)
[✓] No breaking changes to public API

### DIFF
--- a/src/licensedcode/detection.py
+++ b/src/licensedcode/detection.py
@@ -47,6 +47,141 @@ A LicenseDetection combines one or more matches together using various rules and
 heuristics.
 """
 
+
+def combine_expressions_with_exceptions(expressions, licensing=None):
+    """
+    Combine a list of license ``expressions`` into a single expression string,
+    using WITH for license exceptions and AND for regular licenses.
+
+    According to SPDX specification, exceptions must be combined with their
+    base license using the WITH operator, not AND. For example:
+    - "gpl-3.0 WITH gcc-exception-3.1" is valid SPDX
+    - "gpl-3.0 AND gcc-exception-3.1" is NOT valid SPDX
+
+    This function detects when an expression contains only an exception license
+    and combines it with the preceding license using WITH instead of AND.
+    """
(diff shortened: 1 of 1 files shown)


pombredanne:
Thanks but this is not the way to fix. A simple rule would do.

PR #4663, opened 2026-01-04
docs: document Apple Silicon contributor setup in CONTRIBUTING.rst

### LINKED ISSUE #3205: Many scancode dependencies are broken for Mac M1
### Description
This is not necessarily a bug with Scancode, but a handful of the required dependencies are broken and cannot install on Mac M1 platform. This is a huge blocker for anyone using this platform as it is impossible to install scancode using pip. Perhaps the scancode docs should be updated to note that scancode is not usable on this platform?

### How To Reproduce
Try to install scancode on Mac M1 computer. In my case, I am running Ubunutu VM using fusion but the same issues arise when I try to install scancode directly on my mac shell (not in a fusion VM).
```
(ternenv) rose@rose-vm:~/ternenv/tern$ pip install --upgrade pip setuptools wheel
(ternenv) rose@rose-vm:~/ternenv/tern$ pip install scancode-toolkit
```

Specific libraries with issues installing:
bitarray intbitset lxml pyahocorasick pycryptodome cffi

### System configuration

* What OS are you running on?
MacOS Monteray 12.2.1 (M1 chip)

* What version of scancode-toolkit was used to generate the scan file?
Attempting to install scancode_toolkit-31.2.4-cp310-none-any.whl

* What installation method was used to install/run scancode? (pip/source download/other)
pip

thedrishtigupta:
Fixes #3205

### Problem
New contributors on Apple Silicon (M1/M2/M3/M4) following CONTRIBUTING.rst
encounter dependency installation failures when using ARM-native Python.
The contributor guide currently provides no warning or workaround for this
platform, despite the setup failing for Apple Silicon users.

### Root cause
Several development dependencies do not provide ARM64 wheels. When using
ARM-native Python (the default on Apple Silicon), pip attempts to build
these dependencies from source and fails.

### Change in this PR
Adds a warning to CONTRIBUTING.rst immediately before the `./configure --dev`
step explaining:
- why the setup fails on Apple Silicon
- how to use an x86_64 Python installation under Rosetta
- how to configure ScanCode to use that Python explicitly

This aligns the contributor documentation with existing installation docs
and prevents new contributors from hitting a blocking error.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -132,6 +132,29 @@ To set up ScanCode for local development:
    libraries (from the `thirdparty/ directory`), setup the paths, etc.
    See https://virtualenv.pypa.io/en/latest/ for more details.
 
+   .. warning::
+
+   **Apple Silicon (M1/M2/M3/M4) users**
+
+   Due to missing ARM64 wheels for some native dependencies, the contributor
+   setup may fail when using ARM-native Python.
+
+   Use an x86_64 Python installation under Rosetta and configure ScanCode to
+   use it explicitly:
+
+   .. code-block:: bash
+
+      # Install x86_64 Python via Rosetta
+      arch -x86_64 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+      arch -x86_64 /usr/local/bin/brew install python@3.12
+
+      # Tell configure to use it
+      echo "/usr/local/bin/python3.12" > PYTHON_EXECUTABLE
+      ./configure --dev
+
+   See issue #3205 for more context on Apple Silicon compatibility.
+
+
    Run this command to configure ScanCode::
 
         ./configure --dev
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -132,6 +132,29 @@ To set up ScanCode for local development:
...
+   Due to missing ARM64 wheels for some native dependencies, the contributor
+   setup may fail when using ARM-native Python.
+
+   Use an x86_64 Python installation under Rosetta and configure ScanCode to
+   use it explicitly:
+
+   .. code-block:: bash
+
+      # Install x86_64 Python via Rosetta
+      arch -x86_64 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
```
AyanSinhaMahapatra: IMHO we should not include installation instructions for other packages directly (as it might change and our docs will become outdated, might have other requirements/prerequisites). Instead use a link to their installation docs

AyanSinhaMahapatra (changes requested):
Thanks @thedrishtigupta see comments for your consideration.

CONTRIBUTING.rst
thedrishtigupta: Thanks for the feedback, that makes sense.

I can update the warning to avoid embedding Homebrew/Python installation
commands directly, and instead link to the official documentation while
still explaining why Apple Silicon contributors need to use x86_64 Python
under Rosetta.

I’ll push an updated version shortly.

PR #4683, opened 2026-01-15
Delete gpl-2.0_and-unknown-license-reference_1 rule

MarkusObendrauf:
Delete the rule "gpl-2.0_and-unknown-license-reference_1 rule"

From the rule's text::
"It contains only constants, structures, and macros generated from the original header, and thus, contains no copyrightable information."

The text explicitly says that it does not contain copyrightable information, and therefore, we think that the classification as GPL-2.0 is incorrect, and that somebody using this code is not bound to the terms of GPL-2.

Example of a file containing this header:
https://android.googlesource.com/platform/frameworks/av/+/refs/heads/main/media/mtp/f_mtp.h

### Tasks


pombredanne (changes requested):
Thanks, but we cannot merge this as-is. First, we do not delete rules, we mark them as deprecated. 

And second, Google opinion that stripped libc and kernel headers are not copyrightable is just that: an opinion. 

We could make this rule a clue, but I would not deprecate this.

PR #4687, opened 2026-01-16
Add PyPI pyproject.toml test for license, URLs and dependencies

Shreyas12324:
This PR adds a new test case for the PyPI pyproject.toml parser.

The test covers:
- project.license using inline text format
- project.urls (Homepage and Repository)
- project.dependencies with version specifiers

Only test data and test assertions are added.
All relevant tests pass locally.

### Tasks

* [x] PR is descriptively titled
* [x] Tests pass locally
* [x] Commits are in uniquely-named feature branch


AyanSinhaMahapatra:
@Shreyas12324 why was this extra test case required? Does the test cases we already have not cover any data attribute present in this test case? Is this even a real example present in the wild?

Closing as this does not add anything to the test suite more than mere volume and imaginary examples IMHO

PR #4693, opened 2026-01-19
Sync output format docs

ShraddhaSharma3:
The "Configuring scan output formats" tutorial provides only a high-level
overview and did not clearly point readers to the complete list of available
ScanCode output options.

This PR improves the documentation by adding a clear note that references
the authoritative "All Scan Output Options" page, keeping the tutorial
concise while ensuring users can easily find the full and up-to-date
configuration details.


AyanSinhaMahapatra:
There is a reference to the CLI output formats right in the beginning of the page, not sure why you missed this.
Screenshot from 2026-01-22 18-41-55
See https://github.com/aboutcode-org/scancode-toolkit/blob/develop/docs/source/tutorials/configuring-scan-output-formats.rst?plain=1#L8

PR #4473, opened 2025-07-14
Correct license merge tutorials

chinyeungli:
Fixes
#2168 - Update license in documentation
#1980 - Merge tutorials and how-to guides

### Tasks

* [x] Reviewed [contribution guidelines](https://github.com/nexB/scancode-toolkit/blob/develop/CONTRIBUTING.rst)
* [x] PR is descriptively titled 📑 and links the original issue above 🔗
* [ ] Tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR
  Run [tests](https://scancode-toolkit.readthedocs.io/en/latest/contribute/contrib_dev.html#running-tests) locally to check for errors. 
* [x] Commits are in uniquely-named feature branch and has no merge conflicts 📁
* [ ] Updated documentation pages (if applicable)
* [ ] Updated CHANGELOG.rst (if applicable)

Failing tests don't related to this PR.

Signed-off-by: Chin Yeung Li <tli@nexb.com>

PR #4713, opened 2026-01-31
Clarify Python library usage and document current ScanCode limitations

### LINKED ISSUE #3130: No documentation on calling ScanCode as a library
### Description

The ScanCode documentation mentions that it is possible to use it as a Python library, which is great when you want to use it from a Python script. It includes instructions to install it as a library: https://scancode-toolkit.readthedocs.io/en/stable/getting-started/install.html#installation-as-a-library-via-pip However, after that, there is zero documentation available on how to _use_ it as a library, making that choice not very practical.

Just a reference section on the Python API of ScanCode would already be a big improvement over digging through the source. Or, if you folks don't want to support that way of using it, then removing the section of the doc that lists that as a possibility would be a good idea. Thanks!

### Select Category

-  [ ] Inconsistency
-  [ ] New Section Request
-  [x] General Improvement
-  [ ] Typo/Mistakes
-  [ ] Other

vidhiii1711:
Fixes #3130 

### Tasks

* Removed references to non-existent Python scan APIs from library_usage.rst
* Documented the current limitation of ScanCode’s Python usage
* Added a correct, working example showing how to run ScanCode from Python using subprocess
* Included a minimal import example to demonstrate supported module imports
* Updated tests to validate importability without assuming unsupported APIs

Testing

-> Verified ScanCode modules can be imported successfully
-> Relevant tests were run locally
-> Full test suite was not executed due to long runtime
Signed-off-by: Vidhi Singh [vidhiii0909@gmail.com](mailto:vidhiii0909@gmail.com)

### DIFF
--- a/docs/scancode_api_example.py
+++ b/docs/scancode_api_example.py
@@ -1,0 +1,9 @@
+import scancode
+import scancode.api
+
+print("ScanCode modules imported successfully")
+
+
+# ScanCode does not currently expose a stable public Python API
+# This example demonstrates importability only
+
--- a/docs/source/how-to-guides/library_usage.rst
+++ b/docs/source/how-to-guides/library_usage.rst
@@ -1,0 +1,66 @@
+Using ScanCode from Python
+=========================
+
+ScanCode Toolkit is primarily designed to be used as a command-line tool.
+At present, it does **not** expose a stable, documented public Python API
+for invoking scans programmatically.
+
+However, ScanCode can still be integrated into Python-based workflows by
+invoking its command-line interface from Python code. ScanCode modules
+can also be imported, though execution-related APIs are considered internal
+and subject to change.
+
+Installation
+------------
+
+ScanCode Toolkit must be installed with all optional dependencies:
+
+.. code-block:: bash
+
+    pip install scancode-toolkit[full]
+
+
+Using ScanCode via subprocess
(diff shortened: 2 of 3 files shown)


stefan6419846:
Coming from the perspective of an actual API user, the following aspects are not clear to me:

* Why do I need a full installation of SCTK to use the Python-based API?
* If you show that `scancode.api` can be imported in an example, why not showing actual examples of using it?
* I have been using `scancode.api` and other specific modules for years without larger issues. Should we really consider it non-stable? (There are some new attributes from time to time, but this generally does not hurt.)

In general, it seems like the docs proposed for this issue in #4473 seem more reasonable.

vidhiii1711:
@stefan6419846   Thanks for the detailed feedback — that’s very helpful.

I took a conservative approach while documenting this because I couldn’t find an officially documented, supported scan API and wanted to avoid presenting an example that might be misleading for new users.

That said, your point about long-term practical usage of scancode.api makes sense. I’m happy to revise the documentation to include a concrete, real-world example based on commonly used APIs if that’s preferred.

I’ll also revisit the installation section and clarify why a full install is currently recommended.

If you have a specific example or pattern you’d recommend (similar to what’s discussed in #4473), I’d be glad to align the docs with that approach.

PR #4733, opened 2026-02-06
Fix: Weigh programming language tallies by file size (bytes) instead … Fixes #3868

### LINKED ISSUE #3868: SCTK returns "C" rather than "C++" for the Primary Language of whisper.cpp
A recent scan of whisper.cpp-1.6.2.tar.gz from https://github.com/ggerganov/whisper.cpp/archive/refs/tags/v1.6.2.tar.gz returned a Primary Language of "C".  (Scan results attached.)  Given the name of the project, whisper.cpp, that seems very odd to me.  Also please note that GitHub https://github.com/ggerganov/whisper.cpp/tree/v1.6.2  shows the following Language breakdown: 
C++ 74.1%
C 16.2%
Cuda 4.4%
Metal 1.6%
Objective-C 1.5%
CMake 0.6%
Other 1.6%

This example alone is probably not a serious bug, but may be worth investigating. 

[whisper.cpp-1.6.2.tar.gz_scan.zip](https://github.com/user-attachments/files/16354900/whisper.cpp-1.6.2.tar.gz_scan.zip)

RajX-dev:
Fixed an issue where programming languages were tallied by file count instead of file size. This caused projects with many small files (like C headers) to be misidentified as C instead of C++. Changed logic to sum file sizes (bytes) for the tally.


RajX-dev:
Ready for review! @DennisClark @AyanSinhaMahapatra  All checks are passing. 🟢

I have finalized the changes and verified the build. Here is a summary of the fixes included in this PR:

1-Feature Implementation: Switched programming language tallies to use byte counts instead of file counts (Fixes #3868).
2-Bug Fix (Summarizer): Fixed a KeyError crash in get_declared_holders that occurred when a holder existed in key files but was filtered out of the main tallies (0 byte count).
3-Bug Fix (CI/Threading): Patched src/scancode/interrupt.py to handle ValueError when setting signals. This resolves the crash on Python 3.14 (CI environment) where signals were failing because they weren't in the main thread.
4- Test Data: Regenerated the expected JSON and YAML reference files for summarycode, formattedcode, and todo tests to match the new byte-count sorting logic.

RajX-dev:
@AyanSinhaMahapatra should i solve these conflicts or move on to different problem??

PR #4741, opened 2026-02-12
Fix CAL-1.0 false positives on generic GPL headers

Aa-Rho-Hi:
This PR reduces CAL-1.0 false positives caused by generic
“Licensed under the …” fragments that commonly appear in GPL headers.

Changes:
- Increase CAL detection strictness (minimum_coverage, minimum_match_length)
- Anchor detection on stronger CAL identifiers (SPDX / full license name)
- Add regression test using a simple GPL header to prevent cross-matching


pombredanne:
Closing as incorrect. See:
- https://github.com/aboutcode-org/scancode-toolkit/issues/4740#issuecomment-4412012100

PR #4803, opened 2026-03-09
packagedcode: add DotNetDepsJsonHandler to parse .deps.json files

### LINKED ISSUE #4496: Parse and get dependencies from `.deps.json`
nuget `.deps.json` present sometimes beside .dll .pdb files (with the same name)

For example in  https://github.com/snoopwpf/snoopwpf/releases/download/v5.1.0/Snoop.5.1.0.nupkg:

```
to/tools/netcoreapp3.1/Snoop.Core.deps.json
to/tools/netcoreapp3.1/Snoop.Core.dll 
to/tools/netcoreapp3.1/Snoop.Core.pdb
```

[Snoop.Core.deps.json](https://github.com/user-attachments/files/21510151/Snoop.Core.deps.json)

kumarasantosh:
Fixes #4496

NuGet .deps.json files appear alongside .dll and .pdb files in .NET 
projects but were not being parsed by ScanCode. This PR adds a new 
DotNetDepsJsonHandler in nuget.py to handle these files.

The handler detects *.deps.json files, reads the libraries section 
and extracts package names, versions, types and their dependencies. 
Target framework is stored in extra_data.

Tested against a real world fixture from snoopwpf v5.1.0 release 
and a minimal hand-crafted fixture for edge cases.

### Tasks
* [x] Tests pass


tests/packagedcode/data/nuget/deps_json/simple.deps.json
```diff
@@ -0,0 +1,23 @@
+{
```
AyanSinhaMahapatra, line 1: Please add the source link/download URL for this file as a comment, we need to know where the test data is located/comes from.

src/packagedcode/nuget.py
```diff
@@ -265,3 +265,121 @@ def parse(cls, location, package_only=False):
...
+                if target_name == target_framework
+            ]
+            runtime_target_matched = bool(selected_targets)
+        else:
+            selected_targets = []
+
+        if not selected_targets:
+            selected_targets = available_targets
+
+        for lib_key, lib_info in libraries.items():
```
AyanSinhaMahapatra: Can you extract the parsing of dependencies in another function.

src/packagedcode/nuget.py
```diff
@@ -265,3 +265,121 @@ def parse(cls, location, package_only=False):
...
+    default_package_type = 'nuget'
+    description = 'NuGet .deps.json lockfile'
+    documentation_url = 'https://github.com/dotnet/sdk/blob/main/documentation/specs/runtime-configuration-file.md'
+
+    @classmethod
+    def parse(cls, location, package_only=False):
+        with open(location) as loc:
+            try:
+                parsed = json.load(loc)
+            except Exception:
```
AyanSinhaMahapatra: Not required, exception handling is handled in upstream code.

tests/packagedcode/data/nuget/deps_json/Snoop.Core.deps.json
```diff
@@ -0,0 +1,942 @@
...
+    },
+    "WpfAnalyzers/4.1.1": {
+      "type": "package",
+      "serviceable": true,
+      "sha512": "sha512-4+oLAk4gB4Wqn0ozKstB6fY9BafSU6c6GEgWjgOThyzHFBVqxihkS/1OmvEaYJnhjNUA/GOcZ6m2B7Xr2EePNw==",
+      "path": "wpfanalyzers/4.1.1",
+      "hashPath": "wpfanalyzers.4.1.1.nupkg.sha512"
+    }
+  }
+}
```
AyanSinhaMahapatra: We need the full scan results to verify everything works correctly.

Also this file is quite large, can you keep only a subset of dependencies in here.

Be careful as there are a few parallal lists, `libraries` and `targets` which you need to merge, so make sure you test merging data and consistency

AyanSinhaMahapatra: Finding an additional smaller test case and adding that helps too

src/packagedcode/nuget.py
```diff
@@ -265,3 +265,121 @@ def parse(cls, location, package_only=False):
...
+                    dependencies.append(
+                        models.DependentPackage(
+                            purl=str(PackageURL(type='nuget', name=dep_name, version=dep_version)),
+                            extracted_requirement=dep_version,
+                            scope=scope,
+                            is_runtime=True,
+                            is_optional=False,
+                            is_pinned=True,
+                            is_direct=True,
+                        ).to_dict()
```
AyanSinhaMahapatra: `resolved_package` needs to be populated as these are runtime packages which are resolved to versions.

AyanSinhaMahapatra (changes requested):
@kumarasantosh thank you for the PR.
See comments for feedback, some major changes are required, and please apply the same to your other PRs :)

kumarasantosh:
@AyanSinhaMahapatra , thanks for the detailed review - really helpful feedback.

I've gone through all the comments and here's what's been updated:

Removed the local try/except around json.load since you mentioned upstream already handles parse errors. Pulled the dependency parsing out into its own helper function to clean up the main method. Populated resolved_package on the runtime dependencies.

On the test side, moved everything into the existing test_nuget.py rather than a separate file. Trimmed the Snoop fixture down to a smaller consistent subset, regenerated the expected scan output, and added a second smaller real-world fixture as well. Added .ABOUT files to document where each fixture came from — simple.deps.json is hand-crafted and is noted as such. Also added tests for the target fallback behaviour and multi-target merge consistency.

Please take another look when you get a chance — happy to make further changes!

PR #4804, opened 2026-03-10
packagedcode: add CondaEnvironmentYmlHandler to parse environment.yml

### LINKED ISSUE #3694: Add support for Conda environment.yml
environment.yml files contains some interesting data https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#creating-an-environment-from-an-environment-yml-file

Example: https://github.com/Quantco/multiregex/blob/main/environment.yml

kumarasantosh:
Fixes #3694

Conda environment.yml files were not being parsed by ScanCode at all. 
This PR adds CondaEnvironmentYmlHandler in conda.py to fix that.

The handler picks up *environment.yml files and pulls out the 
environment name, channels and conda dependencies with their versions. 
It also handles the pip section inside dependencies and marks those 
packages correctly as pypi type. Channels go into extra_data.

Tested with a real fixture from the multiregex project and a simple 
hand-crafted one to cover edge cases like missing name or empty 
dependencies.

### Tasks
* [x] Tests pass

### DIFF
--- a/src/packagedcode/conda.py
+++ b/src/packagedcode/conda.py
@@ -666,3 +666,79 @@ def get_variables(location):
                 parts = line.split('=')
                 result[parts[0].strip()] = parts[-1].strip().strip('"')
     return result
+
+
+class CondaEnvironmentYmlHandler(BaseDependencyFileHandler):
+    datasource_id = 'conda_environment_yml'
+    path_patterns = ('*environment.yml', '*environment.yaml')
+    default_package_type = 'conda'
+    description = 'Conda environment.yml'
+
+    @classmethod
+    def parse(cls, location, package_only=False):
+        import re
+        with open(location) as fi:
+            conda_data = saneyaml.load(fi.read())
+
+        if not conda_data or not isinstance(conda_data, dict):
+            conda_data = {}
+
+        name = conda_data.get('name')
+        extra_data = {}
+        channels = conda_data.get('channels')
+        if channels:
+            extra_data['channels'] = channels
+
+        dependencies = []
+        deps_data = conda_data.get('dependencies') or []
+        for dep in deps_data:
+            if isinstance(dep, str):
+                match = re.split(r'(>=|<=|==|=|>|<)', dep, maxsplit=1)
+                dep_name = match[0].strip()
+                version = None
+
+                if len(match) > 1:
+                    version = match[2].strip()
+
+                purl = PackageURL(type='conda', name=dep_name)
+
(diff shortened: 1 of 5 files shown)


kumarasantosh:
@pombredanne 

The two failing checks (core_tests misc_and_scancode) look like 
a pre-existing issue — #4732 is failing with the same thing.

My changes only touch conda.py, test_conda.py and the fixture files
so shouldn't be related.

I would really appreciate your feedback on this if you have time!

Thanks!

src/packagedcode/conda.py
```diff
@@ -666,3 +666,79 @@ def get_variables(location):
                 parts = line.split('=')
                 result[parts[0].strip()] = parts[-1].strip().strip('"')
     return result
+
+
+class CondaEnvironmentYmlHandler(BaseDependencyFileHandler):
```
AyanSinhaMahapatra, line 671: Have you actually looked at the code/run a scan on the test file to see this is not already detected somewhat? 
See https://github.com/aboutcode-org/scancode-toolkit/blob/develop/src/packagedcode/conda.py#L357 where we already support this (but this needs enhancement of course)

Yes we should have updated/closed https://github.com/aboutcode-org/scancode-toolkit/issues/3694 when this was added, but you should always do basic checks and read the code before writing code blindly.

tests/packagedcode/test_conda_environment_yml.py
```diff
@@ -0,0 +1,63 @@
+import os
```
AyanSinhaMahapatra, line 1: This is not how we run tests, see https://github.com/aboutcode-org/scancode-toolkit/blob/develop/tests/packagedcode/test_conda.py, where we have sample tests, and that is where the new tests will live.

tests/packagedcode/data/conda/environment_yml/simple-environment.yml
```diff
@@ -0,0 +1,9 @@
+name: testenv
```
AyanSinhaMahapatra, line 1: You need to use real examples seen in the wild, not made up examples.
See examples already in https://github.com/aboutcode-org/scancode-toolkit/tree/develop/tests/packagedcode/data/conda/conda-yaml

AyanSinhaMahapatra (commented):
Closing this PR as the code is not usable at all as the functionality already exists to some extent and might need updating, not completely new code. Even without that, this code was not mergable.'

Please open a new PR if you have some usable code which improves functionality

PR #4807, opened 2026-03-12
Fix the `openssl-exception-lgpl-3.0-plus` primary `spdx_license_key`

sschuberth:
The primary `spdx_license_key` should be the more correct dashed variant, and the other older variant should only be kept for backward compatibility.


sschuberth:
Superseded by https://github.com/aboutcode-org/scancode-toolkit/pull/5221.

PR #4810, opened 2026-03-13
packagedcode: fix gemspec version constants being stored as-is

### LINKED ISSUE #3129: incorrect gem version
We get `pkg:gem/elasticsearch-api@Elasticsearch::API::VERSION` and 
`pkg:gem/elasticsearch-api@Elasticsearch::API::VERSION?uuid=5e6ab586-9674-4f11-b9cb-75f384d8594b`  when parsing a gemspec. See https://github.com/elastic/elasticsearch-ruby/blob/f745eddd3b4332cf365c5df53f482a86a05ae96b/elasticsearch-api/elasticsearch-api.gemspec#L25
We may be able to recover from this.

kumarasantosh:
Fixes #3129 

This updates gemspec parsing so unresolved Ruby version constants such as Elasticsearch::API::VERSION, Excon::VERSION, and Faraday::VERSION are not stored as literal package versions.

This adds an is_ruby_version_constant() helper in spec.py that detects these unresolvable constants and sets version to None instead of storing the literal constant string.

### DIFF
--- a/src/packagedcode/rubygems.py
+++ b/src/packagedcode/rubygems.py
@@ -706,6 +706,7 @@ def party_mapper(role, names=[], emails=[]):
             models.Party(type=models.party_person, email=email, role=role)
             for email in emails
         )
+    return ()
 
 
 def get_parties(gem_data):
--- a/src/packagedcode/spec.py
+++ b/src/packagedcode/spec.py
@@ -133,6 +133,40 @@ PARSER_BY_NAME = {
 }
 
 
+def is_ruby_version_constant(value):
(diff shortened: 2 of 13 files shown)


AyanSinhaMahapatra (changes requested):
Thanks @kumarasantosh, this looks fine to be merged, pending a couple general changes as mentioned in https://github.com/aboutcode-org/scancode-toolkit/pull/4803#pullrequestreview-4113620701, you need to add full scan tests, if you're adding new test files, in addition to unittests.

PR #4878, opened 2026-03-31
Update Pixar regression fixture

### LINKED ISSUE #3241: Validate that we can detect this modified Apache license
The license @ https://github.com/pombredanne/Yc/blob/master/LICENSE.md should be scanned and we should validate that everything is detected correctly.

kumarasantosh:
This PR updates the regression fixture for Fixes #3241 by replacing the earlier notice-style text with the full Yc/LICENSE.md content, improving test fidelity and better reflecting real-world detection scenarios. The expected detection remains pixar, consistent with current behavior on develop. The updated fixture was verified to correctly detect pixar, and the focused data-driven license tests pass successfully in the project venv, with no regressions observed in existing tests.

### DIFF
--- a/tests/licensedcode/data/datadriven/external/pixar-modified-apache.txt
+++ b/tests/licensedcode/data/datadriven/external/pixar-modified-apache.txt
@@ -1,0 +1,172 @@
+# Modified Apache License
+
+_Version 2.0, January 2004_
+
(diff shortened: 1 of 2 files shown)


AyanSinhaMahapatra (approved):
Thanks for the update @kumarasantosh

PR #5009, opened 2026-05-18
Make 'no such option' assertion tolerant of newer click error format

GuillemSeCa:
Fixes the `*_latest_from_pip` CI matrix (ubuntu22/24, macos14, win2019/2022) which currently fails on `develop` because click 8.2 changed the unknown-option error message format:

- click < 8.2:  `Error: No such option: --foo`
- click >= 8.2: `Error: No such option '--foo'. (Did you mean ...)`

The test now asserts the stable substrings (`no such option`, `--json--info`) instead of the exact phrasing, so it works with both versions.

Verified locally: passes with click 8.3.1.

### DIFF
--- a/tests/scancode/test_cli.py
+++ b/tests/scancode/test_cli.py
@@ -730,7 +730,12 @@ def test_scan_errors_out_with_unknown_option():
     test_file = test_env.get_test_loc('license_text/test.txt')
     args = ['--json--info', test_file]
     result = run_scan_click(args, expected_rc=2)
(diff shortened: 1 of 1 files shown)


GuillemSeCa:
I saw the CI was slow so I also made this optional PR to speed it up 1 order of magnitude: #5013

AyanSinhaMahapatra (approved):
@GuillemSeCa Thanks++ for this, merging!

PR #5074, opened 2026-05-27
Fix missing top-level manifest classification in DatafileHandler

Shaurya-dev7:
## Summary

Fix package manifest classification for handlers relying on the default `DatafileHandler` implementation.

Package manifests containing `package_data` bypass the normal classification flow and depend on `get_top_level_resources()`. Since the base implementation was a no-op, manifests like `package.json` were never classified as key files, causing their license evidence to be invisible to the license clarity scorer.

## Fix

* implement a default `get_top_level_resources()` in `DatafileHandler`
* yield the manifest resource itself
* yield sibling legal/readme files using existing classification heuristics
* add regression coverage for an npm package containing only compound SPDX metadata

## Result

The npm compound SPDX regression case now correctly:

* classifies `package.json` as a key file
* detects declared licensing evidence
* computes the expected license clarity score

All related score, classify, and npm tests pass successfully.

### DIFF
--- a/src/packagedcode/models.py
+++ b/src/packagedcode/models.py
@@ -1518,8 +1518,33 @@ class DatafileHandler:
     def get_top_level_resources(cls, manifest_resource, codebase):
         """
         Yield Resources that are considered top-level for a Package type.
+
+        The default implementation yields the manifest resource itself
+        (which is authoritative package-level metadata and should always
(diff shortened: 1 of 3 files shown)


AyanSinhaMahapatra:
Can you link the relevant issue in your commit and the PR description? Fix test failures?

Issue #5077, opened 2026-05-28
[GSoC] Build NER training dataset from annotated rules


Kaushik-Kumar-CEG:
Tasks:
- Parse .RULE files and extract extract required phrase markers
- Propagate annotations (with CLI commands from #3924 ) to unannotated rules (with matching text) to expand the dataset
- Improve composite rules (AND/OR expressions where some aren't marked)
- Convert to BIOES labels
- Split into train/val/test by rule family
- Validate dataset quality

Timeline: Weeks 1-2 (current)

Deliverable: BIOES dataset ready for training (train/val/test splits)

PR #5105, opened 2026-06-03
Add script to extract required phrases from rule files

Kaushik-Kumar-CEG:
references #5077

script that extracts annotated rules (those with `{{ }}` markers) into a BIOES-labeled JSONL dataset for NER model training

What it does:
- Reads all `.RULE` files, skips `is_required_phrase` rules
- Tokenizes using scancode's `required_phrase_splitter` and generates BIOES labels from markers
- Splits 80/10/10 by license expression to prevent data leakage (heavy expressions split per-rule via hash, light ones kept together)
- Outputs train.jsonl, val.jsonl, test.jsonl

Usage:
```
python etc/scripts/dataset_pipeline/build_dataset.py --output-dir dataset-output
```

### Tasks

* [x] Tests pass

disclosure : used Claude to help review and clean up few bugs in script

### DIFF
--- a/etc/scripts/dataset_pipeline/build_dataset.py
+++ b/etc/scripts/dataset_pipeline/build_dataset.py
@@ -1,0 +1,212 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) nexB Inc. and others. All rights reserved.
+# ScanCode is a trademark of nexB Inc.
+# SPDX-License-Identifier: Apache-2.0
+# See http://www.apache.org/licenses/LICENSE-2.0 for the license text.
+# See https://github.com/nexB/scancode-toolkit for support or download.
+# See https://aboutcode.org for more information about nexB OSS projects.
+#
+
+"""Build a BIOES dataset from required phrases marked in license rules."""
+
+from collections import Counter
+import hashlib
+import json
+from pathlib import Path
+import unicodedata
+
+import click
+
+from licensedcode.models import load_rules
+from licensedcode.models import rules_data_dir as default_rules_data_dir
+from licensedcode.required_phrases import get_required_phrase_verbatim
+from licensedcode.tokenize import get_existing_required_phrase_spans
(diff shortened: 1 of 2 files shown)


Kaushik-Kumar-CEG:
@AyanSinhaMahapatra ready for review when you get a chance. ill make the follow up changes to script as well

pombredanne (changes requested):
@Kaushik-Kumar-CEG why not reusing existing code for that https://github.com/aboutcode-org/scancode-toolkit/tree/develop/src/licensedcode like in https://github.com/aboutcode-org/scancode-toolkit/blob/ea42c1d09d38444c6206a245d67ca3d99930c9b0/src/licensedcode/required_phrases.py#L218 ? this script feels redundant work, and it also is missing tests.

Kaushik-Kumar-CEG:
thanks for the review @pombredanne

`collect_is_required_phrase_from_rules` is for rules with `is_required_phrase: yes` which is a different case from what this script handles (inline `{{ }}` phrases which is what the NER model trains on)

skipping `is_required_phrase: yes` rules is intentional. those rules already function as required phrases and don't need `{{ }}` annotation, so they're not in scope for what the model predicts at inference. also tested it in my prototype: including them in training made the model mark entire rule texts as the phrase instead of finding the actual span

on the redundancy, you're right. dropping the local regex in the next commit and using `get_required_phrase_verbatim` for phrase extraction.for BIOES labels in a later commit, `required_phrase_tokenizer` will be reused as well

Ill add the required tests once feature commits lands. thanks :)

Kaushik-Kumar-CEG:
Update: the maintained dataset command now lives in the dedicated `aboutcode-org/scancode-required-phrases` package in [package PR #1](https://github.com/aboutcode-org/scancode-required-phrases/pull/1). I am keeping this PR open for its development and review history while the package stack is reviewed. No duplicate implementation changes are planned here.

PR #5153, opened 2026-06-26
Add timeout to process codebase and plugin steps

AyanSinhaMahapatra:
Reference:
* https://github.com/aboutcode-org/scancode-toolkit/issues/4061

### Tasks

### DIFF
--- a/src/scancode/cli.py
+++ b/src/scancode/cli.py
@@ -68,6 +68,7 @@ from scancode import Scanner
 from scancode.help import epilog_text
 from scancode.help import examples_text
 from scancode.interrupt import DEFAULT_TIMEOUT
+from scancode.interrupt import DEFAULT_PLUGIN_TIMEOUT
 from scancode.interrupt import fake_interruptible
 from scancode.interrupt import interruptible
 from scancode.pool import ScanCodeTimeoutError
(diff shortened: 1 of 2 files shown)


AyanSinhaMahapatra:
Need some review there wrt. the default behaviour there, as these timeouts can cause weird issues/crashes elsewhere, so should we just keep this only as a CLI option which can be enabled to set a timeout, or have a default timeout like we have in the file-level scan timeout. If we have this as a default we also have to decide on a reasonable timeout duration.

src/scancode/interrupt.py
```diff
@@ -45,6 +45,7 @@ class TimeoutError(Exception):  # NOQA
 
 
 DEFAULT_TIMEOUT = 120  # seconds
+DEFAULT_PLUGIN_TIMEOUT = 2400 # seconds
```
pombredanne, line 48: This is way too much? 40 minutes feels forever. And why 40 minutes?

src/scancode/cli.py
```diff
@@ -253,6 +254,14 @@ def default_processes():
          f'[default: {DEFAULT_TIMEOUT} seconds]',
     help_group=cliutils.CORE_GROUP, sort_order=10, cls=PluggableCommandLineOption)
 
+@click.option('--timeout-plugins',
```
pombredanne, line 257: I find that new option confusing... and the default of 2400s super long

src/scancode/cli.py
```diff
@@ -1208,8 +1233,10 @@ def run_scanners(
 
     # TODO: add progress indicator
     # run the process codebase of each scan plugin (most often a no-op)
+    use_threading = processes >= 0
```
pombredanne, line 1236: Are you sure you want that? It seems not used at all

src/scancode/cli.py
```diff
@@ -253,6 +254,14 @@ def default_processes():
          f'[default: {DEFAULT_TIMEOUT} seconds]',
     help_group=cliutils.CORE_GROUP, sort_order=10, cls=PluggableCommandLineOption)
 
+@click.option('--timeout-plugins',
+    type=float,
+    default=DEFAULT_PLUGIN_TIMEOUT,
+    metavar='<seconds>',
+    help='Stop an unfinished codebase processing or post-scan plugin after'
```
pombredanne, line 261: I cannot put my head around this... could we not use a simpler single timeout across the board instead?

pombredanne (changes requested):
Thanks! Could we do better and avoid another new option?

pombredanne:
gentle ping

PR #5251, opened 2026-07-28
Add required phrases to lgpl-3.0_16.RULE

### LINKED ISSUE #4995: LGPL 3 false positive (should be LGPL 2 or 2.1)
### Description

https://github.com/KDE/libqzeitgeist/blob/master/src/logbrowser.cpp

is recognized as LGPL 3:

```
      "license_detections": [
        {
          "license_expression": "lgpl-3.0",
          "license_expression_spdx": "LGPL-3.0-only",
          "matches": [
            {
              "license_expression": "lgpl-3.0",
              "license_expression_spdx": "LGPL-3.0-only",
              "from_file": "logbrowser.cpp",
              "start_line": 4,
              "end_line": 16,
              "matcher": "3-seq",
              "score": 88.68,
              "matched_length": 94,
              "match_coverage": 88.68,
              "rule_relevance": 100,
              "rule_identifier": "lgpl-3.0_16.RULE",
              "rule_url": "https://github.com/nexB/scancode-toolkit/tree/develop/src/licensedcode/data/rules/lgpl-3.0_16.RULE"
            }
          ],
          "identifier": "lgpl_3_0-129dffd6-23d4-eb21-f89e-b5145e4553c9"
        }
      ],

```

What absolutely baffles me is that the exact same license header is used in #4994 where it is not recognized as LGPL 3.

Eljees:
Reference: #4995

## Problem

`lgpl-3.0_16.RULE` is an LGPL-3.0 notice, but none of its text is required, so a partial sequence match reports LGPL-3.0 for an LGPL-2.x notice of the same shape that never mentions version 3.

The reporter's file, [`KDE/libqzeitgeist/src/logbrowser.cpp`](https://github.com/KDE/libqzeitgeist/blob/master/src/logbrowser.cpp), says:

> ...under the terms of the GNU Lesser General Public License as published by the Free Software Foundation; either **version 2** of the License, or (at your option) any later version.

and is detected as `LGPL-3.0-only`.

This follows @pombredanne's suggestion in the issue that the rule should get required phrases.

## Change

The two version-bearing phrases are marked as required:

```
it under the terms of the {{GNU Lesser General Public License version 3}} as
...
You should have received a copy of the {{GNU Lesser General Public License version 3}}
```

Nothing else in the rule changes.

## Verification

Measured with scancode-toolkit 32.5.0, rebuilding the license index (`scancode-reindex-licenses`) between runs.

| input | before | after |
|---|---|---|
| the reported `logbrowser.cpp` | `LGPL-3.0-only` — `lgpl-3.0_16.RULE`, score 88.68 | `LGPL-2.0-or-later AND LGPL-2.1-or-later` — `lgpl-2.0-plus_1.RULE` (100.0) + `lgpl-2.1-plus_113.RULE` |
| a genuine LGPL-3.0 notice of this rule's own shape | `LGPL-3.0-only`, score 100.0 | `LGPL-3.0-only` — `lgpl-3.0_16.RULE`, score 100.0 |

So the false positive resolves to the license the file actually carries, and the rule still matches real LGPL-3.0 notices at full score — the second row is there because required phrases can easily turn a false positive into a false negative, and I wanted that pinned down rather than assumed.

One methodology note that cost me an hour and may be worth knowing: editing a rule inside an installed scancode and rescanning changes **nothing** — the index is cached. I only caught this because a control run with the rule file *deleted entirely* still produced the identical match at the identical score. All numbers above are from runs with an explicit reindex.

I have not run the full `tests/licensedcode` suite; if there are expected-detection fixtures that reference this rule, they'd want a look.


pombredanne:
Did you run tests and add new tests?

pombredanne:
And we are evolving our policy for the use of LLMs. Was this entirely AI-generated? Please join our Slack for discussion.

pombredanne:
Especially when I see this comment:
- https://github.com/ossf/cve-bin-tool/pull/5781#issuecomment-4929445970

Eljees:
Yes — I use an LLM assistant for drafting, and I should have said so up front. Sorry for that, and sorry if my earlier comments read as if every word was mine.

What I'm doing here isn't PR farming. My day job is AppSec false-positive triage of SAST/SCA findings; I'm preparing a conference talk and collecting data on how well models actually do on that kind of work. That only works if the contributions hold up, so I run the tools myself before opening anything.

For this rule specifically:
- current rules: the KDE header matches `LGPL-3.0-only`, score 88.68
- with the required phrases added: `LGPL-2.0-or-later AND LGPL-2.1-or-later`
- control: a genuine LGPL-3.0 notice still matches at 100.0

Both runs after `scancode-reindex-licenses` — without the reindex the license index stays cached and rule edits silently do nothing, which cost me an hour before I caught it by deleting a rule entirely and seeing the same match survive.

I did not add a test file. Tell me where you want it (`tests/licensedcode/data/...`) and I'll add it in this PR.

And for what it's worth — humans have been shipping bad code for decades without any help from models, mine included. :) The disclosure point is fair, though, and I'll state it in my PRs from now on. Happy to join the Slack discussion.

Eljees:
yeah, of course testing it locally, occasionally new here

Eljees:
my code is far more worse, when i write it by hands

Eljees:
sorry for that, frameworks fail sometimes, adjustiing =(

Eljees:
Concretely, on both questions.

**New test.** Added in bf999f78: `tests/licensedcode/data/datadriven/lic4/4995-lgpl-2.0-plus-not-lgpl-3.0.cpp` plus its `.yml`, picked up automatically by `test_detection_datadriven4`. The fixture is the license header verbatim from the reporter's file in #4995 (`KDE/libqzeitgeist/src/logbrowser.cpp`), and it pins the invariant this PR is about: a notice that never mentions version 3 must not resolve to `lgpl-3.0`. It currently detects two exact 100.0-score matches, `lgpl-2.0-plus_1.RULE` (lines 4-14) and `lgpl-2.0-plus_451.RULE` (lines 14-16). No existing fixture needed regeneration, and all the license-data-sensitive suites are green on this PR: datadriven1_2, datadriven3_4, datadriven_ext, every license_validate_*, license_cache, additional_license_combined, and all Windows jobs.

One correction to my own PR description while I am here: it claimed `lgpl-2.0-plus AND lgpl-2.1-plus` based on a manual check on 32.5.0. Against current develop the result is the two `lgpl-2.0-plus` matches above. I inspected the match trace rather than papering over the difference; the version-3 invariant holds either way.

**The red CI is not this PR.** The latest develop build, 18996, ran on 6ba590897 - the exact parent commit of this PR - on 2026-06-25, and fails with the same seven job kinds:

- `core_tests cluecode_license_base`: only `test_match_has_correct_line_positions_in_automake_perl_file`, a +12 token-position shift (`Span(998, 1061)`/`Span(1371, 1393)` vs expected). Build 18996 fails it with bit-identical spans, so it dates to the June update-licenses merge (4e8d71149, #5112) adding rule tokens - the case that test's own docstring warns about. I deliberately did not update those expectations here; that is develop's breakage, not mine, and fixing it in this PR would misattribute the cause.
- posix `all` jobs: `test_scan_info_returns_full_root` and `test_scan_info_returns_correct_full_root_with_single_file`, a missing leading `/` in the expected path. Posix-only, Windows green, failing on 18996 as well. A `.RULE` text edit cannot affect path rooting.
- `core_tests misc_and_scancode`: `test_interrupt.py` and the `no_multiprocess` ignore test dying on `ValueError: signal only works in main thread of the main interpreter` - SIGALRM under pytest-xdist on Python 3.14.6 on the agents.

I reproduced the first two classes locally at this commit in a `python:3.10-slim` container built with the exact `configure --dev` steps, and got the same numbers as CI. Happy to file the automake spans and the posix `full_root` cases as separate issues if that is useful.

**On the LLM question** - fair to ask, and you were right to. Yes, I use an assistant for drafting, and I said so above rather than let you find out. My own unaided code and English are worse than what I produce with it, so I use it and then check the result myself; the tests I run and the evidence above are mine. I am happy to take this to Slack for the policy discussion. If your evolving policy ends up not wanting contributions produced this way, say so and I will close this PR without argument.

Eljees:
Ping — this has been open since 28 July.

To save you the triage: the red CI here is not from this change. Every recent PR in this repo fails the same eight Azure jobs — `core_tests cluecode_license_base`, `macos14_cpython`, `ubuntu22_cpython`, `ubuntu24_cpython` and their `latest_from_pip` variants — regardless of what it touches. #5246 (a Symfony lock test), #5252, #5259 and #5261 are four current examples by four different authors.

The change is in `bf999f78`: required phrases on `lgpl-3.0_16.RULE`, plus a datadriven fixture at `tests/licensedcode/data/datadriven/lic4/4995-lgpl-2.0-plus-not-lgpl-3.0.cpp` and its `.yml`, picked up automatically by `test_detection_datadriven4`. It pins the invariant this PR is about: a notice that never mentions version 3 must not resolve to `lgpl-3.0`. The fixture is the header verbatim from the reporter's file in #4995.

Happy to rebase if it has gone stale, or to close it if the rule change isn't wanted.

Eljees:
The red run on this PR is build 19157 from 28 July, and the failure is not in this branch. Two tests failed:

```
FAILED tests/scancode/test_cli.py::test_scan_info_returns_full_root
FAILED tests/scancode/test_cli.py::test_scan_info_returns_correct_full_root_with_single_file
AssertionError: assert '/home/vsts/work/1/s/tests/scancode/data/info/basic.tgz'
                    == 'home/vsts/work/1/s/tests/scancode/data/info/basic.tgz'
```

Both are about the leading slash in `--full-root` paths, and neither touches a file this branch changes. They were failing on `develop` at the time and were fixed there by 639a1b8 ("Fix full-root path expectations"), which dropped the `.strip('/')` from the two expectations.

I have merged current `develop` in (086e5fc), so the next run happens on a base that carries that fix. The change itself is unchanged: required-phrase markers around "GNU Lesser General Public License version 3" in `lgpl-3.0_16.RULE`, plus the datadriven test for the `lgpl-2.0-plus` file that was matching it.

Eljees:
The merge cleared six of the seven failing jobs - 30 of 32 checks are green on `086e5fc7`. The one left is `core_tests misc_and_scancode`, and I do not think it can come from this branch:

- `develop` is an ancestor of `086e5fc7`, and the complete diff against it is three files: `src/licensedcode/data/rules/lgpl-3.0_16.RULE` and the two datadriven fixtures.
- That job runs with `--ignore=tests/licensedcode`, so both fixtures are outside its selection. Only the RULE is left.
- The RULE change only adds required phrases, so it can make that rule match less, never more or differently. Nothing in the test tree records a match to it - `grep -rl 'lgpl-3.0_16' tests/ --include='*.json' --include='*.yml' --include='*.yaml' --include='*.jsonlines' --include='*.rdf'` returns only the fixture this PR adds - so no expected output can shift.
- `develop` at 3c532f0f is green on all 39 checks of the same pipeline, `misc_and_scancode` included.

I could not reproduce it locally: a large share of that selection shells out to `venv/bin/scancode`, and my sandbox is not a faithful enough copy of the pipeline for a local result to mean anything.

Could someone re-run that one job? If it fails a second time I will set up the pipeline environment properly and find it rather than reason about it from the outside.

Eljees:
Following up on my own promise from 23 August - I said that if it failed again I would set the pipeline environment up properly rather than reason about it from outside, so I did.

Built the branch with `./configure --dev` this time, so the tests that shell out to `venv/bin/scancode` actually work (that is what made my earlier attempt meaningless), and ran the exact `misc_and_scancode` selection from `azure-pipelines.yml`:

```
venv/bin/pytest -n 3 --test-suite=all \
  --ignore=tests/cluecode --ignore=tests/licensedcode \
  --ignore=tests/scancode/test_cli.py --reruns 2
```

| tree | result |
| --- | --- |
| this branch | 14 failed, 4020 passed, 4 skipped, 2 xfailed |
| `develop` | 14 failed, 4020 passed, 4 skipped, 2 xfailed |

The failing sets are **identical, test for test**. All fourteen are artefacts of my sandbox rather than of either tree: eleven are permission tests (`test_is_readable_is_writeable_*`, `test_chmod_*`, `test_copytree_*`, `test_validate_output_file_path_*_non_writable_*`), which cannot fail as they expect when the runner is root, and three need an rpm database library (`test_recognize_rpmdb_sqlite`, `test_collect_installed_rpmdb_xmlish_from_rootfs`, `test_scan_system_package_end_to_end_installed_rpms_fedora_bdb`).

So the selection behaves the same on this branch as on `develop` once the environment is right, which is what the static argument already suggested: `develop` is an ancestor, the whole diff against it is three files, two of them are under `tests/licensedcode` which that job ignores, and nothing in the test tree records a match to `lgpl-3.0_16`.

I have also merged current `develop` (bed47d81) since the branch had fallen six commits behind again - nothing in those six touches the rule or the fixtures. That gives the job a fresh run at the same time, which should settle whether the earlier failure was transient.

Eljees:
Green - 32/32 on bed47d81, `core_tests misc_and_scancode` included. So that one was transient, which matches the local comparison above: the job's selection behaved identically on this branch and on `develop`.

Nothing outstanding from my side; ready whenever someone has time to look at the rule change itself.

AyanSinhaMahapatra:
Thanks, but I would encourage you to not waste tokens like this wholesale for such small and trivial things. Do not use AI to write comments on issues or PRs

PR #3100, opened 2022-09-12
Add license dump option

AyanSinhaMahapatra:
1. Modifies the License and Rule files to have their data as YAML frontmatter.

* .RULE files and .LICENSE files now also contain their data.
* git history is preserved
* loading/dumping methods are modified to support YAML frontmatter
* License/Rule now has different methods to point to their file.

2. Adds a new command line option: `--get-license-data` to:

* Dump license data in JSON, YAML and HTML formats.
* Also dumps the .LICENSE file with text and data as YAML frontmatter.
* Generates an index and a static website to view the data.

This is reusing code originally located at:
https://github.com/nexB/scancode-licensedb

Reference: https://github.com/nexB/scancode-toolkit/issues/2738
Reference: https://github.com/nexB/scancode-toolkit/issues/3049

This also contains PR: https://github.com/nexB/scancode-toolkit/pull/3063 btw.

### Tasks

* [x] Reviewed [contribution guidelines](https://github.com/nexB/scancode-toolkit/blob/develop/CONTRIBUTING.rst)
* [x] PR is descriptively titled 📑 and links the original issue above 🔗
* [x] Tests pass -- look for a green checkbox ✔️ a few minutes after opening your PR
  Run [tests](https://scancode-toolkit.readthedocs.io/en/latest/contribute/contrib_dev.html#running-tests) locally to check for errors. 
* [x] Commits are in uniquely-named feature branch and has no merge conflicts 📁

PR #5259, opened 2026-08-03
Update license scripts to support YAML frontmatter changes

### LINKED ISSUE #3138: Update license scripts for yaml frontmatter
Update scripts in https://github.com/nexB/scancode-toolkit/tree/develop/etc/scripts/licenses to work with the updates made in https://github.com/nexB/scancode-toolkit/pull/3100

sarthak-shubham:
Fixes #3138

## What
Updated `report_license_rules.py` and `buildrules.py` in `etc/scripts/licenses/`
to work correctly with the YAML frontmatter format introduced in #3100, which
added new fields (`is_builtin`, `is_generic`, `is_continuous`, `length`) to the
`License` and `Rule` models.

### report_license_rules.py
- Added `extrasaction='ignore'` to the `csv.DictWriter` calls so the script no
  longer crashes when the models gain fields not present in the CSV column
  headers — also avoids needing to update this script every time a new model
  field is added going forward.
- Added `encoding="utf-8"` to the `open()` call in `write_data_to_csv()`, since
  the default Windows encoding (cp1252) can't handle non-ASCII characters
  (e.g. in author names) present in the license database.
- Added progress logging via `click.echo()` (start, periodic progress every
  5,000 rules, completion message) since exporting 35,000+ rules takes several
  minutes with no prior feedback that the script is working.
- Switched to streaming with `enumerate()` instead of loading all rules into a
  list upfront, to support the progress logging without changing the output.

### buildrules.py
- Added a `click.echo()` message before the duplicate-check step
  (`all_rule_by_tokens()`). This step tokenizes all 35,000+ existing rules to
  check the new rule for duplicates, which takes roughly 10 minutes on a full
  run — expected given the dataset size, but silent without the log.

## Testing
Ran both scripts manually against the current license/rule database —
report_license_rules.py completes end-to-end; buildrules.py correctly parses
an example rule file and generates a new .RULE file (deleted afterward to
avoid committing test artifacts).

I wasn't able to find existing automated tests covering these two scripts —
they appear to be maintainer-facing utility scripts outside the main src/
detection engine and its test suite. Is there a preferred location/pattern
for tests on scripts like these, or should I add something under tests/
following a different structure?

## Note
Saw #3150 referenced in the issue as possibly covering part of this — happy
to adjust if there's overlap.

## Also included
Added an AUTHORS.rst entry and a CHANGELOG.rst note per CONTRIBUTING.rst's
PR guidelines.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -79,6 +79,7 @@ The following organizations or individuals have contributed to ScanCode:
 - Sankha Das @sankha555
 - Saravanan G @SaravananOffl
 - Sarita Singh @itssingh
+- Sarthak Shubham @sarthak-shubham
 - Savino Sguera @savinos
 - Sebastian Roth @ened
 - Sebastian Schuberth @sschuberth
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,13 @@ Next release
   ``licensedcode-data``.
   https://github.com/aboutcode-org/scancode-toolkit/pull/5056
 
+- Fix ``report_license_rules.py`` and ``buildrules.py`` in
+  ``etc/scripts/licenses/`` to work correctly with the YAML frontmatter
+  fields added in #3100.
+  https://github.com/aboutcode-org/scancode-toolkit/issues/3138
+
+
+
 v33.0.0rc1 - 2026-05-14
 ------------------------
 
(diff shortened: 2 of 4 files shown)


sarthak-shubham:
@AyanSinhaMahapatra Hey! @pombredanne suggested on Gitter that I ping you here to collaborate on this PR. I'd love to get your input on it whenever you have a moment. Thanks in advance!

PR #5295, opened 2026-08-30
Fix Poetry lock optional dependency parsing

### LINKED ISSUE #5294: PoetryLockHandler incorrectly marks non-optional packages as optional
### Description

`PoetryLockHandler` incorrectly reports resolved Poetry lockfile packages as optional even when the package has `optional = false` in `poetry.lock`.

For example, the existing ScanCode Poetry fixture contains:

[[package]]
name = "attrs"
version = "21.4.0"
optional = false

However, ScanCode reports the corresponding resolved dependency with:

"is_optional": true

The issue appears to come from the `PoetryLockHandler` implementation in `src/packagedcode/pypi.py`, which currently uses:

is_optional = package.get("is_optional") or True

Poetry uses the `optional` field in the lockfile, not `is_optional`. Additionally, using `or True` causes an explicit `False` value to become `True`.

The expected behavior is for a Poetry lockfile entry with `optional = false` to result in `is_optional = false`, while `optional = true` should result in `is_optional = true`.

### How To Reproduce

The issue can be reproduced using the existing Poetry lockfile fixture:

tests/packagedcode/data/pypi/poetry/univers/poetry.lock

The fixture contains packages with:

optional = false

Run the existing Poetry lock parsing test:

.\venv\Scripts\pytest.exe tests\packagedcode\test_pypi.py::TestPoetryHandler::test_parse_poetry_lock_univers -q

Before the fix, the test fails because the parsed result reports the resolved packages as optional, while the Poetry lockfile specifies `optional = false`.

The behavior can also be reproduced with a minimal Poetry lockfile containing both `optional = false` and `optional = true`. The current implementation incorrectly reports the package with `optional = false` as optional.

### System configuration

* What OS are you running on? Windows
* What version of scancode-toolkit was used to generate the scan file? `commoncode-v32.5.2-14-g441f36f725`
* What installation method was used to install/run scancode? Source checkout with the project's virtual environment
* Python version: 3.14.0

Mars-60:
Fixes #5294

### Tasks

### Description

`PoetryLockHandler` incorrectly reported resolved Poetry lockfile packages as optional even when the package had `optional = false` in `poetry.lock`.

The fix maps Poetry's `optional` field to ScanCode's `is_optional` field:

```python
is_optional = package.get("optional", False)

### CODE REVIEW

File: tests/packagedcode/test_pypi.py [hunk 1]
```diff
@@ -398,6 +398,39 @@ def test_parse_poetry_lock_univers(self):
         expected_loc = self.get_test_loc('pypi/poetry/univers-poetry.lock-expected.json')
         self.check_packages_data(package, expected_loc, regen=REGEN_TEST_FIXTURES)
 
+    def test_parse_poetry_lock_package_optional(self):
```
AyanSinhaMahapatra, line 401: Can you use an actual test file and expected results file like in the tests above?

src/packagedcode/pypi.py
```diff
@@ -806,7 +806,7 @@ def parse(cls, location, package_only=False):
             )
             resolved_package = models.PackageData.from_data(package_data, package_only)
 
-            is_optional = package.get("is_optional") or True
+            is_optional = package.get("optional", False)
```
AyanSinhaMahapatra, line 809: Thanks, good catch

tests/packagedcode/test_pypi.py
```diff
@@ -398,6 +398,39 @@ def test_parse_poetry_lock_univers(self):
...
+''')
+
+        package_data = list(pypi.PoetryLockHandler.parse(test_file))[0]
+        dependencies_by_purl = {
+            dependency['purl']: dependency
+            for dependency in package_data.dependencies
+        }
+
+        assert dependencies_by_purl['pkg:pypi/required-dep@1.0.0']['is_optional'] is False
+        assert dependencies_by_purl['pkg:pypi/optional-dep@2.0.0']['is_optional'] is True
```
AyanSinhaMahapatra: No need to parse results and assert specifically, we run whole file results test with a test and an expected file, just check the expected behaviour and results as a whole

AyanSinhaMahapatra (changes requested):
Thanks @Mars-60 see comments for updating the tests, ready to merge otherwise.

Mars-60:
@AyanSinhaMahapatra ,the Azure core_tests misc_and_scancode check is still failing on the PR. I reproduced the reported interrupt test in a Linux/Python 3.14.7 environment using the xdist-style invocation, and it passed. I don't have permission to rerun the Azure pipeline directly. Could you please rerun the failed check when convenient?

PR #5310, opened 2026-09-11
Accept all PEP 508 operators in python_version markers

### LINKED ISSUE #5309: python_version markers using >, != or ~= are silently dropped from dependency extra_data
### Description

`get_python_version_os()` in `src/packagedcode/pypi.py` silently drops the `python_version` environment marker for some comparison operators, because the list of accepted operators contains `<` twice and is missing `>`:

https://github.com/aboutcode-org/scancode-toolkit/blob/develop/src/packagedcode/pypi.py#L2281

```python
python_version_operators = ['<', '>=', '==', '<=', '<']
```

`<` is duplicated where `>` was presumably intended, so `>` never matches. `!=` and `~=` are absent as well, though both are valid marker operators under PEP 508.

The value returned by this function becomes `extra_data` on the emitted `DependentPackage`:

```python
extra_data = {}
if req.marker:
    platform = get_python_version_os(req.marker)
    if platform:
        extra_data = platform
```

So a dependency declared as `requests; python_version > "3.8"` is reported with an empty `extra_data`, while `requests; python_version >= "3.8"` is reported correctly. The extracted marker data ends up inconsistent and incomplete depending only on which operator the package author happened to use.

### How To Reproduce

Calling the current `develop` implementation over parsed requirements:

```python
from packvers.requirements import Requirement
from packagedcode.pypi import get_python_version_os

for spec in [
    'requests; python_version >= "3.8"',
    'requests; python_version > "3.8"',
    'requests; python_version != "3.8"',
    'requests; python_version <= "3.8"',
    'requests; python_version ~= "3.8"',
]:
    print(spec, '->', get_python_version_os(Requirement(spec).marker))
```

Actual output:

```
requests; python_version >= "3.8"  -> {'python_version': '>= 3.8'}
requests; python_version >  "3.8"  -> {}          <-- dropped
requests; python_version != "3.8"  -> {}          <-- dropped
requests; python_version <= "3.8"  -> {'python_version': '<= 3.8'}
requests; python_version ~= "3.8"  -> {}          <-- dropped
```

Expected: each of these should report the operator and version under `extra_data["python_version"]`.

The same is visible end to end by scanning a `requirements.txt` that contains `requests; python_version > "3.8"` with:

```
scancode --package --json-pp - requirements.txt
```

and inspecting `extra_data` on the resulting dependency.

### Suggested fix

Replace the duplicated `<` with the operators that are actually missing, covering the full set allowed for a marker in PEP 508:

```python
python_version_operators = [
    '<', '<=', '!=', '==', '>=', '>', '~=', '===',
]
```

### System configuration

* Python version: 3.x

Daksha1611:
Fixes #5309

The operator list in `get_python_version_os()` has `<` twice and no `>`:

```python
python_version_operators = ['<', '>=', '==', '<=', '<']
```

So `requests; python_version > "3.8"` never matched and the marker was dropped from the dependency `extra_data`, while `>=` on the same requirement was picked up fine. `!=` and `~=` were missing for the same reason.

Replaced that list with the comparison operators PEP 508 allows in a marker, and added a test covering each of them. `sys_platform` and the no-marker path are untouched.


Shriprasad-P (commented):
## Review

**PR:** Accept all PEP 508 operators in python_version markers

**Touched:** `CHANGELOG.rst`, `src/packagedcode/pypi.py`, `tests/packagedcode/test_pypi.py`

- Addresses a concrete correctness/reliability issue based on the title and diff.
- Includes or touches tests — good for locking the behavior.
- Size looks manageable (+26/-1).

PR #5313, opened 2026-09-11
Remove duplicate keys in alpine DECLARED_TO_SPDX

### LINKED ISSUE #5312: Duplicate keys in alpine DECLARED_TO_SPDX silently discard two license mappings
### Description

`DECLARED_TO_SPDX` in `src/packagedcode/alpine.py` has two duplicate keys. A dict literal keeps only the last value for a repeated key, so the earlier entry of each pair is dead at import time.

`'as-is'`:

- line 1619: `'as-is': 'licenseref-scancode-free-unknown',`
- line 1701: `'as-is': 'licenseref-scancode-unknown-license-reference',`

`'custom:tu-berlin-2.0'`:

- line 1690: `'custom:tu-berlin-2.0': 'licenseref-scancode-unknown-license-reference',`
- line 1715: `'custom:tu-berlin-2.0': 'tu-berlin-2.0',`

Each of the four entries carries its own comment, so the two pairs look like they were added independently without noticing the collision.

This mapping is turned straight into the substitution table applied to every Alpine package license expression:

```python
DECLARED_TO_SPDX_SUBS = {
    licensing.parse(src, simple=True): licensing.parse(tgt, simple=True)
    for src, tgt in DECLARED_TO_SPDX.items()
}
```

So which value wins is decided by position in the file rather than by intent. For `'as-is'` the surviving value is the less specific `licenseref-scancode-unknown-license-reference`, and the `licenseref-scancode-free-unknown` mapping never applies.

### How To Reproduce

```python
import ast, collections

src = open('src/packagedcode/alpine.py').read()
        counts = collections.Counter(
            k.value for k in node.keys
        )
        for key, n in counts.items():
            if n > 1:
```

```
line 1568: duplicate key 'as-is' x2
line 1568: duplicate key 'custom:tu-berlin-2.0' x2
```

### Suggested fix

Drop the dead lines so each key appears once. Which of the two values is actually wanted for each key needs a maintainer decision, since right now it is settled by line order — I have opened a PR that removes the unreachable entries and keeps the currently-winning values, so behaviour does not change.

### System configuration

* Python version: 3.x

Daksha1611:
Fixes #5312

`DECLARED_TO_SPDX` defines `'as-is'` twice (lines 1619 and 1701) and `'custom:tu-berlin-2.0'` twice (lines 1690 and 1715). It is a plain dict literal, so the earlier entry of each pair was silently dropped at import and never applied.

This removes the two unreachable entries and keeps the values that already win, so the mapping is identical at runtime. Whether the surviving value is the right one for each key is a separate question and left to you — for `'as-is'` in particular the entry that was being discarded was the more specific one.


Shriprasad-P (commented):
## Review

**PR:** Remove duplicate keys in alpine DECLARED_TO_SPDX

**Touched:** `CHANGELOG.rst`, `src/packagedcode/alpine.py`

- Addresses a concrete correctness/reliability issue based on the title and diff.
- Security-sensitive change — please double-check edge cases and defaults.
- Size looks manageable (+4/-2).

PR #5315, opened 2026-09-11
Remove unreachable duplicate LGPL key in rubygems license mapping

### LINKED ISSUE #5314: Duplicate LGPL key in rubygems LICENSES_MAPPING makes one mapping unreachable
### Description

`LICENSES_MAPPING` in `src/packagedcode/rubygems.py` defines `'LGPL'` twice, on two consecutive lines (679-680):

```python
    'LGPL': 'lgpl',
    'LGPL': 'lgpl-2.0-plus',
```

Both are entries of the same dict literal, so the first is discarded and every gem declaring a bare `LGPL` resolves to `lgpl-2.0-plus`.

That is worth a second look on its own: an unversioned `LGPL` declaration does not state a version, and `lgpl-2.0-plus` asserts one. The generic `'LGPL': 'lgpl'` line immediately above looks like the mapping intended for the unversioned case, and the table already has separate explicit entries for the versioned spellings. Either way, which one applies is currently decided by line order rather than intent, and the unreachable line is misleading to read.

### How To Reproduce

```python
import ast, collections

src = open('src/packagedcode/rubygems.py').read()
        counts = collections.Counter(
            k.value for k in node.keys
        )
        for key, n in counts.items():
            if n > 1:
```

```
line 658: duplicate key 'LGPL' x2
```

And the effective value:

```python
>>> from packagedcode.rubygems import LICENSES_MAPPING
>>> LICENSES_MAPPING['LGPL']
'lgpl-2.0-plus'
```

The `'lgpl'` entry is unreachable.

### Suggested fix

Remove one of the two lines. I have opened a PR that drops the unreachable one and keeps the current result, so nothing changes at runtime. If a bare `LGPL` should instead stay unversioned as `lgpl`, that is a one-word change on top and I am happy to make it.

### System configuration

* Python version: 3.x

Daksha1611:
Fixes #5314

`LICENSES_MAPPING` has `'LGPL'` on two consecutive lines (679-680), so the first is discarded and a bare `LGPL` gem declaration always resolves to `lgpl-2.0-plus`.

This drops the unreachable line and keeps the value that already applies, so nothing changes at runtime. As noted in the issue, it is worth deciding whether an unversioned `LGPL` should stay unversioned as `lgpl` instead — that is a one-word change on top of this if you want it.


Shriprasad-P (commented):
## Review

**PR:** Remove unreachable duplicate LGPL key in rubygems license mapping

**Touched:** `CHANGELOG.rst`, `src/packagedcode/rubygems.py`

- CI/tooling change — confirm the pipeline still passes on this branch.
- Size looks manageable (+3/-1).

PR #5317, opened 2026-09-11
Remove duplicated autogen.sh entry in default ignores

### LINKED ISSUE #5316: Duplicate autogen.sh entry in commoncode default ignores
### Description

`src/commoncode/ignore.py` lists the same key twice in the default ignores, on two byte-identical consecutive lines (204-205):

```python
    "autogen.sh": "Default ignore: autotools artifact",
    "autogen.sh": "Default ignore: autotools artifact",
```

Nothing misbehaves here — the key and the value are identical, so the dict ends up the same either way. It is just a dead line, and filing it as a small cleanup rather than a bug.

### How To Reproduce

```
$ grep -n '"autogen.sh"' src/commoncode/ignore.py
204:    "autogen.sh": "Default ignore: autotools artifact",
205:    "autogen.sh": "Default ignore: autotools artifact",
```

```python
import ast, collections

src = open('src/commoncode/ignore.py').read()
        counts = collections.Counter(
            k.value for k in node.keys
        )
        for key, n in counts.items():
            if n > 1:
```

```
line 185: duplicate key 'autogen.sh' x2
```

### Suggested fix

Delete one of the two lines. PR attached.

### System configuration

* Python version: 3.x

Daksha1611:
Fixes #5316

`"autogen.sh"` is listed twice in `ignores_Build_scripts`, on two byte-identical lines. This drops the duplicate. The dict is unchanged, so there is no behaviour change and no CHANGELOG entry.


Shriprasad-P (commented):
## Review

**PR:** Remove duplicated autogen.sh entry in default ignores

**Touched:** `src/commoncode/ignore.py`

- Change set looks scoped to the stated goal; please confirm CI is green.
- Size looks manageable (+0/-1).

